PR #124, opened 2016-04-25
Added the default output filename to rs_mapper.

biochem-fan:
As per Harry's request.


graeme-winter:
This seems like a good idea and a harmless change; merged; thank you

PR #290, opened 2017-01-18
OpenMP broken on Windows

Anthchirp:
Hi @jmp1985, @rjgildea 
Could you have a look if this change makes sense?

If happy, could you please click on the green triangle next to _"Merge pull request"_ and choose _"Rebase and merge"_? Thanks :)

This is @oeffner's comment:

> I switched on openmp for nightly builds on windows yesterday. It turns out that the dials.algorithms.indexing module then no longer can be loaded from python. As far as I can see this is a compiler bug. OpenMP is not used anywhere in the dials code unless I'm mistaken. But the linker still tries to include the library at link time but doesn't do it correctly. Trying to load dials.algorithms.indexing then fails.
> 
> A quick and dirty way to rectify this is to #include <omp.h> say in dials\algorithms\indexing\index.h. Instead I tinkered with the dials\algorithms\indexing\SConscript to create a dummy source file that does the same and is linked to the dials.algorithms.indexing module whenever Dials is built on Windows.
> 
> If you think the attached SConscript is OK it would be great if you could commit it to your repo so that libtbx.import_all_ext test will pass without failure again on Windows.

### DIFF
--- a/algorithms/indexing/SConscript
+++ b/algorithms/indexing/SConscript
@@ -23,7 +23,17 @@ env_etc.include_registry.append(
 
 sources = [
     'boost_python/fft3d.cc',
-    'boost_python/indexing_ext.cc']
+    'boost_python/indexing_ext.cc'
+    ]
+
+# Since this code is not using OpenMP a bug in the Microsoft VS2008 compiler means that
+# any /openmp flag must be removed in order to avoid creating a broken library
+if env_etc.compiler == "win32_cl" and "/openmp" in env["SHCCFLAGS"]:
+  ccflags_without_omp = []
+  for f in env["SHCCFLAGS"]:
+    if "/openmp" != f:
+      ccflags_without_omp.append( f )
+  env["SHCCFLAGS"] = ccflags_without_omp
 
 env.SharedLibrary(target='#/lib/dials_algorithms_indexing_ext',
     source=sources,


rjgildea:
I have no idea why openmp would have anything to do with index.h...

jmp1985:
Hi Markus,

We don't use openmp in DIALS. Not sure why it should be included in index.h

Do we have a windows build machine I can use to look at this?

James

________________________________

Anthchirp:
No, not at DLS. Maybe @oeffner can help with that.
FWIW: it certainly sounds like there is something else going wrong - either with the OMP configuration or something conflicting in the dials source - that should be sorted out rather than adding an include when it's not needed.

jmp1985:
I'll have a look at installing DIALS on my old windows laptop then and see if I can reproduce the problem.

________________________________

Oeffner:
Hi all,
Possibly I didn't explain the problem well enough. The problem is that there is a bug in the VS2008 compiler used for Windows builds of Phenix and Dials. See 
http://cci-vm-6.lbl.gov:8010/builders/dials-nightly-intel-windows-x86_64/builds/328/steps/test%20dials/logs/stdio


When enabling OpenMP from bootstrap.py this is done in libtbx/SConscript by appending "/openmp" to CCFLAGS. This is then being copied by SConscripts in Dials I believe. The bug means that because no code in Dials uses OpenMP the linker gets confused and creates a broken library, dials_algorithms_indexing_ext.pyd, that cannot be loaded from python.

I agree that the first workaround with the dummy sourcefile is confusing. I'll try instead to remove the"/openmp" flag from CCFLAGS so that the linker won't attempt to include the OpenMP library altogether.

Rob

Oeffner:
Hi all,

I have now found a cleaner solution to the problem. Essentially, for Windows builds any "/openmp" flag is removed from the compiler flags in dials/algorithms/indexing/SConscript so that the linker doesn't create a broken library. With this change the dials.algorithms.indexing module now loads again without errors from python.

If you think this SConscript is OK it would be great if you would commit it.

Regards,

Rob

```
Import('env')


Import("env_base", "env_etc")

import libtbx, os, os.path

env_etc.dials_indexing_dist = libtbx.env.dist_path("dials")
env_etc.dials_indexing_include = os.path.dirname(env_etc.dials_indexing_dist)
env_etc.dials_indexing_common_includes = [
  env_etc.boost_include,
  env_etc.annlib_dist,
  env_etc.annlib_include[0],
  env_etc.annlib_include[1],
  env_etc.annlib_adaptbx_include[0],
  env_etc.annlib_adaptbx_include[1],
]

env_etc.include_registry.append(
  env=env,
  paths=env_etc.dials_indexing_common_includes)

  
sources = [
    'boost_python/fft3d.cc',
    'boost_python/indexing_ext.cc'
    ]

# Since this code is not using OpenMP a bug in the Microsoft VS2008 compiler means that
# any /openmp flag must be removed in order to avoid creating a broken library
if env_etc.compiler == "win32_cl" and "/openmp" in env["SHCCFLAGS"]:
  ccflags_without_omp = []
  for f in env["SHCCFLAGS"]:
    if "/openmp" != f:
      ccflags_without_omp.append( f )
  env["SHCCFLAGS"] = ccflags_without_omp

    
env.SharedLibrary(target='#/lib/dials_algorithms_indexing_ext',
    source=sources,
    LIBS=['ann']+env["LIBS"])

```

PR #293, opened 2017-01-21
rs_mapper: change box size ("cell parameters")

### LINKED ISSUE #261: rs_mapper - scale of the output
We've been playing with a [viewer][1] that could show both reciprocal lattice spots and a map generated by rs_mapper. So I get spots generated by @rjgildea 's [script](https://github.com/dials/dials_scratch/blob/master/rjg/rl_json.py)
and a map generated by @biochem-fan 's rs_mapper. The first one is in units 1/A, but the latter is not in any particular units?
I was advised to rescale the map by factor 2 / (grid_size * max_resolution), and it seems to work, i.e. it matches the spots found by dials:

rlp_and_map_2

but the max_resolution is not written in the map file.

Any thoughts? Ideally the "unit cell" of the map would be in 1/A and no re-scaling would be necessary, but I don't know if other programs (pymol etc) could handle it.

[1]: http://uglymol.github.io/reciprocal.html

wojdyr:
Andrey tested a reciprocal map in 1/A in coot and it didn't work well.
Coot is not expecting so small cells and can't show them well.
We didn't try PyMOL, but we thought it's better to scale the "cell parameters" by 100.
So the unit used is actually (100A)^-1.

Fixes #261

It can be tested in Reciprocal UM here:
http://uglymol.github.io/reciprocal.html
The spots there are from Dials tutorial data (th_8_2).
To add a map, use rs_mapper (after applying this patch) on the same data and then drag and drop the resulting ccp4 file to that page. The result should look like https://github.com/dials/dials/issues/261#issuecomment-274152537 (use '-' to decrease isolevel to around 0.02 rmsd).

### DIFF
--- a/command_line/rs_mapper.py
+++ b/command_line/rs_mapper.py
@@ -111,7 +111,12 @@ class Script(object):
                 self.process_imageset(imageset)
 
         recviewer.normalize_voxels(self.grid, self.cnts)
-        uc = uctbx.unit_cell((self.grid_size, self.grid_size, self.grid_size, 90, 90, 90))
+        # Let's use 1/(100A) as the unit so that the absolute numbers in the
+        # "cell dimensions" field of the ccp4 map are typical for normal
+        # MX maps. The values in 1/A would give the "cell dimensions" around
+        # or below 1 and some MX programs would not handle it well.
+        box_size = 100 * 2.0 / self.max_resolution
+        uc = uctbx.unit_cell((box_size, box_size, box_size, 90, 90, 90))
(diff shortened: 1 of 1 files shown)


Anthchirp:
Hi,

just a short comment - if you want to accept a pull request that is only a small isolated change it would be best to not click the big green **'merge'**-button, and instead go for the green triangle next to it, and select **'rebase and merge'** from the dropdown.

This is just to keep the repository history tidy. (But no harm done here.)

biochem-fan:
I see. Thanks for pointing it out.

PR #296, opened 2017-01-30
Fix test for export to json.

biochem-fan:


biochem-fan:
Fixed by 3727f85. This PR is no longer necessary.

PR #350, opened 2017-05-22
Enable post-integration filtering of overlapping reflections.

idyoung:
In the final step of integration, check for overlapping foreground between multiple spots or overlapping foreground with neighboring spots' background, if requested, and remove these reflections from the final result. (Default: off.)


graeme-winter:
This looks like something very useful for integration of data with > 1 lattice present, very interested.

PR #467, opened 2017-11-10
Changes to algorithms/refinement/refiner.py

mlxd:
These changes move specific python methods into Boost.Python extension modules to speed up refiner.py

Methods [panel_gp_nparam_minus_nref, unit_cell_nparam_minus_nref, model_nparam_minus_nref] are modified, and show large improvements in performance compared to the bare python code.

Testing of the new extensions has been added to test/command_line/tst_refine.py as test4()


dagewa:
Hi @mlxd, sorry for the delay. I want to look at this either tomorrow or Friday.

dagewa:
I get two new test failures: `libtbx.python test/command_line/tst_idials.py` and `libtbx.python test/algorithms/indexing/tst_index.py 10`. I haven't looked in detail to work out why this is yet, but the stacktrace for the second test suggests something has gone deeply wrong in refinement as the RMSDs are crazy:
```
...
    assert actual <= expected, "%s %s" %(self.rmsds, expected_rmsds)
AssertionError: (122.64471402348383, 176.42199729850896, 0.006744603050930485) (0.33, 0.4, 0.0022)
```

mlxd:
Hi @dagewa, thanks for the help. The changes I have made to `dials/algorithms/refinement/refiner.py` in `def _filter_reflections` (~L395) by sorting the reflection table before returning it seem to be the cause of the RMSD errors in `test/algorithms/indexing/tst_index.py 10`.  This sorting is necessary for the C++ modules to work, so I'll look further into this and push a fix when I find it. I'll examine the `libtbx.python test/command_line/tst_idials.py` when the previous issue is fixed.

dagewa:
Hi @mlxd, I have a feeling it is related to this:
https://github.com/dials/dials/blob/master/algorithms/indexing/indexer.py#L1621

Here the indexer relies on the fact that the result of `refiner.predict_for_indexed()` is in the same order as  the `reflections` passed into refinement. What we could do in that case is undo the sort in the method  `Refiner.predict_for_indexed` (and `Refiner.selection_used_for_refinement`). This would be easy enough using the column `'iobs'`, but at the moment the sort is performed before `'iobs'` is set, which occurs only when the `ReflectionManager` is constructed: https://github.com/dials/dials/blob/master/algorithms/refinement/reflection_manager.py#L186.

I now can't see why I made `_filter_reflections` a method of the `RefinerFactory`. It seems like that functionality belongs within the `ReflectionManager` instead, so that everything related to manipulating the reflection table occurs within one object. So when that is constructed, the full reflection table could be passed in, a copy made keeping the columns of interest, then `'iobs'` set, followed by the sorting.

I haven't tested, but I think that, plus re-sorting on `'iobs'` in those method calls that need it might be a solution here?

mlxd:
Hi @dagewa thanks for the input. I've tried a few changes over the past few days, but none that fix both errors simultaneously. Setting `iobs` followed by sorting sounds like a good solution though. I'll take a look today and try to test it out for expected results.

mlxd:
All tests are now passing again (except for my changes to `tst_refine.py`, which requires some modifications given the changes) for implementing the sorting at https://github.com/dials/dials/blob/master/algorithms/refinement/reflection_manager.py#L221 and adding the sort by `iobs` at https://github.com/dials/dials/blob/master/algorithms/refinement/refiner.py#L2061

I will change the tests for these extensions, and make sure it behaves well before committing anything here. I may ask for some input if you would like `_filter_reflections` to be added to `RefinerFactory`, as I am not sure what would be the best way to handle this (I would rather not simply copy and paste the method without some input). Thanks for the help so far.

dagewa:
Sounds good. I look forward to seeing the changes and am happy to help tidy up.

mlxd:
The added changes seem to fix the errors. Provided that the modified functions in `refiner.py` are always called with the list sorted in the way as expected (by 'id' and next by 'panel') then everything should be fine.

algorithms/refinement/reflection_manager.py
```diff
@@ -185,6 +185,19 @@ def __init__(self, reflections,
     # keep track of the original indices of the reflections
     reflections['iobs'] = flex.size_t_range(len(reflections))
 
+    #Check for monotonically increasing value range. If not, ref_table isn't sorted,
+    # and proceed to sort by id and panel. This is required for the C++ extension
+    # modules to allow for nlogn subselection of values used in refinement.
+    l_id = reflections["id"]
+    id0 = l_id[0]
+    for ii in xrange(1,len(l_id)):
```
dagewa: I was surprised how fast this Python loop is. I tested with a flex.size_t with 1 million elements (not inconceivable for a large synchrotron MX dataset) and this completes in about 0.8s on my desktop

Anthchirp:
merged, with some of the 'clean clutter'-like commits squashed.

Issue #459, opened 2017-10-19
Make DIALS python3 syntax compatible


Anthchirp:
One step on the way to Python3 compatibility is to eliminate Python3 syntax errors. At this point we have 300 files in violation.

This [list of current error messages for DIALS](https://github.com/cctbx/python3/blob/master/dials.log) and the [status across all cctbx modules](https://github.com/cctbx/python3/blob/master/README.md) is updated daily.

This will be mostly print() stuff, however it would be much appreciated to instead of simply sprinkling ((())) all over the place to consider whether it would be more appropriate to, for example, use the [python logging framework](https://docs.python.org/2/library/logging.html) instead.

Anthchirp:
As part of this I would like to unleash the _"Jenkins Python3 monster"_ (quote) on the DIALS repository. We have this currently enabled for [xia2](https://github.com/xia2/xia2).

It means adding a test that enforces Python3 syntax compatibility for all new files, and for all files that are already Python3 syntax compatible. Nothing changes for existing files that are not Python3 syntax compatible.
The aim is to monotonically decrease the number of violating files over time and gently nudge developers to get used to Python3 syntax.

If there are no objections I would like to enable this in about a week, changes are in pull request is #475.

Anthchirp:
This is now live.

PR #475, opened 2017-11-23
Enforce Python3 syntax for all new files

Anthchirp:
See #459. Will merge this in ~ a week if no objections


Anthchirp:
The list of known violations may change until merge-day.
To get an updated list run ```pytest $(libtbx.find_in_repositories dials)/test/test_python3_regression.py```, making sure a python3 executable is available in the path.
Then ```mv $(libtbx.find_in_repositories dials)/.known-python3-violations.log $(libtbx.find_in_repositories dials)/.known-python3-violations```

PR #532, opened 2018-04-10
Removing dials extensions

### LINKED ISSUE #530: Retire DIALS extension framework
Spinoff from #257, discussion and rationale there.

jmp1985:
Remove the dials extension mechanism. This involved looking in certain directories for algorithms and loading them dynamically. Unfortunately, this process contributed to dials taking a while to load due to excessive file system operations. 

The changes, remove the dynamic loading of extensions but preserve the algorithm interface. Package utils is used to define entry points for profile models so that these can be loaded from experiments.json files in dxtbx. As such, there is a branch in dxtbx that has some related changes (https://github.com/cctbx/cctbx_project/tree/removing_dials_extensions) Fixes #530


Anthchirp:
Thank you! I will look at this shortly

Anthchirp:
rebased to fix one merge conflict. have a few suggested changes, running tests now.

Anthchirp:
Another suggestion.

PR #540, opened 2018-04-20
Xtc speedup

asmit3:
Changes in dials sources using features introduced in cctbx/cctbx_project#179

1. imageset is setup using the __class__ method. This helps Just-in-Time reading of the models instead of instantiating all of them at the beginning. 

2. User can specify reference_geometry, which applies geometry from a phil file, or can specify geometry manually using the geometry phil options.  This commit swaps the order these are applied to match the order in xtc_process.  A positive side effect is faster performance since applying the reference geometry now is in the do_work function for composite files, so each image only has the geometry applied once, instead of once per process.

By 
Aaron Brewster and Asmit Bhowmick

### DIFF
--- a/command_line/dials_import.py
+++ b/command_line/dials_import.py
@@ -308,7 +308,6 @@ class ManualGeometryUpdater(object):
     Override the parameters
 
     '''
-    from dxtbx.imageset import ImageSet
     from dxtbx.imageset import ImageSweep
     from dxtbx.model import BeamFactory
     from dxtbx.model import DetectorFactory
@@ -316,7 +315,7 @@ class ManualGeometryUpdater(object):
     from dxtbx.model import ScanFactory
     from copy import deepcopy
     if self.params.geometry.convert_sweeps_to_stills:
-      imageset = ImageSet(data=imageset.data())
+      imageset = imageset.__class__(data=imageset.data())
     if not isinstance(imageset, ImageSweep):
       if self.params.geometry.convert_stills_to_sweeps:
         imageset = self.convert_stills_to_sweep(imageset)
--- a/command_line/dials_import.py
+++ b/command_line/dials_import.py
@@ -308,14 +308,14 @@ class ManualGeometryUpdater(object):
     Override the parameters
 
     '''
-    from dxtbx.imageset import ImageSweep
+    from dxtbx.imageset import ImageSweep, ImageSetFactory
     from dxtbx.model import BeamFactory
     from dxtbx.model import DetectorFactory
     from dxtbx.model import GoniometerFactory
     from dxtbx.model import ScanFactory
     from copy import deepcopy
     if self.params.geometry.convert_sweeps_to_stills:
-      imageset = imageset.__class__(data=imageset.data())
+      imageset = ImageSetFactory.imageset_from_anyset(imageset)
     if not isinstance(imageset, ImageSweep):
       if self.params.geometry.convert_stills_to_sweeps:
         imageset = self.convert_stills_to_sweep(imageset)
--- a/command_line/stills_process.py
+++ b/command_line/stills_process.py
@@ -193,11 +193,10 @@ def do_import(filename):
     raise Abort("Got multiple datablocks from file %s"%filename)
 
   # Ensure the indexer and downstream applications treat this as set of stills
-  from dxtbx.imageset import ImageSet
   reset_sets = []
 
   for imageset in datablocks[0].extract_imagesets():
-    imageset = ImageSet(imageset.data(), imageset.indices())
+    imageset = imageset.__class__(imageset.data(), imageset.indices())
     imageset.set_scan(None)
     imageset.set_goniometer(None)
     reset_sets.append(imageset)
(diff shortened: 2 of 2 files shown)


graeme-winter:
I've asked Nick and Markus to also look at this with a technical hat on, as this looks like rather a fundamental change (in particular, we use the multi reader for Eiger stuff, too)

I understand that the TL;DR version of the description above is essentially "lazy loading" - is this correct? Seems like a useful thing to have...

command_line/stills_process.py
```diff
@@ -332,6 +319,20 @@ def do_work(i, item_list):
         processor = Processor(copy.deepcopy(params), composite_tag = "%04d"%i)
 
         for item in item_list:
+          try:
+            for imageset in item[1].extract_imagesets():
+              update_geometry(imageset)
+          except RuntimeError as e:
+            logger.info("Error updating geometry on item %s, %s"%(str(item[0]), str(e)))
```
Anthchirp: I would suggest *logger.warning* at a minimum. Additionally set keyword *exc_info=True* to propagate exception information.

command_line/stills_process.py
```diff
@@ -332,6 +319,20 @@ def do_work(i, item_list):
...
+            continue
+
+          if self.reference_detector is not None:
+            from dxtbx.model import Detector
```
Anthchirp: suggest moving import from middle of for loop to top of file

command_line/stills_process.py
```diff
@@ -362,12 +363,16 @@ def do_work(i, item_list):
           if len(imagesets[0]) > 1:
             raise Abort("Found a multi-image file. Run again with pre_import=True")
 
+          try:
+            update_geometry(imagesets[0])
+          except RuntimeError as e:
+            logger.info("Error updating geometry on item %s, %s"%(tag, str(e)))
```
Anthchirp: as above

command_line/dials_import.py
```diff
@@ -308,15 +308,14 @@ def __call__(self, imageset):
...
-    from dxtbx.imageset import ImageSet
     from dxtbx.imageset import ImageSweep
     from dxtbx.model import BeamFactory
     from dxtbx.model import DetectorFactory
     from dxtbx.model import GoniometerFactory
     from dxtbx.model import ScanFactory
     from copy import deepcopy
     if self.params.geometry.convert_sweeps_to_stills:
-      imageset = ImageSet(data=imageset.data())
+      imageset = imageset.__class__(data=imageset.data())
```
Anthchirp: should this be passed say an *imagesetfactory* function rather than a class instance?
(If I was to subclass imageset and change the ```__init__``` function this will blow up, right?)

command_line/stills_process.py
phyy-nx: Wouldn't that create an unneeded import for situation where there is no reference detector?

command_line/dials_import.py
phyy-nx: We have subclassed imageset (ImageSetJIT), which is why this change is needed.  There is no factory function of ImageSet that takes ImageSetData instead of filenames, and if there were, it would have to pick between ImageSet, ImageSetJIT and MemImageSet according to some criteria.  Since we already know what the correct class is when ManualGeometryUpdater is called, we can just create a new instance of the class instead.

command_line/stills_process.py
```diff
@@ -345,6 +332,20 @@ def do_work(i, item_list):
         processor = Processor(copy.deepcopy(params), composite_tag = "%04d"%i)
 
         for item in item_list:
+          try:
+            for imageset in item[1].extract_imagesets():
+              update_geometry(imageset)
+          except RuntimeError as e:
+            logger.warning("Error updating geometry on item %s, %s"%(str(item[0]), str(e)))
```
Anthchirp: add ```exc_info=True``` keyword argument to logger.warning()
replace % with ,
```python
logger.warning("Error updating geometry on item %s, %s", str(item[0]), str(e), exc_info=True)
```

command_line/stills_process.py
```diff
@@ -375,12 +376,16 @@ def do_work(i, item_list):
           if len(imagesets[0]) > 1:
             raise Abort("Found a multi-image file. Run again with pre_import=True")
 
+          try:
+            update_geometry(imagesets[0])
+          except RuntimeError as e:
+            logger.warning("Error updating geometry on item %s, %s"%(tag, str(e)))
```
Anthchirp: as above. Replace with
```python
logger.warning("Error updating geometry on item %s, %s", tag, str(e), exc_info=True)
```

Anthchirp:
We spoke to @jmp1985 and he will review the pull request on Monday or Tuesday.

command_line/stills_process.py
```diff
@@ -193,11 +193,10 @@ def do_import(filename):
     raise Abort("Got multiple datablocks from file %s"%filename)
 
   # Ensure the indexer and downstream applications treat this as set of stills
-  from dxtbx.imageset import ImageSet
   reset_sets = []
 
   for imageset in datablocks[0].extract_imagesets():
-    imageset = ImageSet(imageset.data(), imageset.indices())
+    imageset = imageset.__class__(imageset.data(), imageset.indices())
```
jmp1985: I'm not sure, I like calling imageset.__class__  like this, perhaps we could add a classmethod factory function to imageset that instantiates a new imageset of that class. Or something like that.

PR #548, opened 2018-05-01
Travis

Anthchirp:
Initial build and one build per week (triggered per cron job) takes ~80 minutes
Subsequent builds will then speed up to approx ~15 minutes.
Runs on every branch, every pull request. Including this one.
Runs all dials tests that are not marked as slow, and do not depend on any *_regression repository
Uses cctbx master branch. Whenever that is broken it may affect builds here.
Changes in the base build can take up to one week to propagate here.

So do we want this? 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,67 @@
+sudo: false
+
+addons:
+  apt:
+    packages:
+      - libjpeg62
+
+language: python
+python:
+  - "2.7"
+
+cache:
+  directories:
+    - $HOME/build_dials
+
+jobs:
+  include:
+    - stage: build
+      install:
+         # Setup a base installation
+         - ./.travis/setup-base
+
+         # Inject DIALS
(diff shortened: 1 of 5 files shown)


ndevenish:
+1 from me

graeme-winter:
I thought they had to be all done in 50 minutes?

ndevenish:
> I thought they had to be all done in 50 minutes?

Per job. You can split a build into multiple jobs, but it's best done like here, e.g. separate build and test phases. Splitting a massive build into several parts gets a little awkward, and have to be re-queued each time (more of a problem with mac builds, which are more limited in number and therefore queueing time).

@Anthchirp has avoided a lot of the problem by using the pre-built diamond bundle instead of doing everything from scratch.

graeme-winter:
Well it sounds like *a good thing* particularly on the branches etc... is there literally any down side to this beyond the increased chance of actually finding bugs?

PR #562, opened 2018-05-22
Don't keep a copy of symmetrize_reduce_enlarge around

dagewa:
For cctbx/cctbx_project#196

Simple changes to avoid keeping a copy of `symmetrize_reduce_enlarge` in the crystal parameterisation, so that this can be pickled. Set as a pull request because I see on average ~1% increase in run time for a basic test case in dials_regression/refinement_test_data/multi_stills:

`dials.refine combined_experiments.json combined_reflections.pickle`

But this test case is not particularly helpful as it is very fast anyway and is largely set up overhead. It would be useful if this can be tested against a larger XFEL refinement job that involves many more crystals. @phyy-nx if you have such a test case to hand, could you please try it out and check that these changes result in an acceptable increase in run time?

### DIFF
--- a/algorithms/refinement/parameterisation/crystal_parameters.py
+++ b/algorithms/refinement/parameterisation/crystal_parameters.py
@@ -109,11 +109,11 @@ class CrystalUnitCellMixin(object):
     select between versions of the Parameter class"""
 
     # Set up symmetrizing object
-    self._S = symmetrize_reduce_enlarge(crystal.get_space_group())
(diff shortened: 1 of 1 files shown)


phyy-nx:
I ran a 3000 crystal refinement using SparseLevMar at LCLS on the local cluster.  I did 10 trials and recorded the timings.  The mean time increased 2%.  Thumbs up.

PR #567, opened 2018-06-01
move away from .show()

Anthchirp:
following discussion of .show() here is a concrete example of the (alleged) C++ premature myth optimization pattern in action in pure python land, and a remedial patch.

Is this something we want to get rid of - at least in DIALS land?


rjgildea:
Yes - I frequently find it annoying to have to do the StringIO -> show() -> actual string dance, e.g. here where it takes 4 lines of code to do what should probably only be a single line:
https://github.com/dials/dials/blob/b499c6191caa18e04ddaf190321a3da6c14af559/command_line/show.py#L203

Anthchirp:
2:0 motion passed.

PR #598, opened 2018-07-11
Add a progress bar to dials.util

ndevenish:
Here's a progress bar for dials. It's based on the [tqdm](https://github.com/tqdm/tqdm) progress bar, but with the additional feature that the progress bar won't be shown at all if the estimated total time is too short (default: 10 seconds, set by `min_time` option). Some of the configuration defaults have also, been changed; by default it won't show at all for non-TTY, changes column size dynamically and deletes itself when finished.

Usage and most of the functionality is the same as [tqdm](https://github.com/tqdm/tqdm), but for a basic example of how it's used:
```python
  >>> from dials.util import progress
  >>> for i in progress(range(10)):
        ...
```
And here's an example of running on my WIP datablock refactor (without this, it's hard to pin down a single cycle of reading to mark progress) - note that the first import completes in < 10 seconds so doesn't show the bar at all:
output

We could just use the base tqdm package but I thought people would object to progress bars appearing for what is often pretty short processes. The cost for this change is that it isn't quite as optimally efficient as it's parent package.

- [ ] Would this be useful, and is it the kind of thing people think we should put in?


graeme-winter:
I thought @jmp1985 had a progress bar implementation in there somewhere already - however this is a useful and nice user feature so I see no objections to merging (also: will presumably be broadly harmless in cases where is not a tty etc.)

jmp1985:
There is also dials.util.command_line.ProgressBar. It looks like I wrote this back in 2013. As far as I recall, we previously used this during spot finding and integration where we were reading images but eventually decided against using progress bars in those situations. If the new implementation is better, then we should probably remove the old progress bar.

________________________________

libtbx_refresh.py
```diff
@@ -29,6 +29,7 @@
 libtbx.pkg_utils.require('procrunner', '>=0.6')
 libtbx.pkg_utils.require('scipy')
 libtbx.pkg_utils.require('scikit_learn[alldeps]')
+libtbx.pkg_utils.require('tqdm', '==4.23.4')
```
Anthchirp: do we need to lock to a specific version?

util/__init__.py
```diff
@@ -1,7 +1,11 @@
 from __future__ import absolute_import, division, print_function
 
+__all__ = ["progress"]
+
 import sys
 
+from ._progress import progress
```
Anthchirp: do we have to?

I know ```from dials.util.progress import progress``` doesn't look as nice

libtbx_refresh.py
ndevenish: Yes - unfortunately the source package doesn't support turning on/off after a while, only being completely on or totally disabled. I've had to hook into private implementation-specific code, so obviously this has potential to change/break on even minor updates.

util/__init__.py
ndevenish: I suppose it could be argued that unnecessary repetition is totally idiomatic of our codebase, things like:
```
from dxtbx.format.Format import Format
from dxtbx.datablock import Datablock
```
for which at least `Format` has the excuse that it's the same case as the module. But it really doesn't look nice, and package layout should be an implementation detail, right?

phyy-nx:
FYI LCLS developers: you can use `libtbx.pip install tqdm` to get this dependency. I will update the documentation at https://exafel.github.io/docs/psana-cctbx-install to include it.

@Anthchirp could you split the tqdm dependency out of cctbx/cctbx_project#229 and check that in?  

FYI, the conda plans, as described on @bkpoon's cctbxbb post, will make this all easier by allowing projects built on cctbx to add whatever dependencies they'd like. The idea is that there will be a dials-dependencies package that depends on cctbx-dependencies.  Likewise there will be a phenix-dependencies package that depends on both.  Currently there is one list that applies to everyone.  With the new approach, dependencies, can be updated independently by individual projects and conda handles the conflict resolution.  The packages on conda will only specify the minimum api version required for each dependency, eg python 2.7 instead of python 2.7.15, but installer bundles distributed by us will include a manifest that specifies the full list of dependencies, with exact versions and build numbers.  This will be valuable for tracking releases, for developers looking for a quick install path, and for reproducibility in publications.

See
https://github.com/cctbx/conda_build
and
https://anaconda.org/cctbx/cctbx_dependencies

Sample manifest attached to this post:
http://phenix-online.org/pipermail/cctbxbb/2018-September/002003.html

PR #606, opened 2018-08-13
Refactor export mtz

jbeilstenedmands:
This pull request primarily contains changes to the export code, replacing the `_apply_data_filters` function from `export-mtz` with a set of functions/classes in `dials.util.filter_reflections`.
Also affected and updated are `dials.symmetry` and `dials.cosym`, which had called `_apply_data_filters` and `dials.scale`, which calls `dials.export`. As `xia2` relies on the command line options, this also needs updating - I have created a separate branch in `xia2` implementing the changes that should be necessary, which should be merged in when this pull request is merged.

To fix various issues with the exporting code necessitated changing the command line options.
The old options removed are:
`ignore_profile_fitting=False`, `summation=False`, `include_partials=False`, `keep_partials=False`, `scale_partials=True`, `apply_scales=False`.
The new options implemented are:
```
intensity = *prf *sum scale
    .type = choice(multi=True)
    .help = "Choice of which intensities to export. Allowed combinations:
            scale, prf, sum, prf + sum, prf + sum + scale."
combine_partials = True
      .type = bool
      .help = "Combine partials that have the same partial id into one
        reflection, with an updated partiality given by the sum of the
        individual partialities."
partiality_threshold=0.99
      .type = float
      .help = "All reflections with partiality values above the partiality
        threshold will be retained. This is done after any combination of
        partials if applicable."
```
The idea is that one chooses a valid combination of prf, sum and scale, which have a separate algorithm implementations with sensible defaults for the given case. The only thing left to choose is if you want to combine partials together, and above what partiality threshold you want to keep the reflections (feedback on what this value should be, currently set to 0.99 to give same behaviour as before).

I have attempted to maintain the current behaviour as much as possible, with two main use cases in mind:
1. For exporting after integration (for scaling in aimless) one would choose either intensity=prf, sum or prf+sum; filtering will be performed, and lp, dq and partiality corrections will be applied.
2. For exporting after scaling, one should choose intensity=scale; prf and sum intensities will be removed, filtering will be performed and scale factors will be applied.

One issue I would like feedback on is the case of prf+sum, where not all reflections have a prf value, as may be the case for thin-wedge data. 
Currently, if intensity=prf+sum, only reflections with both prf and sum are retained, but this can be changed by setting intensity=sum. Hopefully this can be improved in the near future.

I am sure that this code will require further refining once used in-the-field, however I hope this is a sensible framework to allow easy changes and adaptations.


Anthchirp:
No problem holding for 1.11
Please consider squashing these when merging. If unclear about how then just assign to me once ready.

jbeilstenedmands:
Update:

1. documentation should all be fine now as there were no references to the old options.
2. xia2 changes cleaned up
3. the X4_wide::test_dials xia2 test is the only failing test due to the changes, the tolerance check is not passing due to
-Low resolution limit                             11.21(5%)     11.21(**)    0.64(**)     11.21(0.03) 
>                                                *9.25     *                             *9.25       *

Not sure what's going on here as behaviour should be the same as before, so investigating further

jbeilstenedmands:
I have updated the test_small_molecule expected result file, as it appears only the resolution limits were the difference, other metrics such as cc12, R-factors etc seem the same so probably a change of a reflection or two due to partiality handling.
All xia2 test should now be passing for these branches so I am happy that the changes are ready to be merged in.

jbeilstenedmands:
@Anthchirp Made most of the requested changes, except a few for the mmcif export as I remember this being quite peculiar when I was writing it so would rather keep as-is for now

PR #622, opened 2018-09-28
Probplots in report

benjaminhwilliams:
Dropped the ball on this a while ago.  Picking it up again.

A question before this pull request can be accepted:

- Should this information be in [`dials/command_line/report.py`](https://github.com/dials/dials/blob/probplots_in_report/command_line/report.py) or would it be better for it to live in [`xia2/command_line/report.py`](https://github.com/xia2/xia2/blob/probplots_in_report/command_line/report.py)? See [parallel pull request in `xia2/xia2`](https://github.com/xia2/xia2/pull/255).

Closes xia2/xia2#234.

### DIFF
--- a/command_line/report.py
+++ b/command_line/report.py
@@ -1959,8 +1959,8 @@ class ZScoreAnalyser(object):
     from dials.util.intensity_explorer import IntensityDist
 
     z_scores = IntensityDist(
-        rtable=rlist, elist=experiments).rtable['intensity.z_scores']
-    hist = flex.histogram(z_scores, -5, 5, 100)
+        rtable=rlist, elist=experiments).rtable['intensity.z_score']
+    hist = flex.histogram(z_scores, -10, 10, 100)
 
     return {
       'z_score_histogram': {
@@ -1972,7 +1972,7 @@ class ZScoreAnalyser(object):
         }],
         'layout': {
           'title' : 'Histogram of z-scores',
-          'xaxis' : {'title': 'z-score', 'range': [-5, 5]},
+          'xaxis' : {'title': 'z-score', 'range': [-10, 10]},
           'yaxis' : {'title': 'Number of reflections'},
           'bargap': 0,
         },
--- a/command_line/report.py
+++ b/command_line/report.py
@@ -1896,6 +1896,89 @@ class IntensityAnalyser(object):
       pyplot.close()
 
 
+class ZScoreAnalyser(object):
+  """
+  Analyse the distribution of intensity z-scores.
+
+  z-scores are calculated as z = (I - <I>) / σ, where I is intensity,
+  <I> is a weighted mean and σ is the measurement error in intensity,
+  as calculated by dials.util.intensity_explorer.
+  """
+
+  def __init__(self):
+    from os.path import join
+
+    # Set the required fields
+    self.required = [
+      "intensity.sum.value",
+      "intensity.sum.variance",
+      "xyzobs.px.value",
+    ]
+
+  def __call__(self, rlist, experiments):
+    """
+    :param rlist: A reflection table containing, at the least, the following
+    fields:
+      * ``miller_index``
+      * ``intensity.sum.value``
+      * ``intensity.sum.variance``
+      * ``xyzobs.px.value``
+    :type rlist: `dials_array_family_flex_ext.reflection_table`
+    :return: A dictionary describing a Plotly plot.
+    :rtype:`dict`
+    """
+
+    # Check we have the required fields
+    print("Analysing reflection intensities")
+    if not ensure_required(rlist, self.required):
+      return {}
+
+    selection = rlist['intensity.sum.variance'] <= 0
+    if selection.count(True) > 0:
+      rlist.del_selected(selection)
+      print(u" Removing %d reflections with variance ≤ 0" % \
+            selection.count(True))
+
+    selection = rlist['intensity.sum.value'] <= 0
+    if selection.count(True) > 0:
+      rlist.del_selected(selection)
+      print(u" Removing %d reflections with intensity ≤ 0" % \
+            selection.count(True))
+
+    d = OrderedDict()
+
+    print(" Analysing distribution of intensity z-scores")
+    d.update(self.z_score_hist(rlist, experiments))
+
+    return d
+
+  def z_score_hist(self, rlist, experiments):
+    """
+    Plot a histogram of z-scores.
+    """
+    from dials.util.intensity_explorer import IntensityDist
+
+    z_scores = IntensityDist(
+        rtable=rlist, elist=experiments).rtable['intensity.z_scores']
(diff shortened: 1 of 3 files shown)


command_line/report.py
```diff
@@ -1896,6 +1896,256 @@ def num_foreground_hist(self, rlist):
...
+  """
+  Analyse the distribution of intensity z-scores.
+
+  z-scores are calculated as z = (I - <I>) / σ, where I is intensity,
+  <I> is a weighted mean and σ is the measurement error in intensity,
+  as calculated by dials.util.intensity_explorer.
+  """
+
+  def __init__(self):
+    from os.path import join
```
Anthchirp: No. For many reasons.
1. Never inline import python standard library modules.
2. Don't just import names unless there is a good reason.
3. 'join' has a name conflict with 'join', so there is no good reason.
4. import is unused.

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+
+    from dxtbx.model import ExperimentList
+    from cctbx import miller
+
+    try:
+      assert(isinstance(rtable, flex.reflection_table))
+      assert(isinstance(elist, ExperimentList))
+    except AssertionError:
+      raise TypeError(
+          "Must be called with a reflection table and an experiment list.")
```
Anthchirp:
```
if not isinstance(..) \
    or not isinstance(..):
  raise TypeError(..)
```

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+    uncertainty_name = {
+      'sigma': 'intensity.sum.variance',
+      'stddev': 'intensity.mean.variance'
+    }[uncertainty]
+
+    for key, rtable in self.rtables.items():
+      try:
+        uncertainty_value = flex.sqrt(rtable[uncertainty_name])
+      except KeyError:
+        from logging import warn
```
Anthchirp: python standard import. Off to the top.

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+        uncertainty_value = flex.sqrt(rtable['intensity.sum.variance'])
+        warn(u"""Weighted variances haven't been calculated,
```
Anthchirp: Correct:
```python
import logging
log = logging.getLogger('dials.util.intensity_explorer')
(...)
   def myfunction(..):
      (..)
      log.warn(...)
```

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+      z = (rtable['intensity.sum.value']
+           - rtable['intensity.mean.value']) / uncertainty_value
+      rtable['intensity.z_score'] = z
+
+      self.rtables[key] = rtable
+
+  def _probplot_data(self):
+    """Generate the data for a normal probability plot of z-scores."""
+
+    from scipy import stats as ss
```
Anthchirp: I think ```import scipy.stats``` takes less effort to understand, even though the method is a short one.

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+            label='$z$', bins=100, range=(-5, 5))
+    fig.savefig(self.outfile + '_zhistogram', transparent=True)
+    plt.close()
+
+  def _plot_symmetry_equivalents(self,
+                                 overlay_mean=False,
+                                 overlay_uncertainty=False):
+    """Really just a test plot.  Slow.  You probably don't want to use."""
+
+    from matplotlib import pyplot as plt
```
Anthchirp: import used multiple times. Consider moving to top, leaving out the ```as``` bit and using ```pyplot``` throughout.

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+    plt.close()
+
+  def plot_z_map(self, minimum=0):
+    """
+    Plot a z-score heatmap of the detector.
+
+    Beware, this is only meaningful if the data have a single geometry model.
+    """
+
+    import math
```
Anthchirp: standard library import

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+
+  elist = ExperimentList([Experiment(crystal=Crystal(*crystal_params))])
+
+  return rtable, elist
+
+
+def data_from_pickle_and_json():
+    from dials.util.options import OptionParser, flatten_reflections,\
+      flatten_experiments
+    import libtbx.load_env
```
Anthchirp: drop import

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+
+    dials.util.intensity_explorer indexed.pickle experiments.json
+
+    dials.util.intensity_explorer integrated.pickle integrated_experiments.json
+
+    '''
+
+    # Create the parser
+    usage = ("usage: %s [options] reflections.pickle experiments.json" %
+             libtbx.env.dispatcher_name)
```
Anthchirp: You spelt out the name in the comment above, so can do it here as well. Removes the libtbx dependency.

util/intensity_explorer.py
```diff
@@ -0,0 +1,623 @@
...
+    return rtable, elist
+
+
+if __name__ == "__main__":
+  # FIXME Now ∃ a dials tool for (unmerged MTZ) –> (exp list, refl table),
+  # use it!
+  # TODO Allow determination of output filename root.
+  # FIXME Give a pickle and a json file as arguments:
+
+  import sys
```
Anthchirp: Standard library import.

util/intensity_explorer.py
benjaminhwilliams: I thought you might say that.  I was trying to keep an expensive import out of the way because the methods that use `pyplot` are unlikely to get much use.  In light of that, a better arrangement is to remove these methods altogether.

util/intensity_explorer.py
benjaminhwilliams: Agreed.

benjaminhwilliams:
Naively, I image `.extend(rtable)` is O(len(rtable)), so you have a good point.  I'll think whether it can be trimmed down easily.

util/intensity_explorer.py
Anthchirp: I see. Maybe bundle the mathplotlib-depedent methods into a visualization class in a separate file that takes an ```IntensityDist```?

benjaminhwilliams: Exactly.  Currently living in dials_scratch while I do something else.

benjaminhwilliams:
In answer to my question as to whether this pull request or [this parallel xia2 pull request](https://github.com/xia2/xia2/pull/255) should be merged, it has been decided to merge both.

I am deferring the question of speed improvements (alternatives to `reflection_table`'s `append` method) for now.

PR #675, opened 2018-12-19
Changes to integrate.py to handle memory issues

asmit3:
The work in this pull request stemmed from our failure to use **cctbx.xfel.stripe_experiments** during our recent LU34 beamtime at LCLS. The protein studied, PSII, with it's large unit cell and spots to be integrated to the edges creates a huge number of reflections to be handled by the integration program. **cctbx.xfel.stripe_experiments** refines an experimentlist and the reintegrates the data. During the reintegration step, the experimentlist/reflection table when passed to *integrate.py* is integrated all at once in the current form on the master. This creates a memory problem when there are multiple experiments in the experimentlist, each with 1000s of reflections and cause the program to fail.

  To address this issue, we propose in this pull request to break up the integration into a couple of *for loops* and integrate each experiment separately. In the process, we are able to delete the shoeboxes as we go along which is the biggest memory consumer. This fix allowed us to use stripe_experiments on PSII data. Note that this problem is different from the memory errors sometimes seen when integrating sweeps with high mosaicity. Here we are trying to integrate thousands of stills simultaneously and they can be split trivially. 

Work by Asmit Bhowmick and Aaron Brewster

### DIFF
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -271,6 +271,7 @@ class Script(object):
     # Match reference with predicted
     if reference:
       all_reference = flex.reflection_table()
+      all_predicted = flex.reflection_table()
       for ii in range(len(experiments)):
         preds = predicted.select(predicted['id']==ii)
         refs = reference.select(reference['id']==ii)
@@ -292,6 +293,8 @@ class Script(object):
           logger.info('')
         rubbish.extend(unmatched)
         all_reference.extend(refs)
+        all_predicted.extend(preds)
+      predicted=all_predicted
       reference=all_reference
 
       if len(experiments) > 1:
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -270,22 +270,29 @@ class Script(object):
 
     # Match reference with predicted
     if reference:
-      matched, reference, unmatched = predicted.match_with_reference(reference)
-      assert(len(matched) == len(predicted))
-      assert(matched.count(True) <= len(reference))
-      if matched.count(True) == 0:
-        raise Sorry('''
-          Invalid input for reference reflections.
-          Zero reference spots were matched to predictions
-        ''')
-      elif len(unmatched) != 0:
-        logger.info('')
-        logger.info('*' * 80)
-        logger.info('Warning: %d reference spots were not matched to predictions' % (
-          len(unmatched)))
-        logger.info('*' * 80)
-        logger.info('')
-      rubbish.extend(unmatched)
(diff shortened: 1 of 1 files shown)


jmp1985:
Hi @asmit3 

Sorry for taking a while to reply to this. In principle this is fine; however, I think we need to perform some additional checks to do the appropriate thing. For example, in the case of multiple crystals in a sweep, we would want to integrate all the experiments together so iterating over the experiments in this way would not be the right thing to do. We could add a parameter to explicitly trigger this behaviour i.e. something like integrate_each_experiment_independently=True or perhaps something similar but less verbose!

jmp1985:
Hi Aaron,

As an example, lets say we have a sweep which has lets say 2 crystals in the beam and we have a jointly refined model for these two experiments and each has it's own scan varying crystal model. In this case we can pass the experiment list containing the two experiments which refer to the same imageset and on each image we can predict all the spots from both crystals and we can integrate reflections from both crystals at the same time. This is the only way to properly handle overlaps resulting from reflections from multiple lattices.

This is something that is currently not handled optimally but is currently possible using the 3d_threaded integrator. So we should try to preserve the ability to do this. In that case, when we have multiple experiments, we don't always want to just iterate over each experiment and integrate it separately; I guess there must be cases where you have multiple crystals on a single image for stills too but I don't think we even try to integrate both lattices at the same time in that case - summation integration is problematic anyway with overlaps! 

In summary, I'd like to keep the ability to process all experiments together but I don't mind having a mode to iterate over experiments in this way if that is necessary.

phyy-nx:
Fair enough.  @asmit3 let's add a phil parameter as James suggests leaving the current behavior unchanged.  Then we'll modify stripe_experiments to use this parameter.

dagewa:
Hi @asmit3 @phyy-nx is this PR still relevant, or can it be closed?

asmit3:
I don't see myself addressing the changes requested any time soon. I will leave it to @phyy-nx to confirm whether this is relevant for future beamtimes and take action accordingly.

Anthchirp:
The changes in here don't merge cleanly any more. Feel free to reopen the pull request if you want to follow this up

phyy-nx:
I need to revisit batched ensemble refinement and re-integration more generally because we tried to use it live during an xfel experiment and it was too slow.  We needed more parallelism in the dials.integrate step.  It's a bigger problem than this pull request solves so it's ok to leave it closed for now, but let's keep the branch.  Thanks.

Issue #656, opened 2018-11-22
flatten_reflections should wrangle experiment identifiers gracefully


graeme-winter:
If we have data with identifier eggs and id 0, and another reflection set with identifier spam and id 0, would like it when we flatten these if eggs is id 0 (say) and spam _by magic_ becomes id 1. That way if you have two sets which you put together, you would avoid collisions - when you export these would be unrolled however you like. 

N.B. the numerical id could just as well be mapped to 1702 and 555 to get unique values. 

Currently if you flatten experiments you get a list of experiments, and flattening reflections gives a list with only one _unique_ id in - which makes something like

```
    assert len(experiments) == len(reflections)
```

fail

ndevenish:
I've been looking at this as part of trying to resolve some of the image viewer problems and I somewhat disagree with this approach.

- It's a function that used to literally just do `[x.data for x in reflections]` that now manipulates my data (`unwrap` would have been a better name for it before)
- It's name suggests `List[List[reflection_table]] -> List[reflection_table]` - In no universe do I expect a function called `flatten_reflections` to internally manipulate my data. For a `List[reflection table]` this is a little unclear - maybe, merge them? (but that'd be `merge_reflection_tables`...)
- At the very least this should be a separate, explicit function call in existing scripts. The fact that this is already called isn't good enough a reason to put it in here

If you want to merge tables at some point then this would need to be handled at the point of merge. Otherwise, isn't it only helping code that doesn't know it's using multiple tables, but is using multiple tables? Any code that wasn't broken already would have had to assume that the same ID on different tables could be a different experiment... I'm not sure who/what code this helps other than transforming for the sake of it.

Also, I'm not convinced that you can universally guarantee that number of identifiers in a table is exactly equal to number of experiments, in which case this transformation completely breaks code in a probably silent way. There's literally no way to validate this because the function doesn't have access to the experiment lists or experiments.

If this is was just an intermediate before deprecating the use of id to help stop things that used the id as a lookup index, then it'd be better to just transform all the ID's to > Nexperiments e.g. 100+. 

Basically, not sure what this actually helps by applying to *literally every single reflection table every imported*. It's a leaky abstraction that everywhere that might benefit from needs to be aware of what's going on anyway?

jbeilstenedmands:
[...]

Issue #655, opened 2018-11-22
Post-integration workflow requires assigning identifiers for >1 dataset


jbeilstenedmands:
I'm having a go at analysing some data and it seems that integrate > cosym > symmetry > scale or integrate > symmetry > scale doesn't work for more than one experiment, as the information about which reflections come from which experiment are getting lost when reindexed_reflections are output as a single table from cosym or symmetry.
I have only noticed this now, as usually I do dev.dials.assign_experiment_identifiers > cosym/symmetry. Also this is not an issue at the moment for xia2 where the ids are manually managed. 

I could update cosym and symmetry to call the experiment identifier setting code as a first step as scaling does, but am concerned that this would probably break other things in unexpected ways. Any thoughts?

graeme-winter:
Thoughts are that this sounds like a fairly serious FIXME - probably something when defining the behaviour in an abstract manner then implementing to that abstract specification would be the way forward, in preference to trying to "fix" the problem. 

Happy to help with this process.

jbeilstenedmands:
Actually for this dataset I can't even analyse with experiment identifiers set at the moment, as the identifier map isn't updated if datasets are removed from cosym, but that's a slightly separate issue I can work on now

graeme-winter:
I would think starting from a flow diagram of what we want then implementing to this could be easier? i.e. if we have 
![dials flow chart](https://user-images.githubusercontent.com/1241716/48911114-c80be680-ee69-11e8-9d6b-0fdd8732eb61.png)

for multiple sweeps, what should be happening?

PR #686, opened 2019-01-15
Handle duplicate table id values in flatten reflections (for #656)

jbeilstenedmands:
The change here aims to address some of the issues identified regarding experiment identifiers and id values in reflection tables (see #655 #656 etc).

To briefly recap, if two reflection tables have the 'id' column set to zero (the default), if they are joined into a common table then information is lost about their origin - this is important for multi-dataset work such as in dials.cosym and dials.scale, and many programs do some manual handling of table-id values to get around this.
By implementing these changes at the module level in flatten_reflections, we avoid having to do similar work in each individual program and can stop this issue in general.

I have run the full test suite and the only test which seems to have issues is test/command_line/test_combine_found_spots.py due to this line in the command line script: https://github.com/dials/dials/blob/f77bb95e3cf5e3bb32f0f7f994c1e2f86697514c/command_line/combine_found_spots.py#L200
I was wondering if @jmp1985 or @graeme-winter could comment on the reason behind this assertion and what the general use case of this script is - possibly the assertion to check that no reflections have an id of -1 (i.e. not strong)? If so it seems that it might be okay to update the assertion? 

Although the tests seem to pass I am aware that this behaviour could cause issues in programs that may expect certain table-id values when loading datasets, particularly in odd or unusual cases that may not be covered in tests. If anyone has any suggestions of additional things to check then this would be very useful.


jbeilstenedmands:
Just to be clear, this change does not, and is not intended to, address the issues regarding unique experiment_identifier values, which are unique strings for each experiment (this still needs to be implemented).
Here we are making sure that upon loading data, multiple reflection tables are given different values in the 'id' column, which also acts as a key in reflection_table.experiment_identifiers map (once unique string identifiers are implemented): reflection_table.experiment_identifiers()['id'] = 'unique_string_id'

jmp1985:
@jbeilstenedmands The reason for that assertion I believe is that this script was written with the intention of combining found spots from the same imageset. So I added a check to ensure that all the ids matched. The script itself is a couple of years old and probably not used much so if you update the behaviour and modify the assert I think that will not cause any harm.

jbeilstenedmands:
Okay, I have removed the assertion to reflect the change in behaviour, given that this isn't fundamental to the logic of the code here.

PR #699, opened 2019-01-25
Msgpack support

jmp1985:
Supersedes #535

### DIFF
--- a/util/phil.py
+++ b/util/phil.py
@@ -115,7 +115,10 @@ class ReflectionTableConverters(object):
     if s not in self.cache:
       if not exists(s):
         raise Sorry('File %s does not exist' % s)
-      self.cache[s] = FilenameDataWrapper(s, flex.reflection_table.from_pickle(s))
+      try:
+        self.cache[s] = FilenameDataWrapper(s, flex.reflection_table.from_msgpack_file(s))
+      except:
+        self.cache[s] = FilenameDataWrapper(s, flex.reflection_table.from_pickle(s))
     return self.cache[s]
 
   def from_words(self, words, master):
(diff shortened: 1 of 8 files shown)


graeme-winter:
In order to review this do I need to perform a clean bootstrap to get all the dependencies?

graeme-winter:
> In order to review this do I need to perform a clean bootstrap to get all the dependencies?

```
Grey-Area:cctbx graeme$ libtbx.install msgpack
Using /Users/graeme/svn/cctbx/modules as installation root
Attempting to obtain msgpack using http-tgz...484 kB
success
```

=> 👍

Anthchirp:
just gone through this pull request and cleaned up the commit history.
One question: this does not change the DIALS default behaviour, right? It only adds the capability to read/write .mpack files?

jmp1985:
@Anthchirp You are right, the default behaviour is to write out pickle files. For dials-2.0 I should modify this so that we output msgpack instead.

util/phil.py
```diff
@@ -115,7 +115,10 @@ def from_string(self, s):
     if s not in self.cache:
       if not exists(s):
         raise Sorry('File %s does not exist' % s)
-      self.cache[s] = FilenameDataWrapper(s, flex.reflection_table.from_pickle(s))
+      try:
+        self.cache[s] = FilenameDataWrapper(s, flex.reflection_table.from_msgpack_file(s))
+      except:
```
Anthchirp: Catch a more specific exception here, at a bare minimum make it ```except Exception```.
The difference between this and ```except Exception``` is that this does catch Ctrl+C and ```except Exception``` does not.

jmp1985: I have now fixed this

jmp1985:
@Anthchirp sure, merging into master should be no problem. Then I will make a new commit to dials-2.0 that makes msgpack the default behaviour

PR #703, opened 2019-01-30
Remove out-of-date Windows build instructions

Anthchirp:
any vetos?


graeme-winter:
Gets my vote - wrong instructions are worse than none

PR #708, opened 2019-02-18
Don't override n_macro_cycles

### LINKED ISSUE #707: dials.index real_space_grid_search not indexing all reflections it should
The attached test case, [example.zip](https://github.com/dials/dials/files/2861077/example.zip), comes from a simulated still image that I am able to index using the `real_space_grid_search` method. I am using this command:

```console
dials.index stills.indexer=sweeps datablock.json strong.pickle unit_cell=78.840,78.840,38.290,90.000,90.000,90.000 space_group=P43212 method=real_space_grid_search
```

which indexes most, but not all, reflections:
```console
------------------------------------
| Imageset | #indexed | #unindexed |
------------------------------------
| 0        | 157      | 60         |
------------------------------------
```
None of the found reflections are noise (the image is simulated), so it should be possible to index every reflection. In the reciprocal lattice viewer the unindexed reflections form a band at one side:

band

I can immediately index all the reflections by passing the output of the first job back into `dials.index`:

```console
dials.index experiments.json strong.pickle
...
------------------------------------
| Imageset | #indexed | #unindexed |
------------------------------------
| 0        | 217      | 0          |
------------------------------------
```
I'm trying to work out why I don't get this result the first time through.

rjgildea:
Don't override n_macro_cycles depending on presence/absence of a goniometer object.

Since stills_indexer doesn't do macro_cycles then simply ignore this
parameter rather than bombing out with an error message. Update help
message to make it clear that this parameter isn't used by still_indexer.
Fixes #707.

### DIFF
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -250,8 +250,11 @@ indexing {
       .help = "refine_shells: refine in increasing resolution cutoffs after indexing."
               "repredict_only: do not refine after indexing, just update spot"
               "predictions."
-    n_macro_cycles = Auto
+    n_macro_cycles = 5
       .type = int(value_min=1)
+      .help = "Maximum number of macro cycles of refinement, reindexing all"
+              "reflections using updated geometry at the beginning of each"
+              "cycle. Does not apply to stills.indexer=stills."
     d_min_step = Auto
       .type = float(value_min=0.0)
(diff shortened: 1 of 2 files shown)


dagewa:
Thanks, this makes sense to me and improves the encapsulation of behaviour. :+1:

PR #724, opened 2019-03-05
Define observer/subject pattern and use in scaling to make html report

jbeilstenedmands:
This pull request introduces Subject and Observer classes, based on the observer design pattern, defined in dials.util.observer.py. Creating a pull request to allow people who might have thoughts for general design improvement to offer feedback @ndevenish @Anthchirp .

A subject is an object that can register observers, to be notified in the case of an 'event' of interest. When notified, the observer can inspect the subject to update its state. This can facilitate a range of use cases. My present use case is to facilitate the generation of an html report in dials.scale. I have included the changes to dials.scale code as an example of how this functionality can be used. With this change, dials.scale now also outputs a scaling.html

Details : In my use case, the dials.scale Script class and Scaler class inherit from Subject (see the changes in dials.algorithms.scaler.py) and the scaling observers are only registered at the highest level (in ```if __name__ == "__main__":``` in dials.scale).
If the observers are not registered on the script, then nothing happens, so a different process using the dials.scale Script class is unaffected (e.g. dials.scale_and_filter).

These changes will also allow much of the scaling plotting code to be removed from dials.report and allow the removal of several dials.plot_scaling_* scripts that I have added over time, but I will enact that once merged.

Cheers,
James

### DIFF
--- a/algorithms/scaling/observers.py
+++ b/algorithms/scaling/observers.py
@@ -28,9 +28,6 @@ def register_default_scaling_observers(script):
     script.scaler.register_observer(
         event="performed_outlier_rejection", observer=ScalingOutlierObserver()
     )
-    script.scaler.register_observer(
-        event="performed_error_analysis", observer=ErrorModelObserver()
-    )
 
 @singleton
 class ScalingHTMLGenerator(Observer):
--- a/algorithms/scaling/observers.py
+++ b/algorithms/scaling/observers.py
@@ -1,0 +1,175 @@
+"""
+Observers for the scaling algorithm.
+"""
+from collections import OrderedDict
+from dials.util.observer import Observer, singleton
+from dials.algorithms.scaling.plots import (
+    plot_scaling_models,
+    statistics_tables,
+    plot_outliers,
+    normal_probability_plot,
+)
+from jinja2 import Environment, ChoiceLoader, PackageLoader
+
+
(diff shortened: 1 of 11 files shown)


rjgildea:
I suggest the following patch to util/observer.py:

```
$ git diff util/observer.py
diff --git a/util/observer.py b/util/observer.py
index aedd2a6ce..1d543fe43 100644
--- a/util/observer.py
+++ b/util/observer.py
@@ -4,6 +4,7 @@
 A singleton decorator is also defined.
 """
 
+import functools
 
 def singleton(cls):
     instances = {}
@@ -36,6 +37,7 @@ def notify_event(event):
         """Define a decorator for notifying of a specific event."""
 
         def wrap(method):
+            @functools.wraps(method)
             def notify(self, *args, **kwargs):
                 r = method(self, *args, **kwargs)
                 self.notify(event)
```

This ensures that the original name and docstring of the wrapped method is not lost:
https://docs.python.org/3/library/functools.html#functools.wraps

Output of `print(script.run)` with/without this patch:
```
print(script.run)
<bound method Script.notify of <__main__.Script object at 0x10856ab50>>
print(script.run)
<bound method Script.run of <__main__.Script object at 0x110699510>>
```

jbeilstenedmands:
Thanks @rjgildea, good idea, added this now.

Issue #604, opened 2018-08-03
dials.generate_masks doesn't work with multiple imagesets


dagewa:
Sometimes this is a reasonable thing to want to do:
> because of the limitation of the partial chi goniometer, I collect 360° as two sweeps of 180° around omega, with phi = 0 and phi = 180°.  I can dials.import and process them as one dataset with allow_multiple_sweeps=True.  One little problem is that I cannot simply generate a mask to cover all images, even though the beamstop shadow or unmasked noisy pixels are unchanged throughout the entire dataset.  The error message is "Sorry: datablock must contain exactly 1 imageset".

benjaminhwilliams:
Belatedly picking up on this.  Happy to take it on.  @dagewa, did you previously investigate?

benjaminhwilliams:
Oh sorry, I didn't look at the commit in any detail.  I see this fixes the exact problem you reported, but not the case of multiple imagesets belonging to different `Experiment`s, as in a multi-sweep small-molecule data collection.  I'm going to try to extend `dials.generate_mask`  to accommodate the latter case too.

benjaminhwilliams:
It seems some of the proposed changes in #728 may cause a stealthy regression here, there doesn't seem to be a test.  @dagewa, could you build a test around the original reported problem, or a similar case?  That way, we'll spot it before it happens again.

benjaminhwilliams:
It seems some of the proposed changes in #728 may cause a stealthy regression here, there doesn't seem to be a test.  @dagewa, could you build a test around the original reported problem, or a similar case?  That way, we'll spot it before it happens again.

benjaminhwilliams:
e1f7ef0 added a test for a four-sweep data set, but I don't have test data for the exact case (a split sweep) originally reported by @dagewa, so I can't build a test for that.  @dagewa, are you happy with the fix in e1f7ef0?  If not, or if you have data for your test case, feel free to reopen this issue if/when it is closed by the merging of #728.

dagewa:
Hi @benjaminhwilliams, sorry, didn't see this earlier. I don't think I have data for a test case at hand - the issue originally came from a user and I fixed it "blind"

benjaminhwilliams:
OK, I think the fix should work for your case too.  That's the case on master at least, I haven't double-checked the state of 1.14
  :-/

Issue #691, opened 2019-01-18
Plan out a transition to a functional interface


benjaminhwilliams:
- [x]  Decide on some example boilerplate
- [ ]  Assign responsibility for implementation to main authors of each DIALS command-line program.
- [ ]  Reformat `dials.import`.
- [x] Reformat `dials.generate_mask`.
- [ ]  Reformat `dials.find_spots`.
- [ ]  Reformat `dials.search_beam_position`.
- [x]  Reformat `dials.index`.
- [ ]  Reformat `dials.refine`.
- [ ]  Reformat `dials.two_theta_refine`.
- [ ]  Reformat `dials.refine_bravais_settings`.
- [ ]  Reformat `dials.create_profile_model`.
- [ ]  Reformat `dials.integrate`.
- [ ]  Reformat `dials.symmetry`.
- [x]  Reformat `dials.scale`.
- &c.

Assigning to JBE, MG, BHW.  Would assign to ND, but his leave and mine are neatly consecutive, meaning a five-week delay.

benjaminhwilliams:
Step 1 done.  How do we want to delegate refactoring existing programs?  Above, I've suggested original authors do it, though perhaps it makes sense to divide it between the three of us — thoughts?  Obvs, I'm not going to be able to contribute for three weeks.

benjaminhwilliams:
Agreed boilerplate is at https://github.com/dials/dials/wiki/Boilerplate-(or-the-standard-framework-of-a-DIALS-program).

PR #728, opened 2019-03-11
Convert dials.generate_mask to functional interface

benjaminhwilliams:
Following our discussion around #691, I thought I'd start a few bits of functional interface conversion, specifically the bits that affect i19.screen.  To start tamely, I've done `dials.generate_mask`.  I'd quite like to have it reviewed before I do any of the others, to check that I'm not doing anything controversial.

One particular question — I've converted some `Sorry`s to `ValueError`s, but should I be using `sys.exit()` in this context?

As far as possible, the changes follow boilerplate agreed in [the wiki](../wiki/Boilerplate-(or-the-standard-framework-of-a-DIALS-program)).

### DIFF
--- a/command_line/generate_mask.py
+++ b/command_line/generate_mask.py
@@ -35,15 +35,20 @@ from __future__ import absolute_import, division, print_function
 import sys
 import logging
 import six.moves.cPickle as pickle
+from typing import Tuple, Optional
 
 import dials.util
 import dials.util.log
+from scitbx.array_family import flex
 from iotbx.phil import parse
+from libtbx.phil import scope, scope_extract
 import libtbx.load_env
 from dials.util.options import OptionParser, flatten_experiments
 from dials.util.masking import MaskGenerator
 from dxtbx.format.image import ImageBool
-from dxtbx.model.experiment_list import ExperimentListDumper
+from dxtbx.model.experiment_list import Experiment, ExperimentList, ExperimentListDumper
+
+Masks = Tuple[flex.bool, ...]
 
 help_message = __doc__
 
@@ -76,6 +81,7 @@ phil_scope = parse(
 
 
 def generate_mask(experiments, params):
+    # type: (ExperimentList, scope_extract) -> Tuple[Masks, Optional[ExperimentList]]
     """
     Generate a pixel mask for each image in an experiment.
 
@@ -89,17 +95,13 @@ def generate_mask(experiments, params):
     with the masks applied will be saved to that location.
 
     Args:
-        experiments (:type:`dxtbx_model_ext.ExperimentList`): An experiment list
-            containing only one :type:`dxtbx_model_ext.Experiment` object.
-        params (:type:`libtbx.phil.scope_extract`): Masking parameters, having the
-            structure defined in :data:`dials.util.masking.phil_scope`.
+        experiments: An experiment list containing only one experiment.
+        params: Masking parameters, having the structure defined in :data:`phil_scope`.
 
     Returns:
-        Tuple[:rtype:`scitbx_array_family_flex_ext.bool`]: A tuple containing the
-            generated pixel masks.
-        :rtype:`dxtbx_model_ext.ExperimentList` (optional): A copy of
-            :param:`experiments` with the masks applied.  Only returned if
-            :attr:`params.output.experiments` is set.
+        A tuple containing the generated pixel masks.
+        A copy of :param:`experiments` with the masks applied
+            (optional, only returned if :attr:`params.output.experiments` is set).
     """
     imagesets = experiments.imagesets()
     if len(imagesets) > 1:
@@ -141,20 +143,21 @@ def generate_mask(experiments, params):
 
         return mask, experiments
     else:
-        return mask
+        return mask, None
 
 
 def run(phil=phil_scope, args=None):
+    # type: (scope, list) -> None
     """
     Parse command-line arguments, run the script.
 
     Use the DIALS option parser to extract an experiment list and parameters.  Pass
-    these to :func:`script`.  If :param:`args` is None (default), the option parser
+    these to :func:`script`.  If :param:`args` is `None` (default), the option parser
     defaults to :data:`sys.argv[1:]`.
 
     Args:
-        phil (:type:`libtbx.phil.scope`): PHIL scope for option parser.
-        args (:type:`list`): Arguments to parse.  Defaults to :data:`sys.argv[1:]`.
+        phil: PHIL scope for option parser.
+        args: Arguments to parse.  Defaults to :data:`sys.argv[1:]`.
     """
     # Create the parser
     usage = "usage: %s [options] experiments.json" % libtbx.env.dispatcher_name
--- a/command_line/generate_mask.py
+++ b/command_line/generate_mask.py
@@ -35,26 +35,27 @@ from __future__ import absolute_import, division, print_function
 import sys
 import logging
 import six.moves.cPickle as pickle
-from typing import Tuple, Optional
 
 import dials.util
 import dials.util.log
 from scitbx.array_family import flex
-from iotbx.phil import parse
-from libtbx.phil import scope, scope_extract
+import libtbx.phil as phil
 import libtbx.load_env
 from dials.util.options import OptionParser, flatten_experiments
 from dials.util.masking import MaskGenerator
 from dxtbx.format.image import ImageBool
-from dxtbx.model.experiment_list import Experiment, ExperimentList, ExperimentListDumper
+from dxtbx.model.experiment_list import ExperimentListDumper, ExperimentList
 
-Masks = Tuple[flex.bool, ...]
+try:
+    from typing import List, Optional, Tuple
 
-help_message = __doc__
+    Masks = Tuple[flex.bool, ...]
+except ImportError:
+    pass
 
 log = logging.getLogger('dials.generate_mask')
 
-phil_scope = parse(
+phil_scope = phil.parse(
     """
     output {
         mask = mask.pickle
@@ -81,7 +82,7 @@ phil_scope = parse(
 
 
 def generate_mask(experiments, params):
-    # type: (ExperimentList, scope_extract) -> Tuple[Masks, Optional[ExperimentList]]
+    # type: (ExperimentList, phil.scope_extract) -> Tuple[Masks, Optional[ExperimentList]]
     """
     Generate a pixel mask for each image in an experiment.
 
@@ -96,12 +97,14 @@ def generate_mask(experiments, params):
 
     Args:
         experiments: An experiment list containing only one experiment.
-        params: Masking parameters, having the structure defined in :data:`phil_scope`.
+        params: Masking parameters, having the structure defined in
+            :data:`phil_scope`.
 
     Returns:
         A tuple containing the generated pixel masks.
-        A copy of :param:`experiments` with the masks applied
-            (optional, only returned if :attr:`params.output.experiments` is set).
+
+        A copy of :param:`experiments` with the masks applied (optional,
+        only returned if :attr:`params.output.experiments` is set).
     """
     imagesets = experiments.imagesets()
     if len(imagesets) > 1:
@@ -147,23 +150,20 @@ def generate_mask(experiments, params):
 
 
 def run(phil=phil_scope, args=None):
-    # type: (scope, list) -> None
+    # type: (phil.scope, List[str]) -> None
     """
     Parse command-line arguments, run the script.
 
-    Use the DIALS option parser to extract an experiment list and parameters.  Pass
-    these to :func:`script`.  If :param:`args` is `None` (default), the option parser
-    defaults to :data:`sys.argv[1:]`.
+    Uses the DIALS option parser to extract an experiment list and
+    parameters, then passes these to :func:`generate_mask`.
 
     Args:
         phil: PHIL scope for option parser.
-        args: Arguments to parse.  Defaults to :data:`sys.argv[1:]`.
+        args: Arguments to parse. If None, :data:`sys.argv[1:]` will be used.
     """
     # Create the parser
     usage = "usage: %s [options] experiments.json" % libtbx.env.dispatcher_name
-    parser = OptionParser(
-        usage=usage, phil=phil, epilog=help_message, read_experiments=True
-    )
+    parser = OptionParser(usage=usage, phil=phil, epilog=__doc__, read_experiments=True)
 
     # Parse the command line arguments
     params, options = parser.parse_args(args=args, show_diff_phil=True)
--- a/command_line/generate_mask.py
+++ b/command_line/generate_mask.py
@@ -75,7 +75,7 @@ phil_scope = parse(
 )
 
 
-def script(experiments, params):
+def generate_mask(experiments, params):
     """
     Generate a pixel mask for each image in an experiment.
 
@@ -109,12 +109,12 @@ def script(experiments, params):
         beams = experiments.beams()
         for d in detectors[1:]:
             if not d.is_similar_to(detectors[0]):
-                raise ValueError(
+                sys.exit(
                     "Multiple imagesets are present, but their detector models differ."
                 )
         for b in beams[1:]:
             if not b.is_similar_to(beams[0]):
-                raise ValueError(
+                sys.exit(
                     "Multiple imagesets are present, but their beam models differ."
                 )
 
@@ -178,7 +178,7 @@ def run(phil=phil_scope, args=None):
         sys.exit("Exactly 1 experiment must be specified.")
 
     # Run the script
-    script(experiments, params)
+    generate_mask(experiments, params)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


jbeilstenedmands:
Regarding sys.exit() vs ValueError I think you have it correct at the moment. I believe sys.exit will not give a traceback, which is nice for catching simple user errors at the highest level, but maybe the tracebacks are more helpful deeper within the code to aide debugging.

benjaminhwilliams:
Indeed.  They were `Sorry`s before though, which suggested that suppressed traceback was the intention.  But I don't know if it was a good intention...  Happy to leave this to @Anthchirp's opinion.

command_line/generate_mask.py
```diff
@@ -32,108 +30,157 @@
...
+    """
+    imagesets = experiments.imagesets()
+    if len(imagesets) > 1:
+        # Check beams (for resolution) and detectors are equivalent in each case
+        # otherwise the mask may not be appropriate across all imagesets
+        detectors = experiments.detectors()
+        beams = experiments.beams()
+        for d in detectors[1:]:
+            if not d.is_similar_to(detectors[0]):
+                raise ValueError(
```
rjgildea: I would argue that the user shouldn't be exposed to a full traceback here.

command_line/generate_mask.py
```diff
@@ -32,108 +30,157 @@
...
+                    "Multiple imagesets are present, but their detector models differ."
+                )
+        for b in beams[1:]:
+            if not b.is_similar_to(beams[0]):
+                raise ValueError(
+                    "Multiple imagesets are present, but their beam models differ."
+                )
```
rjgildea: As above, I would argue that the user shouldn't be exposed to a full traceback here.

command_line/generate_mask.py
```diff
@@ -32,108 +30,157 @@
...
-        if params.output.experiments is not None:
-            for imageset in imagesets:
-                imageset.external_lookup.mask.data = ImageBool(mask)
-                imageset.external_lookup.mask.filename = params.output.mask
-            from dxtbx.model.experiment_list import ExperimentListDumper
-
-            print("Saving experiments to {0}".format(params.output.experiments))
-            dump = ExperimentListDumper(experiments)
-            dump.as_file(params.output.experiments)
+def script(experiments, params):
```
ndevenish: Did we specifically agree that the function should be called `script`? Surely something more descriptive like `generate_mask` would be a lot more intuitive e.g.
```python
from dials.command_line.generate_mask import generate_mask

# later...
generate_mask(exp, params)
"""

Comment (rjgildea) [reply command_line/generate_mask.py]: I agree - don't see the point in having a meaningless class/function name `script` proliferated across the source code.

File: command_line/generate_mask.py (hunk 1 above)
Comment (benjaminhwilliams) [reply command_line/generate_mask.py]: OK, I'll use sys.exit instead.

```suggestion
                sys.exit(
```

command_line/generate_mask.py
benjaminhwilliams: Yep, sorry, I was cargo culting.  I'll give it a proper descriptive name.

command_line/generate_mask.py
benjaminhwilliams: Agreed as above.

command_line/generate_mask.py
```diff
@@ -32,108 +30,157 @@
...
+            containing only one :type:`dxtbx_model_ext.Experiment` object.
+        params (:type:`libtbx.phil.scope_extract`): Masking parameters, having the
+            structure defined in :data:`dials.util.masking.phil_scope`.
+
+    Returns:
+        Tuple[:rtype:`scitbx_array_family_flex_ext.bool`]: A tuple containing the
+            generated pixel masks.
+        :rtype:`dxtbx_model_ext.ExperimentList` (optional): A copy of
+            :param:`experiments` with the masks applied.  Only returned if
+            :attr:`params.output.experiments` is set.
```
ndevenish: [**tl;dr** - type is probably wrong for return mask, also return type in general]

Couple of Docstring notes; not really criticism as points not raised before:
- I don't think you need to explicitly link the types as types in the type field. At least, it should know what these are types already? Even if it doesn't, it should just be a matter of parser intelligence. Made redundant by,
- Probably a bigger discussion but I've started moving even more to Python3 style type annotations. The main checking tool is [mypy](http://mypy-lang.org/), which has a standard for [python2](https://mypy.readthedocs.io/en/stable/cheat_sheet.html) _**But** also_ a way to automatically convert the python2-style annotations to real [PEP-484](https://www.python.org/dev/peps/pep-0484/) annotations once the codebase is ready for python3-only, so it's a bit more future-compatible.
- On this note, there are definitely plugins for sphinx/napoleon that will read the type annotations and extract them automatically so that you don't need to define them in the actual docstring text. This means that they only ever need to be written once, and in a static-analysis-verifiable way.

Specific to this code point:
- Assuming that this really does return a tuple of multiple masks rather than a tuple of a single mask, you probably mean `Tuple[scitbx.array_family.flex.bool, ...]` (and actual import path should be fine)
- The return of a type that could be multiple things is
  - In sphinx notation `A or B`
  - In PEP-484 notation: `Union[A, B]`

So, the return type is:
```
Tuple[scitbx.array_family.flex.bool,...] or Tuple[Tuple[scitbx.array_family.flex.bool,...], dxtbx.model.ExperimentList]
or
Union[Tuple[Tuple[scitbx.array_family.flex.bool,...], dxtbx.model.ExperimentList], Tuple[scitbx.array_family.flex.bool,...]]
```
And the full type-annotation type for the function, mypy2-style is:
```python
# type: (dxtbx.model.ExperimentList, libtbx.phil.scope_extract) -> Union[Tuple[Tuple[scitbx.array_family.flex.bool,...], dxtbx.model.ExperimentList], Tuple[scitbx.array_family.flex.bool,...]]
```
or

```python
# mypy python 2-style type annotation:
def generate_mask(
    experiments,  # type: dxtbx.model.ExperimentList
    params  # type: libtbx.phil.scope_extract
):
    # type: (...) -> Union[Tuple[Tuple[scitbx.array_family.flex.bool,...], dxtbx.model.ExperimentList], Tuple[scitbx.array_family.flex.bool,...]]
    """
    Generate a pixel mask for each image in an experiment.

    ...
    """
```
which becomes, in python 3:
```python
def generate_mask(
    experiments: dxtbx.model.ExperimentList, params: libtbx.phil.scope_extract
) -> Union[
    Tuple[Tuple[scitbx.array_family.flex.bool, ...], dxtbx.model.ExperimentList],
    Tuple[scitbx.array_family.flex.bool, ...],
]:
    """
    Generate a pixel mask for each image in an experiment.

    ...
    """
```

This looks... not good, but can generally be simplified by scoped imports and custom return types e.g.
```python
Masks = Tuple[scitbx.array_family.flex.bool, ...]

def generate_mask(
    experiments: ExperimentList, params: libtbx.phil.scope_extract
) -> Union[Masks, Tuple[Masks, ExperimentList]]:
    """Docstring"""
```

This can obviously be reduced further if the return type is simplified to something like always returning a tuple (which is easier to use also) e.g.`Tuple[Masks, Optional[ExperimentList]]`

ndevenish: The other point about not having lots of reST markup in the type field is that google/numpy style docstrings are specifically designed to be easy to read inline in code, which obviously the markup somewhat obfuscates

benjaminhwilliams: Thanks, I got a bit confused when reading the various different specifications, so that's very helpful.  I'll go through again and strip out some chaff.

The returned object is indeed, in general, a tuple of multiple mask objects.  As to always returning a tuple with an optional second value, you have a good point.  I think I tried to over-engineer it a bit because I once got caught out by https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.polyfit.html#numpy-polyfit

command_line/generate_mask.py
ndevenish: I know you'll hate me but I don't think sys.exit is the right thing either. It shows the message, but will also cause an unexpected exit if being used functionally.

Possible alternative: Raise `Sorry` which is an exception but also doesn't stacktrace, or throw something custom and catch it in the `run` function (don't want to hide any exceptions coming from underneath the `generate_mask` function).

command_line/generate_mask.py
ndevenish: Ah, maybe the numpy style that explicitly means multiple/tuple? From the linked document I can't tell if it's one or the other or a tuple of both or what.

The typing module (in 3, but backported) makes stuff nicer, because all these type annotations may look slightly verbose but are completely legal python expressions so statically verifiable, which is why you can do stuff like `Masks = Tuple[flex.bool,...]` and have it work.

ndevenish: But as an ambition, the actual markup the typing should make it unambiguous, and then whatever documentation generator can render it how it likes.

ndevenish:
from Traceback Terror to Exit Exceptionalism?

command_line/generate_mask.py
benjaminhwilliams: Hmm... do we want to specify `IO[str]` or similar for functions that write to disk?

benjaminhwilliams: As in https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#miscellaneous

ndevenish: Not sure if I've misunderstood the question, but that's still for taking parameters/returning values - just if it's a file-like object (e.g. something that has .write, .read, .seek, .flush maybe etc). So, for example, the `show` methods as discussed in https://github.com/cctbx/cctbx_project/issues/275 often take a `log` parameter that it writes to, and they would be of type `IO`. The PEP seems to specify it for *just* things that write to disk but people seem to use it more generally.

Now, if you want to actually encode the behaviour of a function's side effects into the type system, then you start getting into ML style languages and monads and what have you......

benjaminhwilliams:
If @ndevenish is happy with typing & docstrings, I have one question for @Anthchirp — did we arrive at a conclusion as to whether we should be using `libtbx.phil` or `iotbx.phil`?  I haven't delved.

command_line/generate_mask.py
benjaminhwilliams: Ah, yep.  I skim-read and didn't correctly comprehend.  Ignore the question!

rjgildea:
If you need to parse e.g. space groups or unit cells from phil then you’ll need to use iotbx.phil, otherwise libtbx.phil is fine.

ndevenish:
Made a few further tweaks to this in 059a88bd09a77b1a8abf94edfa81cbc30473e8eb.

- Moved phil functions back into a scoped import. Reasoning: It's reasonable to expect any coder looking at dials to know what an `ExperimentList` is, but less so `parse` or `scope` in some random context. Having it prefixed as `phil.parse` makes it obvious what is being called.
- The `typing` module is an optional backport in python2, so we can't necessarily assume it's there as I'm not sure what pulls it in. Protect this with a try block.
- Couple of type and docstring description tweaks. E.g. `list` implies an actual list object, while `List` technically is something very list-like (though the [docs](https://docs.python.org/3/library/typing.html#typing.List) ties them pretty closely). In any case, `List` seems to be preferred in these contexts. Also, lower-case looks weird in types 😜

ndevenish:
re: the sys.exist()/exception thing - also briefly mentioned with @benjaminhwilliams but not explored, technically this only accepts `ExperimentLists` with one experiment. It even checks in the `run` function.

I think this means that it can literally only ever have one imageset/detector/etc, meaning the checks can be removed? It might be worth adding in an assert to the functional function, or, perhaps, switching it to only take a single `Experiment`?

benjaminhwilliams:
OK, you're correct but this raises a broader point of multi-sweep data collections being effectively treated as something of a poor relation, which causes headaches when working with typical small molecule data.  I would prefer to have `generate_mask` ingest an `ExperimentList` and for the mask generation code to be re-written to accommodate multiple imagesets.  That's an issue (#604) outside the scope of this PR though.

ndevenish:
Okay so realised with @benjaminhwilliams that it's not multiple experiments per-se that are the problem but multiple imagesets, so in 25a8faff10026aa42a9df0774a1c7a0f04dbce49 changed docs to reflect that, and check explicitly with an assert. This means that the checks for multiple imageset features were never going to be hit, so removed. 

In the future any better solution will have to work round these problems but with different code anyway so 🤷‍♂️

benjaminhwilliams:
Hmm, @ndevenish I think 25a8faf breaks @dagewa's original fix to #604.  Multiple imagesets can share a single beam and detector, in which case the mask generated is the same for all imagesets.

Proposed solution, either:
 - Revert 25a8faf, merge PR, fix #604 properly later; or
 - Make merging this PR dependent on fixing #604 in this branch.

Since, as I've already mentioned, #604 is somewhat off topic, my preference is for the former.

ndevenish:
Okay, I think I see what has happened. My commit *does* break @dagewa's commit, but only if it wasn't already broken, which it was.

I was basing my logic on the "only one experiment" condition.

Except this condition was only written in https://github.com/dials/dials/commit/747e2410b272d6576ff87086893bd237f26de298#diff-081736f62f9fa5765da636acc43d54c7L83 a couple of months ago when @jmp1985 changed datablocks to experiment lists. But I think this was wrong, because whereas `ExperimentList` cannot contain more than one imageset, I believe `Datablock` could, so changing the check on len() == 1 wasn't correct.

So, yes, there needs to be some reversion but also some rewriting.

benjaminhwilliams:
Ah, good detective work.  OK, I'll do the legwork.

For the record, an `Experiment` can contain only one imageset, but an `ExperimentList` can contain more than one imageset because it can contain more than one `Experiment`.

ndevenish:
Ah yes, obviously, thanks.

command_line/generate_mask.py
```diff
@@ -85,55 +86,70 @@
...
     """
     imagesets = experiments.imagesets()
-    assert len(imagesets) == 1, "Only single imagesets supported"
-
-    imageset = imagesets[0]
+    masks = []
+
+    # Specify the formatting of the output filenames
+    num_imagesets = len(imagesets)
+    pad = len("{:d}".format(num_imagesets))
```
ndevenish: `pad = len(str(num_imagesets))` a bit clearer? I don't think the `:d` actually buys you anything here?

ndevenish: Also this is a more picky comment but I'm not sure that `num_imagesets = len(imagesets)` is saving much - only one less character for the extra indirection.

benjaminhwilliams: Good point, thanks.  I'll push a change when you're done reviewing.

command_line/generate_mask.py
```diff
@@ -85,55 +86,70 @@
...
+    # If there is more than one imageset, append a number to each output mask filename
+    filenames = [
+        "_{:0{:d}d}".format(i + 1, pad).join(os.path.splitext(params.output.mask))
+        if num_imagesets > 1
+        else params.output.mask
+        for i in range(num_imagesets)
+    ]
```
ndevenish: Had to do some mental contortions for this loop - both the `join(os.path.splitext(...` is a bit of an unusual way to do that (though, clever!). Also found the embedded ternary that only happens if there is only one item a little hard to follow.

Maybe something like:
```python
    # Create output mask filenames
    if num_imagesets == 1:
        filenames = [params.output.mask]
    else:
        # If there is more than one imageset, append a number to each output mask filename
        name, ext = os.path.splitext(params.output.mask)
        pad = len(str(num_imagesets))
        filenames = [
            "{name}_{num:0{pad}}{ext}".format(name=name, num=i + 1, pad=pad, ext=ext)
            for i in range(num_imagesets)
        ]
```
or, if you feel the format expansion doesn't need the naming:
```python
        filenames = [
            "{}_{:0{}}{}".format(name, i + 1, pad, ext) for i in range(num_imagesets)
        ]
```
There's possibly an argument that this could be moved off into a small internal function but I'm ambivalent about that in this case.

One other minor point: the `d` typing in format is redundant - if you are passing in a standard integer then it's the default, and it won't e.g. cause a stringified integer or float to convert. I guess there's a question of why it's in the standard if you never need it, but I'd imagine

1. printf-style has it (as printf *needs* the type in the string to correctly decode) and kept for sanity
2. You may want e.g. a custom class that accepts integer-style formatting but defaults to hex or something

As an aside, *knowing* about f-strings really makes me want to write them:
```python
filenames = [f"{name}_{i+1:0{pad}}{ext}" for i in range(num_imagesets)]
```

test/command_line/test_generate_mask.py
```diff
@@ -33,7 +49,6 @@ def test_generate_mask_with_untrusted_rectangle(input_filename):
     assert result["stderr"] == ""
     assert os.path.exists("mask2.pickle")
     assert os.path.exists("masked_experiments.json")
-    from dxtbx.serialize import load
```
ndevenish: Like +++

benjaminhwilliams:
I think this has probably been reviewed enough ;-p

Merging with prejudice.

PR #731, opened 2019-03-14
A first order fix to the shoebox parameters

graeme-winter:
I think this is a useful fix, please merge 🙂not found any counter examples and has good theoretical justification. 

Appreciate that this is not the complete solution.


graeme-winter:
Conflict in here is because I added a check for len(values) to master roughly at the same time for single-frame-sigma_b sums

PR #734, opened 2019-03-15
Generate a cosym report using new observer/subject pattern

rjgildea:


algorithms/symmetry/cosym/plots.py
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    "x": list(hist.slot_centers()),
+                    "y": list(hist.slots()),
+                    "type": "bar",
+                    "name": "Rij histogram",
+                }
+            ],
+            "layout": {
+                "title": "Distribution of values in the Rij matrix",
+                "xaxis": {"title": "$r_{ij}$"},
```
jbeilstenedmands:
```suggestion
                "xaxis": {"title": "r<sub>ij</sub>"},
```

algorithms/symmetry/cosym/observers.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+        with open(filename, "wb") as f:
+            f.write(html.encode("ascii", "xmlcharrefreplace"))
+
+
+@singleton
+class CosymClusterAnalysisObserver(Observer):
+
+    """
+    Observer to record scaling model data and make model plots.
```
jbeilstenedmands: I think this docstring needs updating, also the make_plots docstring in this class

algorithms/clustering/observers.py
```diff
@@ -0,0 +1,48 @@
...
+@singleton
+class UnitCellAnalysisObserver(Observer):
+
+    """
+    Observer to record unit cell clustering data and make plots.
+    """
+
+    def update(self, script):
+        """Update the data in the observer."""
+        self.data["dendrogram"] = script.unit_cell_dendrogram
```
jbeilstenedmands: I'm not sure whether we agreed if cosym should also work on a single dataset. Running on a single dataset crashes at this point with
```AttributeError: 'cosym' object has no attribute 'unit_cell_dendrogram'```
Might be worth adding a try ..., except AttributeError here?

PR #738, opened 2019-03-22
Enable travis builds for 1.14 release candidates

Anthchirp:
should be easy. only need to build on top of the Diamond 1.14 RC builds rather than the 2.0 nightlies.


Anthchirp:
looks alright. Leaving branch until builds have completed.

Issue #714, opened 2019-02-26
DIALS 2 documentation changes


Anthchirp:
Currently we build the documentation from the installer, which takes a considerable amount of time and means we can only update the documentation once a day, unless we also build a new installer.
Building the documentation also depends on DIALS itself, which means it is basically impossible to build it using online tools.

* Move the output files created by ```dev.dials.generate_tutorial_text``` into a separate repository
* See if we can use [ReadTheDocs](https://readthedocs.org/) to build the documentation rather than Jenkins
* Drop explicit documentation dependencies from ```libtbx_config```/```libtbx_refresh.py```

ndevenish:
It'd be nice to use the RTD versioned documentation to have a stable and "dev" version of the documentation. Not that there is a lot of difference right now, but would make it easier in the future.

Anthchirp:
* check if we can move screenshots there, too. Would prefer not having those in the code repository.

Anthchirp:
Discussed with @ndevenish. Outcome so far is to move the entire ```/doc```* tree plus the generated DIALS output to a separate repository which will (or: can) be added to the dials repository as a submodule. The generated DIALS output is maintained/updated by Jenkins, and the submodule link can also be maintained/updated by Jenkins whenever the actual documentation changes.

Then build it and host it on RTD.


* ND prefers writing it as ```doc/```, but we mean the same thing.

Anthchirp:
This is all nice and feasible until you hit the stumbling block of libtbx.sphinx plugins.
There are 3 of them:
#### libtbx.sphinx.pubmed
Reference lookup mechanism. Looks easily extractable into the documentation repository itself, so no problem.
#### libtbx.sphinx.python_string
Imports a file and then extracts a string variable from it. Could be extracted into the documentation repository and the libtbx import mechanism could be replaced by ast parsing which does not require files to be importable, so again not a big problem
#### libtbx.sphinx.phil
Adds dependencies on libtbx and iotbx, requires files to be importable which means it adds dependencies on everything, and this is a big problem.

I can only see three ways out of this:
1. dropping phil and finding another mechanism to do configuration and generate documentation - not going to happen.
2. making DIALS a real installable package - not going to happen for a long time
3. preprocessing the documentation from within a DIALS environment to the replace the phil placeholders with the real information - not sure how to do this and wondering whether that makes the entire process too complex.

It may be easier to come up with our own versioning mechanism on dials.github.io

Anthchirp:
[...]

PR #740, opened 2019-03-24
Move documentation from /doc/ to separate repository

Anthchirp:
Move everything related to documentation to https://github.com/dials/docs
See #714 for background.


Anthchirp:
removing /doc yes. submodule no.

PR #743, opened 2019-04-02
Make msgpack the default output format

jmp1985:
This pull request implements the necessary changes to make msgpack the default output format for all the dials programs. It contains the following changes:

1. Changes from flex.reflection_table.as_pickle to flex.reflection_table.as_msgpack_file throughout.
2. Changes in phil parameters to make default file extension *.mpack rather than *.pickle
3. Added serialization of experiment identifiers to msgpack adapter.
4. Updated all tests

The is a breaking interface change, so before this can be merged changes also need to be made in xia2 and probably other places as well.


graeme-winter:
Please can @rjgildea and @benjaminhwilliams also have a look at this - I can see other tools beyond xia2 being affected by this... good to see though thank you 🙂

jmp1985:
@graeme-winter no problem, I've added them both as reviewers! 😁

phyy-nx:
Hi, I haven't reviewed this yet, but I know it will need a corresponding
set of changes in cctbx/xfel.  We have a lot of implicit searches for the
.pickle extension there.

jmp1985:
Hi @phyy-nx I'm happy to go through cctbx/xfel and make a pull request if that is helpful?

phyy-nx:
That would be very helpful!  I'm particularly concerned about xtc_process,
cspad.cbf_metrology, detector_residuals and plot_run_stats_from_experiments.

jmp1985:
I've now made these changes in cctbx xfel (see https://github.com/cctbx/cctbx_project/pull/317)

array_family/flex.py
```diff
@@ -262,6 +262,18 @@ def from_h5(filename):
...
+    
+    @staticmethod
+    def from_file(filename):
+        """
+        Read the reflection table from either pickle or msgpack
+
+        """
+        try:
+            return reflection_table.from_msgpack_file(filename)
+        except Exception:
```
Anthchirp: Can this exception be more specific?

jmp1985: Changed to use RuntimeError

phyy-nx:
Hi, is there any reason not to always use from_file instead of from_msgpack_file?  Everywhere we can preserve backwards compatibility I think we should.

Anthchirp:
This is not desirable for at least 3 reasons. `from_file` hides RuntimeErrors when reading message pack files making debugging this more complex, it retains all the issues with .pickle files, ie. it will still execute any file that is not a msgpack file, and as far as I am aware retaining .pickle support is part of the reason why we have so many import dependencies which are undesirable in themselves.

I support keeping pickle support if we can agree on a schedule to phase it out, say
* support it for DIALS 2.0 (~May 2019)
* generate DeprecationWarnings in DIALS 2.1 (~August 2019)
* remove the code for DIALS 2.2 (~October 2019)

Anthchirp:
resolved conflicts

phyy-nx:
Backwards compatibility is important, and perhaps more important than the issues you raise.  Here's an example.

In Brewster 2018 I published a wiki and a repo with Jupyter notebooks that reproduces all of my figures.  This work requires extensive use of reflection table pickles.  See:
https://github.com/phyy-nx/dials_refinement_brewster2018/search?q=pickle&unscoped_q=pickle
See especially figure 7.

Not only that, but I deposited both the raw images and the processing results, including all the json and pickles, here:
http://www.cxidb.org/id-81.html
 
Invalidating both the documentation for the publication and the deposition by removing pickle backwards compatibility is something that needs to be carefully considered.

Granted, I recognize that software needs to move on and that we can't be expected held back by old data, but I was hoping to be able to point to that publication for years.  

Of course, this isn't the only example.  We have pickle files going back to 2013 in archive folders at LCLS that we are still referring to.

Anthchirp:
In understand your concerns. The appropriate software to work on .pickle files created in June 2018 is most likely the DIALS 1.10 release, which is and will continue to be available at https://github.com/dials/dials/releases and will read those files forever.

The scenario you describe is indeed a reason why we make releases and why DIALS release versions have specific version numbers that you can point people to, which we report eg. in xia2 output, which will give them the exact software they can use to recreate the output from the same input. DIALS 2.0 will no doubt not give you the same results that DIALS 1.10 did.

phyy-nx:
Right, it's a good point.  And if I could, I would have pointed users to a dials release as part of the publication.  But there is no dials release that can read xtc streams (yet).  That's why I pointed here instead: 
https://exafel.github.io/docs/psana-cctbx-install
And that documentation pulls the live sources.

Likely the answer is I need to change the documentation and point to a custom, frozen build I make myself.

Issue #752, opened 2019-04-10
Suboptimal spot-finding parameters can negatively impact integration


rjgildea:
In some cases integration results appear to be very sensitive to spot-finding parameters.

An example case is for a germanate dataset from DLS beamline I23.

Raw image:
![image](https://user-images.githubusercontent.com/2810624/55895566-5207e200-5bb4-11e9-9754-b67df09ff3b2.png)

Threshold map with default spot-finding parameters:
![image](https://user-images.githubusercontent.com/2810624/55895624-6b109300-5bb4-11e9-802b-426b57709edc.png)

Threshold map with kernel_size=20,20 sigma_b=10 sigma_s=5:
![image](https://user-images.githubusercontent.com/2810624/55895675-88456180-5bb4-11e9-9e2a-adb09780abb4.png)

1) Process with default parameters:
```
$ xia2 anomalous=True read_all_image_headers=False image=/dls/i23/data/2019/nr23017-1/20190302/germ_9keV/data_1_0001.cbf
```
2) Process with larger kernel size for spotfinding:
```
$ cat find_spots.phil 
spotfinder {
  filter.max_spot_size=10000
  threshold {
    dispersion {
      kernel_size = 20 20
      sigma_background = 10
      sigma_strong = 5
    }
  }
}
$ xia2 anomalous=True read_all_image_headers=False dials.find_spots.phil_file=find_spots.phil image=/dls/i23/data/2019/nr23017-1/20190302/germ_9keV/data_1_0001.cbf
```

Comparison of the results show markedly improved I/sigI and merging R-factors for the larger kernel size in spotfinding:
![i_over_sigma_mean](https://user-images.githubusercontent.com/2810624/55895737-af9c2e80-5bb4-11e9-8bb3-24de618fed04.png)
![r_pim](https://user-images.githubusercontent.com/2810624/55895751-b3c84c00-5bb4-11e9-804e-3564f4c4ab6c.png)

Profile model determined using default spot-finding parameters:
sigma b: 0.042689 degrees
sigma m: 0.076268 degrees

Profile model determined using modified spot-finding parameters:
Modified spot-finding parameters:
sigma b: 0.062064 degrees
sigma m: 0.107637 degrees

rjgildea:
With the larger kernel size in spotfinding, the <Ibg> reported by dials.integrate is lower, suggesting that if the kernel size is too small, then signal is leaking into the background:
[...]

PR #758, opened 2019-04-17
Improve spotfinding

jmp1985:
This pull request implements an extended dispersion threshold algorithm for spot finding to try and address https://github.com/dials/dials/issues/752. 

The algorithm does the following steps:

1. Compute the dispersion image
2. Compute the dispersion mask as normal
3. Erode the dispersion mask by half the kernel size
4. Expand the kernel slightly
5. Compute the mean pixel value excluding the dispersion mask
6. Compute the strong mask as normal

The rationale for the algorithm is that the dispersion mask is computed by testing the null hypothesis that all the pixels in the local area are drawn from a Poisson distribution. If this hypothesis is false then the pixel is marked as True in the dispersion mask. 

If a strong pixel is at the edge of the local area then the pixel at the centre of the box is marked True.  This means that the dispersion image tends to have the strong pixels and a band of pixels 1/2 the kernel size around it. Therefore, for true strong spots we can safely erode the dispersion mask by 1/2 the kernel size. This results in a mask which pretty much fits the the spot exactly. 

Finally, we want to only select the strong pixels (sometimes the erodes mask still incorporates weak pixels); however, the problem we had before is that we were using the mean calculated including the strong pixels. This was inevitable if we wanted to do the whole thing with a single summed area table. However, we can recompute the mean excluding the dispersion masked pixels with a slightly expanded kernel to ensure we have plenty of pixels to choose from. This gives actually a pretty good estimate of the background from which we can then threshold the strong pixels as done previously. 

The end result is that we get much better segmentation of the strong spots.  However, the algorithms is a bit slower (currently a lot slower since I have only implemented the debug version - once we are happy that the algorithm is an improvement I will implement a more optimised version). Once optimized I think it will still require more passes of the image data:

1. One pass to compute the initial summed area table.
2. One pass to compute the dispersion mask.
3. Two passes to compute the distance metric for the erosion
4. One pass to compute the eroded dispersion mask
5. One pass to compute the second summed area table
6. One pass to compute the final mask

However, if slower still means better then this may be acceptable.

Here are some examples of some output:
 
The raw data from an I23 image
data_1_0001_image_close

The threshold from the same image
data_1_0001_threshold_close

The spot finding results with strong pixels marked
data_1_0001_pixels

I have also modified the largest spot size to be 1000 pixels (10x10x10) since this will tend to find larger spots. 

Another nice side effect is that the could now have pretty decent background estimates from the spot finding with little extra cost which might also make the change desirable.

Next steps are to test thoroughly, particular on the data that @rjgildea has been looking at.  Then once we are satisfied we need to optimize the code!

### DIFF
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -421,7 +421,7 @@ class SpotDensityFilter(object):
             pyplot.xlim((0, pyplot.xlim()[1]))
             pyplot.ylim((0, pyplot.ylim()[1]))
             pyplot.gca().invert_yaxis()
-            cbar1 = pyplot.colorbar(plot1)
+            pyplot.colorbar(plot1)
             pyplot.axes().set_aspect("equal")
             pyplot.show()
 
@@ -442,7 +442,7 @@ class SpotDensityFilter(object):
             pyplot.xlim((0, pyplot.xlim()[1]))
             pyplot.ylim((0, pyplot.ylim()[1]))
             pyplot.gca().invert_yaxis()
-            cbar1 = pyplot.colorbar(plot1)
+            pyplot.colorbar(plot1)
             pyplot.axes().set_aspect("equal")
             pyplot.show()
 
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -61,7 +61,7 @@ def generate_phil_scope():
                 "to be accepted by the filtering algorithm."
         .type = int(value_min=1)
 
-      max_spot_size = 100
+      max_spot_size = 1000
         .help = "The maximum number of contiguous pixels for a spot"
                 "to be accepted by the filtering algorithm."
         .type = int(value_min=1, allow_none=False)
--- a/algorithms/image/fill_holes/simple.h
+++ b/algorithms/image/fill_holes/simple.h
@@ -62,7 +62,8 @@ namespace dials { namespace algorithms {
     std::size_t width = data.accessor()[1];
 
     // Compute the manhattan distance transform of the mask
-    af::versa< int, af::c_grid<2> > distance = manhattan_distance(mask);
+    af::versa< int, af::c_grid<2> > distance(mask.accessor());
+    manhattan_distance(mask, true, distance.ref());
 
     // Get a list of the pixels to fill
     std::vector<detail::SimpleFillNode> pixels;
--- a/algorithms/image/filter/boost_python/distance.cc
+++ b/algorithms/image/filter/boost_python/distance.cc
@@ -16,11 +16,32 @@ namespace dials { namespace algorithms { namespace boost_python {
 
   using namespace boost::python;
 
+  af::versa< int, af::c_grid<2> > manhattan_distance_wrapper(
+      const af::const_ref<bool, af::c_grid<2> > &src,
+      bool value) {
+    af::versa< int, af::c_grid<2> > dst(src.accessor());
+    manhattan_distance(src, value, dst.ref());
+    return dst;
+  }
+  
+  af::versa< int, af::c_grid<2> > chebyshev_distance_wrapper(
+      const af::const_ref<bool, af::c_grid<2> > &src,
+      bool value) {
+    af::versa< int, af::c_grid<2> > dst(src.accessor());
+    chebyshev_distance(src, value, dst.ref());
+    return dst;
+  }
+
   void export_distance()
   {
     def("manhattan_distance",
-      &manhattan_distance, (
-        arg("data")));
+      &manhattan_distance_wrapper, (
+        arg("data"),
+        arg("value")));
+    def("chebyshev_distance",
+      &chebyshev_distance_wrapper, (
+        arg("data"),
+        arg("value")));
   }
 
 }}} // namespace = dials::algorithms::boost_python
--- a/algorithms/image/filter/boost_python/mean_and_variance.cc
+++ b/algorithms/image/filter/boost_python/mean_and_variance.cc
@@ -74,7 +74,8 @@ namespace dials { namespace algorithms { namespace boost_python {
       arg("image"),
       arg("mask"),
       arg("size"),
-      arg("min_count")));
+      arg("min_count"),
+      arg("ignore_masked") = true));
 
     def("mean_and_variance_filter",
       &make_mean_and_variance_filter<FloatType>, (
--- a/algorithms/image/filter/distance.h
+++ b/algorithms/image/filter/distance.h
@@ -12,6 +12,7 @@
 #define DIALS_ALGORITHMS_IMAGE_FILTER_DISTANCE_H
 
 #include <dials/array_family/scitbx_shared_and_versa.h>
+#include <dials/error.h>
 
 namespace dials { namespace algorithms {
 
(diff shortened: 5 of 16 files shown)


jmp1985:
One thing to note is that I found that the kernel size of 3x3 actually performed well in the majority of cases; however, one CCD dataset I had with really massive spots did better with a kernel of 5x5. I don't think it should require a kernel any larger than that.

ndevenish:
> However, if slower still means better then this may be acceptable.

It's my understanding that we've explicitly rejected improving the spotfinding in the past because speed was a much higher priority. I'm assuming this will sit as an alternate method, but do you have any quantification of what the actual improvements are?

rjgildea:
Anecdotally the spot-finding results look a lot better, e.g. kernel_size=3,3 for image 144 of the i23 germanate dataset:
Before:
image
 After:
image
Comparison of merging statistics with the old algorithm (b=before) and the new algorithm (a=after):
i_over_sigma_mean
r_pim
chart
Possibly still some improvements to be made by increasing the kernel size slightly, but the changes are a lot smaller than before.

rjgildea:
For i03_thaum_eiger dataset there seems a lot less dependence of integration results on kernel size :
i_over_sigma_mean
r_pim
chart-1

rjgildea:
Similarly for I04_Thaum_eiger:
i_over_sigma_mean
r_pim
chart-2

rjgildea:
I04 thaumatin pilatus (i04_bag_training):
i_over_sigma_mean
r_pim
chart-3

rjgildea:
VMXi thaumatin:
i_over_sigma_mean
r_pim
chart-4

jmp1985:
@rjgildea Thanks for doing these tests. The results look positive.

jmp1985:
I've now added a faster implementation and done a crude benchmark using the bag training data. 

The old spot finding algorithm is obviously faster but extracts fewer spots, mainly due to filtering out small spots < 3 pixels. Since the spots extracted from the new algorithm are generally larger, this is less of a problem. However, 4 spots were filtered due to having more than 1000 pixels. The new algorithm takes approximately 1.6 times as long to run but as can be seen above, gives much better results

**Old**
Saved 116082 reflections to strong.mpack
Time Taken: 122.089928

**New**
Saved 126782 reflections to strong.mpack
Time Taken: 198.506274

jmp1985:
The average of ten runs:

**Old:**
Time taken: 124.1504409 seconds

**New**
Time taken: 175.5127746 seconds

graeme-winter:
Fails for me with Eiger data in xia2:

```
Grey-Area with :) $ cat DEFAULT/NATIVE/SWEEP1/index/2_dials.find_spots.log 
DIALS 2.dev.258-g67c4870f5

output {
  reflections = "2_SWEEP1_strong.pickle"
  experiments = "2_SWEEP1_experiments.json"
}
spotfinder {
  write_hot_mask = True
  hot_mask_prefix = "2_hot_mask"
  scan_range = 1 488
  filter {
    min_spot_size = 3
  }
  mp {
    nproc = 8
  }
  threshold {
    dispersion {
      min_local = 0
    }
  }
}
input {
  experiments = /Users/graeme/Projects/chef-dI/improved-spotfinding/with/DEFAULT/NATIVE/SWEEP1/index/1_SWEEP1_experiments.json
}

Configuring spot finder from input parameters
Finding strong spots in imageset 0


Finding spots in image 1 to 488...
Setting chunksize=61
Extracting strong pixels from images
 Using multiprocessing with 8 parallel job(s)



easy_mp crash detected; subprocess trace: ----
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/job_scheduler.py", line 64, in job_cycle
    value = target( *args, **kwargs )
  File "/Users/graeme/svn/cctbx/modules/dials/util/mp.py", line 191, in __call__
    result.append(self.func(i))
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 285, in __call__
    result = self.function(task)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 147, in __call__
    threshold_mask = self.threshold_function.compute_threshold(im, mk)
  File "/Users/graeme/svn/cctbx/modules/dials/extensions/dispersion_extended_spotfinder_threshold_ext.py", line 66, in compute_threshold
    return self._algorithm(image, mask)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/threshold.py", line 194, in __call__
    algorithm = threshold.DispersionExtendedThreshold(
'module' object has no attribute 'DispersionExtendedThreshold'
----------------------------------------------


  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 224, in <module>
    halraiser(e)
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 222, in <module>
    script.run()
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 153, in run
    reflections = flex.reflection_table.from_observations(experiments, params)
  File "/Users/graeme/svn/cctbx/modules/dials/array_family/flex.py", line 208, in from_observations
    return find_spots(experiments)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 770, in __call__
    table, hot_mask = self._find_spots_in_imageset(imageset)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 867, in _find_spots_in_imageset
    r, h = extract_spots(imageset[j0:j1])
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 478, in __call__
    return self._find_spots(imageset)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 582, in _find_spots
    callback=process_output,
  File "/Users/graeme/svn/cctbx/modules/dials/util/mp.py", line 272, in batch_multi_node_parallel_map
    preserve_exception_message=True,
  File "/Users/graeme/svn/cctbx/modules/dials/util/mp.py", line 172, in multi_node_parallel_map
    preserve_exception_message=preserve_exception_message,
  File "/Users/graeme/svn/cctbx/modules/dials/util/mp.py", line 45, in parallel_map
    preserve_exception_message=preserve_exception_message,
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/easy_mp.py", line 628, in parallel_map
    result = res()
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/result.py", line 119, in __call__
    self.traceback( exception = self.exception() )
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/stacktrace.py", line 117, in __call__
    self.raise_handler( exception = exception )
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/stacktrace.py", line 134, in raise_with_traceback
    raise_(type(exception), exception, self.traceback)
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/mainthread.py", line 100, in poll
    value = target( *args, **kwargs )
  File "/Users/graeme/svn/cctbx/modules/dials/util/mp.py", line 88, in __call__
    preserve_exception_message=self.preserve_exception_message,
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/easy_mp.py", line 628, in parallel_map
    result = res()
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/result.py", line 119, in __call__
    self.traceback( exception = self.exception() )
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/stacktrace.py", line 117, in __call__
    self.raise_handler( exception = exception )
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/stacktrace.py", line 147, in raise_with_stacktrace
    self.stacktrace( exception = exception )
  File "/Users/graeme/svn/cctbx/modules/cctbx_project/libtbx/scheduling/stacktrace.py", line 88, in __call__
    raise exception
AttributeError: Please report this error to dials-support@lists.sourceforge.net: 'module' object has no attribute 'DispersionExtendedThreshold'
```

graeme-winter:
@rjgildea asks "did you type Make"

No, I did not

rjgildea:
FYI: for testing both algorithms:

```$ dials.find_spots -c -e2 | grep algorithm
    algorithm = dispersion *dispersion_extended helen single```

**Comment by graeme-winter** [comment]
On understanding that we will add a background estimate in the future, agree is a good idea.

File: algorithms/spot_finding/factory.py [hunk 1]
```diff
@@ -61,7 +61,7 @@ def generate_phil_scope():
                 "to be accepted by the filtering algorithm."
         .type = int(value_min=1)
 
-      max_spot_size = 100
+      max_spot_size = 1000
```
ndevenish: Slightly worried that this might have a negative effect on the standard spotfinding?

rjgildea: max_spot_size=100 is overly conservative, especially in the context of smaller eiger pixels (remember this is the number of spot pixels in 3 dimensions).

ndevenish: And I suppose rings picked up will either be noise or eliminated by the other filters.

ndevenish:
Do we really want to move to this as default immediately?

Otherwise, I see no reason why not to merge this, it mostly just inserts new things without affecting the existing structure.

graeme-winter:
@ndevenish makes spot finding slower but better - that said I am happy to see this merged (size change & all)

re: sizes - could have 100 for 2D spots, 1000 for 3D?

PR #760, opened 2019-04-19
Kapton correction for single/multi-panel detectors

asmit3:
During our beamtime at SACLA in Feb, 2019, where the drop on tape (DOT) sample delivery system was used, it was realized that a new tool is needed to correct for the kapton tape shadow that affects integrated intensities. The existing kapton tool in DIALS works only for single panel detectors. Here we present a new implementation of the kapton correction making use of dxtbx objects. The kapton tape itself is represented by 6 dxtbx detector objects (to form a volume) and we have made use of various detector functions to get the path length of the x-rays through the tape. Furthermore, no assumptions regarding single/multi-panel detectors are necessary because the mathematics has been coded up in a general way.
The code reproduces  older kapton results on the Rayonix at LCLS using the same set of parameters. More importantly, it now allows us to use it on multi-panel detectors like the Octal-MPCCD.

### DIFF
--- a/algorithms/integration/SConscript
+++ b/algorithms/integration/SConscript
@@ -35,3 +35,8 @@ env.SharedLibrary(
     source=["bayes/boost_python/ext.cc"],
     LIBS=env["LIBS"],
 )
+env.SharedLibrary(
+    target="#/lib/dials_algorithms_integration_kapton_ext",
+    source=["boost_python/kapton_ext.cc"],
+    LIBS=env["LIBS"],
+)
--- a/algorithms/integration/kapton_2019_correction.py
+++ b/algorithms/integration/kapton_2019_correction.py
@@ -1,0 +1,648 @@
+from __future__ import division, print_function, absolute_import
+
+import math
+
+from dials.algorithms.shoebox import MaskCode
+from dials.array_family import flex
+from scitbx.matrix import col
+from dials.algorithms.integration.kapton_correction import get_absorption_correction
+import logging
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+
+
+class KaptonTape_2019(object):
+    """Class for defining Kapton tape using dxtbx models and finding the path through the tape traversed by s1 vector"""
+
+    def __init__(
+        self,
+        height_mm,
+        thickness_mm,
+        half_width_mm,
+        rotation_angle_deg,
+        wavelength_ang=None,
+    ):
+        from scitbx.matrix import col
+
+        self.height_mm = h = height_mm  # plugin controlled
+        self.thickness_mm = t = thickness_mm  # plugin controlled
+        self.half_width_mm = w = half_width_mm  # plugin controlled
+        self.tape_depth_mm = a = half_width_mm * 20.0  # Big number in mm
+        self.angle_rad = rotation_angle_deg * math.pi / 180.0  # plugin controlled
+        self.wavelength_ang = wavelength_ang
+        self.num_pixels = 5000  # number of pixels to put in fictitious kapton faces
+        # Now set up the kapton physical model using dxtbx detector objects
+        #
+        # determine absorption coeff (mm-1) through kapton for a given X-ray energy
+        G = get_absorption_correction()
+        attenuation_length_mm = G(self.wavelength_ang)
+        self.abs_coeff = 1 / attenuation_length_mm
+        #
+        def create_kapton_face(ori, fast, slow, image_size, pixel_size, name):
+            """ Create a face of the kapton as a dxtbx detector object """
+            from dxtbx.model import Detector
+
+            d = Detector()
+            p = d.add_panel()
+            p.set_local_frame(fast.elems, slow.elems, ori.elems)
+            p.set_pixel_size((pixel_size, pixel_size))
+            p.set_image_size(image_size)
+            p.set_trusted_range((-1, 2e6))
+            p.set_name("KAPTON_%s" % name)
+            return d
+
+        # Set up the bounding box of the kapton
+        rA = col((t / 2, a / 2, -w))
+        rB = col((t / 2, a / 2, w))
+        rC = col((t / 2, -a / 2, w))
+        rD = col((t / 2, -a / 2, -w))
+        rE = col((-t / 2, a / 2, -w))
+        rF = col((-t / 2, a / 2, w))
+        rG = col((-t / 2, -a / 2, w))
+        rH = col((-t / 2, -a / 2, -w))
+        # Now add a rotation
+        # rotation about z-axis
+        rot_axis = col((0, 0, 1))
+        rot_mat = rot_axis.axis_and_angle_as_r3_rotation_matrix(
+            self.angle_rad, deg=False
+        )
+
+        rA = rot_mat * rA
+        rB = rot_mat * rB
+        rC = rot_mat * rC
+        rD = rot_mat * rD
+        rE = rot_mat * rE
+        rF = rot_mat * rF
+        rG = rot_mat * rG
+        rH = rot_mat * rH
+        # Now add an offset to all the points in the direction normal to the ABCD plane
+        rAB = rA - rB
+        rCB = rC - rB
+        # Get normal in the direction of the drop
+        drop_normal = rCB.cross(rAB).normalize()
+        offset = -(h + t / 2) * drop_normal
+        rA = rA + offset
+        rB = rB + offset
+        rC = rC + offset
+        rD = rD + offset
+        rE = rE + offset
+        rF = rF + offset
+        rG = rG + offset
+        rH = rH + offset
+        # Store these edge points if needed later
+        self.edge_points = [rA, rB, rC, rD, rE, rF, rG, rH]
+        # Now set up the 6 faces
+        faces = []
+        px_off = 0
+        fast = (rA - rB).normalize()
+        slow = (rC - rB).normalize()
+        pixel_size = self.half_width_mm * 2.0 / self.num_pixels
+        image_size = (
+            self.num_pixels + px_off,
+            int(self.tape_depth_mm / pixel_size) + px_off,
+        )
+        # Face YZ0
+        ori = rB
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "yz0"))
+        # face YZ1
+        ori = rF
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "yz1"))
+        #
+        fast = (rB - rF).normalize()
+        slow = (rG - rF).normalize()
+        pixel_size = self.thickness_mm / self.num_pixels
+        image_size = (
+            self.num_pixels + px_off,
+            int(self.tape_depth_mm / pixel_size) + px_off,
+        )
+        # Face XY0
+        ori = rF
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "xy0"))
+        # face XY1
+        ori = rE
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "xy1"))
+        #
+        fast = (rA - rE).normalize()
+        slow = (rF - rE).normalize()
+        pixel_size = self.thickness_mm / self.num_pixels
+        image_size = (
+            self.num_pixels + px_off,
+            int(self.half_width_mm * 2.0 / pixel_size) + px_off,
+        )
+        # Face XZ0
+        ori = rE
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "xz0"))
+        # face XY1
+        ori = rH
+        faces.append(create_kapton_face(ori, fast, slow, image_size, pixel_size, "xz1"))
+        #
+        self.faces = faces
+
+    def get_kapton_path_mm(self, s1):
+        """ Get kapton path length traversed by an s1 vecto. If no kapton intersection or just touches the edge,
+            then None is returned"""
+        intersection_points = []
+        px = []
+        # determine path length through kapton tape
+        for face in self.faces:
+            try:
+                px = map(
+                    int, face[0].get_ray_intersection_px(s1)
+                )  # faces are single panel anyway
+                if (
+                    px[0] < 0
+                    or px[1] < 0
+                    or px[0] > face[0].get_image_size()[0]
+                    or px[1] > face[0].get_image_size()[1]
+                ):
+                    continue
+                intersection_points.append(
+                    face[0].get_lab_coord(face[0].get_ray_intersection(s1))
+                )  # faces are single panel anyway
+            except RuntimeError as e:
+                pass
+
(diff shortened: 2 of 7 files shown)


asmit3:
@ndevenish this branch is no more conflicts after fixing the blackened SConsript issue. Any other concerns here ?

Anthchirp:
Since you're rebasing anyway would suggest squashing commits. Or alternatively squash everything on merging.

ndevenish:
Hi, yes there are a few issues, but Friday/Monday is holiday in UK so only those of us checking out-of-hours will have seen it.

asmit3:
@ndevenish could you please list the issues you have noticed ?

ndevenish:
Firstly, wondering why you've gone for a series of detector panels rather than, say, directly defining a volume. I know we have lots of code to deal with shadowing/dynamic shadowing already, did you consider exploiting that at all? @rjgildea is probably better placed to comment on this.

Code-wise, please can you:
- Write tests for the new functionality - Ideally, unit tests without the need for extra data, but this can be sorted if essential
- Run `black` on the python code to format it
- Find and resolve clutter; our pre-commit uses flake8 for this but I've run it for you:
```
$  flake8 --select=W291,W292,W293,F401 algorithms/integration/kapton_2019_correction.py algorithms/integration/kapton_correction.py
algorithms/integration/kapton_2019_correction.py:7:1: F401 'iotbx.phil' imported but unused
algorithms/integration/kapton_2019_correction.py:8:1: F401 'scitbx.matrix.sqr' imported but unused
algorithms/integration/kapton_2019_correction.py:9:1: F401 'dials.algorithms.integration.kapton_correction.absorption_phil_scope' imported but unused
algorithms/integration/kapton_2019_correction.py:26:9: F401 'scitbx.matrix.sqr' imported but unused
algorithms/integration/kapton_correction.py:19:76: W291 trailing whitespace
```
And in C++:
- Please at a bare minimum document the intended effect of the C++ functions
- Is there a reason for multiple separate namespaces - even the closing then reopening of the same namespace - not being a header file and only ever being intended as a single translation unit I'm not sure the usual purposes apply. Also, what role is the anonymous namespace around the `kapton_init_module` playing - I'm not entirely familiar with boost::python so if this is a requirement/recommendation of it then no problem, but I'm not 100% sure how the usual purpose applies here.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+        #for s1 in s1_flex:
+        #  kapton_path_mm.append(self.get_kapton_path_mm(s1))
+        # determine absorption correction
+        if kapton_path_mm is not None:
+            absorption_correction = 1/flex.exp(-self.abs_coeff * kapton_path_mm) # unitless, >=1
+        return absorption_correction
+
+    def zone_of_point_in_kapton_shadow(self,s1):
+        """Given an s1 vector, determine whether it is in the increasing/decreasing or no kapton shadow regime"""
+        pass
```
ndevenish: This function appears not implemented. Would it be more appropriate to remove, or raise `NotImplementedError`?

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+
+    def distance_of_point_from_line(self, r0, r1, r2):
```
ndevenish: This doesn't use `self`. If the maths are intrinsically linked to the Kapton class, then maybe it would be better as a static method, or otherwise a standalone function

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+            and then determines absorption correction for each s1 vector """
+        kapton_path_mm=flex.double()
+        kapton_faces=self.faces
+        from dials.algorithms.integration import get_kapton_path_cpp
+        # new style, much faster
+        kapton_path_mm=get_kapton_path_cpp(kapton_faces, s1_flex)
+        # old style, really slow
+        #for s1 in s1_flex:
+        #  kapton_path_mm.append(self.get_kapton_path_mm(s1))
+        # determine absorption correction
```
ndevenish: Since this "old style" is new code we probably don't need to keep this as a reminder

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+            self.kapton_params_maxes = [[self.kapton_params[i] + self.kapton_params_sigmas[j]
+                                  if j==i else self.kapton_params[i]
+                                  for i in xrange(4)] for j in xrange(4)]
+            self.kapton_params_mins  = [[max(self.kapton_params[i] - self.kapton_params_sigmas[j], 0.001)
+                                  if j==i else self.kapton_params[i]
+                                  for i in xrange(3)] + [a] for j in xrange(3)] + \
+                                  [[self.kapton_params[i] for i in xrange(3)] + [a - sig_a]]
+
+    def __call__(self, plot=False):
+        def correction_and_within_spot_sigma(params_version, variance_within_spot=True):
```
ndevenish: Again - why does this need to be an embedded function? It doesn't appear to capture anything except `self`, so could it just be a normal member? (with a private prefix e.g. `_correction_and_within_spot_sigma` if only intended for internal use)

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+        sorted_idx=flex.sort_permutation(dlist, reverse=True)
+
+        edge_pts=[edges[dlist_idx[sorted_idx[0]][0]], edges[dlist_idx[sorted_idx[1]][0]], edges[dlist_idx[sorted_idx[0]][1]], edges[dlist_idx[sorted_idx[1]][1]]]
+
+        self.detector_edges=edge_pts
+       # Now get the maximum extent of the intersection of the rays with the detector
+        all_ints=[]
+        kapton_path_list = []
+
+        def not_present_in_list(mylist, datapoint, tolerance):
```
ndevenish: Another embedded function that could be standalone (my main objection to these is that if it's embedded then anyone reading your code needs to parse the whole function to check that it doesn't capture anything - the operating assumption should be that deep in a package there should a reason for it).

algorithms/integration/SConscript
```diff
@@ -35,3 +35,8 @@ env.SharedLibrary(
     source=["bayes/boost_python/ext.cc"],
     LIBS=env["LIBS"],
 )
+env.SharedLibrary(
+    target='#/lib/kapton_ext',
```
ndevenish: the output target should be called `dials_algorithms_integration_kapton_ext` to match the other libraries at this level

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+        # Get line intersection between 1-3 and a-b
+        r12_ab=get_line_intersection(m12,c12,mab,cab)
+        # Get line intersection between 0-2 and c-d
+        r03_cd=get_line_intersection(m03,c03,mcd,ccd)
+        # Get line intersection between 1-3 and c-d
+        r12_cd=get_line_intersection(m12,c12,mcd,ccd)
+        # returns pairs of x,y
+        return [(r03_ab[0], r03_ab[1], r12_ab[0], r12_ab[1]), (r03_cd[0], r03_cd[1], r12_cd[0], r12_cd[1])]
+
+class image_kapton_correction(object):
```
ndevenish: Less important, but something we've been trying to catch ourselves doing; This class appears to exist only to:
```python
some_action = image_kapton_correction(param, param, configuration, configuration)
result = some_action()
```
this isn't a class, it's a function. Is there any reason why it can't just be `result = image_kapton_correction(....)` We're actively trying to slowly refactor this pattern our where unnecessary.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+                    x_int, y_int=panel.get_lab_coord(panel.get_ray_intersection(s1))[0:2]
+                    int_point=(x_int, y_int, detz)
+                    # Arbitrary tolerance of couple of pixels otherwise these points were getting clustered together
+                    tolerance=min(panel.get_pixel_size())*2.0
+                    if not_present_in_list(all_ints, int_point, tolerance):
+                        all_ints.append(int_point)
+                    kapton_path_list.append(kapton_path_mm)
+                    #all_ints.append(edge_point)
+                except Exception as e:
+                    if 'DXTBX_ASSERT' not in str(e):
```
ndevenish: This catchall exceptions just swallows the error and print - a) is this really the intended behaviour and b) This should use the (existing) logger instead of printing

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+        intersection_points = []
+        px=[]
+        # determine path length through kapton tape
+        for face in self.faces:
+            try:
+                px=map(int,face[0].get_ray_intersection_px(s1)) # faces are single panel anyway
+                if px[0] <0 or px[1] < 0 or px[0]>face[0].get_image_size()[0] or px[1]>face[0].get_image_size()[1]: continue
+                intersection_points.append(face[0].get_lab_coord(face[0].get_ray_intersection(s1))) # faces are single panel anyway
+            except Exception as e:
+                if 'DXTBX_ASSERT' not in str(e):
```
ndevenish: See other `DXTBX_ASSERT` comment.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
...
+                    print (str(e))
+
+        if len(intersection_points) == 0:
+            return 0.0
+        if len(intersection_points) ==1:
+            print ('Ray not intersecting 2 faces.Either s1 vector does not intersect kapton or just touches edge of kapton tape \
```
ndevenish: This should also use the logger. Looks like a warning?

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,538 @@
+from __future__ import division, print_function
```
ndevenish: This should also `__future__` import `absolute_import`!

Anthchirp:
I don't want to add too much to this, but I encourage you to have a look under the "show all checks" link at the bottom of the pull request:
image
There you will find the LGTM analysis report:
image
While this test did not fail (I don't think it ever does) [it did highlight 14 problems](https://lgtm.com/projects/g/dials/dials/rev/pr-1b5ef5f14d3bbabf921b1ed1a7c6210b7dec2aca) in the new code. That number should ideally be 0 for all pull requests.

asmit3:
> Firstly, wondering why you've gone for a series of detector panels rather than, say, directly defining a volume. I know we have lots of code to deal with shadowing/dynamic shadowing already, did you consider exploiting that at all? @rjgildea is probably better placed to comment on this.
> 
We are interested in calculating path length through a solid, not necessarily shadowing, so it made sense to us to use dxtbx tools.

> Code-wise, please can you:
> * Write tests for the new functionality - Ideally, unit tests without the need for extra data, but this can be sorted if essential
> * Run `black` on the python code to format it
> * Find and resolve clutter; our pre-commit uses flake8 for this but I've run it for you:
> 

A kapton test has been added to our xfel_regression repo which will be tested nightly at LBL. The code in this pull request has been blacked, clean cluttered, comments added (C++) & comments on specific portions of the code have been addressed as deemed fit.
We would like to merge this into master ASAP.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,648 @@
...
+                    px[0] < 0
+                    or px[1] < 0
+                    or px[0] > face[0].get_image_size()[0]
+                    or px[1] > face[0].get_image_size()[1]
+                ):
+                    continue
+                intersection_points.append(
+                    face[0].get_lab_coord(face[0].get_ray_intersection(s1))
+                )  # faces are single panel anyway
+            except Exception as RuntimeError:
```
Anthchirp: correctness: this captures all exceptions and redefines RuntimeError.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,648 @@
...
+
+        for ii, edge_point in enumerate(self.edge_points):
+            s1 = edge_point.normalize()
+            kapton_path_mm = self.get_kapton_path_mm(s1)
+            for panel in detector:
+                try:
+                    x_int, y_int = panel.get_lab_coord(panel.get_ray_intersection(s1))[
+                        0:2
+                    ]
+                except Exception as RuntimeError:
```
Anthchirp: correctness: this captures all exceptions and redefines RuntimeError.

algorithms/integration/kapton_2019_correction.py
```diff
@@ -0,0 +1,648 @@
...
+                    all_ints.append(int_point)
+                kapton_path_list.append(kapton_path_mm)
+        # Use the idea that the extreme edges of the intersection points are end points of the diagonal and will be the
+        # top 2 max dimension among all end points
+        dlist = flex.double()
+        dlist_idx = []
+        n_edges = len(all_ints)
+        for ii in range(n_edges - 1):
+            for jj in range(ii + 1, n_edges):
+                pt_1 = col(all_ints[ii])
```
Anthchirp: performance: can go into the outer loop

Anthchirp:
In DIALS, dxtbx, and xia2 we now keep the tests with the code (but not any test-related data files).
Having tests in xfel_regression is equivalent to not having tests: they are not run by Travis, not run by Jenkins, not run by anyone in DIALS East, and nobody knows they're there.

Please move the tests into the DIALS repository. If the tests rely on any data files in xfel_regression you can reference them there and skip the test with an appropriate message if the files are not present on the system. That way the tests may still not be run, but they will show up as skipped and tell developers what they need to do to run them. Plus the data files can be made publicly available at a later date, there is work in progress on this.

ndevenish:
It's important to have a test, even if it's explicitly skipped without data for now - tests of dials code should be within dials, otherwise you need "special knowledge" about how to actually run the tests.

With the `except Exception as RuntimeError:` lines fixed and a test moved in (even if skipped without data for now - it lets the developer know) I'm happy for this to be merged.

asmit3:
@Anthchirp @ndevenish I have added a kapton test to dials which is marked to be skipped by pytest. Bugs have been fixed. Thanks for feedback on the code.

ndevenish:
Done. One other question though; the two `except RuntimeError as e: pass` clauses; what is this catching/where does it come from? Still a little nervous with ignoring errors.

asmit3:
We added the RuntimeError catching only around the block of code where a dxtbx panel function is used (to make it as limited as possible). In the kapton code when xrays do not intersect a kapton face (here a dxtbx panel object), a DXTBX_ASSERT error is thrown in C++ which translates to a RuntimeError in python land.

ndevenish:
Ah, makes sense - though that seems an odd way of indicating a miss.

Issue #627, opened 2018-10-04
More than one Eiger data set referenced from single master file; test / implement


graeme-winter:
See e.g.

https://zenodo.org/record/1442922

31 / 38 data sets from Eiger collected from one arming so one master file - no idea how these look, will investigate once home. Data from Br soaked lysozyme.

graeme-winter:
https://github.com/keitaroyam/yamtbx/blob/master/doc/eiger-en.md

for details (as yet incomplete)

Unpacking to /dls/mx-scratch/dials/dials-627/ccp4school2018_bl32xu/lysbr2-demo-03/data/  @ Diamond

graeme-winter:
Data seem to import correctly but as a 3100 image sweep

Testing xia2 with:

```
xia2 atom=Br image=../ccp4school2018_bl32xu/lysbr2-demo-03/data/demo-03-multi_master.h5:1:3100:100
```

since this should treat each 100 image block independently - dials.import could scan the OMEGA table to determine that these are in fact multiple sweeps - current outcome:

```
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 1.dev.2857-gf589b6da7
The following parameters have been modified:

input {
  datablock = <image files>
}

--------------------------------------------------------------------------------
  format: <class 'dxtbx.format.FormatHDFEigerNearlyNexus.FormatEigerNearlyNexus'>
  num images: 3100
  num sweeps: 1
  num stills: 0
--------------------------------------------------------------------------------
Writing datablocks to datablock.json
```

graeme-winter:
N.B. can also look at ntriggers / nimages in master file and "fake up" the :a:b:n stuff

graeme-winter:
```
Grey-Area data :) $ h5ls -r demo-04-multi_master.h5 |grep ntrigger
/entry/instrument/detector/detectorSpecific/ntrigger Dataset {SCALAR}
Grey-Area data :) $ h5ls -r demo-04-multi_master.h5 |grep nimages
/entry/instrument/detector/detectorSpecific/nimages Dataset {SCALAR}
/entry/instrument/detector/detectorSpecific/summation_nimages Dataset {SCALAR}
Grey-Area data :) $ h5dump -d /entry/instrument/detector/detectorSpecific/ntrigger demo-04-multi_master.h5 
HDF5 "demo-04-multi_master.h5" {
[...]

PR #774, opened 2019-05-02
DIALS 1.14.4

Anthchirp:
* HDF5: support for multiple datasets referenced from a single master file (#627)
* Add support for Diamond Light Source VMXm beamline data
* `dials.reciprocal_lattice_viewer`: support for old format `.pickle` files
* `dials.export_mtz`: handle split sweeps
* `xia2`: Provide more useful output when called programs fail (xia2/xia2#300)
* `dials.report`: include additional plots after scaling
* `dials.scale`: Fix possibility for recursion error in outlier rejection, generate scaling.html output report.


Anthchirp:
@jbeilstenedmands have to say not a massive fan of basically copying commits over from master rather than backporting them. For example removing `dials.plot_scaling_*` scripts from a released version in a patch update breaks semantic versioning. They could have just stayed. Refactoring changes are high-risk no-gain on a release branch and will break other people using the API in semantic versioning, etc. I know that scaling is very much in flux so it is unlikely that there are users out there, but still... moving code between xia2 and dials sounds very much sounds like a recipe for disaster.
Going forward would prefer to keep the stable build stable with maintenance fixes only, the odd new feature is fine, adding deprecation warnings is fine. Anyone wanting to use the new features can be pointed to a nightly build.

jbeilstenedmands:
Hi @Anthchirp ,
I agree with your point about removing the plotting scripts, I can add these back in. I didn't actually remove any code from xia2 for dials-next, just adding the code that dials needs for plotting. I'm aware that including the refactoring code introduces some risk of errors, however some of the changes are necessary for the new features and I judged that this was safer than picking partial changes from this. The main motivator behind this is providing the updated report plots for ccp4/dui post-scaling.

Anthchirp:
@benjaminhwilliams @jbeilstenedmands can I have one/more lines for the release notes please?
Feel free to edit directly into https://github.com/dials/dials/pull/774#issue-275294946

PR #775, opened 2019-05-03
Pythonic reserved names

graeme-winter:
Avoid python reserved names as variables in source code - have avoided wx which seems to overload the use of id by default but otherwise we have a few hand-tweaked names which should be idiomatically sensible.

### DIFF
--- a/util/image.py
+++ b/util/image.py
@@ -39,9 +39,7 @@ class reader:
 
         # Check the type of the element to ensure it's a binary
         # otherwise raise an exception
-        dtype = self.cbf_handle.get_typeofvalue()
-        if dtype.find("bnry") > -1:
-
+        if "bnry" in self.cbf_handle.get_typeofvalue():
             # Read the image data into an array
             image_string = self.cbf_handle.get_integerarray_as_string()
             image = flex.int(numpy.fromstring(image_string, numpy.int32))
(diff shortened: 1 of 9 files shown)


graeme-winter:
Obviously squash and merge would be tidier

util/image.py
```diff
@@ -62,7 +62,7 @@ def get_data(self):
 
 if __name__ == "__main__":
     import sys
-    import numpy
+    import numpy  # noqa: F401
```
ndevenish: Is this necessary at all?

util/intensity_explorer.py
```diff
@@ -392,10 +392,10 @@ def data_from_unmerged_mtz(filename):
     rtable["id"] = flex.int(rtable.size(), 0)
 
     # Now generate a corresponding experiment list.
-    id = flex.vec3_double([(1, 0, 0), (0, 1, 0), (0, 0, 1)])
+    _id = flex.vec3_double([(1, 0, 0), (0, 1, 0), (0, 0, 1)])
```
ndevenish: This appears to be using miller index as an id - so `indices` might be better? or even just `ids`...

ndevenish:
I rewrote some of the names to be what I thought were better choices. e.g. things like instead of renaming `file`→`f`, rename to `filename` because that's what it is. 

I'm fine with everything now.

PR #776, opened 2019-05-08
Add check-merge-conflict pre-commit hook.

jbeilstenedmands:
Add the check-merge-conflict precommit hook, to check that
"<<<<<<< HEAD" etc hasn't been left in a file, which I have accidentally done before and know of others who have also.
While some of these cases should be picked up by the check-ast hook for python, this can be missed if within a docstring or non-python file.


Anthchirp:
no concerns noted within a week

Issue #795, opened 2019-05-23
matplotlib backend set in /algorithms/spot_finding/per_image_analysis.py


asmit3:
#### Context
While trying to debug code in ```dials.stills_process```, I was unable to show plots using matplotlib when I put a breakpoint in the code (atleast not interactively). The reason seems to stem from https://github.com/dials/dials/blob/63d2fef817489c9f68da68ee7e3d808d5ae72252/algorithms/spot_finding/per_image_analysis.py#L8-L15

Here the matplotlib backend is set as part of a global import which turns off interactive plotting. Thus when a program imports something from this file, the global import lines for matplotlib turn off interactive plotting (this is what's happening with ```dials.stills_process```). In ```dials.stills_process``` this is happening indirectly through imports made from ```indexer.py``` --> ```max_cell.py```

#### Question
I was wondering if there is a reason for declaring the backend at the top in ```per_image_analysis.py```? Can anything be done to not set the backend here and instead do so when plotting is done within the program ? 

Just for demonstration, here is a barebones program that does not currently show plots. This is similar to the case in ```dials.stills_process```

```
from libtbx.phil import parse

phil = parse("""
  include scope dials.algorithms.indexing.indexer.index_only_phil_scope
""", process_includes=True).extract()
from matplotlib import pyplot as plt; plt.plot(range(10), range(10)); plt.show()
```

Anthchirp:
I think it is reasonable to move the entire `import matplotlib` down to the actual plotting code blocks. Then I would also drop the `try:`/`except:`/`Sorry` bit as the import is required and any `ImportError` can be dealt with upstream. That way we wouldn't even import `matplotlib` when the code in `per_image_analysis.py` is used without plotting.
@rjgildea, @ndevenish what do you think?

Anthchirp:
Doing this would make mocking harder, but I don't see any test directly importing this file, so there is no immediate downside as far as I can see.

ndevenish:
Reopening, because it seems this is throwing lots of warnings because matplotlib versions/support and backends. Need to have another look.

Might be worth backing out change and putting the other one in for now.

ndevenish:
I've switched to the moving commit 5f595583c4e7e3dc494f261de1b6d85abf8b83b6 which hopefully solves without the issues of the other one.

PR #796, opened 2019-05-25
Per-image-analysis matplotlib move

graeme-winter:
As requested in #795 and following suggestion from @Anthchirp 

Remove import from top of file; move to local imports with no Sorry (if no matplotlib should be handled by calling code)

### DIFF
--- a/algorithms/spot_finding/per_image_analysis.py
+++ b/algorithms/spot_finding/per_image_analysis.py
@@ -274,8 +274,10 @@ def estimate_resolution_limit(reflections, imageset, ice_sel=None, plot_filename
 
     if plot_filename is not None:
         import matplotlib
+
         matplotlib.use("Agg")
         from matplotlib import pyplot
+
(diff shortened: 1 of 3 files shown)


rjgildea:
Perhaps it would be better to set the backend at a single point in the command line program just before this line here:
https://github.com/dials/dials/blob/33fa5472920dbca6332b2391ea2951bf2a96b7c3/command_line/spot_counts_per_image.py#L146
```
import matplotlib
matplotlib.use("Agg")
```
rather than liberally sprinkling this at every point where matplotlib is used. This would mean that it is up to the calling program to decide the backend, rather than enforcing it on users of the code.

graeme-winter:
That would be a lot tidier; was just aiming to reproduce existing behaviour - wilco

ndevenish:
I've done this in an alternate way in #800. The output from @rjgildea's example is identical.

PR #802, opened 2019-06-03
Updates to dxtbx and dials to allow image viewer regarding load_models

dermen:
This PR arose from working with XTC Formats, where the detector models are loaded on the fly and in order to load all models one must parse entire XTC files, which is time consuming and not necessary for stills image display. 

Current image viewer requires that all detector and beam models be loaded prior to display. In which cases is this necessary ? Maybe one would display a scan vector for rotation series data ? If most of the time folks would rather just load models on the fly, then perhaps the default value should be False in the image viewer, but for now it is as it was, set t as load_models=True for all image formats

### DIFF
--- a/util/options.py
+++ b/util/options.py
@@ -549,7 +549,7 @@ class PhilCommandParser(object):
         try:
             load_models = params.load_models
         except AttributeError:
-            load_models = True  # NOTE: this is defaulted in Importer as True already
+            load_models = True
 
         # Try to import everything
         importer = Importer(
--- a/util/options.py
+++ b/util/options.py
@@ -275,7 +275,7 @@ class Importer(object):
         :param compare_goniometer:
         :param scan_tolerance:
         :param format_kwargs:
-        :param load_models: whether to load all models for expLists
+        :param load_models: Whether to load all models for ExperimentLists
         :param propagate_geom: for stills, whether to propagate the initial images geometry to all experiments
           (mainly used for imageviewer)
         :return: Unhandled arguments
--- a/util/options.py
+++ b/util/options.py
@@ -188,6 +188,7 @@ class Importer(object):
         compare_goniometer=None,
         scan_tolerance=None,
         format_kwargs=None,
+        load_models=True,
     ):
         """
         Parse the arguments. Populates its instance attributes in an intelligent way
@@ -207,7 +208,7 @@ class Importer(object):
         :param read_experiments_from_images: Try to read the experiments from images
         :param check_format: Check the format when reading images
         :param verbose: True/False print out some stuff
-
+        :param load_models: whether to load models from every experiment
         """
 
         # Initialise output
@@ -227,6 +228,7 @@ class Importer(object):
                 compare_goniometer,
                 scan_tolerance,
                 format_kwargs,
+                load_models,
             )
 
         # Second try to read experiment files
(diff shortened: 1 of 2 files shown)


graeme-winter:
Conceptually we recognise the need for this however

 - this fix is kind-of nasty in the option parser
 - we need a way more generally parse PHIL strings before importing data

we have this problem in other places and use an environment variables to overcome this, which is more explicit

Could we consider `DIALS_DELAY_LOAD_MODELS=1` as a workaround? Then resolve this properly?

graeme-winter:
See e.g. `DIALS_EXPORT_DO_NOT_CHECK_FORMAT` in export.py

dermen:
@graeme-winter , Environment variable works for me, basically anything that allows the image viewer to work with XTC streams (or other huge formats).

phyy-nx:
FormatXTC is a very interesting problem.  It stresses all of our software in ways we didn't anticipate.  Not only is dials.image_viewer essentially broken with FormatXTC, dxtbx.image_average, dxtbx.radial_average, dials.import, and other programs simply don't work.

@dermen and I are reviewing this pull request tomorrow.  Thanks all for your feedback.  We'll be in touch.

phyy-nx:
(reminder, for production LCLS processing we use cctbx.xfel.xtc_process, which uses a CBF intermediate layer and is its own beast. We'd like to deprecate it though at some point and use just FormatXTC)

graeme-winter:
Understand your problem here with the XTC stream (of consciousness) and your motivation for the fix. I would like to be able to get to a place where we have a proper solution for things like this, in preference to something which feels like a "hack"

In principle - will merging this in prevent us from looking for a proper solution? Once you have told your users in newsletter articles to do `$thing` we will never be allowed to change the mechanism because `$thing` will be in print somewhere...

From your options presented above, 3 is desirable for this reason...

Actually the more I look at it, the more option 3 makes sense - a lazy image set _knows_ it should be lazy 

Or have I missed something?

ndevenish:
As long as we don't have:
> will merging this in prevent us from looking for a proper solution?

Then it's probably okay to put this in. My main concern is that it's making OptionParser change behaviour based on magically having a phil parameter present in the specification, which isn't exactly intuitive. If it's intended as a short-term solution then it's probably okay, otherwise it might be better to either
a) Make it part of the universal OptionParser API
or
b) Make a special OptionParser subclass/overrideable behaviour just for image_viewer

util/options.py
```diff
@@ -261,14 +263,22 @@ def try_read_experiments_from_images(
...
         Try to import images.
 
         :param args: The input arguments
         :param verbose: Print verbose output
+        :param compare_beam:
+        :param compare_detector:
+        :param compare_goniometer:
+        :param scan_tolerance:
+        :param format_kwargs:
+        :param load_models: whether to load all models for expLists
```
ndevenish:
```suggestion
        :param load_models: Whether to load all models for ExperimentLists
```

util/options.py
```diff
@@ -293,6 +303,7 @@ def try_read_experiments_from_images(
             compare_goniometer=compare_goniometer,
             scan_tolerance=scan_tolerance,
             format_kwargs=format_kwargs,
+            load_models=load_models,  # default is True in from_filenames
```
ndevenish:
```suggestion
            load_models=load_models,
```

util/options.py
```diff
@@ -535,6 +546,11 @@ def parse_args(
             scan_tolerance = None
             format_kwargs = None
 
+        try:
+            load_models = params.load_models
+        except AttributeError:
+            load_models = True  # NOTE: this is defaulted in Importer as True already
```
ndevenish:
```suggestion
            load_models = True
```

util/options.py
```diff
@@ -207,7 +208,7 @@ def __init__(
         :param read_experiments_from_images: Try to read the experiments from images
         :param check_format: Check the format when reading images
         :param verbose: True/False print out some stuff
-
+        :param load_models: whether to load models from every experiment
```
ndevenish: Since this just passes through to `try_read_experiments_from_images`, it should have the same description. At the moment it's different.

phyy-nx:
No reason we can't investigate 3 as an option in the future and then deprecate load_models.  I agree the load_models approach is a short term solve.

@dermen has made the suggested changes.  Voting to merge.

graeme-winter:
@ndevenish please could you confirm that the changes you requested have been applied?

phyy-nx:
Ok to merge?

ndevenish:
Seems good

PR #805, opened 2019-06-05
Index refactor

rjgildea:
1. Refactor basis vector search code into strategies in `dials.algorithms.indexing.basis_vector_search.strategies`
    - fft1d, fft3d, real_space_grid_search
    - take as input a list of reciprocal lattice vectors and return a list of candidate basis vectors
2. Refactor basis vector combination code into `algorithms.indexing.basis_vector_search.combinations`
    - Implemented as a sequence of iterators that can be chained together
    - candidate_orientation_matrices: finds suitable combinations of basis vectors in P1
    - filter_known_symmetry: filters output of previous to only accept those models that match known symmetry
    - filter_similar_orientations
3. Refactor assign_indices to a strategy pattern in `dials.algorithms.indexing.assign_indices`
    - assign_indices_global, assign_indices_local
4. Move non-primitive basis code to separate module `dials.algorithms.indexing.non_primitive_basis` with tests
5. New `dials.algorithms.indexing.model_evaluation`
    - Move/rename SolutionTrackers: `ModelRankWeighted`, `ModelRankFilter`
    - move run_one_refinement inner function to new `ModelEvaluation` class
6. Attempt to encapsulate handling of known symmetry into external class `dials.algorithms.indexing.symmetry.SymmetryHandler`, with tests
7. New `dials.algorithms.indexing.lattice_search` module
    - Single `BasisVectorSearch` inheriting from `indexer_base` to consolidate previous `fft1d`, `fft3d` and `real_space_grid_search` classes. Code that is specific to a basis vector search now lives in this class instead of `indexer_base`.
    - Add simple test for `BasisVectorSearch` class
8. Move basis vector optimisation code to new module `dials.algorithms.indexing.basis_vector_search.optimise` with tests
9. Replace scan_range parameter with image_range
10. Refactor command line script to a functional interface
    - add joint_indexing=False option to allow indexing multiple independent experiments in a single dials.index process
    - include parallel indexing using `concurrent.futures` backport
11. Some refactoring of indexer.index() method to minimise code duplication with stills_indexer


algorithms/indexing/stills_indexer.py
```diff
@@ -960,19 +808,9 @@ def refine(self, experiments, reflections):
...
-    pass
-
-
-class stills_indexer_real_space_grid_search(
-    stills_indexer, indexer_real_space_grid_search
-):
-    pass
-
-
-class stills_indexer_fft3d(stills_indexer, indexer_fft3d):
```
phyy-nx: Is the stills fft3d indexer gone?

phyy-nx:
@rjgildea this is very cool.  Happy to see the stills_indexer redundancy with the indexer resolved!!

Questions:
Is the stills fft3d indexer gone?
Can the index_reflections function still exist, but throw a deprecation warning?  We have other projects not in dials or cctbx that are using index_reflections directly.  This seems like it's an API change that merits some time to adjust to.

Otherwise, I'll run the xfel regression tests with this when I get a chance.

rjgildea:
> Is the stills fft3d indexer gone?

Yes, there is now a single BasisVectorSearch (and StillsBasisVectorSearch) class which selects the basis search vector strategy based on the value of params.indexing.method:

https://github.com/dials/dials/blob/da74231881cca84b1cc13f420957fe9084708b6c/algorithms/indexing/lattice_search.py#L128-L157

https://github.com/dials/dials/blob/da74231881cca84b1cc13f420957fe9084708b6c/algorithms/indexing/stills_indexer.py#L815-L816

phyy-nx:
test_stills_process.py fails for me on this line:
https://github.com/dials/dials/blob/index-refactor/test/command_line/test_stills_process.py#L141

That test is looking at the number of integrated reflections, and on the index-refactor branch it fails with image 3 have 518 reflections.  On master this test passes.  To me this indicates a change in functionality.

Haven't gone deeper to figure out what changed yet.

This is on an lcls interactive node, rh7.

rjgildea:
> test_stills_process.py fails for me on this line:
> https://github.com/dials/dials/blob/index-refactor/test/command_line/test_stills_process.py#L141
> 
> That test is looking at the number of integrated reflections, and on the index-refactor branch it fails with image 3 have 518 reflections. On master this test passes. To me this indicates a change in functionality.
> 
> Haven't gone deeper to figure out what changed yet.
> 
> This is on an lcls interactive node, rh7.

Interestingly the test passes on both master and the branch on RHEL6, however I noticed from comparing the output that in the branch it was refining one more crystal model than requested. After fixing this in 9d80920bb7aed61d1e002552c6b3dea94bdb7fbf the output appears to be more or less identical on both branches.

rjgildea:
> Can the index_reflections function still exist, but throw a deprecation warning? We have other projects not in dials or cctbx that are using index_reflections directly. This seems like it's an API change that merits some time to adjust to.

@phyy-nx I've added a deprecated index_reflections function in 72c163f15

rjgildea:
@phyy-nx are you happy that your tests now behave as expected? I'm hoping to merge this early next week if there are now objections.

phyy-nx:
Yup.  Ship it.  Happy with this and the cctbx pull request.

Thanks,
-Aaron

dermen:
@rjgildea , just out of curiosity, what was the main motivation behind this refactor ? Was the old code broken in any way ? I only ask because I am using a copy of the older real space grid search code (two_color_indexer.py in the cci module  cxi_xdr_xes), and I am wondering if my results generated with the old code are bad, and do I need to re-write the code to comply with this refactor. Thanks!

ndevenish:
The primary purpose of refactoring is to improve the code/code design without altering the behaviour. This can be beneficial for many reasons, such as new unanticipated requirements, clearing [technical debt](https://martinfowler.com/bliki/TechnicalDebt.html), tending to [code](https://en.wikipedia.org/wiki/Software_rot) [rot](https://henrikwarne.com/2017/04/28/code-rot/), [readability](https://software.ac.uk/resources/guides/writing-readable-source-code) - making the code easier to understand (and thus reason about) and, normally, you understand the problem much more after you've written it than when you were first writing it. It should be a natural part of any large software project. It's the tests let you do this with the reassurance that you haven't broken anything. See also; general thesis of [Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882).

Note that saying code needs to be refactored **explicitly** *isn't* a criticism of the original code; needs, requirements and understanding of the problem change over time, and overengineering in the first place can be as bad as or worse than something badly written.

As far as behaviour changes,I believe several bugs were noticed and fixed during the refactor, and features were added that the new design enabled but the old design obstructed. If there is some API interface you were reasonably using that is broken then obviously we'd like to see if that could be sorted. Do your tests still pass?

Was there some reason that you had to duplicate the code in another module rather than use it directly? Dials not being available as an import?

dermen:
@ndevenish  @rjgildea  I inherited the code from another coworker who has since left, so I did not write it originally, just added to it - so I cant say for certain why it was duplicated.  I suspect the original real space grid search API  prevented the desired functionality, hence why it was copied rather than inherited from, but I am not sure. In any case, the tests stopped working which is why I came here for more information. Im happy using an older commit in dials to keep the code working, but the bugs you hinted at cause room for concern now.  In any case, I will work on moving the code into the public domain , where it will live in cctbx/rstbx. Then I will point you to it. I actually added some features to the ported grid search that might be nice to implement in the dials grid search as well (for example searching for basis vectors on a rough hemisphere as opposed to a smooth one as a kind of refinement). Just so you know what to expect, the code I am using that is a copy of real space grid search is a multi-wavelength, single crystal indexing class which inherits from dials indexer_base and stills_indexer .

dermen:
Also, probably the most useful thing now, however, would be a report of the bugs discovered in the refactor

Issue #507, opened 2018-02-09
Discussion topic: experiment.id != (0,1, ..., (n-1))


graeme-winter:
Currently when indexing > 1 sweep we have 

dials.index db1 refl1 db2 refl2 

and the _order_ defines the experiment id - this is obviously fragile. Therefore, the concept of having the dataset id or experiment id be a unique ID rather than an integer counting from 0 would help to reduce the chance of this.

Proposal:

datablock.id is created which is a unique identifier (perhaps random number, better a hash derived from pixels) which is then used as the id column in the derived strong pickle. Then, this can be used to reliably match up reflections with source

Same applies to experiments...

Thoughts?

phyy-nx:
Another use case to consider is images referenced by timestamp, such as we
get from xfel data.  I find I can easily lose track of which shot an
experiment id came from after a few dials.combine_experiments, and an
arbitrary hash wouldn't help me find my way back.  However, if there were a
hash in string form that was application specific, that would be very
helpful as I could use my timestamp as the hash.

graeme-winter:
Reflection table has lookup in dictionary => points to same value as experiment.id

graeme-winter:
@jmp1985 says sounds sensible

graeme-winter:
Needs to be on serialise / de-serialise that this is used.... mostly working algorithms should not be changed as the experiment id (working wise) should still be a small integer

graeme-winter:
@nksauter @phyy-nx you have opinions about this, happy for us to go ahead? Really need this for multi-experiment handling in scaling (and @jmp1985 going to start looking at this real soon)

dagewa:
Just a note to say that I am sure that refinement code will need changes following this. The assumption that the order of experiments determines the experiment id is made in a few places, I think. Current tests won't pick this up unless that assumption is violated in the tests.

phyy-nx:
I also have made the assumption that the order of the experiments determines the experiment ID in many places. I wouldn’t want to get rid of that assumption. I would want however to add a hash or string that would help us always be able to connect an experiment with its source.

If we do feel it’s important to break this assumption, I would need time to look for where it’s relied on in the XFEL sources. It would be any loop like this (and there are many loops like this in the source code).

for exp_id, experiment in enumerate(experiments):
  refls = reflections.select(reflections[‘id’] = exp_id)

-Aaron

jmp1985:
[...]

PR #819, opened 2019-06-13
change default experiment identifier

asmit3:
This is parallel pull request with cctbx/dxtbx#39 . Since after spotfinding the experiment identifier was set to an empty string in stills_indexer, these changes (a) ensure the identifier is preserved through indexing and subsequently integration (b) append tags based on number of lattices indexed. The tags based on number of lattices ensure that each experiment identifier for LCLS experiments is unique, even in situations where there are multiple lattices. 

Example of changes
```
In [1]: from dxtbx.model.experiment_list import ExperimentListFactory

In [2]: spotfinder_expts=ExperimentListFactory.from_json_file('idx-mfx_data_00001_experiments.json')

In [3]: indexing_expts=ExperimentListFactory.from_json_file('idx-mfx_data_00001_refined_experiments.json')

In [4]: spotfinder_expts[0].identifier
Out[4]: '2016-12-13T21:10Z22.805'

In [5]: indexing_expts[0].identifier
Out[5]: '2016-12-13T21:10Z22.805-00'
```
As being discussed in the dxtbx PR, the changes are specifically for stills. Hoping that something similar will be done for sweeps to maintain consistency in behaviour.

### DIFF
--- a/algorithms/indexing/stills_indexer.py
+++ b/algorithms/indexing/stills_indexer.py
@@ -137,13 +137,25 @@ class StillsIndexer(Indexer):
             n_lattices_previous_cycle = len(experiments)
 
             # index multiple lattices per shot
+            # Append tags to identifiers  and assign them if identifier
+            # from spotfinding is not an empty string
+            # For example for XTC streams when read with FormatXTC
+            # 2016-12-13T21:10Z19.804-00 where -00 gets added at the end.
             if len(experiments) == 0:
                 experiments.extend(self.find_lattices())
                 if len(experiments) == 0:
                     raise DialsIndexError("No suitable lattice could be found.")
+                if self.experiments[0].identifier != "":
+                    experiments[0].identifier = self.experiments[
+                        0
+                    ].identifier + "-%02d" % (len(experiments) - 1)
             else:
                 try:
                     new = self.find_lattices()
+                    if self.experiments[0].identifier != "":
+                        new.identifier = self.experiments[0].identifier + "-%02d" % (
+                            len(experiments) - 1
+                        )
                     experiments.extend(new)
                 except Exception as e:
                     logger.info("Indexing remaining reflections failed")


ndevenish:
I'm extremely concerned that using *information* for the identifier will lead to relying on that information. The intent for this field when put in was that it would be some autogenerated internal index field, or hash, or UUID. If you want to represent some representation of the history of processing it'd be better to use a field designed for it.

As I understand it, this it for output identification purposes, can you not just lookup the file-timestamp on the source files when you are inspecting the experiment?

graeme-winter:
+1 for UUID or hash or some other arbitrary value. In database terms this identifier should be a meaningless primary key - any attempt to encode useful information in there is unlikely to end well. 

You already illustrate this above - 

```
In [5]: indexing_expts[0].identifier
Out[5]: '2016-12-13T21:10Z22.805-00'
```

It will be the case that you may never have > 99 lattices (I agree, it is unlikely) - however what if the timestamp used was `2016-12-13T21:10Z22.805-PDT` and you then get a ValueError converting PDT to an int?

Whatever you encode in here, someone will rely on it and the form of it - which will eventually go wrong. A hash like ec513ced86906fa2d514720216b99a9a or a UUID (hashes are good as you can repeat the construction of the value so if you re-import the data you end up with the same ID) means that people will not try to encode information in them.

Anthchirp:
hashes are probably not as good as they might seem.
Firstly you create the expectation that same data leads to same hash, inviting people to rely on it - something you predicted in the previous sentence will eventually go wrong.
Secondly you might include the same data twice (say, same stills image?). I understood @jbeilstenedmands comment in the other place to mean "when a collision occurs this will blow up in new and exciting ways". So while including data twice may not be something you aspire to, this outcome is probably definitely not what you want to happen in that situation.

graeme-winter:
Fair enough 🙂UUID then

asmit3:
Thanks all for the comments and appreciate the discussion. 

1) The use of hashes/UUID is a great solution in situations where 100s of events might have taken place at the same time (like credit card transactions). Since the arrow of time always points in one direction and each LCLS experiment (i.e proposal) is unique in itself, the use of time as an identifier for formatXTC is OK. Essentially the timestamp is a UUID.

2) The identifier is a string so I am not sure we would be casting to an int at any point (besides hashes are also not castable to int AFAIK).  Currently, the identifier is an empty string and since the commits here and in dxtbx are specific to formatXTC, it won't actually affect any other stills processing (and definitely not sweeps). 

3) Benefits to us: The use of timestamp as an identifier in stills is beneficial for us in multiple ways (a) debugging (b) feedback to experimentalists (c) syncing/processing data collected on multiple detectors at different frequencies.

Let me be more clear as to point 3c.  We collected an experiment at LCLS using two XRD detectors and a spectrometer.  One of them were running at 30 Hz, one of them was sometimes running at 30 and other times running at 120Hz, and the third was running at 120Hz.  This means we have three imagesets, each with a different number of images.  We need to match events together, and image number is insufficient, since they don't sync up.  For this case, using the LCLS timestamp allows us to sync them up.

While this is a very specific case, the point is that the format class should be able to define it's own set of identifiers, for whatever reason.  If the importer reads those identifiers and puts them in the experiment list, they become useful for downstream analysis.

We are happy to merge the 2 PR (since they address our specific concerns for stills) if you feel a separate conversation is to be had regarding doing identifier assignment for sweeps. We can also wait out till next week when @jbeilstenedmands said he will take a look at the sweeps identifier.

rjgildea:
I'm concerned that this commit further diverges the stills indexer from the standard indexer. Any solution would ideally be applicable to both stills and sweeps (as the issue of tracking identifiers is not unique to stills), without any duplication of code. It would probably be better to wait to address the question of how to carry identifiers forward through indexing until we have resolved cctbx/dxtbx#39 more generally.

ndevenish:
So we discussed this today and think we know a way you can use them how you want.

We will set the identifier arbitrarily on import, and preserve that - you are completely free to relabel it and use for information storage however you wish in stills_process - but in the main tools there will be no support for treating it as anything except conserved arbitrary data. Indexing will regenerate for separate lattices, but you can re-relabel them after indexing (which would need to happen in the timestamp scheme anyway).

Would this accommodate your use case?

asmit3:
Are there any incoming commits/PRs to set the identifier arbitrarily as discussed in this thread (like using a UUID) ? Without looking at the implementation of setting the identifier, it's difficult to tell if my use case can be accommodated.

graeme-winter:
@asmit3 nope, the idea being we discuss what we need to do _before_ implementation to avoid the feeling from others that they are being railroaded into an implementation they do not want. See suggestions at https://github.com/dials/dials/issues/507#issuecomment-503792130

My feeling with this is that we _should not_ encode information in the identifier - if we want human readable things then they should be tagged separately. The identifier should only be a unique primary key - this is a long established (though still debated) topic in database land e.g.

https://www.techrepublic.com/article/the-great-primary-key-debate/

If you encode information in the primary key you _will_ come unstuck sometime. That said, the contract proposed by @ndevenish is that we faithfully track whatever primary key is used throughout the system (modulo creation of new experiments in indexing) so if you want to have a primary key with meaning, then fine, great but we won't pull you out of the swamp when you get stuck 😉

So, with regards to this actual PR - we do not think the default identifier for all experiments should be changed to be a timestamp, but we agree that this should be an _option_ - meanwhile the default for rotation sets should be the arbitrary string @ndevenish mentioned above e.g. a UUID or similar. FWIW there are facilities around the place (PF BL-1A) which operate with two detectors synchronised, collecting two data streams on different detectors. In this world we would have two distinct experiments (since the detectors are different) with precisely the same time stamps => not unique.

asmit3:
@graeme-winter @ndevenish I am a little confused now. Are you proposing (a) DIALS assigns an initial identifier by default and then we can change it within stills_process _or_ (b) We can assign a custom identifier (like timestamp in this case) from the very beginning (i.e import) using some kind of flag which will be False by default. I would of course prefer (b) as it gives us flexibility.

jbeilstenedmands:
Hi @asmit3 . Sorry I dropped the ball on this a while ago but coming back to it now. I think (b) is what we were proposing. I've pushed a branch i've just started working on to the branch 'expt_identifiers' in dials for you to take a look. I haven't implemented the timestamp identifier yet, happy to let you do this/use your ideas from previously.
Caveat that I've only tested the import on a few simple datasets and this is a work in progress but thought it would be useful to share now.

asmit3:
@jbeilstenedmands that sounds good. I will wait for the pull request and then put in the timestamp specific implementation within the framework you will have.

graeme-winter:
> @graeme-winter @ndevenish I am a little confused now. Are you proposing (a) DIALS assigns an initial identifier by default and then we can change it within stills_process _or_ (b) We can assign a custom identifier (like timestamp in this case) from the very beginning (i.e import) using some kind of flag which will be False by default. I would of course prefer (b) as it gives us flexibility.

@asmit3 `dials.stills_process` does not use the command-line `dials.import` yes? So the command-line one can use a UUID as the default (with or without scope for override, a different discussion perhaps) and stills process when it performs the logical import _step_ can assign whatever it likes as the id, yes?

asmit3:
@graeme-winter The confusion has been cleared now. Will wait to see how this work develops. Cheers.

Issue #812, opened 2019-06-10
Cleanup command_line directory


rjgildea:
1) Contain excessive amounts of code
    - ~28.2k lines of code (out of a total of 91.4k lines of code for the whole project)
    - where appropriate, code should ideally be moved out of the command_line directory

2) Almost half (57 out of 129) of the programs are hidden behind a "dev.dials" dispatcher
    - unlikely to be used (or tested) externally, e.g. via xia2
    - CCP4 do not distribute dev.dials.* dispatchers anyway
    - most likely one-off test scripts that would be better off moved to another repository, e.g. dials_scratch

3) Pollutes the code coverage statistics:
    - https://codecov.io/gh/dials/dials/tree/master/command_line
    - 72 (out of 129) files with 0% test coverage
    - Overall test code coverage for this directory of 37.27% (compared to 62.89% for the whole project)
    - 10,009 (out of 17,108) lines of untested code
    - lack of tests means the scripts are likely to become stale over time as a result of major refactors such as the recent datablock -> experiments move, pickle -> msgpack, or e.g. the recent indexing refactor.

4) Pollutes the LGTM alerts:
    - https://lgtm.com/projects/g/dials/dials/latest/files/?sort=name&dir=ASC&mode=heatmap
    - 92 alerts for command_line directory (out of a total of 227 for the whole project)
    - risks masking genuine errors in code that is actually used in DIALS

dagewa:
@phyy-nx any objections if I remove `stills_detector_hybrid_refine.py`?

rjgildea:
> stills_detector_hybrid_refine

This does at least have a test - with 85% code coverage!

phyy-nx:
Heh, indeed.  I'm not using it at all though.  Everything goes through
dials.refine.

dagewa:
@jbeilstenedmands  @jmp1985 Should `dials.compute_delta_cchalf` be a method of `dials.filter_reflections` (which in turn might be renamed `dials.filter`) rather than a separate program?

jbeilstenedmands:
@dagewa I'm split on this one. While practically these are having a similar outcome i.e. filtering reflections, compute_delta_cchalf is a whole algorithm in itself rather than simple filtering and so perhaps belongs as a distinct program. Although I don't like the name, and
```dials.filter method=deltacchalf``` would be quite nice. Hmmm.

dagewa:
[...]

PR #822, opened 2019-06-19
Rename output files

dagewa:
This branch alters output filenames to meet the proposals made in #784 and #785.

DIALS tests pass, but corresponding changes would be required in xia2, DUI, cctbx.xfel, and possibly IOTA and PRIME. In addition, parts of dials_scratch and dials_research could be updated for consistency. I'm asking for review now to check everyone is comfortable with this interface for DIALS 2.0. There were not many comments on #785, so this is a second chance to raise any objections. Assuming all are happy, change branches should be made in the affected projects to be merged alongside this. Given the large number of files this touches, it would be preferable if this is done reasonably soon to avoid merge conflicts.

@phyy-nx I noticed that `dials.stills_process` controls its own output filenames. I have not touched these here beyond changing the filename extensions, so they are still in the longer form (`integrated_experiments_filename = %s_integrated_experiments.expt`). If you prefer to shorten these please feel free to push to the branch.

Issue #162, opened 2016-06-17
Refine: interval_width_degrees=Auto would be useful


graeme-winter:
i.e. to avoid 

```
Sorry: Too few reflections to parameterise Crystal1 unit cell
Try modifying refinement.parameterisation.auto_reduction options
```

for data with small unit cell

this case setting = 60 worked

alt: explain #parameters used so we can estimate externally right #intervals to use

dagewa:
My intention is to solve this and #158 at the same time by performing an analysis of the input residuals. With a measure of what frequency of change is required to sufficiently model real changes, then the degree of smoothing can be set appropriately, as well as the block width over which reflections can be assumed to come from the same model for outlier rejection.

graeme-winter:
Sounds like a nice solution :+1:

rjgildea:
@dagewa issue #158 appears to have been resolved, but this is still open - is there now an interval_width_degrees=Auto option?

dagewa:
almost... the code is there to do the calculation, but I have not been able to test it on lots of datasets to make sure it gives sensible results

dagewa:
Example starting from data created by xia2_regression test `xia2_regression.X4_wide.dials`:

```
dials.refine  '9_refined_experiments.json' '9_refined.pickle' 'scan_varying=True' 'close_to_spindle_cutoff=0.020000' 'outlier.algorithm=auto' 'output.experiments=junk.json' 'output.reflections=junk.pickle' 'reflections_per_degree=100' 'unit_cell.smoother.interval_width_degrees=Auto' 'orientation.smoother.interval_width_degrees=Auto' debug_centroid_analysis=True
dev.dials.plot_centroid_analysis centroid_analysis_reflections.pickle
```
**x-residual analysis**
![x-residual-analysis](https://user-images.githubusercontent.com/11502083/27386357-7a761050-568d-11e7-8ee5-2424c2c73fd6.png)
**y-residual analysis**
![y-residual-analysis](https://user-images.githubusercontent.com/11502083/27386361-7fa9ef60-568d-11e7-8d69-3a8cb1d65802.png)
**phi-residual analysis**
![phi-residual-analysis](https://user-images.githubusercontent.com/11502083/27386368-83c323a0-568d-11e7-9d84-f85577569f88.png)

In this case, the narrowest interval width is determined for the x-residual analysis, so this is the value that will be set for scan-varying refinement. Red lines are at frequencies corresponding to 54, 36 and 18 degrees.

[...]

PR #824, opened 2019-06-20
Cleanup dials/command_line directory

rjgildea:
This pull requests removes (most) untested dials/command_line programs with a `dev.dials.*` dispatcher name set. A parallel pull request will be made in the dials_scratch repository to move these programs to dials_scratch/command_line.

In my opinion if the code is worth having in the main dials repository, then it should:

1. have a test
2. probably not hide behind a `dev.dials.*` (these aren't distributed in the CCP4 version of dials anyway)

See also https://github.com/dials/dials/issues/812 for further discussion of reasons behind this (points from issue duplicated below):
1) Contain excessive amounts of code
    - ~28.2k lines of code (out of a total of 91.4k lines of code for the whole project)
    - where appropriate, code should ideally be moved out of the command_line directory

2) Almost half (57 out of 129) of the programs are hidden behind a "dev.dials" dispatcher
    - unlikely to be used (or tested) externally, e.g. via xia2
    - CCP4 do not distribute dev.dials.* dispatchers anyway
    - most likely one-off test scripts that would be better off moved to another repository, e.g. dials_scratch

3) Pollutes the code coverage statistics:
    - https://codecov.io/gh/dials/dials/tree/master/command_line
    - 72 (out of 129) files with 0% test coverage
    - Overall test code coverage for this directory of 37.27% (compared to 62.89% for the whole project)
    - 10,009 (out of 17,108) lines of untested code
    - lack of tests means the scripts are likely to become stale over time as a result of major refactors such as the recent datablock -> experiments move, pickle -> msgpack, or e.g. the recent indexing refactor.

4) Pollutes the LGTM alerts:
    - https://lgtm.com/projects/g/dials/dials/latest/files/?sort=name&dir=ASC&mode=heatmap
    - 92 alerts for command_line directory (out of a total of 227 for the whole project)
    - risks masking genuine errors in code that is actually used in DIALS


Anthchirp:
on a lighter note I would suggest removing the files as @DiamondLightSource-build-server so you don't mess up your personal project commit line contribution statistics :wink:

rjgildea:
> ### dev.dials.generate_bias_lookup_table
> is implied to be used to generate source code in
> https://github.com/dials/dials/blob/fbbdbe7fdb5c28874cd0fecd4314ee71fa137ea5/algorithms/image/centroid/bias.h#L22-L23

With the parallel PR in https://github.com/dials/dials_scratch/pull/3 the command dev.dials.generate_bias_lookup_table would still exist if you have dials_scratch configured. Alternatively, if the only purpose of command this is to generate some numbers to copy into some source code, I'm not sure it really warrants a command line dispatcher. Perhaps we could move the code into dials/algorithms/image/centroid i.e. next to the code where the output is used in bias.h?

rjgildea:
> ### dev.dials.plot_centroid_analysis
> is referred to in this phil parameter
> https://github.com/dials/dials/blob/fbbdbe7fdb5c28874cd0fecd4314ee71fa137ea5/algorithms/refinement/parameterisation/configure.py#L109-L115

@dagewa will likely have an opinion here - is this command needed/useful? If so, would it be possible to add a test and move out of the `dev.dials.*` namespace?

rjgildea:
> ### dev.dials.merge_reflection_lists
> has test in
> https://github.com/dials/dials/blob/fbbdbe7fdb5c28874cd0fecd4314ee71fa137ea5/test/command_line/test_merge_reflection_lists.py#L30-L32
> 
> 
> which should either be also removed or the command retained. The same is true for...
> ### dev.dials.sort_reflections
> which has tests in
> 
> https://github.com/dials/dials/blob/fbbdbe7fdb5c28874cd0fecd4314ee71fa137ea5/test/command_line/test_sort_reflections.py#L7-L9
> 
This pull request purposely hasn't touched these files as they appear to have tests. It is probably a separate discussion as to whether these are worth keeping or moving out of the `dev.dials.*` namespace.

dagewa:
> > ### dev.dials.plot_centroid_analysis
> > is referred to in this phil parameter
> > https://github.com/dials/dials/blob/fbbdbe7fdb5c28874cd0fecd4314ee71fa137ea5/algorithms/refinement/parameterisation/configure.py#L109-L115
> 
> @dagewa will likely have an opinion here - is this command needed/useful? If so, would it be possible to add a test and move out of the `dev.dials.*` namespace?

I think this command is really a development tool for #162, on which my work stalled. It'd be fine to move to dials_scratch and remove that bit in the dials.refine phil help string

Anthchirp:
re **dev.dials.generate_bias_lookup_table**:
if the code generates code for the main repository then I think it should also live in the main repository, maybe in a `/contrib/` directory as it does not have to be packaged in the future. Agree that it does not need a dispatcher.

re the last 2: my mistake, all good.

rjgildea:
> re **dev.dials.generate_bias_lookup_table**:
> if the code generates code for the main repository then I think it should also live in the main repository, maybe in a `/contrib/` directory as it does not have to be packaged in the future. Agree that it does not need a dispatcher.

See 9ebcecf8e40ec12a6dae0f1cb6867aaa3d871e8e - it made (some) sense to me to place this code next to the place where its output is used.

Anthchirp:
FYI this conflicts with #822 and #822 should have priority. Similarly stuff should only be copied over to dials_scratch after #822 is merged

rjgildea:
I have synched the output file renaming changes from dials/dials#822 with the files moved over to dials_scratch in https://github.com/dials/dials_scratch/pull/3/commits/36351de4ef509e68d26d4509025fac71dc739b64. Any further objections before I go ahead and (squash 'n') merge?

dagewa:
None from me

Issue #215, opened 2016-09-21
dials.check_indexing_symmetry needs sanity checks


rjgildea:
Needs sanity checks if nref is too small to give any sensible answer:

```
nproc=`libtbx.show_number_of_processors`
dials.import $(libtbx.find_under_build xia2_regression)/test_data/i04_bag_training/th_8_2_*.cbf
dials.find_spots datablock.json nproc=$nproc scan_range=1,5 scan_range=268,272 scan_range=536,540
dials.index datablock.json strong.pickle
dials.check_indexing_symmetry experiments.json indexed.pickle grid_search_scope=1 symop_threshold=0.7
```

output:

```
Checking HKL origin:

dH dK dL   Nref    CC
-1  0 -1     20 0.769
-1  1  0     20 0.863
 0  0  0     41 -0.052
 0  1 -1     14 0.719
 1  0  0     20 0.898
 1  0  1     18 0.765
```

dagewa:
With the merge of #825, the SE is printed alongside the CCs if the SE is significant at the level of the number of digits reported. This makes it possible for the user to assess the relevance of results from the grid search in `test_P1_crystal_indexing`

In addition, within `test_crystal_pointgroup_symmetry`, the acceptable element indicator `***` is only added to rows where the CC is greater than the `symop_threshold` at the 95% confidence level.

rjgildea:
@dagewa perhaps I'm missing something, but it isn't clear to me how this helps for the specific case above. This output still suggests that a reindexing of +0,-1,+1 or +0,+1,-1 is ok, even though the initial indexing (i.e. 0,0,0) is correct in this case? I.e. a CC of 0.903 +/- 0.012 and 0.901 +/- 0.010?
```
Checking HKL origin:

dH dK dL   Nref    CC
-1 -1 -1     28 0.07(4)
-1 -1  0     18 0.01(6)
-1  0 -1     12 0.04(10)
-1  0  0     18 0.77(3)
-1  0  1     28 0.17(4)
-1  1 -1     24 0.14(4)
-1  1  1     20 0.14(5)
 0 -1 -1     18 0.25(6)
 0 -1  0     20 0.80(2)
 0 -1  1     18 0.903(12)
 0  0  0     20 0.28(5)
 0  1 -1     20 0.901(10)
 0  1  1     28 0.05(4)
 1 -1 -1     24 0.57(3)
 1  0 -1     14 0.50(6)
 1  0  0     10 0.83(4)
[...]

PR #825, opened 2019-06-20
Report standard error of correlation coefficients.

dagewa:
This is intended to address #215. Standard errors are printed
alongside the correlation coefficients. These can be large when
the number of reflections is small. To enact as a sanity check,
the correlation coefficient itself is not compared with
symop_threshold, but a lower bound given by (CC - SE) is
calculated first then this is compared with symop_threshold.

### DIFF
--- a/command_line/check_indexing_symmetry.py
+++ b/command_line/check_indexing_symmetry.py
@@ -11,6 +11,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import math
 import logging
 import sys
 
@@ -123,12 +124,16 @@ def standard_error_of_pearson_cc(corr_coeffs, n):
 
     numerator = 1 - flex.pow2(corr_coeffs)
     denominator = n - 2
-    if min(denominator) < 1:
-        raise ValueError("Too few reflections to calculate a standard error")
-    return numerator / denominator.as_double()
+    sel = denominator < 1
+    denominator.set_selected(sel, 1)
+    vals = numerator / denominator.as_double()
+    vals.set_selected(sel, float("NaN"))
+    return vals
 
 
 def format_cc_with_standard_error(cc, se, decimal_places=3):
+    if math.isnan(se):
+        return "?"
     minimum = pow(0.1, decimal_places)
     if se >= minimum:
         cc_str = format_float_with_standard_uncertainty(cc, se, minimum=1.0e-3)
--- a/command_line/check_indexing_symmetry.py
+++ b/command_line/check_indexing_symmetry.py
@@ -28,6 +28,7 @@ from dials.util.options import OptionParser
 from dials.util.options import flatten_reflections, flatten_experiments
 from dials.util import log
 from dials.util.version import dials_version
+from libtbx.utils import format_float_with_standard_uncertainty
 
 logger = logging.getLogger("dials.command_line.check_indexing_symmetry")
 
@@ -128,6 +129,15 @@ def standard_error_of_pearson_cc(corr_coeffs, n):
     return numerator / denominator.as_double()
 
 
+def format_cc_with_standard_error(cc, se, decimal_places=3):
+    minimum = pow(0.1, decimal_places)
+    if se >= minimum:
+        cc_str = format_float_with_standard_uncertainty(cc, se, minimum=1.0e-3)
+    else:
+        cc_str = "{val:0.{dp}f}".format(val=cc, dp=decimal_places)
+    return cc_str
+
+
 def get_symop_correlation_coefficients(miller_array, use_binning=False):
     corr_coeffs = flex.double()
     n_refs = flex.int()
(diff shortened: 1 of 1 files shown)


graeme-winter:
@dagewa struggling to get this in the output - from what I can tell the output of the code in the branch is identical to the code in master?

graeme-winter:
Oh wait, comparing not cc but cc - 2 * se => will not see difference in my case... would make sense to output the SE if computed?

dagewa:
It does report the standard error, but only when that value is greater than 3 d.p. (which is what the value is rounded to anyway on report). Otherwise you get very ugly output where the standard error is close to zero. I'll try to find a good test case to show output...

dagewa:
Following the original test case in #215 on this branch:

```bash
dials.import $(dials.data get -q thaumatin_i04)/*.cbf
dials.find_spots imported_experiments.json \
  nproc=$(libtbx.show_number_of_processors) \
  scan_range=1,5 scan_range=268,272 scan_range=536,540
dials.index imported_experiments.json strong.pickle 
dials.check_indexing_symmetry indexed_experiments.json indexed.pickle grid=1
```

gives this table in the output:

```
dH dK dL   Nref    CC
-1 -1 -1     28 0.07(4)
-1 -1  0     18 0.01(6)
-1  0 -1     12 0.04(10)
-1  0  0     18 0.77(3)
-1  0  1     28 0.17(4)
-1  1 -1     24 0.14(4)
-1  1  1     20 0.14(5)
 0 -1 -1     18 0.25(6)
 0 -1  0     20 0.80(2)
 0 -1  1     18 0.903(12)
 0  0  0     20 0.28(5)
 0  1 -1     20 0.901(10)
 0  1  1     28 0.05(4)
 1 -1 -1     24 0.57(3)
 1  0 -1     14 0.50(6)
 1  0  0     10 0.83(4)
 1  0  1     18 0.39(5)
 1  1  0     20 0.79(2)
 1  1  1     14 0.27(8)
```
so the standard error is shown if it is significant at the level of reporting for the CC.

dagewa:
However, I have noticed that if there are too few reflections to calculate a standard error the program now fails with this error:
```
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/check_indexing_symmetry.py", line 367, in <module>
    run(sys.argv[1:])
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/check_indexing_symmetry.py", line 361, in run
    test_P1_crystal_indexing(reflections, experiment, params)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/check_indexing_symmetry.py", line 306, in test_P1_crystal_indexing
    ses = standard_error_of_pearson_cc(ccs, n_refs)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/check_indexing_symmetry.py", line 127, in standard_error_of_pearson_cc
    raise ValueError("Too few reflections to calculate a standard error")
ValueError: Too few reflections to calculate a standard error
```
This occurs because now the logic is to compare against `symop_threshold` at 95% confidence that the CC is above that value, but of course that can't be done when there are so few reflections that the SE can't be calculated. Rather than fail outright, I suppose it would be better to report that the particular CC cannot be calculated reliably.

dagewa:
With that last change, an example that previously crashed with `ValueError` now gives similar output to master:
```
dials.import $(dials.data get -q centroid_test_data)/*.cbf
dials.find_spots imported_experiments.json
dials.index strong.pickle imported_experiments.json
dials.check_indexing_symmetry indexed.pickle indexed_experiments.json grid=1
```
on this branch:
```
dH dK dL   Nref    CC
 0  0  0      0 0.000
 0  0  1     12 0.28(9)
 0  1  1     10 0.09(12)
```
on master:
```
dH dK dL   Nref    CC
 0  0  0      0 0.000
 0  0  1     12 0.284
 0  1  1     10 0.087
```

graeme-winter:
With that in mind, happy to merge

dagewa:
Thanks :-) I will squash and merge then

graeme-winter:
👍 thank you

rjgildea:
> dH dK dL   Nref    CC
> -1 -1 -1     28 0.07(4)
> -1 -1  0     18 0.01(6)
> -1  0 -1     12 0.04(10)
> -1  0  0     18 0.77(3)
> -1  0  1     28 0.17(4)
> -1  1 -1     24 0.14(4)
> -1  1  1     20 0.14(5)
>  0 -1 -1     18 0.25(6)
>  0 -1  0     20 0.80(2)
>  0 -1  1     18 0.903(12)
>  0  0  0     20 0.28(5)
>  0  1 -1     20 0.901(10)
>  0  1  1     28 0.05(4)
>  1 -1 -1     24 0.57(3)
>  1  0 -1     14 0.50(6)
>  1  0  0     10 0.83(4)
>  1  0  1     18 0.39(5)
>  1  1  0     20 0.79(2)
>  1  1  1     14 0.27(8)

Perhaps I'm missing something, but it isn't clear to me how this helps for the specific issue in #215.  This output still suggests that a reindexing of +0,-1,+1 or +0,+1,-1 is ok, even though the initial indexing (i.e. 0,0,0) is correct in this case?

PR #828, opened 2019-06-28
Generate output file names using a template string

Anthchirp:
Instead of having many options to describe the output file naming template simply expose a generic templating mechanism instead. This mechanism supports [every format string feature](https://pyformat.info).

Call examples with a set of 12 combined experiments:

`dials.split_experiments "output.template=\"some_{prefix}_{index:0{maxindexlength}d}_suffix.{extension}\""` results in
```
  some_experiments_01_suffix.json
  ..
  some_reflections_12_suffix.pickle
```

`dials.split_experiments "output.template=\"fixed_{index:05d}.{extension}\""` results in
```
  fixed_00000.json
  ..
  fixed_00012.pickle
```


RBolotovsky:
Hi Markus,

I do like the idea of a general template.  I would suggest:
1. Set .expert_level = 2
2. Have the help string explain the _default_ values for "prefix" (modifiable through the legacy parameters output.experiments_prefix and output_reflections_prefix?) and "extension" (non-modifiable "json" and "pickle" strings?).
3. What are the minimum bits that must be present in the template? For example, what happens if the user creates a template with no {extension}? The python script raises a RunTime error, right?

Cheers,
Robert

Anthchirp:
Hi,

1. makes sense
2. makes sense, will look into this during working hours
3. the code generating the files currently requires `.{extension}` to be at the end of the template, otherwise it will get confused in what format it should create the output and raise something `RuntimeError`ish. Apart from that the template string is unconstrained, therefore the minimum valid template string at this time is `.{extension}`

Anthchirp:
assuming the tests pass I will merge this shortly then

PR #838, opened 2019-07-04
Add a CONTRIBUTING/standard/conventions document

ndevenish:
This is a draft document to outline basic requirements for contributing; While writing it the scope expanded somewhat - it started as a basic conventions document but @Anthchirp pointed out that CONTRIBUTING is probably a more natural and common place to put those; at the moment the aim is to:
- Make it clear that we welcome contributions
- Give a very basic guide to getting started
- Outline the essential and preferred code standards/conventions

I've tried to write conventions with explanations rather than just trying to be prescriptive; I've also tried to avoid veering into general coding discussion, although obviously some of these are targeted at common anti-patterns we've been uncomfortable with for a while.

In the spirit of the document, I welcome open discussion on this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,7 +70,7 @@ message that the code in question is special and care should be taken.
   setting the name to None). Matplotlib is an exception to this guideline -
   because its startup logic defines the backend it uses, this can be imported
   inline. There are also a few exceptions to help avoid circular imports that
-  are hard to remove. Further, unique imports are encouraged within code branches
+  are hard to remove. Further, unique imports are allowed within code branches
   or within optional functions/classes that are not always used when the file is
   imported.  This reduces the runtime import load for functionality that isn't
   universally used.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ message that the code in question is special and care should be taken.
 - Try not to do `from <module> import *` imports - it makes it hard to trace
   where definitions are coming from, and turns off many useful diagnostics in
   static analysis tools. Exceptions are allowed for modules that purely import
-  from an extension to wrap the interface, if it would be excessively verbose.
+  from an extension to wrap the interface, if it would be excessively verbose otherwise.
 - Make pull requests a clean representation of the implementation of a feature
   — meandering commit history is okay for WIP branches if it's supposed to be
   squashed down to one commit eventually, but development dead-ends and back
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,142 @@
+# Contributing to Dials
+
+We're happy to consider contributions from outside sources; whether in the form
+of code, tickets, documentation or even just typo corrections! We're aware that
+Dials can at times appear to have relatively indiosyncratic code, and some of
+this is supporting obscure/niche use cases - so, if in the event that you are
+planning any large scale work that you will want to merge back in, please
+contact us beforehand so that we can discuss any potential impacts this might
+have.
+
+Listed in this document are code standards and conventions that you should try
+to adhere to, some of which are essential and others that are just encouraged.
+The intention is that all of the code should try to converge towards these.
+
+## Configuring Dials for development
+
+1.  Install Dials and it's cctbx_project dependencies for development; For
+    Linux/macOS, the current best way to create a fresh installation of Dials
+    and all of it's dependencies is with the following commands:
+    ```
+    git clone https://github.com/cctbx/cctbx_project modules/cctbx_project
+    ln -s modules/cctbx_project/libtbx/auto_build/bootstrap.py
+    python bootstrap.py hot base update build --builder=dials --use-conda
+    ```
+2.  Activate the environment with `source <root>/build/setpaths.sh`. This will
+    need to be done every time you work on dials code.
+3.  Install pre-commit hooks with `libtbx.precommit install`. This will install
+    hooks for Dials, dxtbx and xia2.
+
+The Dials repository is now checked out in `<root>/modules/dials`. During
+development, run tests with pytest with `libtbx.pytest --regression` to ensure
+that all of the tests still pass.
+
+If you update or change dependencies, you may occasionally need to regenerate
+the static libtbx ecosystem by running `libtbx.refresh` - this will also
+generate an alias to any executable scripts installed by `pip` as
+`libtbx.<script>`. Using these helps ensure that you are using the bundled
+python version instead of your system version.
+
+## Essential Code Development Standards
+
+- **Please install the pre-commit hooks**. (`libtbx.precommit install` if using
+  the libtbx ecosystem). These use the [pre-commit] library and ensures that
+  various sanity checks are run before commit, including formatting, syntax
+  compatibility, basic flake8 checks, lack of conflict markers and file size
+  limits. Basically, most of the essential rules will be checked automatically
+  by this.
+- **We format python code with [black]**. This means that while writing code
+  you don't have to worry about laying things out neatly, because black will
+  take care of the formatting. You should try to commit code formatted with
+  black (the pre-commit hook will help do this for you), but if for some reason
+  you can't, the whole codebase is auto-cleaned once a week. Most IDE's and
+  editors have support for running formatters like black frequently or
+  automatically.
+- **Avoid introducing new pre-commit flake8 warnings** - if you feel that it's
+  appropriate to violate a warning, mark it up explicitly with a [noqa]
+  comment. Probably the most common cause of this are "F401 - module imported
+  or unused", which happens when importing packages to collect into a single
+  namespace for other imports (though declaring `__all__` avoids this issue).
+  The pre-commit hooks will pick up the most important of these, but please try
+  to resolve any other valid warnings shown with a normal run of flake8. The
+  configuration in the repository turns off any that disagree with black's
+  interpretation of the rules or standard practice in our repositories.
+- **Code should be written with python 3 syntax** - This means "new" exception
+  syntax, print functions, new-style classes, and a smattering of other changes
+  — the pre-commit hooks check for this.
+- **Commit messages should be descriptive**; See [How to Write a Git Commit
+  Message] for a long explanation of how and why this matters, or just skip to
+  [The Seven Rules] for general guidelines (and they are guidelines for us, not
+  rules per se). The first line should be a concise summary, ideally not over
+  50 characters but never over 72. Follow this first line with an empty line.
+  Remember that someone may be looking at your commit in several years time,
+  trying to work out the reason for your commit and wondering what on earth you
+  were thinking. That someone may be you.
+- **We format C++ code with [clang-format]**. We use a configuration for style
+  broadly compatible with what our existing prevailing style was. We don't
+  require that everyone has clang-format installed - the weekly cleaning job
+  will pick it up if you don't - but if you do, remember to run with
+  `-style=file` to pick up our configuration.
+
+### Less strict guidelines
+
+These should be followed wherever possible - but remember the [Zen of Python]'s
+"Practicality beats purity.", and [PEP8]'s "A Foolish Consistency is the
+Hobgoblin of Little Minds" - it's okay to stray from the rules if you have a
+good reason for it, but "I prefer it this way" isn't a strong enough reason -
+there is real value to a standard style, because diverging from it sends a
+message that the code in question is special and care should be taken.
+
+- **Try to write code compatible with python 2 and 3** - we have both
+  [six] and [future] available for help with this, but try to use future to
+  write idiomatic python 3 wherever possible. If you are unsure how to do this,
+  ask for help. Once we are fully python 3 compatible, this will become a
+  requirement, but for now we don't expect most people to be developing on top
+  of python 3. If you get used to it now, it'll be less of a shock when we
+  require and test it.
+- Err on the side of [PEP8] when making any style decision not covered by
+  black. In particular, **use PEP8 as a guide for naming** when you aren't sure
+  the correct form to use.
+- **All imports should go at the top of the file** - not inline in a function.
+  This makes it very easy to reason about the dependencies of a module, makes
+  it hard to make conflicting definitions, and avoids duplicate imports
+  scattered throughout the file (one file used to have over 30 separate imports
+  of the `sys` package). If an import is for an optional dependency, then it
+  should be imported in a try/except block, with some fallback to identify the
+  missing case (such as setting the name to `None`. **Matplotlib is an
+  exception** to this rule - because it's startup logic defines the backend it
+  uses, this is permitted to be imported inline. There are still a few
+  exceptions to help avoid circular imports that are hard to remove. Anything
+  that is a python standard library import should be at the top of the file,
+  without question. However, _you don't need to fix this for existing code_.
+- **Don't create classes with one function**. If your class is an `__init__`
+  and a single 'do' function (or a `__call__`), then it can probably be more
+  concisely expressed as a function.
+- Avoid classes that work by `__call__` unless you have a good reason for it to
+  act as a functor. A named function to do the action instead is almost always
+  clearer with a proper name.
+- **Write docstrings**. We have a mix of styles at the moment, but new
+  docstrings should try to follow [Google-style] - it has a good balance
+  between clarity and length.
+- Try not to do `from <module> import *` imports - it makes it hard to trace
+  where definitions are coming from, and turns off may useful diagnostics in
+  static analysis tools. Exceptions are allowed for modules that purely import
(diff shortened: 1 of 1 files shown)


dagewa:
Personally, I much prefer DIALS to Dials (:bike: :derelict_house: )

ndevenish:
> Personally, I much prefer DIALS to Dials

I flip-flopped on this and never felt really happy with either 'Dials' or 'dials', and was trying to avoid shouting something repeatedly, but you are probably right

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+contact us beforehand so that we can discuss any potential impacts this might
+have.
+
+Listed in this document are code standards and conventions that you should try
+to adhere to, some of which are essential and others that are just encouraged.
+The intention is that all of the code should try to converge towards these.
+
+## Configuring Dials for development
+
+1.  Install Dials and it's cctbx_project dependencies for development; For
```
rjgildea:
```suggestion
1.  Install Dials and its cctbx_project dependencies for development; For
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+    Linux/macOS, the current best way to create a fresh installation of Dials
+    and all of it's dependencies is with the following commands:
```
rjgildea:
```suggestion
    and all of its dependencies is with the following commands:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+If you update or change dependencies, you may occasionally need to regenerate
+the static libtbx ecosystem by running `libtbx.refresh` - this will also
+generate an alias to any executable scripts installed by `pip` as
+`libtbx.<script>`. Using these helps ensure that you are using the bundled
+python version instead of your system version.
+
+## Essential Code Development Standards
+
+- **Please install the pre-commit hooks**. (`libtbx.precommit install` if using
+  the libtbx ecosystem). These use the [pre-commit] library and ensures that
```
rjgildea:
```suggestion
  the libtbx ecosystem). These use the [pre-commit] library and ensure that
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+
+If you update or change dependencies, you may occasionally need to regenerate
+the static libtbx ecosystem by running `libtbx.refresh` - this will also
+generate an alias to any executable scripts installed by `pip` as
+`libtbx.<script>`. Using these helps ensure that you are using the bundled
+python version instead of your system version.
+
+## Essential Code Development Standards
+
+- **Please install the pre-commit hooks**. (`libtbx.precommit install` if using
```
jmp1985: Does this really need to be essential? I have found pre-commit to be very flaky on my machine. I.e. it has never worked for me with the DIALS repository.

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
+# Contributing to Dials
+
+We're happy to consider contributions from outside sources; whether in the form
+of code, tickets, documentation or even just typo corrections! We're aware that
+Dials can at times appear to have relatively indiosyncratic code, and some of
```
dagewa: This comes across too negative in my opinion. Rather than "idiosyncratic" and "obscure/niche" I would just say that "DIALS addresses a wide range of use cases - so, if in the event..."

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  the libtbx ecosystem). These use the [pre-commit] library and ensures that
+  various sanity checks are run before commit, including formatting, syntax
+  compatibility, basic flake8 checks, lack of conflict markers and file size
+  limits. Basically, most of the essential rules will be checked automatically
+  by this.
+- **We format python code with [black]**. This means that while writing code
+  you don't have to worry about laying things out neatly, because black will
+  take care of the formatting. You should try to commit code formatted with
+  black (the pre-commit hook will help do this for you), but if for some reason
+  you can't, the whole codebase is auto-cleaned once a week. Most IDE's and
```
dagewa:
```suggestion
  you can't, the whole codebase is auto-cleaned once a week. Most IDEs and
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  black. In particular, **use PEP8 as a guide for naming** when you aren't sure
+  the correct form to use.
+- **All imports should go at the top of the file** - not inline in a function.
+  This makes it very easy to reason about the dependencies of a module, makes
+  it hard to make conflicting definitions, and avoids duplicate imports
+  scattered throughout the file (one file used to have over 30 separate imports
+  of the `sys` package). If an import is for an optional dependency, then it
+  should be imported in a try/except block, with some fallback to identify the
+  missing case (such as setting the name to `None`. **Matplotlib is an
+  exception** to this rule - because it's startup logic defines the backend it
```
dagewa:
```suggestion
  exception** to this rule - because its startup logic defines the backend it
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  and a single 'do' function (or a `__call__`), then it can probably be more
+  concisely expressed as a function.
+- Avoid classes that work by `__call__` unless you have a good reason for it to
+  act as a functor. A named function to do the action instead is almost always
+  clearer with a proper name.
+- **Write docstrings**. We have a mix of styles at the moment, but new
+  docstrings should try to follow [Google-style] - it has a good balance
+  between clarity and length.
+- Try not to do `from <module> import *` imports - it makes it hard to trace
+  where definitions are coming from, and turns off may useful diagnostics in
```
dagewa:
```suggestion
  where definitions are coming from, and turns off many useful diagnostics in
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  static analysis tools. Exceptions are allowed for modules that purely import
+  from an extension to wrap the interface, if it would be excessively verbose.
```
dagewa: If what would be excessively verbose?

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  comment. Probably the most common cause of this are "F401 - module imported
+  or unused", which happens when importing packages to collect into a single
+  namespace for other imports (though declaring `__all__` avoids this issue).
+  The pre-commit hooks will pick up the most important of these, but please try
+  to resolve any other valid warnings shown with a normal run of flake8. The
+  configuration in the repository turns off any that disagree with black's
+  interpretation of the rules or standard practice in our repositories.
+- **Code should be written with python 3 syntax** - This means "new" exception
+  syntax, print functions, new-style classes, and a smattering of other changes
+  — the pre-commit hooks check for this.
```
Anthchirp: I don't think so.

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+  without question. However, _you don't need to fix this for existing code_.
+- **Don't create classes with one function**. If your class is an `__init__`
+  and a single 'do' function (or a `__call__`), then it can probably be more
+  concisely expressed as a function.
+- Avoid classes that work by `__call__` unless you have a good reason for it to
+  act as a functor. A named function to do the action instead is almost always
+  clearer with a proper name.
+- **Write docstrings**. We have a mix of styles at the moment, but new
+  docstrings should try to follow [Google-style] - it has a good balance
+  between clarity and length.
```
Anthchirp: After reading 'clean code' I am now less convinced this is a good thing than I was before.

CONTRIBUTING.md
Anthchirp:
```suggestion
  from an extension to wrap the interface, if it would be excessively verbose otherwise.
```

CONTRIBUTING.md
ndevenish: I thought we used to have explicit AST checks, and assumed that the lack of "python3" interpreter explicitly was a reflection of the fact that all our hooks are run on python 3 now?

Also, black fails for most of these anyway, I think...?

CONTRIBUTING.md
ndevenish: Docstrings in general, or this style in particular? It's probably google/numpy/sphinx, and sphinx is almost unreadable for in-line code (where you read most of them). Somewhat less useful if using type annotations but there's still many complex and interestingly-parametered functions out there

Anthchirp: docstrings in general. ie. no docstring is better than
```python
def wibblyfy(sprocket):
  """a function to wibblyfy a sprocket

  Args:
    sprocket (object): a sprocket

  Returns: 
    object: a wibblified sprocket or something else depending on the sprocket wibblification outcome
  """
```

CONTRIBUTING.md
Anthchirp: pre-commit black runs on Python 3 and does assume correct syntax. And we do run check-ast, which may be stricter. But that still doesn't check for new-style classes or right-shifted print functions.

CONTRIBUTING.md
Anthchirp: Yes. (And if it doesn't work then that is a bug that should be reported please.)

ndevenish: We just had this discussion - turns out @jmp1985 had the *old* precommit hook/tool installed. With the new-style it works fine!

CONTRIBUTING.md
ndevenish: Ah, "The pre-commit hooks check for this" was referring to " **Code should be written with python 3 syntax**" - technically right shifted print functions are python 3 syntax :)

CONTRIBUTING.md
ndevenish: I agree, it should be wrapped to column limit.....

In those cases where it truly is entirely obvious and redundant (which probably matches small do-one-thing-well unit testable functions) I'd argue that it would happily come under "Having a good reason to veer from the guidelines".

Also, even though most of the time the code will be read, I still think the goal of having readable html-generated documentation is a good idea, especially since we have been actively working/talking about tidying up the API interfaces - in which case everything exposed as an API should have this anyway (but maybe not needing the full formal structure). Many internal functions probably _could_ get away with the one-line docstring.

(Typing and type annotations is a separate discussion for when we are fully python 3, but I think much of the same argument applies)

Anthchirp: having said that - I do agree with the idea of settling on one format. And since we do not have type annotations [in 99% of cases. I am aware of your leadership role in this, @jbeilstenedmands :)] these comments are useful. I did notice this in my recent dxtbx work and a few more comments would likely have made that job easier. So let's keep it as it is, and reconsider this particular can of worms once we have dropped Python 2 and after we have established a culture of type annotation :)

CONTRIBUTING.md
ndevenish: Also, while maybe ontologicallly confusing, this entry is a "please" whilst the running of black and flake8 is a separate entry in essential and also "try to" because if you really can't/don't want to install precommit/doing an essential fix on an unfamiliar computer, then running them separately is fine, but if you don't, expect your code to be changed by someone else (or the automatic tools) downstream to be compliant.

As another example, if you commit a 100mb file because you didn't have the precommit installed, expect @Anthchirp to frown at you and your commit to be reverted.

ndevenish:
Note to self: Something something towncrier

Anthchirp:
where do tests go. `/test/ or `/code_under_test/`?

Anthchirp:
alternatively:  add a paragraph that these things are not 'rules set in stone' and can themselves be changed by pull request, then add towncrier and discuss test locations in another PR 😏

rjgildea:
> where do tests go. `/test/ or `/code_under_test/`?

My preference would be for tests to be close to the code being tested.

dagewa:
> > where do tests go. `/test/ or `/code_under_test/`?
> 
> My preference would be for tests to be close to the code being tested.

I agree and want to get around doing this for refinement tests.

ndevenish:
I think the traditional argument for having separate test folders are 
- Ease of location for finding
- Easy to strip out when distributing

Pytest makes the former mostly unnecessary (although, faster collection?) and we've never done the latter. The main thing I can think of arguing for a `/test[s]?/` folder is integration tests e.g. things that test on the `command_line` whole-program level.

jmp1985:
I am very much in favour of keeping tests in a separate tests/ directory. I think it keeps things tidier in the main codebase and I think that code and tests are two different things and should be logically separated.

dagewa:
Having tests close to the code makes it easier to think of tests as developer "documentation" rather than being hidden away somewhere else. It also means I can just run pytest in one sub directory to get all the tests relevant to, say, refinement. At the moment this partially possible, because we have a structure under `test/` that mirrors the project - `tests/algorithms/refinement/` etc. However, I didn't complete that structure. So there is no `test/algorithms/refinement/parameterisation/` for example. I'd rather not bother maintaining the structure under test/, and just move the tests close to the code.

If there are many tests that might "clog up" a code directory, I suppose these could be moved to a `test/` subdirectory under the code directory.

Anthchirp:
I find @dagewa's argument that having the tests next to the tested code allows you to run pytest in the same directory convincing. As it stands my opinion is **don't care** leaning towards **next to code**.

Advance warning: this is likely to change in the future to a strong **separate from code** when we start packaging our software properly. This is a while off though.

Anthchirp:
So I've updated https://github.com/cctbx/dxtbx/pull/62 to use `import *` in the case where we a file only imports & exports symbols. Is this what we want?

graeme-winter:
@nksauter as @ndevenish mentioned at the top of this PR "In the spirit of the document, I welcome open discussion on this."

graeme-winter:
@phyy-nx would welcome your opinion on this, when you get a moment to look at it 🙂

Thanks in advance 👍

dagewa:
This is a useful short read: https://opensource.guide/how-to-contribute/#orienting-yourself-to-a-new-project

> While not every project has a CONTRIBUTING file, its presence signals that this is a welcoming project to contribute to

ndevenish:
Excellent - this has been here a month, so probably long enough to merge.

It would be good to get agreement from someone on the other side of the pond before merging; but if there aren't any comments then that's okay too.

CONTRIBUTING.md
```diff
@@ -85,28 +44,36 @@ good reason for it, but "I prefer it this way" isn't a strong enough reason -
 there is real value to a standard style, because diverging from it sends a
 message that the code in question is special and care should be taken.
 
+- **Code should be written with python 3 syntax** - This means "new" exception
```
ndevenish: This is a little confusing now being moved under "Guidelines" because it will cause test failures and build breakage

CONTRIBUTING.md
```diff
@@ -0,0 +1,143 @@
...
+  otherwise.
+- **Common imports should go at the top of the file**.  This makes it very easy
+  to reason about the dependencies of a module, makes it hard to make
+  conflicting definitions, and avoids duplicate imports scattered throughout
+  the file. If an import is for an optional dependency, consider using a
+  try/except block, with some fallback to identify the missing case (such as
+  setting the name to None). Matplotlib is an exception to this guideline -
+  because its startup logic defines the backend it uses, this can be imported
+  inline. There are also a few exceptions to help avoid circular imports that
+  are hard to remove. Further, unique imports are encouraged within code branches
```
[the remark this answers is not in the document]
phyy-nx: Ok

CONTRIBUTING.md
```diff
@@ -0,0 +1,143 @@
...
+  or within optional functions/classes that are not always used when the file is
+  imported.  This reduces the runtime import load for functionality that isn't
+  universally used.
```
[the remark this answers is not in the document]
phyy-nx: I'd like to keep it. Anything to reduce imports on clusters is important.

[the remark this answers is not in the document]
ndevenish: If that's true then zipping the packages is going to buy you a much, much larger effect.

Fiddling with deep import-tree imports isn't going to give you anything - `strace -eopen dials.stills_process` (python2) calls open 8200 times uniquely and `stat` 2800 times - that's without doing _anything_ - an extra package here or there is going to be immeasurable. All you can really buy with import reduction is CLI startup time for fast-feedback, and that' better dealt with closer to the main entry point.

Perhaps more relevantly, the sentence doesn't change the guideline at all.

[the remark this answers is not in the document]
ndevenish: e.g. vaguely `zip -r dials.zip dials && mv dials _dials && export PYTHONPATH=$(pwd)/dials.zip && strace -eopen libtbx.python -mdials.command_line.stills_process` reduces the unique open() calls by about 300 (python2, one measurement)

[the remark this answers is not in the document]
phyy-nx: Yes, and nersc developed shifter to try and solve this problem too.  No one import will solve the problem, but inlining the big, optional ones helps, and describing why helps contributors.

At least that's what I think.

graeme-winter:
Given we have agreement on this from @phyy-nx as well as others, merging.

PR #853, opened 2019-07-17
Add client/server to stills_process

phyy-nx:
Recent experiment at SACLA really highlighted the need for client/server in dials.stills_process so I've ported it from xtc_process into here.  I've made it the only option because it's nearly always better than striping (if n > 2).

### DIFF
--- a/command_line/stills_process.py
+++ b/command_line/stills_process.py
@@ -506,8 +506,10 @@ class Script(object):
 
             log.config(params.verbosity, info=info_path, debug=debug_path)
 
-            if size <= 2: # client/server only makes sense for n>2
-                subset = [item for i, item in enumerate(iterable) if (i + rank) % size == 0]
+            if size <= 2:  # client/server only makes sense for n>2
+                subset = [
+                    item for i, item in enumerate(iterable) if (i + rank) % size == 0
+                ]
                 do_work(rank, subset)
             else:
                 if rank == 0:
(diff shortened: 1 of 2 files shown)


phyy-nx:
(56d3a94 is a tag-along)

Anthchirp:
Could you please run black over it before merging?

command_line/stills_process.py
```diff
@@ -506,8 +506,41 @@ def do_work(i, item_list):
...
+                      item = comm.recv(source=0)
+                      if item == 'endrun':
+                          print("Rank %d recieved endrun"%rank)
+                          break
+                      print("Rank %d beginning processing"%rank)
+                      try:
+                          do_work(rank, [item])
+                      except Exception as e:
+                          print("Rank %d unhandled exception processing event"%rank, str(e))
+                      print("Rank %d event pocessed"%rank)
```
ndevenish: `s/pocessed/processed/`

command_line/stills_process.py
```diff
@@ -506,8 +506,41 @@ def do_work(i, item_list):
...
+                      print("Rank %d waiting for response"%rank)
+                      item = comm.recv(source=0)
+                      if item == 'endrun':
+                          print("Rank %d recieved endrun"%rank)
+                          break
+                      print("Rank %d beginning processing"%rank)
+                      try:
+                          do_work(rank, [item])
+                      except Exception as e:
+                          print("Rank %d unhandled exception processing event"%rank, str(e))
```
ndevenish: Unless there is some MPI subtlety I'm not aware of, Is it really intended behaviour that general errors in workers get ignored without sending a message back to the server to log? Feel like this will just cause remote hanging and debugging frustrations - not to mention silent dropping of tasks.

command_line/stills_process.py
```diff
@@ -506,8 +506,41 @@ def do_work(i, item_list):
...
+                else:
+                    # client process
+                    while True:
+                      # inform the server this process is ready for an event
+                      print("Rank %d getting next task"%rank)
+                      comm.send(rank,dest=0)
+                      print("Rank %d waiting for response"%rank)
+                      item = comm.recv(source=0)
+                      if item == 'endrun':
+                          print("Rank %d recieved endrun"%rank)
```
rjgildea:
```suggestion
                          print("Rank %d received endrun"%rank)
```

command_line/stills_process.py
phyy-nx: This is a great question.  The original design philosophy of stills_process is that the per-event processing function must not ever fail, because if it does it would stop the entire processing job (at least that's what happens in MPI).  The assumption is that if an individual image fails to process that's ok and if large numbers of images are failing to process for a systemic reason, that that would be detected by fast feedback reports such as what is in the XFEL GUI.

Moving the do_work function of stills_process into a try/except for MPI processing better aligns with that philosophy.  Note it's exactly how the non-mpi version works currently since easy_mp.multi_core_run squashes exceptions as well.

ndevenish: An image failing to process feels different from a physical crash in the processing, though I guess the result is the same either way and would eventually be noticed by a user. I was thinking in terms of "User runs as server, clients on cluster connect to it" but it looks more like server/client architecture within a set of cluster jobs.

> Note it's exactly how the non-mpi version works currently since easy_mp.multi_core_run squashes exceptions as well

That's a good point - I forgot this. In that case, it's not at risk of breaking anything that wouldn't already be broken.

PR #855, opened 2019-07-17
Speed up dials.refine_bravais_settings.

### LINKED ISSUE #586: dials.refine_bravais_settings outlier rejection and speed
I was thinking about the issue of speed of `dials.refine_bravais_settings` raised at the LMB meeting yesterday. At the moment, `dials.refine_bravais_settings` performs Tukey outlier rejection for each refinement in a different setting. Although this is reasonably quick, it can take a significant portion of the total runtime of a refinement job.

Thinking about it further, outlier rejection might even be problematic for cases where the metric fit and rmsds are rather high. It may be better to initially select only the reflections that were `used_in_refinement` from indexing, and then not do any further outlier rejection at all for the setting refinement jobs. We could go even further, and completely determine the subset of reflections to use for refinement in advance of any of the refinement jobs. That would mean passing less data around to each `easy_mp` job, and also guarantee that the same reflections are used for refinement for each case.

dagewa:
RBS does two rounds of refinement, changing outlier rejection from
rather weak to normal tukey. This change removes the first round
of outlier rejection entirely, relying on the used_in_refinement
flag from the input instead. This speeds up RBS, and the results
are very similar (as the second round of refinement is unchanged).

Fixes #586


dagewa:
I did some (very limited) testing in #586, suggesting the improvement can be significant.

PR #861, opened 2019-07-19
Python 3: lib2to3.fixes.fix_dict

Anthchirp:
This pull request is based on the `lib2to3.fixes.fix_dict` fixer. In Python 3 dictionary methods `.keys()`, `.values()`, etc. no longer return lists. The automatic fixer just wraps every call in `list()` to restore previous behaviour.


Anthchirp:
Pull request is ready for review. Full test suite passes on Python 2 and no regressions on Python 3.
I left some flake8 and lgtm warnings untouched due to them being more involved (hence ❌ Travis complaining)
Some un-pythonic constructs remain due to #862.

PR #865, opened 2019-07-22
Python 3: libfuturize.fixes.fix_xrange_with_import

Anthchirp:
This pull request is based on the `Libfuturize.fixes.fix_xrange_with_import` fixer. In Python 3 `xrange()` is gone and `range` returns a `range` object, which basically is an iterator. The automatic fixer wraps every `range()` call in `list()` to restore previous behaviour and adds a `from builtins import range` line to each file using `range()` to change its behaviour to that in Python 3.


Anthchirp:
This is ready. Tests pass on Python 2 and no regressions on Python 3.

PR #869, opened 2019-07-24
Python 3: fix dictionary iteration from C++

Anthchirp:
We call `dict.iteritems()` and `.next()` on its return value from within C++.
Both `.iteritems()` and `.next()` will be gone in Python 3.

As far as I can tell a _good_ solution for the 5 instances we do this would be to iterate over the dictionary, then use dictionary key access to get the relevant value.

However I could not figure this out, so I did something dumb instead: I call `six.iteritems()` on the dictionary, and `six.next()` on the return value. This is not optimal for many reasons, least of all the function lookup in the inner loop. However as far as I know the dictionaries in question have a fixed size and are relatively small, so the impact should be limited. This solution does have the benefit of working in both Python 2 and 3.

Maybe someone with better domain knowledge can come up with a better solution?

### DIFF
--- a/array_family/boost_python/flex_reflection_table.cc
+++ b/array_family/boost_python/flex_reflection_table.cc
@@ -909,10 +909,11 @@ namespace dials { namespace af { namespace boost_python {
       // Extract the columns
       dict columns = extract<dict>(state[3]);
       DIALS_ASSERT(len(columns) == ncols);
-      object iterator = boost::python::import("six").attr("iteritems")(columns);
+      object items = list(columns.items());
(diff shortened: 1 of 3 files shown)


Anthchirp:
Discussed this with @jmp1985 and came up with above variant. This is not optimal, as it creates a `list(.items())`, but it works. If we ever figure out how to do a proper iteration then we can do that, until then JMP says this should only copy some ~50 pointers, so the cost should be limited.

Anthchirp:
will merge if/when tests pass

PR #872, opened 2019-07-24
add LoggingContext manager as suggested by @Anthchirp

rjgildea:
Add a LoggingContext manager as described in https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging

### DIFF
--- a/util/log.py
+++ b/util/log.py
@@ -165,24 +165,16 @@ def print_banner(force=False, use_logging=False):
 
 class LoggingContext(object):
     # https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging
-    def __init__(self, logger, level=None, handler=None, close=True):
+    def __init__(self, logger, level=None):
         self.logger = logger
         self.level = level
-        self.handler = handler
-        self.close = close
 
     def __enter__(self):
         if self.level is not None:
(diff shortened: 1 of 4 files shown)


util/log.py
```diff
@@ -161,3 +161,28 @@ def print_banner(force=False, use_logging=False):
         logging.getLogger("dials").info(_banner)
     else:
         print(_banner)
+
+
+class LoggingContext(object):
+    # https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging
+    def __init__(self, logger, level=None, handler=None, close=True):
+        self.logger = logger
```
Anthchirp: 👍 
One question though - we don't tend to keep loggers from other places around, so it probably would make more sense for us to pass the name to the context manager as string and then do
```python
def __init__(self, logger, level=None, handler=None, close=True):
    self.logger = logging.getLogger(logger)
```
which makes the call for other loggers neater.

Alternatively could make it so that
```
self.logger = logging.getLogger(logger) if isinstance(logger, six.string_types) else logger
```
that would keep the call for the file's own logger neat as well.

I don't know if we ever need `handler=` and `close=`, so might consider removing those until we do && have a test for them :)

PR #882, opened 2019-08-13
Refactor to add LatticeSearch class

dagewa:
Extract methods for searching for and selecting the best crystal model
into a LatticeSearch class. Child classes, like BasisVectorSearch
implement the find_candidate_crystal_models method.

With this change it becomes possible to create new indexing methods
that construct crystal models directly rather than finding a set of
basis vectors first. No such algorithms are implemented yet, but it
is envisaged that they could also utilise a strategy pattern.

### DIFF
--- a/algorithms/indexing/lattice_search.py
+++ b/algorithms/indexing/lattice_search.py
@@ -113,62 +113,15 @@ basis_vector_search_phil_scope.adopt_scope(
 )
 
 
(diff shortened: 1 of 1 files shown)


dagewa:
This isn't everything required to get new indexing methods like the ED stills algorithm into dials.index, but it is a start. I wanted to get this rearrangement of the class hierarchy reviewed first to be sure it seems like the right approach.

PR #883, opened 2019-08-16
ignore all forms of refinement in stills indexer

dermen:
This is useful when one wants to use stills
indexer yet only wants the orientation matrices
and does not wish to bother with refinement
of orientation matrix or nave parameters

To add to this, I am currently using code that involves using dials to grab the orientation matrices after spot finding... Its using an modification of real space grid search, where one has multiple beams per crystal (for e.g. two color MAD images) ..  

You can see the basic idea [here](https://github.com/dermen/cxid9114/blob/master/indexing/msi.py)

Currently stills_indexer fails with ```refinement_protocol = repredict_only``` because nave parameter refinement that is used for the predictions does not handle the case of two beams and one crystal (and besides that I don't wish to use the nave parameters predictions even if they were to work). Further, refinement has a high failure rate with the *multiple beams / one crystal* type of experiments.. Therefore I wish to just snag the orientation matrix and get out of the indexer...

### DIFF
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -169,12 +169,13 @@ indexing {
     .type = bool
     .expert_level = 1
   refinement_protocol {
-    mode = *refine_shells repredict_only
+    mode = *refine_shells repredict_only None
       .type = choice
       .expert_level = 1
       .help = "refine_shells: refine in increasing resolution cutoffs after indexing."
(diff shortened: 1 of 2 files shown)


dermen:
Was getting this from travis

```
Running flake8 on changed files

algorithms/indexing/indexer.py:961:25: F821 undefined name 'sgtbx'

The command ".travis/run-flake8" exited with 1.
```

Seems unrelated to the pull request..

PR #884, opened 2019-08-16
Low res spot match indexing

dagewa:
This branch extends #882 by actually implementing a `LatticeSearch` strategy using entry points along the same lines as done for the existing `BasisVectorSearch` strategies. The method implemented is particularly designed for use with electron diffraction still images, where low resolution reflections are available and a known space group and unit cell can be provided. This algorithm was originally developed in [indigo.py](https://github.com/dials/dials_scratch/blob/master/dgw/indigo.py), but this PR moves it to DIALS for future development. At the moment, a single test is added, to ensure indexing using the new algorithm works as expected for a simulated test image.

### DIFF
--- a/algorithms/indexing/lattice_search_strategies.py
+++ b/algorithms/indexing/lattice_search_strategies.py
@@ -7,7 +7,7 @@ import abc
 import logging
 
 import copy
-from math import pi, sqrt
+import math
 import cctbx.miller
 from dials.array_family import flex
 import operator
@@ -16,7 +16,7 @@ from scitbx.math import superpose, least_squares_plane
 from dxtbx.model import Crystal
 from libtbx import phil
 
-TWO_PI = 2.0 * pi
+TWO_PI = 2.0 * math.pi
 FIVE_DEG = TWO_PI * 5.0 / 360.0
 
 logger = logging.getLogger(__name__)
@@ -127,7 +127,7 @@ candidate_spots
     n_spots = 10
     .type = int
 
-    dstar_tolerance = 4.0
+    d_star_tolerance = 4.0
     .help = "Number of sigmas from the centroid position for which to"
             "calculate d* bands"
     .type = float
@@ -272,7 +272,7 @@ class LowResSpotMatch(Strategy):
         )
         rt = flex.reflection_table()
         rt["miller_index"] = hkl_list.indices()
-        rt["dstar"] = 1.0 / hkl_list.d_spacings().data()
+        rt["d_star"] = 1.0 / hkl_list.d_spacings().data()
         rt["rlp_datum"] = self.Bmat.elems * rt["miller_index"].as_vec3_double()
         self.candidate_hkls = rt
 
@@ -285,7 +285,7 @@ class LowResSpotMatch(Strategy):
         hkl_list_p1 = hkl_list.expand_to_p1()
         rt = flex.reflection_table()
         rt["miller_index"] = hkl_list_p1.indices()
-        rt["dstar"] = 1.0 / hkl_list_p1.d_spacings().data()
+        rt["d_star"] = 1.0 / hkl_list_p1.d_spacings().data()
         rt["rlp_datum"] = self.Bmat.elems * rt["miller_index"].as_vec3_double()
         self.candidate_hkls_p1 = rt
         return
@@ -295,19 +295,19 @@ class LowResSpotMatch(Strategy):
     indices. Each observation will record its d* value as well as
     tolerated d* bands and a 'clock angle'"""
 
-        spot_dstar = reflections["rlp"].norms()
+        spot_d_star = reflections["rlp"].norms()
         if self._params.candidate_spots.limit_resolution_by == "n_spots":
             n_spots = self._params.candidate_spots.n_spots
             n_spots = min(n_spots, len(reflections) - 1)
-            dstar_max = flex.sorted(spot_dstar)[n_spots - 1]
-            self._params.candidate_spots.d_min = 1.0 / dstar_max
+            d_star_max = flex.sorted(spot_d_star)[n_spots - 1]
+            self._params.candidate_spots.d_min = 1.0 / d_star_max
 
         # First select low resolution spots only
-        spot_dstar = reflections["rlp"].norms()
-        dstar_max = 1.0 / self._params.candidate_spots.d_min
-        sel = spot_dstar <= dstar_max
+        spot_d_star = reflections["rlp"].norms()
+        d_star_max = 1.0 / self._params.candidate_spots.d_min
+        sel = spot_d_star <= d_star_max
         self.spots = reflections.select(sel)
-        self.spots["dstar"] = spot_dstar.select(sel)
+        self.spots["d_star"] = spot_d_star.select(sel)
 
         # XXX In what circumstance might there be more than one experiment?
         detector = experiments.detectors()[0]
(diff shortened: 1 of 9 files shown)


phyy-nx:
Hi, I'm assigning @dermen as a reviewer here, as he has two color code he's working on that might be affected.  My quick read makes me think he won't be affected, but it makes sense to test it.

FWIW, the goal is to move the two color indexer somewhere out of our cci repos soon with tests to help it keep pace with changes.

dagewa:
Sounds good. BTW it was pointed out to me that this new algorithm is quite similar to `small_cell`. I'd be interested to chat with you at some point about possibly reusing parts of `small_cell` and/or getting a version of that into DIALS too.

phyy-nx:
👍

dermen:
This change doesnt build our cci code which inherits from the realspace grid search strategy

graeme-winter:
Waiting on comments from @rjgildea ?

rjgildea:
@dagewa does this depend on #882 or does this include the changes from #882?

dagewa:
@rjgildea it includes the changes, so if this is merged, #882 can be closed.

algorithms/indexing/lattice_search_strategies.py
```diff
@@ -0,0 +1,611 @@
...
+
+            result.append(g)
+
+        return result
+
+    @staticmethod
+    def _fit_U_from_superposed_points(reference, other):
+
+        # Add the origin to both sets of points
+        origin = flex.vec3_double(1)
```
rjgildea: Would it not be neater/clearer to simply do here:
```
        reference.append((0, 0, 0))
        other.append((0, 0, 0))
```

algorithms/indexing/lattice_search_strategies.py
```diff
@@ -0,0 +1,611 @@
...
+        # Ideal relps from the known cell
+        other = flex.vec3_double([e["rlp_datum"] for e in vertices])
+
+        U = self._fit_U_from_superposed_points(reference, other)
+        UB = U * self.Bmat
+
+        if self._params.bootstrap_crystal:
+
+            # Attempt to index the low resolution spots
+            from dials_algorithms_indexing_ext import AssignIndices
```
rjgildea: Ideally I'd prefer not to import directly from the extension here, but I see the current implementation exposed in https://github.com/dials/dials/blob/master/algorithms/indexing/assign_indices.py expects reflections/experiments as input, which you don't currently have at this point...

algorithms/indexing/lattice_search_strategies.py
```diff
@@ -0,0 +1,611 @@
...
+
+            # Use the result to get a new UB matrix
+            reference = self.spots["rlp"].select(sel)
+            other = self.Bmat.elems * hkl_vec
+            U = self._fit_U_from_superposed_points(reference, other)
+            UB = U * self.Bmat
+
+        # Calculate RMSD of the fit
+        other_rotated = U.elems * other
+        rms = reference.rms_difference(other_rotated)
```
rjgildea: Again, is there anything to be gained by assigning `U.elems * other` to a separate variable only to use it in the following line?

algorithms/indexing/lattice_search_strategies.py
```diff
@@ -0,0 +1,611 @@
...
+                The experimental geometry models
+
+        Returns:
+            A list of candidate crystal models.
+
+        """
+        pass
+
+
+class CompleteGraph(object):
```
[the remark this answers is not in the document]
dagewa: Perhaps. I was going to make more of the graph class, but in the end all I needed was a total weight, so it is fairly trivial. I also needed to implement my own `__hash__` function to allow efficient filtering of effectively duplicate graphs via `set`. networkx might provide solutions, but I'd have to look into that.

dagewa:
> Perhaps now the lattice_search module should be made into a sub-package, with lattice_search_strategies moved to dials.algorithms.indexing.lattice_search.strategies?

I agree, but it takes a bit of disentangling with `LatticeSearch` and `BasisVectorSearch` currently both defined in `lattice_search.py`. Possibly these should each move to their sub-packages under `lattice_search/` and `basis_vector_search/`. I'd prefer if that was done on a new branch.

dagewa:
> Would you expect this to work well/at all on rotation images? Sweeps? I've tested this on x4wide dataset and it doesn't perform spectacularly on a single image, indexing only 12% of reflections (compared to 92% for fft3d)

The defaults are certainly intended for ED geometry stills, but the algorithm isn't strictly limited to stills. In fact, in this case - annoyingly - including a single extra spot to seed the search allows it to index 91% of reflections (`n_spots=11` rather than `n_spots=10`):

```
dials.index imported.expt strong.refl space_group=P422 unit_cell=42.46,42.46,39.85,90,90,90 indexing.method=low_res_spot_match n_spots=11
...
Refined crystal models:
model 1 (477 reflections):
Crystal:
    Unit cell: (42.56(3), 42.56(3), 39.94(3), 90.0, 90.0, 90.0)
    Space group: P 4 2 2
    U matrix:  {{-0.8769, -0.4797, -0.0312},
                { 0.0385, -0.1347,  0.9901},
                {-0.4792,  0.8670,  0.1366}}
    B matrix:  {{ 0.0235,  0.0000,  0.0000},
                { 0.0000,  0.0235,  0.0000},
                { 0.0000,  0.0000,  0.0250}}
    A = UB:    {{-0.0206, -0.0113, -0.0008},
                { 0.0009, -0.0032,  0.0248},
                {-0.0113,  0.0204,  0.0034}}

--------------------------------------------------
| Imageset | # indexed | # unindexed | % indexed |
--------------------------------------------------
| 0        | 477       | 46          | 91.2%     |
--------------------------------------------------
```

I found I had to make similar changes to index single images taken from the `thaumatin_i04` data set too. Generally these changes make the algorithm take longer to run.

benjaminhwilliams:
Not of very much importance, but you can be reassured that the remaining Flake8 warning that's causing the Travis PR check to fail (`algorithms/indexing/indexer.py:982:25: F821 undefined name 'sgtbx'`) is fixed in #885.

dagewa:
> Not of very much importance, but you can be reassured that the remaining Flake8 warning that's causing the Travis PR check to fail (`algorithms/indexing/indexer.py:982:25: F821 undefined name 'sgtbx'`) is fixed in #885.

Ah thanks. I wondered whether to fix that here, but I'm happy that #885 is its proper home.

algorithms/indexing/test_lattice_search.py
```diff
@@ -99,3 +99,37 @@ def test_stills_indexer_BasisVectorSearch_i04_weak_data(
...
+            experiment.imageset.data(), experiment.imageset.indices()
+        )
+        experiment.imageset.set_scan(None)
+        experiment.imageset.set_goniometer(None)
+        experiment.scan = None
+        experiment.goniometer = None
+    params = phil_scope.fetch().extract()
+    params.indexing.method = indexing_method
+    params.indexing.basis_vector_combinations.max_refine = 5
+    if unit_cell is not None:
```
rjgildea: `unit_cell` and `space_group` are required input for this indexing method right? So in the check for if unit_cell/space_group is not None is not necessary.

algorithms/indexing/lattice_search_strategies.py
dagewa: Yes, that's the issue. I'm not yet sure how useful the non-default option to trigger this import is. It seemed to occasionally help, but not enough to make it default

algorithms/indexing/test_lattice_search.py
dagewa: Yes, copy-pasta leftovers. An alternative would be to merge the `test_stills_indexer_LatticeSearch_i04_weak_data` and `test_stills_indexer_BasisVectorSearch_i04_weak_data` tests, by figuring out whether to import a `StillsIndexerBasisVectorSearch` or a `StillsIndexerLatticeSearch`

dagewa: I've merged the tests, which reduces code duplication. It also exercises the possibility that `BasisVectorSearch` gets passed an indexing method it is not aware of

PR #885, opened 2019-08-16
Deprecation warnings

benjaminhwilliams:
While working in xia2, I noticed a deprecation warning, so I did a little semi-automatic tidy-up.  This PR does the following, based on the advice in the deprecation warnings:
* Replace `dials.util.halraiser` with `dials.util.show_mail_on_error` context manager;
* Replace `dxtbx.model.experiment_list.ExperimentListDumper` and `dxtbx.serialize.dump` with `as_file` method of experiment list;
* Replace `dials.algorithms.indexing.index_reflections` with `dials.algorithms.indexing.assign_indices.AssignIndicesGlobal`;
* Remove some unused imports and variables to satisfy flake8.

Any objections?

There is a  corresponding PR in xia2 (https://github.com/xia2/xia2/pull/325), and others to follow in dials_scratch & dxtbx.

### DIFF
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -963,22 +963,6 @@ class Indexer(object):
                 xs.sites_mod_short()
                 f.write(xs.as_pdb_file())
 
-    def debug_write_ccp4_map(self, map_data, file_name):
-        from iotbx import ccp4_map
-
-        gridding_first = (0, 0, 0)
-        gridding_last = map_data.all()
-        labels = ["cctbx.miller.fft_map"]
-        ccp4_map.write_ccp4_map(
-            file_name=file_name,
-            unit_cell=self.fft_cell,
-            space_group=sgtbx.space_group("P1"),
-            gridding_first=gridding_first,
-            gridding_last=gridding_last,
-            map_data=map_data,
-            labels=flex.std_string(labels),
-        )
-
     def export_as_json(
         self, experiments, file_name="indexed_experiments.json", compact=False
     ):
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -982,10 +982,8 @@ class Indexer(object):
     def export_as_json(
         self, experiments, file_name="indexed_experiments.json", compact=False
     ):
-        from dxtbx.serialize import dump
-
         assert experiments.is_consistent()
-        dump.experiment_list(experiments, file_name)
+        experiments.as_file(file_name)
 
     def export_reflections(self, reflections, file_name="reflections.pickle"):
         reflections.as_file(file_name)
--- a/command_line/sort_reflections.py
+++ b/command_line/sort_reflections.py
@@ -42,28 +42,28 @@ class Sort(object):
         phil_scope = parse(
             """
 
-      key = 'miller_index'
-        .type = str
-        .help = "The chosen sort key. This should be a column of "
-                "the reflection table."
+            key = 'miller_index'
+                .type = str
+                .help = "The chosen sort key. This should be a column of "
+                        "the reflection table."
 
-      reverse = False
-        .type = bool
-        .help = "Reverse the sort direction"
+            reverse = False
+                .type = bool
+                .help = "Reverse the sort direction"
 
-      output = sorted.refl
-        .type = str
-        .help = "The output reflection filename"
+            output = sorted.refl
+                .type = str
+                .help = "The output reflection filename"
 
-    """
+            """
         )
 
         # The script usage
         usage = (
             """
-      usage: %s [options] observations.refl
+            usage: %s [options] observations.refl
 
-    """
+            """
             % libtbx.env.dispatcher_name
         )
 
@@ -103,10 +103,10 @@ class Sort(object):
         reflections = reflections.select(perm)
 
         if options.verbose > 0:
-            print("Head of sorted list " + attr + ":")
+            print("Head of sorted list " + params.key + ":")
             n = min(len(reflections), 10)
             for i in range(n):
-                print(reflections[i][attr])
+                print(reflections[i][params.key])
 
         # Save sorted reflections to file
         if params.output:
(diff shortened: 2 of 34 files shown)


benjaminhwilliams:
Had to leave work before the checks finished.  I'll tidy up the Flake8 problems on Monday.

benjaminhwilliams:
@phyy-nx, I notice your :+1:.  Since I naturally have CCTBX as part of the DIALS build, my conversions caught lots of instances of the same sort of stuff in `cctbx_project/xfel`.  Want me to open a corresponding PR there, or just leave it alone?

graeme-winter:
@benjaminhwilliams humbly suggest you leave those ones to the XFEL folks, since they will have a lot of other ongoing work there so high risk of conflicts 🙂

algorithms/indexing/indexer.py
```diff
@@ -948,29 +948,11 @@ def _debug_write_reciprocal_lattice_points_as_pdb(
...
-        ccp4_map.write_ccp4_map(
-            file_name=file_name,
-            unit_cell=self.fft_cell,
-            space_group=sgtbx.space_group("P1"),
-            gridding_first=gridding_first,
-            gridding_last=gridding_last,
-            map_data=map_data,
-            labels=flex.std_string(labels),
-        )
-
```
benjaminhwilliams: This method isn't used anywhere and isn't tested (or documented, not that that's exceptional in DIALS).  It's also broken, with an unused import, so I suggest hacking it out and burning it.

command_line/sort_reflections.py
```diff
@@ -102,10 +103,10 @@ def run(self):
         reflections = reflections.select(perm)
 
         if options.verbose > 0:
-            print("Head of sorted list " + attr + ":")
+            print("Head of sorted list " + params.key + ":")
             n = min(len(reflections), 10)
             for i in range(n):
-                print(reflections[i][attr])
+                print(reflections[i][params.key])
```
benjaminhwilliams: `attr` is defined nowhere in the whole file.  Judging by the original commit, which predates the use of PHIL and the DIALS option parser, it seems the intended behaviour is to use `params.key`.

The fact that the broken behaviour wasn't caught in a test is also fixed in this commit.

algorithms/indexing/indexer.py
benjaminhwilliams: This method isn't used anywhere and isn't tested (or documented, not that that's exceptional in DIALS). It's also broken, with an unused import, so I suggest hacking it out and burning it.

benjaminhwilliams:
Isn't Git fun?

benjaminhwilliams:
The final Flake8 warning `command_line/analyse_background.py:68:13: F841 local variable 'mask' is assigned to but never used` will be resolved by #889.

doc/examples/read_experiment_and_data/example_experiment_data.py
```diff
@@ -116,18 +118,16 @@ def run(self):
             axis = matrix.col(goniometer.get_rotation_axis_datum())
             F = matrix.sqr(goniometer.get_fixed_rotation())
             S = matrix.sqr(goniometer.get_setting_rotation())
-            R = S * axis.axis_and_angle_as_r3_rotation_matrix(angle, deg=True) * F
+            R = (
+                S * axis.axis_and_angle_as_r3_rotation_matrix(angle, deg=True) * F
+            )  # noqa F841
         else:
-            R = matrix.sqr((1, 0, 0, 0, 1, 0, 0, 0, 1))
+            R = matrix.sqr((1, 0, 0, 0, 1, 0, 0, 0, 1))  # noqa F841
```
ndevenish: I guess all these `# noqa: F841` are because it's an example file for how to process this; I'd suggest instead of doing noqa, doing something like adding a `print` of the values to use them - this would be demonstrative of usage. There _are_ places in the code where things are unfortunately called in order for their side-effects, so an example probably shouldn't give that impression.

doc/examples/read_experiment_and_data/example_experiment_data.py
```diff
@@ -52,7 +54,7 @@ def __init__(self):
         )
 
     def run(self):
-        from dials.array_family import flex  # import dependency
+        from dials.array_family import flex  # noqa F401 import dependency
```
rjgildea: Is this _really_ an import dependency? If so, it ideally _shouldn't_ be, I'd have thought?

ndevenish: I agree - I'm so used to seeing this everywhere without being able to do anything that I basically ignore them now. I _think_ in some cases it's to ensure that the dials flex libraries are loaded so that the reflection_table et al can be unpickled by dxtbx, in which case it _might_ be appropriate in an example source. We should consider trying to eliminate this pattern long term, though.

benjaminhwilliams: Oh, I agree, I just went with the original commenter.  I haven't investigated.

benjaminhwilliams: @graeme-winter, was there a specific reason for this import?

rjgildea: @ndevenish I think the recommended way to load a reflection file should now be via `flex.reflection_table.from_file()`, especially if we are moving towards message pack instead of pickle.

doc/examples/read_experiment_and_data/example_experiment_data.py
benjaminhwilliams: Yep, sure, I'll do that.  I wasn't very clear on what this example file was trying to show, so I was really just trying to muffle all complaints from it with minimum effort.

doc/examples/read_experiment_and_data/example_experiment_data.py
Anthchirp: most likely this is no longer the case

benjaminhwilliams: :+1:

doc/examples/read_experiment_and_data/example_experiment_data.py
benjaminhwilliams: :+1:

PR #886, opened 2019-08-16
split findbasisvectors into functions

dermen:
I have been wanting to do this for some time. It helps when inheriting from this RealSpaceGridSearch class.. 

Im not sure about the rstbx flex import... Dials tests passed

Thoughts ?

### DIFF
--- a/algorithms/indexing/basis_vector_search/strategies.py
+++ b/algorithms/indexing/basis_vector_search/strategies.py
@@ -493,7 +493,7 @@ class RealSpaceGridSearch(Strategy):
 
     phil_scope = phil.parse(real_space_grid_search_phil_str)
 
-    def __init__(self, max_cell, target_unit_cell, params=None, *args, **kwargs):
+    def __init__(self, max_cell, target_unit_cell, params=None, n_unique_vec=30, *args, **kwargs):
         """Construct a real_space_grid_search object.
 
         Args:
@@ -507,6 +507,13 @@ class RealSpaceGridSearch(Strategy):
             max_cell, params=params, *args, **kwargs
         )
         self._target_unit_cell = target_unit_cell
+        cell_dimensions = self._target_unit_cell.parameters()[:3]
+        self.unique_cell_dimensions = set(cell_dimensions)
+        self.n_unique_vec = n_unique_vec
+
+    def _compute_functional(self, vector):
+        two_pi_S_dot_v = 2 * math.pi * self._rec_latt_vec.dot(vector)
+        return flex.sum(flex.cos(two_pi_S_dot_v))
 
     def find_basis_vectors(self, reciprocal_lattice_vectors):
         """Find a list of likely basis vectors.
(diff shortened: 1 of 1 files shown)


dagewa:
Hi @dermen, the DIALS tests pass even if the `from rstbx.array_family import flex` is removed, so it appears that is not needed.

This looks like a straightforward refactor, but I will leave to @rjgildea to review as the author here.

rjgildea:
@dermen thanks for the pull request - refactoring of this code was probably necessary, and is a good idea in principle, but I think taking a slightly more functional approach would be better in this case. I'll have a look at it myself using your changes as an inspiration and possibly make another pull request in a few days.

rjgildea:
I'm not sure of the performance implications of replacing one loop over ~16k items:
https://github.com/dials/dials/blob/f148d48db4672c4d4a1e940c164599c98012da03/algorithms/indexing/basis_vector_search/strategies.py#L543-L548
with two loops over ~16k items:
https://github.com/dials/dials/blob/35639ca6aa0912347ab08d809b6d81927f442c68/algorithms/indexing/basis_vector_search/strategies.py#L555
https://github.com/dials/dials/blob/35639ca6aa0912347ab08d809b6d81927f442c68/algorithms/indexing/basis_vector_search/strategies.py#L562-L567

rjgildea:
@dermen can you explain the motivation behind the desire to refactor the code? Do you wish to override a specific part of the algorithm in a subclass?

dermen:
@rjgildea  regarding the SST.angles, how about ``` self._directions = (ang.dvec for ang in SST.angles) ``` ? I dont think its a bottleneck either way based on my tests

And yes, a subclass is the motivation , one that would replace ```self._directions_on_hemisphere()``` and ```self._evaluate_functional_at_vectors()``` .. I want to keep the ```find_unique_vectors```  portion..

rjgildea:
It would be useful to know more details about precisely how you expect these functions to differ. I'm particularly curious about how `self._directions_on_hemisphere()` would differ, as it is otherwise a very thin wrapper around `rstbx.dps_core.SimpleSamplerTool`.

dermen:
Trying out different sample schemes. the SST samples are concentric circles along the hemisphere [see here](https://drive.google.com/open?id=1RBKNNB1z85zVyXHGyNWOW87hsFSQv13k)- note the blue and orange are two different grid spacings, one more fine than the other. Notice the samples  don't change much (or at all) near the poles when decreasing the grid spacing.. contrary to say a spiral-based sample approach [see here](https://drive.google.com/open?id=1y9mnxMhZ0qLwcyxm6mfc8dgTkjt_rYUV).

dermen:
Basically I wanted a way to try out different angular samples on the hemisphere at e.g. very fine spacing, and I was playing around with sampling *rough* hemispheres at millions of points, with unit cell lengths drawn from a normal distribution,  as opposed to smooth hemispheres (fixed unit cell lengths). This involved increasing the number of samples into the millions.. but it seemed to work really nicely on the two color data..  Presumably you have looked at something like this before ?

rjgildea:
@dermen I have made a more in-depth refactoring in https://github.com/dials/dials/pull/895 - does this provide the flexibility in the `RealSpaceGridSearch.find_basis_vectors()` function that you require?

dermen:
#895  looks greatm exactly what I need. I will close this one

PR #894, opened 2019-08-22
No sigi

graeme-winter:
Cope with absence of SIGI or SIGIMEAN in MTZ file


jbeilstenedmands:
I think a distinction should be made depending on whether the mtz file is merged or unmerged. If merged, then you don't need a variance for calculating <I>, however if it is unmerged then you really do need the sigmas to properly calculate <I> for each group, so perhaps an error should be thrown. I'll work on something and add to the PR.

PR #895, opened 2019-08-23
Index basis vector strategy refactor

rjgildea:
General refactoring of basis_vector_search subpackage:
- Refactor basis vector search strategies to individual modules
- Rename strategies.py -> strategy.py
- Export strategies in top level basis_vector_search/__init__.py to simplify imports
- Move utility functions to utils.py module
- Extract `group_vectors()` to standalone function in utils.py
- Remove "_strategy" suffix from entry point names

Refactoring of `RealSpaceGridSearch.find_basis_vectors()` function
- Use `group_vectors()` in RealSpaceGridSearch
- New `compute_functional() static method
- New `search_directions` and `search_vectors` properties which return an iterator of search directions/vectors respectively
- New score_vectors() method
- `max_vectors` param instead of hardcoded value


algorithms/indexing/basis_vector_search/strategy.py
```diff
@@ -0,0 +1,41 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import abc
+
+
+class Strategy(object):
+    """A base class for basis vector search strategies."""
+
+    __metaclass__ = abc.ABCMeta
```
Anthchirp: This is Python 2 only syntax. Do we really need metaclasses here, or would a 
```Python
def override_this_function():
    raise NotImplementedError()
```
suffice?

dermen:
@rjgildea  looks good to me, my tests passed. Thanks for this!

rjgildea:
General refactoring of basis_vector_search subpackage:
- Refactor basis vector search strategies to individual modules
- Rename strategies.py -> strategy.py
- Export strategies in top level basis_vector_search/__init__.py to simplify imports
- Move utility functions to utils.py module
- Extract `group_vectors()` to standalone function in utils.py
- Remove "_strategy" suffix from entry point names

Refactoring of `RealSpaceGridSearch.find_basis_vectors()` function
- Use `group_vectors()` in RealSpaceGridSearch
- New `compute_functional() static method
- New `search_directions` and `search_vectors` properties which return an iterator of search directions/vectors respectively
- New score_vectors() method
- `max_vectors` param instead of hardcoded value

PR #898, opened 2019-08-27
simplifying parts of the reeke algorithm

Anthchirp:
This came up as part of getting rid of `cmp()` for Python 3 (#897), but really warrants its own PR.

The changes in this pull request result in mathematically identical code, which raised some questions with the original code, which likely contains at least one bug.

### DIFF
--- a/algorithms/spot_prediction/reeke.py
+++ b/algorithms/spot_prediction/reeke.py
@@ -260,8 +260,8 @@ class reeke_model:
 
         # Find distances between p = 0 and the plane passing through the
         # centre of the Ewald sphere
-        dp_beg = abs(v_beg.dot(self._source))
-        dp_end = abs(v_end.dot(self._source))
+        dp_beg = v_beg.dot(self._source)
+        dp_end = v_end.dot(self._source)
 
         # There are two planes of constant p that are tangential to the Ewald
         # sphere, on either side of the sphere. The smaller in magnitude of p
@@ -274,19 +274,24 @@ class reeke_model:
         # The correct sign is determined by whether the plane normal vector is
         # more closely parallel or antiparallel to the beam direction.
 
-        sign = cmp(v_beg.dot(self._source), 0)
+        sign = cmp(dp_beg, 0)
 
         limits = [
-            (sign * s * (self._source.length() + s * dp_beg) / p_dist) for s in (-1, 1)
+            (sign * s * (self._source.length() + s * abs(dp_beg)) / p_dist)
+            for s in (-1, 1)
         ]
 
         self._ewald_p_lim_beg = tuple(sorted(limits))
 
-        sign = cmp(v_end.dot(self._source), 0)
+        sign = cmp(dp_end, 0)
 
         limits = [
-            (sign * s * (self._source.length() + s * dp_end) / p_dist) for s in (-1, 1)
+            (sign * s * (self._source.length() + s * abs(dp_end)) / p_dist)
+            for s in (-1, 1)
         ]
+
+        dp_beg = abs(dp_beg)
+        dp_end = abs(dp_end)
 
         self._ewald_p_lim_end = tuple(sorted(limits))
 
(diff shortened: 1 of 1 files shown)


graeme-winter:
@Anthchirp 

https://codegolf.stackexchange.com/questions/49778/how-can-i-use-cmpa-b-with-python3

graeme-winter:
`cmp(a,b)` => `((a > b) - (a < b))` apparently

dagewa:
Thanks, I need some time to look through the Mosflm implementation again to work out what's happening with the potential bugs.

It is worth noting that this Python implementation is not actually used in anywhere other than tests these days, AFAIK. Spot prediction now uses the C++ version from here: https://github.com/dials/dials/blob/86aed53cfe3575290c91ae5fc87cb2a314cd2bd5/algorithms/spot_prediction/reeke_index_generator.h#L657

At some point in the past it seemed like a good idea to keep the Python version around after converting to C++, perhaps it was because the Python algorithm was (marginally) easier to follow conceptually. However, I'm no longer sure it is a good idea to keep this code. At least, if real bugs are found here then the equivalent parts of the C++ version should also be changed.

Anthchirp:
oh okay, if the code is effectively dead then I think we should remove it.

graeme-winter:
+1

though we should as David says review the C++

algorithms/spot_prediction/reeke.py
```diff
@@ -274,21 +274,19 @@ def _p_limits(self):
         # The correct sign is determined by whether the plane normal vector is
         # more closely parallel or antiparallel to the beam direction.
 
-        sign = cmp(dp_beg, 0)
-
-        limits = [
-            ((sign * s * self._source.length()) + dp_beg) / p_dist
-            for s in (-1, 1)
-        ]
+        if dp_beg == 0:
```
dagewa: I'm checking the original algorithm again now. The magnitudes of `dp_beg` and `dp_end` are always >> 0, so `limits = [0,0]` will never arise. I'm noting this here for when any changes are translated over to the C++ implementation.

Anthchirp: that would explain why the mistake I found made no difference - both values would always have the same (positive) sign

dagewa: I meant |`dp_beg`| >> 0, but `sign` could still be {-1,1}, just never 0

algorithms/spot_prediction/reeke.py
```diff
@@ -303,15 +297,31 @@ def _p_limits(self):
         assert abs(sin_theta) <= 1.0  # sanity check
         sin_2theta = math.sin(2.0 * math.asin(sin_theta))
 
+        # FIXME
```
dagewa: I believe this is correct, though messy. The signs of `dp_end` and `dp_beg` will always be the same

dagewa: so it was unnecessary in the original code to redefine the `sign`

algorithms/spot_prediction/reeke.py
```diff
@@ -303,15 +297,31 @@ def _p_limits(self):
...
+        # I assume the sign trickery in the following block was done in error
+        # and should not really happen: The original code redefined 'sign' in
+        # the block above this code and as a result ends up attaching the sign
+        # of dp_end to the absolute value of db_beg. I have faithfully retained
+        # this behaviour below.
+        # I suspect in reality the abs() in e=... should be dropped and the
+        # 'sign' in the final expression replaced by 1, similar to the second
+        # block dealing with dp_end.
```
dagewa: This would be correct - drop `abs` and use the sign of the value instead

dagewa:
I will make the extra changes to the Python algorithm so we have a final, clean representation. Then I can look at tidying the C++ version equivalently. Finally, can remove the Python algorithm.

Anthchirp:
Sounds all good to me 👍

Issue #862, opened 2019-07-21
unpythonic data structures: experiment_identifiers, reflection_tables


Anthchirp:
This is how a python dictionary behaves:
```python
>>> x = {0:"0", 1:"1"}
>>> print(x, list(x.keys()), list(x.values()), list(x))
({0: '0', 1: '1'}, [0, 1], ['0', '1'], [0, 1])
```

This is how a `experiment_identifiers()` return value behaves:
```python
>>> x = reflection_table.experiment_identifiers()
>>> print(x, list(x.keys()), list(x.values()), list(x))
(<dials_array_family_flex_ext.experiment_id_map object at 0x7f6be4709758>, [0, 1], ['0', '1'], [(0, '0'), (1, '1')])
```

If the data structure pretends to be a dictionary, which the `.keys()` and `.values()` methods imply, then the default iterable should be working on `.keys()` like a regular dictionary in python.
The current default iterable could be renamed `.items()`, which `experiment_identifiers()` does not currently offer.

The same is true for `reflection_table`.

As this is an API change I suggest to decide on this before the DIALS 2.0 release.

On a related note: `.experiment_identifiers()` appears to be a function that takes no arguments. This feels wrong. Any chance this could be changed to a property?

dagewa:
These sound like useful changes. `experiment_identifiers` stuff is fairly safe, `reflection_table` will need lots of testing across various projects

dagewa:
Just (re)discovered `reflection_table['missing_key']` raises `RuntimeError`, not `KeyError`

Anthchirp:
About the `RuntimeError` vs `KeyError`. As far as I can tell the only implemented exception that we can raise from C++ is `RuntimeError`. No obvious idea how we could raise anything else.

ndevenish:
Should the KeyError thing have a `newsfragment`? Seems like something that would be useful to document. (Maybe we need an API category, though I think 'bug' would be appropriate for this)

Anthchirp:
already written up. https://github.com/dials/dials/pull/908/files#diff-5340eab23dd2989dc9277a1ee45a2a6cR8-R9

ndevenish:
Ah, sorry, missed that because it didn't go in with the `KeyError` change

Anthchirp:
That's because I forgot :smile:

PR #908, opened 2019-08-30
Make reflection tables behave more like 'dict's

Anthchirp:
Patch for the changes discussed in #862 

The `.values()`/`.items()` bits are exceedingly harmless as they just add new methods.

Changing the default iterator as described means that all existing
```python
for r in reflections:
   r["bbox"] ...
for r1, r2 in zip(reflections1, reflections2):
   ...
```
would become
```python
for r in reflections.rows():
   r["bbox"] ...
for r1, r2 in zip(reflections1.rows(), reflections2.rows()):
   ...
```

This is not a massive change, and arguably the new form is more explicit, but due to its impact I just want to confirm that we go ahead with this.

Tests are broken right now, as I haven't made these changes to the code base yet.


graeme-winter:
I like this

When I iterate through a reflection list I do not know whether I will get per-column or per-reflection iteration - this makes it explicit which I like 👍

asmit3:
Does this PR change the default access convention for reflection_table from row major to column major ? Not sure if I interpret reflection tables to behave similar to dictionaries. Reflection_tables have append, extend and delete methods that don't exist for dictionaries.

Anthchirp:
What do you mean by _"default access convention"_? Element access was and remains column access. The default iterator changes from `.keys()` to `.rows()`.
[I've extended the test included in the PR to illustrate this.](https://github.com/dials/dials/blob/76174a9f005c2125758cc60a9c3b9e02a8b2bbca/test/array_family/test_reflection_table.py#L22-L46)

Currently you have the situation that eg. `"A" in reflection_table` is `True`, but `"A" in list(reflection_table)` and `"A" in set(reflection_table)`, etc. are `False`. This I found surprising and counterintuitive.

asmit3:
By default access convention, I meant the default iterator. Maybe  I don't fully understand then how reflection_tables are stored in memory - my impression was the access was done in a row-major form (i.e per-reflection basis) and the default iterator in master reflects that - Please correct me if I am misinterpreting here.

graeme-winter:
The data in memory are stored column-major, and in most cases (when dealing with reflection data certainly) accessed primarily as a `dict` to get data e.g. `hkl = refl["miller_index"]` - the fact that you can _also_ index it as `refl[i]` makes for delightful ambiguity when it comes to what will happen when you iterate through `refl` - which access pattern will it follow? 🎲

The proposal from Markus above is to make this _explicit_ by adding a method which will give you an iterator through the reflections. 

I like explicit - it saves guessing.

ndevenish:
It's definitely a form of dictionary because you almost exclusively want to access it via some key, as a mapping of either column->row, rownumber->row (or boolean mask -> row). I think row/column major is a slight red herring because it's not stored as a fixed layout 2D array; each column is stored as a separate instance, and as @graeme-winter pointed out, you can `__getitem__` in both 'row' and 'column' forms.

FWIW I checked and yes, pandas sensibly does the iterator this way.

As a minor aside, I'm not sure `values` totally makes sense for the reflection table and am finding it hard to think of a use case; In a `dict` when `.values` is useful you are often using it to store e.g. a lookup table `Mapping[KeyType, ValueType]` where the key is derived from the value and you want to access all of the values, like a set. This makes no sense for a column: table mapping where you cannot work out what column you have from the contents of the column without adding e.g. a "column name" field to the flex array.

No harm really, in having it though.

Anthchirp:
I think you may be right in that `.values()` may not be generically useful for reflection tables.

I thought of that mainly in the context of experiment identifiers, where I could see a use for it. If it is not _useful_ then I agree it should better be left out.

Thanks for checking re. pandas

asmit3:
Ok good to know. Thanks for the clarification everyone.

Anthchirp:
To test for existing code using the default iterator the best way I found is to deliberately (and temporarily) break the iterator, for example with
```patch
--- a/array_family/boost_python/flex_table_suite.h
+++ b/array_family/boost_python/flex_table_suite.h
@@ -1177,7 +1177,7 @@ namespace dials { namespace af { namespace boost_python { namespace flex_table_s
         .def("__delitem__", &delitem_column<flex_table_type>)
         .def("__delitem__", &delitem_row<flex_table_type>)
         .def("__delitem__", &delitem_slice<flex_table_type>)
-        .def("__iter__", make_iterator<key_iterator<flex_table_type> >::range())
+        .def("__iter__", &has_key<flex_table_type>)
         .def("cols", make_iterator<column_iterator<flex_table_type> >::range())
         .def("items", make_iterator<column_iterator<flex_table_type> >::range())
         .def("rows", make_iterator<row_iterator<flex_table_type> >::range())
```

With this I found a place in scaling where we apparently interchangeably use reflection tables and lists or lists of reflection tables. I suspect this is a bug, but need to check with @jbeilstenedmands first.

phyy-nx:
Hi, I've only been peripherally following this, so forgive me if I'm not up
to speed.  If I understand it, the syntax:

for i in refl_table:

Will now be iterating over columns instead of rows, correct?

What will len(refl_table) return then?  We use len(refl_table) all the
time.  If the answer is number of rows, isn't that inconsistent with the
new iterator?

jmp1985:
I like this change; however, I also agree with @phyy-nx that len(reflections) is an issue. len(reflections) returning the number of reflections makes intuitive sense to me but is not the same behaviour as a dictionary where len would be the number of keys

graeme-winter:
reflection_table.size() => give the no. of reflections?

rjgildea:
For comparison with pandas dataframe behaviour:
```
>>> df
...
[727901 rows x 16 columns]
>>> len(df)
727901
>>> len(df.keys())
16
>>> df.size
11646416
```

Anthchirp:
I could live with 
`len(A) == len(A.rows()) != len(A.keys()) == len(list(A))`.

Given that this seems to be consistent with the pandas world view then that is probably not even a bad thing.

phyy-nx:
So in this example, len(list(A)) == len(A), so the default iterator remains
over rows and not keys?

graeme-winter:
@Anthchirp N.B. this will require code changes in xia2 e.g. 

```
            refl = flex.reflection_table.from_file(spot_filename)
            if not len(refl):
                raise RuntimeError("No spots found in sweep %s" % xsweep.get_name())
```

in `DialsIndexer.py`

phyy-nx:
My grepping this morning has found three instances in xfel where we need to change from 

`for r in refls:`

to

`for r in refls.rows():`

Plus there are other projects using dials that will be affected.  Please wait to merge this until I give the go ahead?  Thanks.

phyy-nx:
Ok, all clear here.  Thanks.

Anthchirp:
Just to note as I got bitten by this:
`len(refl) == len(list(refl.rows())) == refl.size()` but `len(refl.rows())` does not work and raises an exception. If that is not desirable then it should likely be discussed in another ticket.

dagewa:
Should we have called it `refl.iterrows()`?

graeme-winter:
Pretty sure this is a non-problem - `len(generator)` usually does not work...

Anthchirp:
> Should we have called it `refl.iterrows()`?

since `.iter*` names are deprecated - no.
Possibly after moving to Python3 only we can discuss changing `.rows()` into [view objects](https://docs.python.org/3.2/library/stdtypes.html#dictionary-view-objects), which then would allow a `len()` operation, but until then @graeme-winter is probably correct that this is a non-issue.

Issue #637, opened 2018-10-19
Let dials export to xds_ascii format for multi-dataset files


jbeilstenedmands:
I has been suggested that the xds_ascii option in dials.export should allow exporting of multi-dataset files (mainly relevant for post-scaling), to support workflows where downstream processing uses files in this format. Both XDS and XSCALE write in xds_ascii format, so these files can support multiple datasets.
I shall ask around to get a good example so we are confident of the format, unless anyone has some?

jbeilstenedmands:
I now have example files I can use so I will have a go at this

graeme-winter:
Not sure what the use case is here but I suggest we have a chat - 

 - XDS will write out XDS_ASCII.HKL, one per "run"
 - XSCALE will write out a format similar to this which can have > 1 "run" in

but they are subtly different...

PR #911, opened 2019-09-05
Exporting of multi-dataset and still experiments to XDS_ASCII

ndevenish:
Was asked to help export some still data to XDS_ASCII this, and lots of things fell over, so this fixes the immediate problems.

- Stills have been handled by writing a neutral (1,0,0) rotation axis and zero oscillation values. Not knowing the use cases for this, I don't know if this violates assumptions.
- If multiple experiments, an index will be inserted into the filename before the extension and written to a separate file.
- Also, a more useful error if the profile column doesn't exist

This issue was raised in #637, which this fixes.

Suggestions on testing/validating this output are welcome, I don't know any still/multilattice datasets in dials-data yet.

### DIFF
--- a/util/export_xds_ascii.py
+++ b/util/export_xds_ascii.py
@@ -2,11 +2,16 @@ from __future__ import absolute_import, division, print_function
 
 import copy
 import logging
+import os
 
+from cctbx.miller import map_to_asu
+from dials.array_family import flex
+from dials.util import Sorry
 from dials.util.filter_reflections import (
-    filter_reflection_table,
     FilteringReductionMethods,
+    filter_reflection_table,
 )
+from rstbx.cftbx.coordinate_frame_helpers import align_reference_frame
 from scitbx import matrix
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


util/export_xds_ascii.py
```diff
@@ -13,21 +21,43 @@ def export_xds_ascii(integrated_data, experiment_list, params, var_model=(1, 0))
...
+        experiment: The experiment list entry to export
+        params: The PHIL configuration object
+        var_model:
+    """
     # export for xds_ascii should only be for non-scaled reflections
     assert any(
         [i in integrated_data for i in ["intensity.sum.value", "intensity.prf.value"]]
     )
+    # Handle requesting profile intensities (default via auto) but no column
+    if "profile" in params.intensity and not "intensity.prf.value" in integrated_data:
```
ndevenish:
```suggestion
    if "profile" in params.intensity and "intensity.prf.value" not in integrated_data:
```

Anthchirp: note that I've changed all those on master today.

ndevenish: Right - this was new code on the branch anyway, but I forgot that pre-commit flake8 wasn't the complete list.

Issue #913, opened 2019-09-06
Discussion on per-file license/header comment blocks


Anthchirp:
Some files have headers such as
```python
#!/usr/bin/env python
#
# -*- Mode: Python; c-basic-offset: 2; indent-tabs-mode: nil; tab-width: 8 -*-
#
# $Id
#
# {filename}.py
#
#  Copyright (C) {year in the past} Diamond Light Source {or otherwise}
#
#  Author: {name}
#
#  This code is distributed under the BSD license, a copy of which is
#  included in the root directory of this package.
```

I tend to remove these, but as [@dagewa says](https://github.com/dials/dials/commit/edf059ef86e3ee1794349296d1f99104d6a2bbb9#commitcomment-34976534) I don't recall there being a recorded discussion on this.

So here is my reasoning:
1. In most cases the shebang (`#!`) does nothing as we mostly only run scripts from `/command_line`.
2. If you need to configure your editor (`-*- ...`) then you can do that outside of the source code file. Apart from that the settings are wrong. This is managed by black anyway.
3. We are not in SVN any more (`$Id`).
4. The license header is in the top license file, so unless this is different there is no signal here
5. The up-to-date author information is in `git blame`.

graeme-winter:
The overall copyright should include everyone who has contributed - individual files are ~ useless - and unlikely to be copied on their own, so (for example with `xia2`) I have been removing this boiler plate. Almost always this is out of date anyways.

rjgildea:
There are (were) also many that had the wrong filename or wrong author, which were obviously just copied and pasted from another file.

ndevenish:
All of these points are 100% correct and I entirely applaud the removal of these chunks.

The only marker that should be up there is `# coding: utf-8` (or whatever the emacs version is if people prefer) and even that can be dropped once python 3.

Should we systematically remove them?

dagewa:
I agree with the points above that the comment block headers are problematic and think that systematic removal make sense, rather than piecemeal deletion within other commits.

I should note though that previously STFC advised me to assert copyright and authorship like this:
```
__author__ = "David Waterman"
__copyright__= "Copyright 2019 United Kingdom Research and Innovation”
__license__ = "BSD 3-clause"
```
That was in the context of cctbx contributions, where as a minor contributor it was deemed I did not have enough control over top level assertions and should be rigorous about this in each new file (however, I don't think I have actually committed a whole new file to cctbx since getting that advice).

[...]

PR #914, opened 2019-09-06
Update LICENSE

dagewa:
- STFC --> UKRI
- current year


dagewa:
See #913

Anthchirp:
I think a year range is more appropriate, as not all content is necessarily from 2019.

PR #915, opened 2019-09-06
add AUTHORS file

graeme-winter:
👍

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-Contributors include::
+Contributors include:
 
 Aaron Brewster
 Asmit Bhowmick
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,28 @@
+Contributors include::
+
+Aaron Brewster
+Asmit Bhowmick
+Benjamin Williams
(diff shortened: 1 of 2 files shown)


graeme-winter:
@rjgildea commit which I think is a _very good idea_ - we should also add similar wording to CONTRIBUTING as pytest have e.g. "Unless your change is a trivial or a documentation fix (e.g., a typo or reword of a small section) please add yourself to the AUTHORS file, in alphabetical order."

AUTHORS
```diff
@@ -0,0 +1,28 @@
+Contributors include::
```
Anthchirp:
```suggestion
Contributors include:
```

rjgildea:
As used in projects such as [django](https://github.com/django/django/blob/master/AUTHORS) and [pytest](https://github.com/pytest-dev/pytest/blob/master/AUTHORS).

AUTHORS
ndevenish: Depends. I think it's like this in pytest because they include it in their ReST documentation directly, and that makes it a literal block

PR #916, opened 2019-09-06
Cleanup file headers

ndevenish:
For further discussion from #913. This cleans up the file headers, using [this script](https://gist.github.com/ndevenish/cb9ed83742d3c5102cc1aef9de1fb24a). Precisely:

- Removes all header lines except when possibly relevant (ignores e.g. `FIXME`, libtbx dispatcher instructions, pytest comment in conftest and a couple of hand-tuned files)
- Removes all shebang unless in `command_line/`, `setup.py` or `util/show_version.py`
- Makes all coding instruction `# coding: utf-8`
- Removes editor commands otherwise

### DIFF
--- a/algorithms/background/gmodel/modeller.py
+++ b/algorithms/background/gmodel/modeller.py
@@ -1,14 +1,3 @@
-#!/usr/bin/env python
-#
-# algorithm.py
-#
-#  Copyright (C) 2013 Diamond Light Source
-#
-#  Author: James Parkhurst
-#
-#  This code is distributed under the BSD license, a copy of which is
-#  included in the root directory of this package.
-
 from __future__ import absolute_import, division, print_function
 
 import logging
--- a/algorithms/background/median/algorithm.py
+++ b/algorithms/background/median/algorithm.py
@@ -1,14 +1,3 @@
-#!/usr/bin/env python
-#
-# algorithm.py
-#
-#  Copyright (C) 2013 Diamond Light Source
-#
-#  Author: James Parkhurst
-#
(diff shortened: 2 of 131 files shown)


util/show_version.py
```diff
@@ -1,6 +1,4 @@
 #!/usr/bin/python
-#
-# this is used by the build scripts - it should not normally need to be
```
Anthchirp: bit too eager

dagewa:
When deleting `# Author: David Waterman.` or similar, please replace with ` __author__ = "David Waterman"`. This is a guideline that comes from my employers.

dagewa:
I'll check those guidelines are relevant for this case and report back.

ndevenish:
I've left in the lines saying "Author: David Waterman". Does that solve your concern?

dagewa:
I've had a reply from the IP team now, which is very reasonable. They would be satisfied if all copyright owners and authors are listed in the LICENSE file, and that this file is kept up to date. So if AUTHORS is referred to from LICENSE, all is well and I don't need any per-module claims.

I had to check - historically licensing has been a significant issue and it is better that I follow organisational guidelines stringently. This is easy to do when the guidelines are in fact quite relaxed.

ndevenish:
Awesome - it's so much easier with sane approaches :) - and yes, much easier to keep up-to-date centrally.

Will re-clean then push.

ndevenish:
After discussion, I've also removed the shebangs from `command_line/` - they are never (and probably shouldn't be?) called that way, only through dispatchers or `libtbx.python -m<path>`

Issue #858, opened 2019-07-19
Update the boilerplate


Anthchirp:
https://github.com/dials/dials/wiki/Boilerplate-(or-the-standard-framework-of-a-DIALS-program)
has gone slightly stale:
* black
* experiments/reflections
* cPickle -> `from six.moves import cPickle as pickle`
* upcoming changes for #832

dagewa:
Noted https://github.com/dials/dials/issues/463#issuecomment-507622616 but I should have put the issue in then

Anthchirp:
I'm wondering whether this should go into the main source tree (not `/command_line/` of course)

dagewa:
@ndevenish's example is much better than the existing entry in the wiki, so I'm going to replace it there.

rjgildea:
The current "gold standard" boilerplate example now lives here:
https://github.com/dials/dials/blob/master/doc/examples/boilerplate.py
I.e. somewhere it can be tested and included in the docs (see https://github.com/dials/dials/pull/1042). Perhaps we should remove the boilerplate from the wiki entirely to avoid confusion.

dagewa:
Right, I missed that. Yes, I'll edit the page in the wiki to be "writing a DIALS program" and point to this.

PR #917, opened 2019-09-06
dials.sweep_to_stills

phyy-nx:
New program: dials.sweep_to_stills 

Converts a sweep to a series of stills by using the goniometer and scan to create a stills Experiment for every scan point in the scan. Includes an oscillation_angle/2 offset to the A matrices to account for how scans are recorded in DIALS.

Also includes reflection splitting. Idea is that the indexed reflections from the sweep need to be split into slices so that each experiment gets all the spots that impinged on it, even if the centroid is on the next phi step.

Also also adds parameter check_reference_geometry=True, expert_level=2, to dials.import

### DIFF
--- a/command_line/sweep_to_stills.py
+++ b/command_line/sweep_to_stills.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python
+#
+# LIBTBX_SET_DISPATCHER_NAME dials.sweep_to_stills
+
+from __future__ import absolute_import, division, print_function
+
+import logging
+
+from dxtbx.model.experiment_list import Experiment, ExperimentList
+from scitbx.matrix import sqr, col
+from dxtbx.model import MosaicCrystalSauter2014
+from cctbx import crystal_orientation
+from dials.model.data import Shoebox
+from dials.algorithms.refinement.prediction.managed_predictors import (
+    ExperimentsPredictorFactory,
+)
+from dials.util import show_mail_on_error
+from libtbx.phil import parse
+
+logger = logging.getLogger("dials.command_line.stills_process")
+
+help_message = """
+
+Utility script to split a sweep into a set of stills.
+
+Example::
+
+  dials.sweep_to_stills sweep.expt sweep.refl
+
+"""
+
+# The phil scope
+phil_scope = parse(
+    """
+output {
+  experiments = stills.expt
+    .type = str
+    .help = "Filename for the experimental models that have been converted to stills"
+  reflections = stills.refl
+    .type = str
+    .help = "Filename for the reflection tables with split shoeboxes (3D to 2D)"
+  domain_size_ang = None
+    .type = float
+    .help = "Override for mosaic angle. If None, use the crystals mosaic angle, if"
+            "available"
+  half_mosaicity_deg = None
+    .type = float
+    .help = "Override for mosaic angle. If None, use the crystals mosaic angle, if"
+            "available"
+}
+"""
+)
+
+
+class Script(object):
+    """A class for running the script."""
+
+    def __init__(self):
+        """Initialise the script."""
+        from dials.util.options import OptionParser
+        import libtbx.load_env
+
+        # The script usage
+        usage = (
+            "usage: %s [options] [param.phil] filenames" % libtbx.env.dispatcher_name
+        )
+
+        # Create the parser
+        self.parser = OptionParser(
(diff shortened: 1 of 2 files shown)


dagewa:
The functionality sounds useful. Please include tests in this repository for any new command line programs.

ndevenish:
I'm curious what specifically this is useful for?

Also, it'd be great if you could add a `newsfragments/917.feature` file with a short note saying this has been added and what it's for. (this is [towncrier](https://github.com/hawkowl/towncrier), which we're using to attempt to be better at building useful release notes)

Anthchirp:
Please see the [program boilerplate](https://github.com/dials/dials/wiki/Boilerplate-(or-the-standard-framework-of-a-DIALS-program)) for an updated template for new tools.
Although slightly out of date (#858) it is still more modern than the `Script()` helper class.

dagewa:
Some electron diffraction datasets (such as the one described in [this paper](https://doi.org/10.1073/pnas.1809978115)) were collected as a series of discrete tilt steps on a single crystal. You could index and refine the whole dataset as a sweep, but would want to convert to stills for integration.

I imagine the LBL use case is something different though. It would be interesting to hear what that is.

command_line/sweep_to_stills.py
```diff
@@ -0,0 +1,230 @@
...
+                new_experiment.imageset = experiment.imageset.as_imageset()[
+                    i_scanpoint : i_scanpoint + 1
+                ]
+                new_experiments.append(new_experiment)
+
+                # Each reflection in a 3D shoebox can be found on multiple images. Slice the reflections
+                # such that any reflection on this scan point is included with this image
+                new_id = len(new_experiments) - 1
+                subrefls = refls.select((i_scanpoint >= z1) & (i_scanpoint < z2))
+                for refl in subrefls:  # this will break when #908 is merged
```
rjgildea: All the more reason for requiring a test when the pull request contains comments like this!
Would it be possible to re-write this loop to avoid breaking when #908 is merged?

command_line/sweep_to_stills.py
```diff
@@ -0,0 +1,230 @@
...
+            )
+            goniometer_axis = col(experiment.goniometer.get_rotation_axis())
+            step = experiment.scan.get_oscillation()[1]
+
+            refls = reflections.select(reflections["id"] == expt_id)
+            _, _, _, _, z1, z2 = refls["bbox"].parts()
+
+            # Create an experiment for each scanpoint
+            for i_scanpoint in range(*experiment.scan.get_array_range()):
+                # The A matrix is the goniometer setting matrix for this scan point times the scan varying A matrix at this scan point
```
rjgildea: Could you wrap the comments to max 88 characters line length to match the max line length of the code?

command_line/sweep_to_stills.py
```diff
@@ -0,0 +1,230 @@
...
+    ExperimentsPredictorFactory,
+)
+from dials.util import show_mail_on_error
+from libtbx.phil import parse
+
+logger = logging.getLogger("dials.command_line.stills_process")
+
+help_message = """
+
+Utility script to split a sweep into a set of stills.
```
rjgildea:
```suggestion
Split a sweep into a set of stills.
```
The first 3 words add no useful information to the program description.

command_line/sweep_to_stills.py
```diff
@@ -0,0 +1,230 @@
...
+  reflections = stills.refl
+    .type = str
+    .help = "Filename for the reflection tables with split shoeboxes (3D to 2D)"
+  domain_size_ang = None
+    .type = float
+    .help = "Override for mosaic angle. If None, use the crystals mosaic angle, if"
+            "available"
+  half_mosaicity_deg = None
+    .type = float
+    .help = "Override for mosaic angle. If None, use the crystals mosaic angle, if"
```
rjgildea: Are the help strings for this and the above parameter intentionally identical?

phyy-nx:
Thanks all, I've addressed as many code review comments as I can for right now.  Was particularly pleased about using set_A directly.  Thanks @rjgildea.  I don't have time to fix the boiler point stuff now, sorry.

I'll send more about the usecase particulars when I can.  Appreciate the curiosity :)

phyy-nx:
(I'll merge tomorrow if that's ok)

PR #918, opened 2019-09-09
Refactor dials.array_family.flex

Anthchirp:
to make the source of imported objects clearer.

### DIFF
--- a/array_family/flex.py
+++ b/array_family/flex.py
@@ -17,6 +17,13 @@ import six.moves.cPickle as pickle
 from dials.util import Sorry
 from scitbx import matrix
 
+# Note: Right at the end of this file all names from
+#         cctbx.array_family.flex and
+#         dials_array_family_flex_ext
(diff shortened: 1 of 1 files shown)


Anthchirp:
ND says:
- `from cctbx.array_family import flex` _in this file_ is probably okay, because only flex and makes code in this file easier to read
- This is acceptable way of dealing with the problem, but _it needs massive flashing caveat signs at the top in the imports_. The only reason I _know_ that the imports chain is because I opened the file and saw `from cctbx.array_family.flex import *`
- Might it be better to separate this stuff out to a `_dials_flex.py` or something and then have `flex.py` be _explicitly_ namespace combination from cctbx, dials_ext etc so there isn’t the confusion?

Anthchirp:
I'll follow recommendation 2 then for now (add a note).
If we have time at some point in the future (:laughing:) we should probably implement recommendation 3 and refactor things out into a separate file.

Issue #921, opened 2019-09-10
Change dials.export to not require data verification by default


graeme-winter:
Only `dials.export format=best` requires it so we should move this to a different program. The `best` output also performs calculations and is not much used, so I think this is low cost with a high benefit that exporting data will _just work_ - a common annoyance for users.

graeme-winter:
Side effect: retire `DIALS_EXPORT_DO_NOT_CHECK_FORMAT` 🙂

graeme-winter:
Working on this in `dials-921` branch.

ndevenish:
See also: #484, #639

Issue #832, opened 2019-07-02
Retire *.debug.log


dagewa:
Somewhere there was a discussion about removing the *.debug.log from DIALS programs and simply using `logging` level features to obtain debug output only when requested. I can't find that discussion now, so creating a new ticket for it.

dagewa:
It's in the [DIALS 2.0 Goals](https://docs.google.com/document/d/14MSdeeHrd6_0c5l08EksggJXQg5NlJMUnw2P2bGXbnY/edit) too.

rjgildea:
Re-opening, as we still appear to have many `debug=*.debug.log` parameters in command line programs:
```
$ find . -name "*.py" | xargs grep "debug\.log"
./command_line/find_hot_pixels.py:        debug="dials.find_hot_pixels.debug.log",
./command_line/symmetry.py:  debug_log = dials.symmetry.debug.log
./command_line/compute_delta_cchalf.py:    debug_log = 'dials.compute_delta_cchalf.debug.log'
./command_line/model_background.py:    debug_log = 'dials.model_background.debug.log'
./command_line/search_beam_position.py:  debug_log = "dials.search_beam_position.debug.log"
./command_line/import_stream.py:    debug_log = 'dials.import_stream.debug.log'
./command_line/export.py:    debug_log = dials.export.debug.log
./command_line/complete_full_sphere.py:            debug="dials.complete_full_sphere.debug.log",
./command_line/integrate.py:    debug_log = 'dials.integrate.debug.log'
./command_line/stills_process.py:            debug="dials.process.debug.log",
./command_line/check_indexing_symmetry.py:  debug_log = dials.check_indexing_symmetry.debug.log
./command_line/two_theta_refine.py:    debug_log = dials.two_theta_refine.debug.log
./command_line/cosym.py:  debug_log = dials.cosym.debug.log
./command_line/find_shared_models.py:    debug_log = 'dials.find_shared_models.debug.log'
./command_line/dials_import.py:    debug_log = 'dials.import.debug.log'
./command_line/find_spots.py:    debug_log = 'dials.find_spots.debug.log'
```

PR #922, opened 2019-09-10
Fixes for #921 - split off dials.export format=best

graeme-winter:
Made a second program `dials.export_best` and removed option for `dials.export` so that it does not need access to the format object (so `check_format=False` and can run without images)

### DIFF
--- a/command_line/export_best.py
+++ b/command_line/export_best.py
@@ -1,0 +1,173 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+import sys
+
+from libtbx.phil import parse
+from libtbx import Auto
+
+logger = logging.getLogger("dials.command_line.export_best")
+
+help_message = """
+This program is used to export the results of dials processing for the strategy
+program BEST.
+
+"""
+
+phil_scope = parse(
+    """
+
+  intensity = *auto profile sum scale
+    .type = choice(multi=True)
+    .help = "Choice of which intensities to export. Allowed combinations:
+            scale, profile, sum, profile+sum, sum+profile+scale. Auto will
+            default to scale or profile+sum depending on if the data are scaled."
+
+  debug = False
+    .type = bool
+    .help = "Output additional debugging information"
+
+  best {
+
+    prefix = best
+      .type = str
+      .help = "The prefix for the output file names for best"
+              "(.hkl, .dat and .par files)"
+
+    n_bins = 100
+      .type = int(value_min=1)
+      .help = "Number of resolution bins for background estimation"
+
+    min_partiality = 0.1
+      .type = float(value_min=0, value_max=1)
+      .help = "Minimum partiality of reflections to export"
+
+  }
+
+  output {
+
+    log = dials.export_best.log
+      .type = path
+      .help = "The log filename"
+
+    debug_log = dials.export_best.debug.log
+      .type = path
+      .help = "The debug log filename"
+
+  }
+"""
+)
+
+
+class BestExporter(object):
+    def __init__(self, params, experiments, reflections):
+        """
+        Initialise the exporter
+
+        :param params: The phil parameters
+        :param experiments: The experiment list
+        :param reflections: The reflection tables
+        """
+
+        # Check the input
+        if not experiments:
+            raise Sorry("BEST exporter requires an experiment list")
+        if not reflections:
+            raise Sorry("BEST exporter require a reflection table")
+
+        # Save the stuff
+        self.params = params
+        self.experiments = experiments
+        self.reflections = reflections
+
+    def export(self):
+        """
+        Export the files
+        """
+        from dials.util import best
+
+        experiment = self.experiments[0]
+        reflections = self.reflections[0]
+        partiality = reflections["partiality"]
+        sel = partiality >= self.params.best.min_partiality
+        logger.info(
+            "Selecting %s/%s reflections with partiality >= %s",
+            sel.count(True),
+            sel.size(),
+            self.params.best.min_partiality,
+        )
+        if sel.count(True) == 0:
+            raise Sorry(
+                "No reflections remaining after filtering for minimum partiality (min_partiality=%f)"
+                % (self.params.best.min_partiality)
+            )
+        reflections = reflections.select(sel)
+
+        imageset = experiment.imageset
+        prefix = self.params.best.prefix
+
+        best.write_background_file(
+            "%s.dat" % prefix, imageset, n_bins=self.params.best.n_bins
+        )
+        best.write_integrated_hkl(prefix, reflections)
+        best.write_par_file("%s.par" % prefix, experiment)
+
+
+if __name__ == "__main__":
+    from dials.util.options import (
+        OptionParser,
+        flatten_experiments,
+        flatten_reflections,
+    )
+    from dials.util.version import dials_version
+    from dials.util import log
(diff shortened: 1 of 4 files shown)


graeme-winter:
FWIW I updated the test and made sure they all pass before making PR

command_line/export_best.py
```diff
@@ -0,0 +1,173 @@
...
+    .type = choice(multi=True)
+    .help = "Choice of which intensities to export. Allowed combinations:
+            scale, profile, sum, profile+sum, sum+profile+scale. Auto will
+            default to scale or profile+sum depending on if the data are scaled."
+
+  debug = False
+    .type = bool
+    .help = "Output additional debugging information"
+
+  best {
```
rjgildea: We probably don't need the `best` scope if it is now a separate program - i.e. move the parameters in this scope to the top level scope.

command_line/export_best.py
```diff
@@ -0,0 +1,173 @@
...
+
+  }
+
+  output {
+
+    log = dials.export_best.log
+      .type = path
+      .help = "The log filename"
+
+    debug_log = dials.export_best.debug.log
```
rjgildea: I thought we were no longer creating debug.log files? (see https://github.com/dials/dials/issues/832)

command_line/export_best.py
```diff
@@ -0,0 +1,173 @@
...
+phil_scope = parse(
+    """
+
+  intensity = *auto profile sum scale
+    .type = choice(multi=True)
+    .help = "Choice of which intensities to export. Allowed combinations:
+            scale, profile, sum, profile+sum, sum+profile+scale. Auto will
+            default to scale or profile+sum depending on if the data are scaled."
+
+  debug = False
```
rjgildea: I don't think this debug parameter is used anywhere.

command_line/export_best.py
```diff
@@ -0,0 +1,173 @@
...
+help_message = """
+This program is used to export the results of dials processing for the strategy
+program BEST.
+
+"""
+
+phil_scope = parse(
+    """
+
+  intensity = *auto profile sum scale
```
rjgildea: I don't think this `intensity` parameter is used in the best export.

command_line/export_best.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if not params.input.experiments and not params.input.reflections:
+        parser.print_help()
+        sys.exit()
+
+    # Get the experiments and reflections
+    experiments = flatten_experiments(params.input.experiments)
+    reflections = flatten_reflections(params.input.reflections)
+
+    # do auto intepreting of intensity choice:
```
rjgildea: As above, I don't think `params.intensity` is used in the `BestExporter`.

newsfragments/921.feature
```diff
@@ -0,0 +1,3 @@
+Moved `dials.export format=best` to `dials.export_best` as that one needed
```
rjgildea: As I understand this is intended to form the release notes - so could it be made a more user-friendly description of the change? I think just moved to standalone dials.export_best command would suffice, the user doesn't need to know about the underlying reasons for the change in this case.

command_line/export_best.py
graeme-winter: Done - moved to top level scope and made the change in `xia2` to reflect this.

command_line/export_best.py
graeme-winter: We are; have now removed this in both `export` and `export_best`

command_line/export_best.py
graeme-winter: Could be passed through `params.debug` to other code? Will look.

command_line/export_best.py
graeme-winter: It _could_ be though... could imagine using it with scaled data for partial data sets (e.g. in an i19 context)

command_line/export_best.py
```diff
@@ -0,0 +1,165 @@
...
+    .type = choice(multi=True)
+    .help = "Choice of which intensities to export. Allowed combinations:
+            scale, profile, sum, profile+sum, sum+profile+scale. Auto will
+            default to scale or profile+sum depending on if the data are scaled."
+
+  debug = False
+    .type = bool
+    .help = "Output additional debugging information"
+
+  prefix = best
```
rjgildea: I suppose now we don't have the specific `best` scope, the `prefix` parameter should strictly be in the `output` scope, since it controls the names of output files.

command_line/export_best.py
graeme-winter: Also as above.

command_line/export_best.py
rjgildea: Yes, but we shouldn't have unused program parameters that _might_ hypothetically be used in the future. If we decide to add such functionality at some point in the future then we can re-add the parameters to `dials.export_best`.

This is the code that extracts the intensities and writes them to `best.hkl`, and it doesn't currently use any parameter to control the choice of intensities:
https://github.com/dials/dials/blob/142bd8994a92bb9e50a5c6e9f08801e2118ed265/util/best.py#L21-L44

command_line/export_best.py
rjgildea: I'm pretty sure it isn't - and if you remove the parameter and the tests still pass, then it probably isn't used :)

command_line/export_best.py
graeme-winter: Removed

command_line/export_best.py
graeme-winter: Removed

newsfragments/921.feature
graeme-winter: I think we can tidy that up when we get to it, but I would leave in some of the details as _some_ end users know about the environment variable nonsense so making it clear that this is no longer needed would be useful. However if you disagree feel free to revise.

ndevenish:
Do we need a separate program? How often do we think `best` is used? Couldn't we either default to `check_format=False` and then use the environment variable to _activate_ `check_format`, or just do it as a configuration?

Or, before loading the optionparser, manually look for "format=best" in `sys.argv` and pass `check_format` based on that (or, strip all non-phil params out of the args, parse that and then use the result of _that_ to parse with the input files). It's a workaround, but AFAIK this is basically the only place we have this problem.

graeme-winter:
@ndevenish personally I am happy with a different program as it imposes the constraint on future exported formats that they have to think twice before needing anything from the images (hint: they won't)

graeme-winter:
For completeness:

 - checking for `format=best` in option parser adds debt
 - removing an environment variable makes the API cleaner

graeme-winter:
@rjgildea re: `info` 👍

ndevenish:
As much debt as a whole separate program that does the same thing except for one thing, and just adds more command-line-program-clutter?

graeme-winter:
> As much debt as a whole separate program that does the same thing except for one thing, and just adds more command-line-program-clutter?

Strictly, no - `export_best` does extra non-export things like computing background from images. `export` just exports integrated or scaled data in a number of formats. I hold with the distinction, and think that this will help users in the future. 

Finally, `export format=best` _sounds_ to the uninitiated like what they probably want, right? I want the _best_ format. We could even deprecate `export_best` to a `dev.` since generally unused, or migrate to xia2.

ndevenish:
I'll buy separating with the aim to make removal easier.

PR #931, opened 2019-09-16
Fixes for dials#929

graeme-winter:
Compute number of digits needed for template - int(math.ceil(math.log10(n_images))) + 1 if exactly 10^n

### DIFF
--- a/command_line/merge_cbf.py
+++ b/command_line/merge_cbf.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import binascii
+import math
 import sys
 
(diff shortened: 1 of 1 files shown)


Anthchirp:
`len(str(n_images))`?

Anthchirp:
```
out_image = "{prefix}{number:0{digits}d}.cbf".format(prefix=out_prefix, number=i_out+1, digits=n_digits)
```

graeme-winter:
@Anthchirp 👍 what happens if you look at the problem as a programmer not a mathematician 🙂 both sensible suggestions...

PR #934, opened 2019-09-16
Multiscan

kmdalton:
I've added a new phil parameter to bypass one of the configuration sanity checks in dials.refine. The new parameter is called **allow_multiple_scans_per_crystal**. It is **boolean**, defaults to **False**, and has **expert_level = 2**. If True, dials.refine will allow a single scan-varying crystal and goniometer to be associated with multiple scans. 

The use case for this parameter is rotation series of still images wherein multiple images were acquired per angle.

### DIFF
--- a/algorithms/refinement/parameterisation/configure.py
+++ b/algorithms/refinement/parameterisation/configure.py
@@ -107,6 +107,12 @@ phil_str = (
               "of the serialized model small. At the moment, this only has"
               "an effect for crystal unit cell (B matrix) errors."
 
+    allow_multiple_scans_per_crystal = False
+      .type = bool
+      .help = "If scan-varying refinement is done, allow multiple scans"
+              "per crystal.                                             "
+      .expert_level = 2
+
     debug_centroid_analysis = False
       .help = "Set True to write out a file containing the reflections used"
               "for centroid analysis for automatic setting of the  scan-varying"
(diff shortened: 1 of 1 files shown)


dagewa:
Hi @kmdalton, I'm trying to understand your use case. It sounds like you are constructing "scans" after the fact from a tilt series, but instead of a single tilt series making a single scan, you actually have N tilt series, because at each angle you collected N images. Is this correct?

In that case, each "scan" is not continuous in dose, because every image is followed by N-1 images until the next image in the same scan is recorded. As each scan is not a continuous data collection event we have to think about whether scan-varying refinement is appropriate. It's not clear to me if you are combining all the data from N scans in a single refinement job, or want to refine the data from a single scan but use the resulting model for each scan.

I'm just thinking there may be ways to get what you want using the multi-experiment manipulation tools in DIALS, but I need to have a better understanding of what you want to achieve.

kmdalton:
@dagewa, your description of the data acquisition strategy is correct. I am combining all the data from the N scans into a refinement job with a single scan-varying crystal model. I have been using this strategy to analyze data, and it agrees very well with the results of dials.stills_process. 

I don't believe there is any way to achieve this sort of refinement without overriding the check in algorithms/refinement/parameterisation/configure.py

dagewa:
I guess that your N scan and goniometer models are actually identical, right? Then it might be possible to run `dials.combine_experiments` with `reference_from_experiment.scan=0 reference_from_experiment.goniometer=0` prior to refinement to force each experiment to actually share a goniometer and a scan model. After refinement, `dials.split_experiments` should separate out the models again.

I'm still not completely sure if that is what you want though. It would be helpful to see a test case consisting of some experiments and reflections files and the commands you are using. Anyway, if we end up merging this PR to add a new feature to dials.refine then we will need to also add a test case.

kmdalton:
@dagewa, you are totally right! With a bit of help from @phyy-nx, I figured out how to avoid tripping this error. Thanks for your feedback. I do not think there is any need to merge this request.

dagewa:
Ah good, I'm glad you've got a way of getting this to work with the existing tools.

PR #936, opened 2019-09-17
Speed up .mtz export, #924

### LINKED ISSUE #924: dials.export very slow for large numbers of batches
This is a known thing but I could not find a ticket for it - seems that adding batches is linear time on `n_batches` making it quadratic time overall. This is a "feature" of how the MTZ libraries are implemented. 

batch write time

Some hacking around in `iotbx.mtz` can give a factor of ~ 2 speed up (see graph) however this is still `O(n^2)` behaviour.

Anthchirp:
Instead of running O(n) calls to _add_batch(), which runs an O(n) check
and then runs another C library function running another O(n) check for
a total complexity of O(2n²)...

...run one custom function that inserts all batch information at the
same time for O(n) cost in time.

Fixes #924 if used together with the branch of same name in cctbx_project

### DIFF
--- a/util/export_mtz.py
+++ b/util/export_mtz.py
@@ -125,29 +125,23 @@ class MADMergedMTZWriter(MergedMTZWriter):
         )
 
 
-def _add_batch(
+def _add_batch_list(
+    image_range,
     mtz,
     experiment,
     wavelength,
     dataset_id,
-    batch_number,
-    image_number,
+    batch_offset,
     force_static_model,
 ):
-    """Add a single image's metadata to an mtz file.
-
-    Returns the batch object.
-    """
-    assert batch_number > 0
+    """Add batch metadata to an mtz file."""
 
     # Recalculate useful numbers and references here
-    # We ignore panels beyond the first one, at the moment
-    panel = experiment.detector[0]
-
-    if experiment.goniometer:
-        axis = matrix.col(experiment.goniometer.get_rotation_axis())
-    else:
-        axis = 0.0, 0.0, 0.0
+    n_batches = image_range[1] - image_range[0] + 1
+    phi_start = flex.float(n_batches)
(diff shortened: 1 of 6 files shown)


util/export_mtz.py
```diff
@@ -125,135 +125,107 @@ def add_dataset(
...
-    o.set_time1(0.0).set_time2(0.0).set_title("Batch {}".format(batch_number))
-    o.set_ndet(1).set_theta(flex.float((0.0, 0.0))).set_lbmflg(0)
-    o.set_alambd(wavelength).set_delamb(0.0).set_delcor(0.0)
-    o.set_divhd(0.0).set_divvd(0.0)
+    i0 = image_range[0]
+    for i in range(n_batches):
+        if experiment.scan:
+            phi_start[i], phi_range[i] = experiment.scan.get_image_oscillation(i + i0)
+        else:
+            phi_start[i], phi_range[i] = 0.0, 0.0
```
Anthchirp: is `flex.float()` guaranteed to be 0-initialised? If so could drop the `else:` clause

graeme-winter: @Anthchirp by default no but we can force it to be... good suggestion. `flex.float(size, init_value)` will do it

Anthchirp:
This seems to produce viable .mtz files which, according to `mtzdmp` are identical to the original ones, however `xxd` shows a few bit differences which I would like to investigate before merging.

graeme-winter:
There are unused areas in MTZ files which are there to "leave space for the future"

Do you get differences at the same place with repeat runs of one version of the code? Possibly this is just uninitialised space being written to file.

Anthchirp:
it's the last bit of the mantissa in phi_range and phi_end, so basically an irrelevant rounding error in the float-flexes. I'm happy with that.

PR #944, opened 2019-09-20
Attenuation tidy up

dagewa:
Starting to tidy up after changes made in #943. 

Two more tests fail after this:
- test/algorithms/integration/test_parallel_integrator.py:156
- test/algorithms/integration/test_parallel_integrator.py:180

To be investigated...


dagewa:
It seems these two extra tests aren't failing after all on this branch. So I'll open for review

PR #947, opened 2019-09-25
Use Prim MST instead of Kruskal in "local" indexing

rjgildea:
Profiling of the `AssignIndicesLocal` method has identified the calculation of the minimum spanning tree (MST) as a bottleneck when the number of reflections is large:
https://github.com/dials/dials/blob/d764cb97fcfc2e7d3ccf88781c36236666d00cda/algorithms/indexing/index.h#L266-L270

The existing code used Kruksal's algorithm, however apparently Prim's algorithm can be faster when you have a dense graph
with many more edges than vertices:
https://www.quora.com/What-is-the-difference-in-Kruskals-and-Prims-algorithm

This pull request uses the Boost Graph Library `prim_minimum_spanning_tree` algorithm in place of `kruskal_minimum_spanning_tree`.

Comparative timings:

| #ref | Kruskal (s) | Prim (s) |
| -----| ---------- | -------- |
| 82| 0.001 | 0.001 |
| 229| 0.005 | 0.002 |
| 592| 0.015 | 0.008 |
| 1568| 0.035 | 0.028 |
| 4609| 0.047 | 0.041 |
| 13890| 0.169 | 0.131 |
| 42677| 0.665 | 0.461 |
| 176019| 3.472 | 2.303 |
| 520986| 13.672 | 10.350 |
| 1719853| 69.767 | 36.060 |

### DIFF
--- a/algorithms/indexing/index.h
+++ b/algorithms/indexing/index.h
@@ -22,7 +22,7 @@
 #include <dials/array_family/scitbx_shared_and_versa.h>
 #include <dials/error.h>
 #include <boost/graph/adjacency_list.hpp>
-#include <boost/graph/kruskal_min_spanning_tree.hpp>
+#include <boost/graph/prim_minimum_spanning_tree.hpp>
 #include <boost/graph/depth_first_search.hpp>
 #include <boost/property_map/property_map.hpp>
 #include <boost/graph/connected_components.hpp>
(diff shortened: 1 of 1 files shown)


dagewa:
The timings look good. I suppose you always get the same result?

rjgildea:
> The timings look good. I suppose you always get the same result?

`pytest algorithms/indexing/test_assign_indices.py` passes 🙂

rjgildea:
To satisfy my own curiosity, I wrote a simple [script](https://github.com/dials/dials_scratch/blob/master/rjg/index_assignment_robustness.py) to test the robustness of the `global` and `local` index assignment algorithms against incorrect beam centre. This showed that `local` index assignment has a much larger radius for successfully assigning correct indices than the `global` method. Comparison between `master` (Kruskal's MST) and `local_indexing` (Prim's MST) branches showed identical results.
```
$ dials.python $(libtbx.find_in_repositories dials_scratch)/rjg/index_assignment_robustness.py $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.{expt,refl}
```
image

PR #948, opened 2019-09-26
Allow choice of threshold algorithms

dagewa:
The two dispersion algorithms share the same parameters but give
different results. This change allows the user to explore those
differences in the image viewer and to save out find_spots.phil
parameters including the algorithm choice.


dagewa:
As as aside, I noticed that a `DispersionExtendedThresholdDebug` (or `DispersionThresholdDebug`) is created every time one of the image type buttons is clicked, which contributes to slow navigation between the types of images. It would be better to create this only when the spot finding parameters are changed, and use the result to provide access to the various types of images, which are then already calculated.

PR #951, opened 2019-10-01
Add systematic absences checking for dials.symmetry.

jbeilstenedmands:
Incorporate systematic absences code into dials.symmetry.

Off by default (for now), trigger with check_absences=True.

Currently only checks for screw axes in MX space groups, hopefully will implement all space groups in near future.


dagewa:
Just wondering, is there a good reason to have `check_absences=False` as default for now? It's not going to cause any problems to go straight for `check_absences=True`, right?

graeme-winter:
Agree with comments from @rjgildea as this makes it more similar to POINTLESS, which is what the end user will likely be expecting...

jbeilstenedmands:
Sounds good, will work on implementing this and also changes to xia2 in separate PR

rjgildea:
Trying to review this now:
```
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl}
  File "/Users/rjgildea/software/cctbx/build/../modules/dials/command_line/symmetry.py", line 26, in <module>
    from dials.algorithms.symmetry.absences.run_absences_checks import (
ImportError: No module named run_absences_checks
Richards-MacBook-Pro-2:311 rjgildea$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl}
  File "/Users/rjgildea/software/cctbx/build/../modules/dials/command_line/symmetry.py", line 26, in <module>
    from dials.algorithms.symmetry.absences.run_absences_checks import (
ImportError: No module named run_absences_checks
```

I also note that this now has some conflicts - would it be possible to rebase onto master resolving conflicts?

command_line/symmetry.py
```diff
@@ -55,6 +65,21 @@
   .type = float
   .help = "Use only reflections with a partiality above this threshold."
 
+mode = *laue_plus_absences laue_only absences_only
+  .type = choice
```
rjgildea: I wonder if this parameter could be improved slightly, as the current options seem a bit verbose. Two possibilities:
1)
```
mode = *laue *absences
  .type = choice(multi=True)
```
This would translate to the following command lines:
```
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode=absences
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode=laue
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode="*absences *laue"
```
The last line here is a bit cumbersome as it necessitates both quoting and selecting both modes with an asterisk (although this is the default anyway).

2)
```
mode = *full laue absences
  .type = choice
```
Which translates to the following command lines:
```
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode=absences
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode=laue
$ dials.symmetry DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl} mode=full
```

jbeilstenedmands: I like option 2 :)

rjgildea: Another alternative would be:
```
laue_group = auto
  .type = space_group
systematic_absences = True
  .type = bool
```
Command lines:
```
# Use input laue_group, do systematic absences check
$ dials.symmetry scaled.{expt,refl} laue_group=None systematic_absences=True

# Use laue_group from command line, do systematic absences check
$ dials.symmetry scaled.{expt,refl} laue_group=P222 systematic_absences=True

# Switch off systematic absences check
$ dials.symmetry scaled.{expt,refl} systematic_absences=False

# Laue group + systematic absences check (default)
$ dials.symmetry scaled.{expt,refl} laue_group=auto systematic_absences=True
```
This option would be more closely aligned with the [LAUEGROUP](http://www.ccp4.ac.uk/html/pointless.html#lauegroup) and [SYSTEMATICABSENCE](http://www.ccp4.ac.uk/html/pointless.html#sysabs) keywords in pointless.

command_line/symmetry.py
```diff
@@ -64,87 +95,147 @@
...
+        logger.info("")
+
+        if (params.d_min is Auto) and (result is not None):
+            d_min = result.intensities.resolution_range()[1]
+        elif params.d_min is Auto:
+            d_min = resolution_filter_from_reflections_experiments(
+                reflection_tables,
+                experiments,
+                params.min_i_mean_over_sigma_mean,
+                params.min_cc_half,
```
rjgildea: Making a note for myself: we should probably only need to do the resolution filtering once at the beginning, rather than once before Laue group determination and once before systematic absences checks.

command_line/symmetry.py
```diff
@@ -64,87 +95,147 @@
...
+    experiments.as_file(params.output.experiments)
+    if params.laue_group is not None:
         logger.info(
             "Saving %s reindexed reflections to %s",
-            len(reindexed_reflections),
-            self._params.output.reflections,
+            len(reflection_tables[0]),
+            params.output.reflections,
+        )
+        reflection_tables[0].as_file(params.output.reflections)
```
rjgildea: Is it a potential cause for confusion that a reflection file is only exported `if params.laue_group is not None`? Since in a lot of cases the output reflections will be identical to the input file anyway if the cb_op is identity, would it not be simplest just to always re-export the reflections? We could always check if the `params.output.reflections is not None` to allow the user to choose not to re-export the reflections.

jbeilstenedmands: Yes I think that's a good idea to keep a consistent behaviour between options, I'll update

jbeilstenedmands:
I've realised there is an issue with how we treat multi-dataset cases, as currently the code is only doing systematic absences checks on the first dataset. Although I'm not clear how we should be treating this for unscaled data.
For scaled data, assuming all data has been scaled together, then you just merge and do the test on the merged data. For integrated data, for a single dataset, merging for the absences test is probably okay as differences in scale factors across a dataset are typically not huge, but what about for multiple datasets, where there could be large differences in scale?

graeme-winter:
This is a completely reasonable question. 

I think we should do it independently on the individual sweeps then form a synthesis of the results - would this make sense?

rjgildea:
> I've realised there is an issue with how we treat multi-dataset cases, as currently the code is only doing systematic absences checks on the first dataset. Although I'm not clear how we should be treating this for unscaled data.
> For scaled data, assuming all data has been scaled together, then you just merge and do the test on the merged data. For integrated data, for a single dataset, merging for the absences test is probably okay as differences in scale factors across a dataset are typically not huge, but what about for multiple datasets, where there could be large differences in scale?

Isn't normalisation supposed to correct for this?

graeme-winter:
As Phil would say, you end up having to scale the data before you can perform the calculations in one way or another. 

That said, if we are looking for the difference between there / not there _probably_ the overall scales will have little influence.

jbeilstenedmands:
I've updated the code so that the absence analysis is consistent for all data input types - scaled, or unscaled, single or multi-dataset. The absences analysis is performed on all datasets at once. Normalisation isn't appropriate as the absence check is done on I/sigma. Also that implies that for large scale differences between datasets there should be limited effect as I/sigma is invariant. We could think about doing the test on individual datasets then combining (but then it's not clear how to deal with disagreements), but I think we should merge as in and then can investigate this later as a lower priority.

jbeilstenedmands:
I've updated the code so that the absence analysis is consistent for all data input types - scaled, or unscaled, single or multi-dataset. The absences analysis is performed on all datasets at once. Normalisation isn't appropriate as the absence check is done on I/sigma. Also that implies that for large scale differences between datasets there should be limited effect as I/sigma is invariant. We could think about doing the test on individual datasets then combining (but then it's not clear how to deal with disagreements), but I think we should merge as in and then can investigate this later as a lower priority.

PR #952, opened 2019-10-01
Simplify usage of flex.reflection_table methods.

rjgildea:
`flex.reflection_table.centroid_px_to_mm()` and `flex.reflection_table.map_centroids_to_reciprocal_space()` now take an `ExperimentList` as input instead of individual dxtbx models, which simplifies usage, removing the need to iterate over experiments/imagesets everywhere the methods are used.

### DIFF
--- a/array_family/flex.py
+++ b/array_family/flex.py
@@ -14,6 +14,8 @@ import dials_array_family_flex_ext
 import libtbx.smart_open
 import six
 import six.moves.cPickle as pickle
+from dials.algorithms.centroid import centroid_px_to_mm_panel
+
 from dials.util import Sorry
 from scitbx import matrix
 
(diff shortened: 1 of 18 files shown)


phyy-nx:
Hi, this will affect xfel/small_cell/small_cell.py and xfel/command_line/detector_residuals.py. I can't future proof this one so a concurrent pull request needs to be prepared.

array_family/flex.py
```diff
@@ -14,6 +14,8 @@
 import libtbx.smart_open
 import six
 import six.moves.cPickle as pickle
+from dials.algorithms.centroid import centroid_px_to_mm_panel
+
```
Anthchirp: empty line should be in front of the `from dials ...` :)

PR #955, opened 2019-10-04
Use a mask from dxtbx in dials.rs_mapper

biochem-fan:
See discussion in https://github.com/cctbx/dxtbx/issues/102. I also removed FIXMEs and added more checks.

This is my first commit since the introduction of new coding guidelines. I made sure pre-commit hooks do not complain, but feel free to improve to code. A long standing concern since I wrote this program five years ago is the lack of tests. What kind of tests is appropriate for this program?

If there are interests, it is trivial to improve the speed of this program by parallelization.

### DIFF
--- a/algorithms/rs_mapper/ext.cpp
+++ b/algorithms/rs_mapper/ext.cpp
@@ -23,12 +23,10 @@ namespace recviewer { namespace ext {
       for (size_t y = 0; y < ylim; y++) {
         xy[0] = x;
         xy[1] = y;
-        try {
-          if (panel.get_resolution_at_pixel(s0, xy) > maxres) {
-            ret.push_back(xy);
-          }
-        } catch (...) {
-          // FIXME: How can I avoid this? Want to ignore beam center where res = INF.
+
+        // get_resolution_at_pixel() no longer returns INF, so this is safe
+        if (panel.get_resolution_at_pixel(s0, xy) > maxres) {
+          ret.push_back(xy);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


biochem-fan:
Thanks for reviewing.

> run dials.rs_mapper against images with a mask

Do you have a nice test dataset? 

For example we might be able to use the one in https://github.com/cctbx/dxtbx/blob/master/tests/format/test_FormatHDF5EigerNearlyNexusSPring8.py, but I don't know how long a test may take and I don't have access the DLS cluster.

> parallel execution would be useful to add

Sure, I will work on it when I have time.

> Sometimes it is difficult to get the maps from rs_mapper to display well in Coot or Pymol.

The most common cause of this is fixed by this PR; huge numbers in gap pixels of PILATUS/EIGER are now masked out and do not skew intensity histograms.

dagewa:
There is a small dataset that can be used in a test:
```bash
dials.import $(dials.data get -q thaumatin_eiger_screen)/Therm_6_1_master.h5
dials.rs_mapper imported.expt ignore_mask=true
mapdump mapin rs_mapper_output.mrc <<<GO | grep "Maximum density"
```
gives
```
           Maximum density ................................. 65535.00000
```
while
```
dials.rs_mapper imported.expt
mapdump mapin rs_mapper_output.mrc <<<GO | grep "Maximum density"
```
gives
```
           Maximum density .................................  6330.33350
```

I'm happy to add this test myself.

biochem-fan:
Great! Thank you very much for merging.

PR #956, opened 2019-10-05
Fix change of basis operation in dials.symmetry

benjaminhwilliams:
After correctly determining the best space group and unit cell parameters, `dials.symmetry` sometimes then writes incorrect ones to the output experiment list file.

I think that for historical reasons, the change-of-basis operation from the input unit cell to the best solution unit cell was previously of the form
(minimum ↦ best) ∘ (input ↦ minimum)
(where minimum ≡ primitive perhaps?), though (minimum ↦ best) was misleadingly labeled `cb_op_inp_best`, more suggestive of the form (input ↦ best).

Now that `cb_op_inp_best` really is (input ↦ best), the change-of-basis operation is incorrect and needs tidying up.  In its current form, it results in an incompatibility between the final space group and an erroneous unit cell for some combinations of input and best symmetries.  This goes undetected because, before saving to the experiment list file, the unit cell is always automatically 'corrected' to an average of the set of unit cells generated by application of each of the symmetry operations in the space group to the specified (potentially erroneous) unit cell.  Equivalently, the unit cell is forced to abide by the symmetry of a space group that is correct except that has a different basis to the cell.

The symptom of this is that the correctly reported unit cell in the `dials.symmetry` log file may be at odds with an incorrect-symmetry-averaged unit cell stored in the experiment list file.  For example, a monoclinic cell with space group P2/m (which, in the conventional setting P 1 2/m 1, would have β ≠ 90°) is assigned the correct space group in the conventional setting (P 1 2/m 1) but a unit cell consistent with the unconventional setting (P 1 1 2/m, with γ ≠ 90°), due to the incorrect change-of-basis operation.  Before saving to the experiment list, the unit cell is then forcibly made consistent by applying the correct space group symmetry to the incorrect unit cell and taking an average.  This results in a unit cell that is consistent with the space group symmetry (γ becomes 90° and b & c are altered accordingly), but is clearly not correct.

This slipped through the net because the `dials.symmetry` tests are very basic and the only tested system happens to be one in which (input ↦ minimum) = 𝟙, so the problem does not manifest.  Also, this problem took a long time to debug because the sparseness of docstrings and comments in `sgtbx` (which we cannot help) is matched here in this corner of DIALS.

In addition to patching the dodgy change-of-basis operator in `dials.command_line.symmetry`, I propose that @jbeilstenedmands, @rjgildea and I
  - add some proper tests and
  - add some decent inline developer documentation to make sense, for newcomers and oldcomers alike, of what the code is trying to represent.

Being something of a bumbling idiot in this area, I could do with a little guidance as to what's needed in terms of tests and what the code is actually trying to do.

### DIFF
--- a/test/command_line/test_symmetry.py
+++ b/test/command_line/test_symmetry.py
@@ -3,6 +3,10 @@ from __future__ import absolute_import, division, print_function
 import os
 
 import procrunner
+import pytest
+from cctbx import sgtbx
+
+from dials.array_family import flex
 
 
 def test_symmetry(dials_regression, tmpdir):
@@ -21,3 +25,44 @@ def test_symmetry(dials_regression, tmpdir):
     assert not result.returncode and not result.stderr
     assert tmpdir.join("symmetrized.refl").check()
     assert tmpdir.join("symmetrized.expt").check()
+
+
+from dials.algorithms.symmetry.cosym._generate_test_data import (
+    generate_experiments_reflections,
+)
+
+
+def test_symmetry_R3m_I2m(tmpdir):
+    os.chdir(tmpdir.strpath)
+    space_group = "C 2"
+    unit_cell = (53.173, 61.245, 69.292, 90.0, 93.04675, 90.0)
+    space_group = sgtbx.space_group_info(space_group).group()
+    experiments, reflections, _ = generate_experiments_reflections(
+        space_group=space_group,
+        unit_cell=unit_cell,
+        sample_size=1,
+        map_to_minimum=False,
+    )
+    experiments.as_json("tmp.expt")
+    expt_file = tmpdir.join("tmp.expt").strpath
+    joint_table = flex.reflection_table()
+    for r in reflections:
+        joint_table.extend(r)
+    joint_table.as_pickle("tmp.refl")
+    refl_file = tmpdir.join("tmp.refl").strpath
+
+    # okay so generated in C 1 m 1, but lattice looks like R-3m, s*o reindex
+    # to this first
+
+    command = ["dials.symmetry", expt_file, refl_file]
+    result = procrunner.run(command, working_directory=tmpdir.strpath)
+    assert not result.returncode and not result.stderr
+    assert tmpdir.join("symmetrized.refl").check(file=1)
+    assert tmpdir.join("symmetrized.expt").check(file=1)
+    from dxtbx.serialize import load
+
+    expts = load.experiment_list(
+        tmpdir.join("symmetrized.expt").strpath, check_format=False
+    )
+    for v, expected in zip(expts[0].crystal.get_unit_cell().parameters(), unit_cell):
+        assert v == pytest.approx(expected)
--- a/algorithms/symmetry/cosym/_generate_test_data.py
+++ b/algorithms/symmetry/cosym/_generate_test_data.py
@@ -20,6 +20,7 @@ def generate_experiments_reflections(
     sample_size=100,
     map_to_p1=False,
     twin_fractions=None,
+    map_to_minimum=True,
 ):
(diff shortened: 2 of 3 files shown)


jbeilstenedmands:
These changes look good to me, and I think the idea of setting the space group as P1 first before applying basis changes is the right approach. For adding tests, it should be possible to use/adapt some functions in `algorithms/symmetry/cosym/_generate_test_data.py` to generate a range of inputs with different unit cells. I tried to start with a complex example and got lost, so starting simple is probably the best way to go :)

jbeilstenedmands:
I've just added a test here for starters, using the unit cell of beta lactamase, which is C2 space group and also undergoes the issue that @benjaminhwilliams was seeing last week. This test fails on master and passes on this branch. I could do with some help with the test dataset generation as it appears to only generate in the asu at the moment.

test/command_line/test_symmetry.py
```diff
@@ -21,3 +25,44 @@ def test_symmetry(dials_regression, tmpdir):
...
+from dials.algorithms.symmetry.cosym._generate_test_data import (
+    generate_experiments_reflections,
+)
+
+
+def test_symmetry_R3m_I2m(tmpdir):
+    os.chdir(tmpdir.strpath)
+    space_group = "C 2"
+    unit_cell = (53.173, 61.245, 69.292, 90.0, 93.04675, 90.0)
+    space_group = sgtbx.space_group_info(space_group).group()
```
benjaminhwilliams: Would `sgtbx.space_group(space_group)` not suffice here?  Equivalent but a little clearer.  I've never been entirely clear what `sgtbx.space_group_info` is really for...  It's some ancient bit of Ralf's stuff and it has no docstring.

benjaminhwilliams: Oh no, hang on, there's some weirdness here:
```Python
>>> import sgtbx
>>> print(sgtbx.space_group('C 2').type().universal_hermann_mauguin_symbol())
P 1 2 1 (c,2*a+c,b)
>>> print(sgtbx.space_group_info('C 2').group().type().universal_hermann_mauguin_symbol())
C 1 2 1
```

graeme-winter: to the uninitiated this smells like a bug?

graeme-winter: Anyhow - for the purposes of the calculation the long winded route through `space_group_info` _works_ so we are not in a bad place?

graeme-winter:
I think this will fix a problem I have outstanding right now on master - with data in symmetry F m -3 m where the processing comes out as complete nonsense, which should be fixed by this branch.

benjaminhwilliams:
OK, I think we can leave the campsite tidying to a subsequent PR.  I've raised an issue to that end.

graeme-winter:
This makes the processing work very nicely:

```
For AUTOMATIC/DEFAULT/NATIVE                Suggested   Low    High  Overall
High resolution limit                           0.60    1.62    0.60    0.60
Low resolution limit                            5.58    5.58    0.61    5.58
Completeness                                  100.0   100.0   100.0   100.0
Multiplicity                                   31.0    36.5     5.8    30.7
I/sigma                                        86.2   140.7    21.2    85.1
Rmerge(I)                                     0.053   0.048   0.063   0.053
Rmeas(I)                                      0.054   0.049   0.068   0.054
Rpim(I)                                       0.010   0.010   0.025   0.010
CC half                                       0.999   0.999   1.000   0.999
Total observations                             6608     620      58    6622
Total unique                                    213      17      10     216
Assuming spacegroup: F m -3 m
Unit cell (with estimated std devs):
11.16797(12) 11.16797(12) 11.16797(12)
90.0         90.0         90.0        

```

PR #957, opened 2019-10-07
remove dead code

Anthchirp:


phyy-nx:
Hi, please restore example_frame_plugin_hidden.py.  We use the plugin method all the time for kapton correction parameter tweaking for the xfel tape drive and the example of how to make more plugins is useful.

Thanks.

Anthchirp:
Would renaming it to .txt work for you?

phyy-nx:
I guess?  Curious as to why?

dagewa:
I've not looked at the example, but just wondering would it be possible to write it as a test instead - in the spirit of tests being the best documentation?

phyy-nx:
Mmmm... It's a gui widget, and GUIs are notoriously hard to write tests for.

phyy-nx:
(not that we shouldn't have a comprehensive set of tests for our GUIs, just
noting that such a thing is tricky :)

Anthchirp:
> I guess? Curious as to why?

because it is currently covered by static code analysis and raises code duplication alerts there. In addition to that it skews code coverage checks.

The file remains of course available at https://github.com/cctbx/cctbx_project/blob/master/rstbx/slip_viewer/example_frame_plugin_hidden.py so maybe a reference to that would be sufficient?

phyy-nx:
Gotcha.  Renaming to txt seems like a good solution then.

phyy-nx:
👍

PR #967, opened 2019-10-16
Refactor index command line program

rjgildea:
- Simple function instead of class
- Use sys.exit instead of Sorry

### DIFF
--- a/command_line/index.py
+++ b/command_line/index.py
@@ -2,8 +2,10 @@ from __future__ import absolute_import, division, print_function
 
 # DIALS_ENABLE_COMMAND_LINE_COMPLETION
 
+import concurrent.futures
 import copy
 import logging
+import sys
 
 import iotbx.phil
 from dxtbx.model.experiment_list import ExperimentList
@@ -14,7 +16,8 @@ from dials.util.slice import slice_reflections
 from dials.util.options import OptionParser
 from dials.util.options import flatten_reflections
 from dials.util.options import flatten_experiments
-from dials.util import Sorry
+from dials.util import log
+from dials.util.version import dials_version
 
 logger = logging.getLogger("dials.command_line.index")
 
@@ -106,7 +109,7 @@ refinement {
 working_phil = phil_scope.fetch(sources=[phil_overrides])
 
 
-def index_experiments(experiments, reflections, params, known_crystal_models=None):
+def _index_experiments(experiments, reflections, params, known_crystal_models=None):
     idxr = indexer.Indexer.from_parameters(
         reflections,
         experiments,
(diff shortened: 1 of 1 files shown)


command_line/index.py
```diff
@@ -119,106 +122,81 @@ def index_experiments(experiments, reflections, params, known_crystal_models=Non
...
+    reflections = reflections[0]
+
+    for expt in experiments:
+        if (
+            expt.goniometer is not None
+            and expt.scan is not None
+            and expt.scan.get_oscillation()[1] == 0
+        ):
+            expt.goniometer = None
+            expt.scan = None
```
benjaminhwilliams: Can we have a comment above that describes in plain English the case that this is designed to catch?

command_line/index.py
```diff
@@ -119,106 +122,81 @@ def index_experiments(experiments, reflections, params, known_crystal_models=Non
...
+        known_crystal_models = experiments.crystals()
+    else:
+        known_crystal_models = None
+
+    if len(reflections) == 0:
+        raise ValueError("No reflection lists found in input")
+    elif len(reflections) == 1:
+        reflections[0]["imageset_id"] = reflections[0]["id"]
+    elif len(reflections) > 1:
+        assert len(reflections) == len(experiments)
```
benjaminhwilliams: Is it not also necessary to make this assertion if `len(reflections) == 1`?

command_line/index.py
```diff
@@ -119,106 +122,81 @@ def index_experiments(experiments, reflections, params, known_crystal_models=Non
...
-                    refl["imageset_id"] = flex.size_t(len(refl), 0)
-                    futures.append(
-                        pool.submit(
-                            index_experiments,
-                            ExperimentList([expt]),
-                            refl,
-                            copy.deepcopy(params),
-                            known_crystal_models=known_crystal_models,
-                        )
+def index(experiments, reflections, params):
```
benjaminhwilliams: DIALS and xia2 are riddled with functions called `index`, `indexer`, etc..  Here, it's definitely appropriate, but to make it clear what _this_ function does (it's the main one after all), I think it would be quite handy to have a docstring.

command_line/index.py
rjgildea: No, because in this case the input could multiple experiments, but with the reflections corresponding to those experiments concatenated into a single reflection table (rather than a list of reflection tables one per experiment).

benjaminhwilliams: Ah, OK, thanks.

command_line/index.py
benjaminhwilliams: @dagewa, I see you've been digging around here today.  Any chance you could distil your new wisdom into an explanatory comment?

PR #970, opened 2019-10-17
Fix indexing of stills for ED

graeme-winter:
If you are going to delete goniometer and scan (which I would prefer you did not) convert to an ImageSet from an ImageSequence

Fixes tests as restores previous behaviour. Later discussion as to whether this is _correct_


dagewa:
Just for the record, not ED-specific!

command_line/index.py
```diff
@@ -147,6 +147,10 @@ def __init__(self, experiments, reflections, params):
                 and expt.scan is not None
                 and expt.scan.get_oscillation()[1] == 0
             ):
+                # if removing things an ImageSequence needs, change to ImageSet
+                from dxtbx.imageset import ImageSetFactory
```
rjgildea: Probably should float import to top of file.

dagewa:
I found I can trigger the problem also like this:
```
dials.import $(dials.data get -q x4wide)/X4_wide_M1S4_2_0001.cbf\
  scan.oscillation=0,0
dials.index strong.refl imported.expt stills.indexer=sequences
dials.show indexed.expt 
```

phyy-nx:
2 reasons as to why the scan/gonio are deleted during indexing stills (these reasons may no longer apply).  1) There are assumptions in place in the stills indexer/integrator that lack of a scan/gonio is the definition of whether something is a still.  Deleting them early ensures everything downstream knows what is going on.  For example, it guarantees the stills integrator is used for a sweep with the oscillation angle set to zero, which should be treated as a still.  2) It ensures that any code that assumes a sweep will fail since the gonio/scans are None. This is more of a safety check.

If these reasons are no longer applicable, we can discuss not deleting the scan/gonio, as long as the stills indexer and integrator remain automatically triggered for oscillation=0.

Issue #938, opened 2019-09-17
volume integrator: (fix && add tests) || remove


Anthchirp:
flake8 tells me that nobody is using the volume integrator. So should we remove it?

How does flake8 know? Well, the relevant class
https://github.com/dials/dials/blob/b0ca531d3784f935a54d5642755ae8d510ceb3b8/algorithms/integration/integrator.py#L1623-L1624
is called `IntegratorVolume`, which
https://github.com/dials/dials/blob/b0ca531d3784f935a54d5642755ae8d510ceb3b8/algorithms/integration/integrator.py#L1352-L1357
is derived from (or shall we say: identical to) `ImageIntegrator`, which references `Parameters` in `__init__()`:
https://github.com/dials/dials/blob/b0ca531d3784f935a54d5642755ae8d510ceb3b8/algorithms/integration/image_integrator.py#L607
The only issue here is that `Parameters` is neither defined nor imported in the file, so this constructor is an unlikely to succeed.

dagewa:
Correct
```
$ dials.integrate indexed.expt indexed.refl integrator=volume

...

Traceback (most recent call last):
  File "/home/david/sw/cctbx/build/../modules/dials/command_line/integrate.py", line 664, in <module>
    script.run()
  File "/home/david/sw/cctbx/build/../modules/dials/command_line/integrate.py", line 343, in run
    integrator = IntegratorFactory.create(params, experiments, predicted)
  File "/home/david/sw/cctbx/modules/dials/algorithms/integration/integrator.py", line 1636, in create
    return IntegratorClass(experiments, reflections, params)
  File "/home/david/sw/cctbx/modules/dials/algorithms/integration/image_integrator.py", line 577, in __init__
    self.params = Parameters.from_phil(params.integration)
NameError: Please report this error to dials-support@lists.sourceforge.net: global name 'Parameters' is not defined
```

This looks a lot like incomplete work. I would suggest to remove.

PR #975, opened 2019-10-18
Remove volume integrator

Anthchirp:
Given that the class constructor has been broken for nearly 2 years (since 62affd0737b0e0d34696316ce87f4b2faba84832) I guess it is safe to say that nobody will miss it?


Anthchirp:
Turns out I found this before and DGW suggested in #938 that we should remove, so removing.

Issue #978, opened 2019-10-21
dials.generate_mask trusted_range bug


dagewa:
The behaviour of `dials.generate_mask` appears to be to combine untrusted area definitions with the un`trusted_range` mask of the first image of a sequence. This causes overloaded spots on image 1 to be persistently masked throughout a dataset, like a flash burn. This example makes it obvious:
```bash
cat > mask.phil <<+
untrusted {
  rectangle = 0 1375 1246 1300
}
+
dials.import $(dials.data get -q x4wide)/*.cbf trusted_range=-1,100
dials.generate_mask imported.expt mask.phil
dials.import $(dials.data get -q x4wide)/X4_wide_M1S4_2_00*.cbf mask=pixels.mask
```
Now see the strong pixels from the first image masked throughout the whole dataset:
![Screenshot from 2019-10-21 21-15-15](https://user-images.githubusercontent.com/11502083/67241275-0a0f2500-f44b-11e9-84f2-42fabd02a4e5.png)

dagewa:
Probable culprit https://github.com/dials/dials/blob/9780dfd90bc729dd3aa901a01edcfca436df18c7/util/masking.py#L144-L145

graeme-winter:
Good spot. Suspect that this was an attempt in $past to find and mask hot pixels. Ouch.

graeme-winter:
Verified bug but pointy finger of blame misplaced... `print(1/0)` just after this line does not cause crash.

graeme-winter:
```
Silver-Surfer util :) [master] $ rm -rfv masking
masking/__pycache__/__init__.cpython-36.pyc
masking/__pycache__
masking/__init__.pyc
masking
```

- explains condusion - d'oh

graeme-winter:
Branched to `trusted-pixel-range-mask` for fix.

graeme-winter:
We should discuss how this interacts with multi-panel detectors. Also for large image sets could have better way of doing this e.g. with random sampling?

dagewa:
I don't understand - I still see the same behaviour with the test case above while on the trusted-pixel-range-mask branch.

graeme-winter:
```
  551  dials.import $(dials.data get -q x4wide)/*.cbf trusted_range=-1,100
  552  dials.generate_mask imported.expt mask.phil
  553  dials.import $(dials.data get -q x4wide)/X4_wide_M1S4_2_00*.cbf mask=pixels.mask
  554  dials.image_viewer imported.expt 
```

[...]

PR #980, opened 2019-10-22
Putative fix for dials.generate_mask trusted_range bug (#978)

graeme-winter:
Loop is efficient for single-panel detectors and resolves the issue as published. Will be expensive for large image sets but that is the expected behaviour I would expect. Could do with refactoring for multi-panel detectors as the loop is inside out.

Draft PR to allow discussion as to best solution.

### DIFF
--- a/test/command_line/test_generate_mask.py
+++ b/test/command_line/test_generate_mask.py
@@ -1,10 +1,11 @@
 from __future__ import absolute_import, division, print_function
 
 import six.moves.cPickle as pickle
-import procrunner
 import pytest
 
-from dials.command_line.generate_mask import generate_mask, phil_scope
+from dials.command_line.generate_mask import generate_mask, run, phil_scope
+from dials.command_line.dials_import import Script as ImportScript
+from dials.command_line.dials_import import phil_scope as import_phil_scope
 from dxtbx.serialize import load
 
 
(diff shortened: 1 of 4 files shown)


rjgildea:
Could we add a regression test to cover this?

graeme-winter:
Yeh, I can write one once we’re happy with the changes.

dagewa:
If I read the commit right, this masks pixels that are consistently outside the trusted range over the whole image set, right? If so, could the loop terminate early as soon as it is determined that there are no pixels that are always outside the trusted range?

graeme-winter:
@dagewa that is true, but if there is one which was always outside the trusted range for frames 1...N-1 we don't know if frame N has it as OK

graeme-winter:
But can count, and if all True obviously break - good suggestion

dagewa:
I added a regression test, so I'm marking this as ready for review.

util/masking.py
```diff
@@ -149,11 +149,22 @@ def generate(self, imageset):
...
         for index, (im, panel) in enumerate(zip(image, detector)):
 
-            # Create the basic mask from the trusted range
             if self.params.use_trusted_range:
+                trusted_mask = None
                 low, high = panel.get_trusted_range()
-                imd = im.as_double()
-                mask = (imd > low) & (imd < high)
+                trusted_mask = None
+                for image_index in imageset.indices():
```
rjgildea: Could we include a comment here explaining how the trusted range is determined in people-words (this description would probably also be worth including in the commit message when the PR is squash/merged)?

test/command_line/test_generate_mask.py
```diff
@@ -153,3 +153,48 @@ def test_generate_mask_function_with_untrusted_rectangle(input_experiment_list,
...
+    # https://github.com/dials/dials/issues/978
+    image_files = dials_data("x4wide").listdir("*.cbf", sort=True)
+
+    # Import as usual
+    procrunner.run(
+        ["dials.import", "trusted_range=-1,100", "output.experiments=no-overloads.expt"]
+        + [f.strpath for f in image_files],
+        working_directory=tmpdir,
+    )
+    procrunner.run(
```
rjgildea: I think it was discussed recently that it might be better to avoid the additional `procrunner.run()` level of indirection and instead call e.g. `dials.command_line.generate_mask.run()` directly. See e.g. https://github.com/dials/dials/commit/456755e334800c5bc2b5301c5300081c883f113c#diff-a57154b54de916bc084f9ff480d38c61R58

PR #989, opened 2019-10-28
Dials util mtz

graeme-winter:
Copy `iotbx.mtz.__init__` in here and change references - means we can change the source of this later on to prevent further issues in PHENIX. I think I have found all the places in the code base where this is used. N.B. further commits to follow as (unsurprisingly) there are many flake8 errors in this code.

### DIFF
--- a/command_line/cluster_unit_cell.py
+++ b/command_line/cluster_unit_cell.py
@@ -51,13 +51,12 @@ def run(args):
     if len(experiments) == 0:
         if len(args):
             import os
-            from iotbx.reflection_file_reader import any_reflection_file
+            from dials.util import mtz
 
             for arg in args:
                 assert os.path.isfile(arg), arg
-                reader = any_reflection_file(arg)
-                assert reader.file_type() == "ccp4_mtz"
-                arrays = reader.as_miller_arrays(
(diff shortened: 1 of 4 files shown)


graeme-winter:
Flake8 complaints: 

- util/mtz/__init__.py:10:1: F403 'from iotbx_mtz_ext import *' used; unable to detect undefined names
- util/mtz/__init__.py:10:1: F401 'iotbx_mtz_ext.*' imported but unused
- util/mtz/__init__.py:27:11: E401 multiple imports on one line
- util/mtz/__init__.py:521:13: F841 local variable 'isym' is assigned to but never used
- util/mtz/__init__.py:1365:5: F841 local variable 'uppercase' is assigned to but never used

graeme-winter:
Once we have moved everything we need over to this we can revert the change set in the `cctbx_project` branch in https://github.com/cctbx/cctbx_project/pull/391

graeme-winter:
Something which is a shame here is that the `flake8` fixes above are real and avoid nasty aliasing issues - genuine campsite cleaning which would have improved the basic cctbx code base ¯\_(ツ)_/¯

Issue #994, opened 2019-11-01
dials.refine: if multiple experiments in input with 1 crystal and scan varying, please duplicate not crash


graeme-winter:
If I index a multi-sweep data set with a single crystal - as common for e.g. chemical crystallography - when I come to refine with `dials.refine indexed.* scan_varying=auto` it complains with:

```
Sorry: A single scan-varying crystal model cannot be refined when associated with more than one scan or goniometer
```

Obviously what I may expect is for the crystal to be duplicated and then `scan_varying` refinement be performed. 

At the moment we handle this in dials.index with `split_experiments=true` (a hack I added) but @rjgildea suggested this is better handled in refinement and I am minded to agree.

Example data attached. 

[indexed.zip](https://github.com/dials/dials/files/3797061/indexed.zip)

graeme-winter:
At the moment at the end of index we have 

```
    # Save experiments
    if params.output.split_experiments:
        logger.info("Splitting experiments before output")
        indexed_experiments = ExperimentList(
            [copy.deepcopy(re) for re in indexed_experiments]
        )
    logger.info("Saving refined experiments to %s" % params.output.experiments)
    assert indexed_experiments.is_consistent()
    indexed_experiments.as_file(params.output.experiments)
```

So it does exactly what it says on the tin - which may or may not be correct / desired, but does "work" - that said adding this behaviour to the _input_ of dials.refine would be ugly. 🤔 this is not obvious. 

Should it -
 - duplicate experiments in dials.index by default as above
 - duplicate only the _crystal_ in dials.index
 - duplicate the _crystal_ on input to dials.refine
 - split the experiment completely as above on dials.refine

I am kinda-sorta tempted to think that no. 3 is probably the correct solution here.

graeme-winter:
WIP in split_crystals_scan_vary branch.

rjgildea:
I think logically this belongs in `dials.refine`. Modifying the _output_ of `dials.index` based upon what you _may_ want to do in downstream programs seems like the wrong way to go about it. If `dials.index` has done joint refinement on the crystal model, then it makes sense for `dials.index` to output a joint crystal model. If this is handled by `dials.refine`, then you would potentially be able to do further joint static refinement, split static refinement, or split scan-varying refinement on the same output files, without having to re-run `dials.index` to accomplish such a task.

graeme-winter:
This is now fixed.

dagewa:
We were going to add a test?
[...]

PR #995, opened 2019-11-01
For #994 - split experiments

graeme-winter:
If len(crystals) == 1 but len(experiments) > 1 split experiments to allow scan varying refinement.

Draft PR to allow discussion of how best to solve this - this change duplicates each model which is probably not 100% what we want but does work. Means detectors are _not_ constrained to be the same (likewise beam) - 

```
Silver-Surfer split :) $ dials.show refined.expt |grep _axis
  fast_axis: {0.00299689,0.86451,0.502607}
  slow_axis: {0.999995,-0.00210866,-0.00233566}
  fast_axis: {0.00223049,0.864408,0.502786}
  slow_axis: {0.999991,-0.000121619,-0.00422713}
  fast_axis: {-0.00445467,0.864301,0.502955}
  slow_axis: {0.999989,0.0044793,0.00115946}
  fast_axis: {0.000307802,0.999999,0.00151444}
  slow_axis: {0.999995,-0.000312514,0.00311193}
```

First three should probably be identical...

### DIFF
--- a/command_line/refine.py
+++ b/command_line/refine.py
@@ -385,11 +385,18 @@ def run(args=None, phil=working_phil):
         )
 
     if params.refinement.parameterisation.scan_varying is not False:
-        # duplicate crystal if necessary for scan varying
-        if len(experiments) > len(experiments.crystals()):
-            logger.info("Splitting crystal models before refinement")
-            for e in experiments:
-                e.crystal = copy.deepcopy(e.crystal)
+        # duplicate crystal if necessary for scan varying - will need
+        # to compare the scans with crystals - if not 1:1 will need to
+        # split the crystals
+
+        crystal_has_scan = {}
+        for j, e in enumerate(experiments):
+            if e.crystal in crystal_has_scan:
+                if e.scan is not crystal_has_scan[e.crystal]:
+                    logger.info("Separating crystal model for experiment %d" % j)
+                    e.crystal = copy.deepcopy(e.crystal)
+            else:
+                crystal_has_scan[e.crystal] = e.scan
 
     # Run refinement
     experiments, reflections, refiner, history = run_dials_refine(
--- a/command_line/refine.py
+++ b/command_line/refine.py
@@ -370,15 +370,6 @@ def run(args=None, phil=working_phil):
 
     logger.info(dials_version())
 
-    if params.refinement.parameterisation.scan_varying is not False:
-        # duplicate crystal if exactly one and > 1 scan
-        if len(experiments) > 1 and len(experiments.crystals()) == 1:
-            logger.info("Splitting crystal models before refinement")
-            c = experiments.crystals()[0]
-            crystals = [copy.deepcopy(c) for e in experiments]
-            for j, e in enumerate(experiments):
-                e.crystal = crystals[j]
-
     # Log the diff phil
     diff_phil = parser.diff_phil.as_str()
     if diff_phil:
@@ -392,6 +383,15 @@ def run(args=None, phil=working_phil):
             "circumstances. It is not recommended for typical data processing "
             "tasks.\n"
         )
+
+    if params.refinement.parameterisation.scan_varying is not False:
+        # duplicate crystal if exactly one and > 1 scan for scan_varying
+        if len(experiments) > 1 and len(experiments.crystals()) == 1:
+            logger.info("Splitting crystal models before refinement")
+            c = experiments.crystals()[0]
+            crystals = [copy.deepcopy(c) for e in experiments]
+            for j, e in enumerate(experiments):
+                e.crystal = crystals[j]
 
     # Run refinement
     experiments, reflections, refiner, history = run_dials_refine(
--- a/command_line/refine.py
+++ b/command_line/refine.py
@@ -15,6 +15,7 @@ Examples::
 """
 
 from __future__ import absolute_import, division, print_function
+import copy
 import sys
 import logging
 from time import time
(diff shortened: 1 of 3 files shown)


graeme-winter:
Figured out how to make this work correctly to only duplicate the crystal if scan_varying is not False 

```
Silver-Surfer split :) $ dials.show refined.expt |grep _axis
  fast_axis: {0.000264975,0.86441,0.502788}
  slow_axis: {0.999999,0.000473157,-0.00134048}
  fast_axis: {0.000264975,0.86441,0.502788}
  slow_axis: {0.999999,0.000473157,-0.00134048}
  fast_axis: {0.000264975,0.86441,0.502788}
  slow_axis: {0.999999,0.000473157,-0.00134048}
  fast_axis: {0.00029133,0.999992,0.00395859}
  slow_axis: {1,-0.000288118,-0.000811937}
```

now the detector etc. are correctly fixed for joint refinement => 🙂

jbeilstenedmands:
Why do we only want to do the 'crystal splitting' if scan varying is True? Even if doing static refinement, don't we want to refine the crystals individually? Would be nice to be consistent between the two cases or else this could lead to extra confusion about the workflow?

graeme-winter:
@jbeilstenedmands yes and no. Yes, because it makes sense what you say. No because this is an API change and maybe just maybe you have a specific interest in refining exactly one static unit cell - am thinking of e.g. small wedge of data 0-3° and 90-93° say and wanting one refined crystal.

But I welcome opinions on this. 

I'm just trying to make a thing work 😉

graeme-winter:
Turns out not splitting the detectors etc. slightly improves the data quality at the end.

command_line/refine.py
```diff
@@ -383,6 +384,13 @@ def run(args=None, phil=working_phil):
             "tasks.\n"
         )
 
+    if params.refinement.parameterisation.scan_varying is not False:
+        # duplicate crystal if necessary for scan varying
+        if len(experiments) > len(experiments.crystals()):
+            logger.info("Splitting crystal models before refinement")
```
dagewa: I don't think this message is very meaningful. What is 'splitting a crystal model'? I would prefer a more verbose but explicit description like `Ensuring a separate crystal model exists for each scan in refinement` or similar.

command_line/refine.py
```diff
@@ -383,6 +384,13 @@ def run(args=None, phil=working_phil):
             "tasks.\n"
         )
 
+    if params.refinement.parameterisation.scan_varying is not False:
+        # duplicate crystal if necessary for scan varying
+        if len(experiments) > len(experiments.crystals()):
```
dagewa: This condition is really being used as a proxy for detecting when a crystal is associated with more than one scan. It may be better to actually test for that, as otherwise this behaviour will also be triggered in situations where you don't want to do this. These situations are mostly hypothetical, but not totally unreasonable. Two detectors for one scan is an example - the two-EIGERs-in-a-V-shape, or a separate high angle detector. The code that includes the `Sorry` message mentioned in #994 already does this.

graeme-winter: @dagewa the correct test is therefore to look to see if #scans > #crystals and split according to this? Since we don't have test coverage for more esoteric combinations (such as I am trying to add support for) I don't know the space of "allowed solutions" - though your 2 detector at the same time combination makes sense.

command_line/refine.py
graeme-winter: Literally hold no opinion here 🙂`dials.refine` output is entirely your call

graeme-winter:
Re: test cases from indexing onwards - not sure best way to compose however here are the input for this: 

[994.zip](https://github.com/dials/dials/files/3802865/994.zip)

```
dials.index strong.refl imported.expt 
```

will do what you want.

graeme-winter:
For provenance - these are derived from https://zenodo.org/record/51405

command_line/refine.py
graeme-winter: Made more explicit:

```
Separating crystal model for experiment 1
Separating crystal model for experiment 2
Separating crystal model for experiment 3
```

command_line/refine.py
graeme-winter: I think I resolved this in the most recent commit - 

```
        # duplicate crystal if necessary for scan varying - will need
        # to compare the scans with crystals - if not 1:1 will need to
        # split the crystals

        crystal_has_scan = {}
        for j, e in enumerate(experiments):
            if e.crystal in crystal_has_scan:
                if e.scan is not crystal_has_scan[e.crystal]:
                    logger.info("Separating crystal model for experiment %d" % j)
                    e.crystal = copy.deepcopy(e.crystal)
            else:
                crystal_has_scan[e.crystal] = e.scan
```

This _should_ duplicate the crystal iff > 1 scan associated with that crystal.

command_line/refine.py
dagewa: I still think `Separating crystal model` is not any more meaningful than `Splitting crystal model` to the user because it does not explain _why_ the action is taken.

command_line/refine.py
dagewa: That looks good to me :+1:

command_line/refine.py
dagewa: I would settle with `Copying crystal model for scan-varying refinement of experiment N`, which at least hints at why, and accurately describes what happens (a copy) rather than 'separating' which is ambiguous.

rjgildea: _Duplicating_ crystal model...?

rjgildea:
> Re: test cases from indexing onwards - not sure best way to compose however here are the input for this:
> 
> [994.zip](https://github.com/dials/dials/files/3802865/994.zip)
> 
> ```
> dials.index strong.refl imported.expt 
> dials.refine indexed.expt indexed.refl
> ```
> 
> will do what you want.

I think we already have these in `dials.data` (albeit with dials 1.X-style file naming):
https://github.com/dials/data/blob/master/dials_data/definitions/l_cysteine_dials_output.yml

command_line/refine.py
graeme-winter: Please make it as verbose as you like - personally I think no user will ever read it anyway but I recognise your territory here @dagewa

graeme-winter:
Going to hold test cases for the moment...

PR #1003, opened 2019-11-07
dials.symmetry: export numbers in JSON

Anthchirp:
export numbers as numbers
instead of strings containing numbers


Anthchirp:
@rjgildea working on it already.

PR #1007, opened 2019-11-11
Datablocks are not a thing - imported experiments are

graeme-winter:
Simple replacement of `datablock.expt` to `imported.expt` which is the default filename from `dials.import`


graeme-winter:
I think this conflicts with other doc updates => should probably close without merging

Sorry

PR #1010, opened 2019-11-14
Parallel bad pixels

graeme-winter:
Made the calculation parallel. Removed progress bars as not sure useful. 

Closeed / compact change set but would welcome a quick peruse if anyone interested. 

@noemifrisina you may find that some of the calculations in here are useful. Essentially it looks at signal pixels and if one is signal for every image it is probably duff.

### DIFF
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -34,7 +34,7 @@ image_range = None
   .help = "Image range for analysis e.g. 1,1800"
 nproc = 1
   .type = int(value_min=1)
-  .help = "Number of processors to use for bad pixel finding"
+    .help = "The number of processes to use."
 
 output {
     mask = pixels.mask
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -136,7 +136,7 @@ def run(args):
 
     if params.images:
         if min(params.images) < first or max(params.images) > last:
-            raise Sorry("image outside of scan range")
+            sys.exit("Image outside of scan range")
         images = params.images
 
     # work around issues with HDF5 and multiprocessing
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -46,7 +46,10 @@ output {
 
 def find_constant_signal_pixels(imageset, images):
     """Find pixels which are constantly reporting as signal through the
-    images in imageset."""
+    images in imageset: on every image the pixel dispersion index is computed,
+    and signal pixels identified using the default settings. A map is then
+    calculated of the number of times a pixel is identified as signal: if this
+    is >= 50% of the images (say) that pixel is untrustworthy."""
 
     panels = imageset.get_detector()
     assert len(panels) == 1
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -33,7 +33,7 @@ image_range = None
   .type = ints(value_min=0, size=2)
   .help = "Image range for analysis e.g. 1,1800"
 nproc = 1
-  .type = int
+  .type = int(value_min=1)
   .help = "Number of processors to use for bad pixel finding"
 
 output {
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -143,15 +143,15 @@ def run(args):
     imageset.reader().nullify_format_instance()
 
     n = int(math.ceil(len(images) / params.nproc))
-    work = [images[i : i + n] for i in range(0, len(images), n)]
-    assert len(images) == sum([len(chunk) for chunk in work])
+    chunks = [images[i : i + n] for i in range(0, len(images), n)]
+    assert len(images) == sum([len(chunk) for chunk in chunks])
 
     total = None
 
     with concurrent.futures.ProcessPoolExecutor(max_workers=params.nproc) as p:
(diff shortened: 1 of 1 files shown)


command_line/find_bad_pixels.py
```diff
@@ -30,6 +32,9 @@
 image_range = None
   .type = ints(value_min=0, size=2)
   .help = "Image range for analysis e.g. 1,1800"
+nproc = 1
+  .type = int
```
rjgildea:
```suggestion
  .type = int(value_min=1)
```
We can't have fewer than 1 process

command_line/find_bad_pixels.py
```diff
@@ -30,6 +32,9 @@
...
+  .help = "Number of processors to use for bad pixel finding"
```
rjgildea:
```suggestion
    .help = "The number of processes to use."
```
Probably strictly _processes_ not _processors_, and might as well use the help string as used elsewhere.

command_line/find_bad_pixels.py
```diff
@@ -40,48 +45,9 @@
...
-        start, end = params.image_range
-        params.images = list(range(start, end + 1))
-
-    if params.images:
-        if min(params.images) < first or max(params.images) > last:
-            raise Sorry("image outside of scan range")
-        images = params.images
+def find_constant_signal_pixels(imageset, images):
+    """Find pixels which are constantly reporting as signal through the
+    images in imageset."""
```
rjgildea: This might be a good place to include a brief description of the algorithm you are using to find the bad pixels.

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+        read_experiments_from_images=True,
+        epilog=help_message,
+    )
+
+    params, options = parser.parse_args(show_diff_phil=True)
+
+    experiments = flatten_experiments(params.input.experiments)
+    if len(experiments) != 1:
+        parser.print_help()
+        print("Please pass an experiment list\n")
```
rjgildea:
```suggestion
        sys.exit("Please pass an experiment list\n")
```
and then no need for the `return` statement below?

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+        return
+
+    imagesets = experiments.imagesets()
+
+    if len(imagesets) != 1:
+        raise Sorry("Please pass an experiment list that contains one imageset")
+
+    imageset = imagesets[0]
+    detectors = imageset.get_detector()
```
rjgildea: Doesn't this give a list of panels, not detectors? In which if it's a panel, call it a panel!

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+    experiments = flatten_experiments(params.input.experiments)
+    if len(experiments) != 1:
+        parser.print_help()
+        print("Please pass an experiment list\n")
+        return
+
+    imagesets = experiments.imagesets()
+
+    if len(imagesets) != 1:
+        raise Sorry("Please pass an experiment list that contains one imageset")
```
rjgildea:
```suggestion
        sys.exit("Please pass an experiment list that contains one imageset")
```

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+    first, last = imageset.get_scan().get_image_range()
+    images = range(first, last + 1)
+
+    if params.images is None and params.image_range is not None:
+        start, end = params.image_range
+        params.images = list(range(start, end + 1))
+
+    if params.images:
+        if min(params.images) < first or max(params.images) > last:
+            raise Sorry("image outside of scan range")
```
rjgildea:
```suggestion
            sys.exit("Image outside of scan range")
```

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+        images = params.images
+
+    # work around issues with HDF5 and multiprocessing
+    imageset.reader().nullify_format_instance()
+
+    n = int(math.ceil(len(images) / params.nproc))
+    work = [images[i : i + n] for i in range(0, len(images), n)]
```
rjgildea:
```suggestion
    chunks = [images[i : i + n] for i in range(0, len(images), n)]
```
Would `chunks` be a better variable name here? Given that below you do `for chunk in work`

command_line/find_bad_pixels.py
graeme-winter: True - will pop that into the TODO stack

command_line/find_bad_pixels.py
```diff
@@ -133,29 +94,84 @@ def run(args):
...
+        if min(params.images) < first or max(params.images) > last:
+            sys.exit("Image outside of scan range")
+        images = params.images
+
+    # work around issues with HDF5 and multiprocessing
+    imageset.reader().nullify_format_instance()
+
+    n = int(math.ceil(len(images) / params.nproc))
+    chunks = [images[i : i + n] for i in range(0, len(images), n)]
+    assert len(images) == sum([len(chunk) for chunk in chunks])
```
Anthchirp:
```suggestion
    assert len(images) == sum(len(chunk) for chunk in chunks)
```
avoid explicit list construction

PR #1011, opened 2019-11-14
Don't use libtbx.env.dispatcher_name

rjgildea:
- Float imports
- Sorry -> sys.exit

### DIFF
--- a/command_line/find_shared_models.py
+++ b/command_line/find_shared_models.py
@@ -1,9 +1,19 @@
 from __future__ import absolute_import, division, print_function
 
+import datetime
 import logging
-from dials.util import tabulate
+from collections import Counter
+from time import time
 
+from libtbx.phil import parse
+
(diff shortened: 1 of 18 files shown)


command_line/find_shared_models.py
```diff
@@ -38,14 +46,8 @@ class Script(object):
...
         """Initialise the script."""
-        from dials.util.options import OptionParser
-        import libtbx.load_env
-
         # The script usage
-        usage = (
-            "usage: %s [options] [param.phil] "
-            "models.expt " % libtbx.env.dispatcher_name
-        )
+        usage = "usage: dials.find_shared_models [options] [param.phil] " "models.expt "
```
Anthchirp: just as comment: iirc this is equivalent to
```suggestion
        usage = "dials.find_shared_models [options] [param.phil] models.expt"
```
because the OptionParser adds the `usage: ` bit automatically

PR #1014, opened 2019-11-15
Import still ImageSequences as N experiments

graeme-winter:
In parallel with 

https://github.com/cctbx/dxtbx/pull/117

Trying as @rjgildea suggested to do "the right thing" on import - 

 - this starts in finite time (good)
 - it does iterate over experiments (bad)


graeme-winter:
This change recreates the "find spots takes forever to start" feature :-/

graeme-winter:
With the changes in https://github.com/cctbx/dxtbx/pull/118 now loads before the heat death of the universe which is a good thing. This PR includes code which is sufficient to find spots on a long experiment list against one underlying image set correctly and restores the ascii art output e.g. 

```
Histogram of per-image spot count for imageset 237:
20473 spots found on 238 images (max 934 / bin)
*                                                           
*    * *    *  *   *                                        
**   ****   ** *   *  *   *                                 
**   ****   ****   ****   *  *                              
**   ****   ****   ****   ** *   *                          
**  *****   ****  *****  *** ** **   *                      
**  *****  *****  ****** *** ** **   *                      
**  *****  ****** ****** ****** **  **                      
*** ****** ****** ****** ****** **  ** **                   
************************************************************
1                         image                          238

```

There is a certain amount of boiler plate here which is duplicated and could probably be improved by pulling into a central location. 

N.B. there is a large extant issue here: the image viewer canna cope with this as a concept - need to get it to understand that > 1 experiment can point at an image set. Had a look at this for an hour and failed.

ndevenish:
> N.B. there is a large extant issue here: the image viewer canna cope with this as a concept - need to get it to understand that > 1 experiment can point at an image set. Had a look at this for an hour and failed.

Does this mean that currently image_viewer fails to open these files?

algorithms/spot_finding/finder.py
```diff
@@ -745,19 +746,41 @@ def __call__(self, experiments):
...
+        for experiment in experiments:
+            if experiment.imageset not in imagesets:
+                imagesets.append(experiment.imageset)
+                imageset_scan[experiment.imageset] = copy.deepcopy(experiment.scan)
+            else:
+                imageset_scan[experiment.imageset] += experiment.scan
+
+        for imageset in imagesets:
+            imageset.set_scan(imageset_scan[imageset])
+
```
jmp1985: I think the above lines will actually end up modifying the scan contained in the experiment since you are not deep copying here. If this what you intended to do?

ndevenish: I think it does - this also bothers me, but i think in _current_ usage this is never written out or reused after spotfinding?

command_line/find_spots.py
```diff
@@ -135,11 +135,24 @@ def run(self, args=None):
...
         for i, experiment in enumerate(experiments):
-            ascii_plot = spot_counts_per_image_plot(
-                reflections.select(reflections["id"] == i)
-            )
+            if experiment.imageset not in imagesets:
+                imagesets.append(experiment.imageset)
+                imageset_experiment[experiment.imageset] = [i]
+            else:
+                imageset_experiment[experiment.imageset].append(i)
+
```
jmp1985: You create imageset_experiment but it doesn't appear to be used for anything. You could probably simplify these changes by doing the following . 

You probably don't need to maintain a list of imagesets separate to the dictionary of (imageset, experiment) pairs. Perhaps change this to something like:

```
from collections import defaultdict
imageset_experiment = defaultdict(list)
for i, experiment in enumerate(experiments):
    imageset_experiment[experiment.imageset].append(i)
```

Then instead of the lines below you could have:

```
        for imageset in imageset_experiment.keys():
            selected = flex.bool(reflections.nrows(), False)
            for i in imageset_experiment[imageset]:
                selected.set_selected(reflections["id"] == i, True)
                ascii_plot = spot_counts_per_image_plot(reflections.select(selected))
```

jmp1985: I'm happy to make these changes

algorithms/spot_finding/finder.py
jmp1985: It is possible to write out experiments after spot finding which contain a mask with hot pixels. I don't know if this is used in practice but in principle if the experiments are messed up then it could possible cause problems. I think we should try to ensure that the experimental models are consistent because this kind of thing can cause confusion down the line.

graeme-winter:
When I tested this the output looked correct... I wrote out an experiment list and manually inspected

algorithms/spot_finding/finder.py
graeme-winter: The imageset scan is never written to disk - only the experiment ones. The only thing sent to disk is the imageset indices:

```
  "imageset": [
    {
      "__id__": "ImageSequence", 
      "template": "/Users/graeme/data/vmxi-grid/image_34016_master.h5", 
      "single_file_indices": [
        0, 
        1, 
        2, 
        3, 
        4, 
        5, 
        6, 
        7, 
        8, 
        9, 
        10, 
        11, 
----8<-----
        7371, 
        7372, 
        7373, 
        7374, 
        7375, 
        7376, 
        7377, 
        7378, 
        7379
      ], 
      "mask": "", 
      "gain": "", 
      "pedestal": "", 
      "dx": "", 
      "dy": "", 
      "params": {
        "dynamic_shadowing": "Auto", 
        "multi_panel": false
      }
    }
  ], 
```

I also worried about this...

command_line/find_spots.py
graeme-winter: Thanks for the suggestion - I'm happy to make the changes as I would like to fully understand everything which is happening here (bus # increment)

graeme-winter:
Also:

Does not do *the right thing* on Pilatus grid scans:

```
Setting spotfinder.filter.min_spot_size=3
Configuring spot finder from input parameters
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 200, in <module>
    script.run()
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 128, in run
    reflections = flex.reflection_table.from_observations(experiments, params)
  File "/Users/graeme/svn/cctbx/modules/dials/array_family/flex.py", line 198, in from_observations
    return find_spots(experiments)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 760, in __call__
    imageset.set_scan(imageset_scan[imageset])
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: dxtbx Internal Error: /Users/graeme/svn/cctbx/modules/dxtbx/imageset.h(367): DXTBX_ASSERT(index < scans_.size()) failure.
```

=> something else to investigate, yay.

graeme-winter:
Does _the right thing_ on pilatus now => think we have something worth discussion now

```
Found 190 strong pixels on image 226
Found 387 strong pixels on image 227
Found 248 strong pixels on image 228
Found 228 strong pixels on image 229
Found 230 strong pixels on image 230

Extracted 5799 spots
Removed 2907 spots with size < 3 pixels
Removed 0 spots with size > 1000 pixels
Calculated 2892 spot centroids
Calculated 2892 spot intensities
Filtered 2578 of 2892 spots by peak-centroid distance

Histogram of per-image spot count for imageset 229:
2578 spots found on 230 images (max 482 / bin)
                                *                           
                          *     *                           
                      *   *     *                           
                      *   *     *                           
                      *   *     *      *                    
                     **   *     *      *                    
                     **   *     *      *                    
                     **   *     *      *                    
             *       **   *     **     *                    
             *       **  ***   ***     **   *               
1                         image                          230

Saved 2578 reflections to strong.refl
Time Taken: 23.278084
```

graeme-winter:
Unsurprisingly a lot of tests now fail... usually those which have `stills` in name - e.g.

```
            # Fix up the experiment ID's now
            table["id"] = flex.int(table.nrows(), -1)
            for i, experiment in enumerate(experiments):
                if experiment.imageset is not imageset:
                    continue
>               z0, z1 = experiment.scan.get_array_range()
E               AttributeError: 'NoneType' object has no attribute 'get_array_range'

/dls/science/users/gw56/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py:771: AttributeError
```

graeme-winter:
With last couple of hacks to `dxtbx` heat death branch getting full tests passing => review please 🙂

graeme-winter:
Note to self: this explodes if you try to import frame NN to NN+j not starting at 1 (if files 1...NN-1 not on disk)

graeme-winter:
> Note to self: this explodes if you try to import frame NN to NN+j not starting at 1 (if files 1...NN-1 not on disk)

Fixed in https://github.com/cctbx/dxtbx/pull/118/commits/77b8ba4774de2f11c519da540073965f8413e10f

graeme-winter:
#info I would like to see this merged real soon

dagewa:
There's a request for some tests to be added, plus @jmp1985 made some comments that could be read as requesting changes. Once these points are resolved I think it is ready.

algorithms/spot_finding/finder.py
graeme-winter: This code is now gone... 🤔

command_line/find_spots.py
graeme-winter: Done 👍

graeme-winter:
Added tests as requested - valid feedback thank you.

Using centroid test data and importing as still sequence highlighted issue in dxtbx, also fixed.

graeme-winter:
@dagewa added tests 🙂

graeme-winter:
👍 appreciate the feedback and help thank you - will squash merge in a mo

PR #1015, opened 2019-11-16
New AutoReduce

### LINKED ISSUE #990: dials.refine leaves panels behind in multipanel refinement
A problem we've had for some time, but I thought I'd create an issue for it since I hit it again.  Refining the JF16M. 256 panels arranged in 4 level hierarchy, a lot like the cspad.  Refinement params for level 2:

```
refinement {
  parameterisation {
    auto_reduction {
      min_nref_per_parameter = 3
      action = fail fix *remove
    }
    beam {
      fix = *all in_spindle_plane out_spindle_plane wavelength
    }
    detector {
       hierarchy_level = 2
       fix_list = Group1Tau1,Tau2,Tau3
       constraints {
         parameter = Dist
       }
    }
  }
  refinery {
    engine = *SparseLevMar
  }
  reflections {
    outlier {
      algorithm = None
      separate_experiments = False
      separate_panels = True
    }
  }
}
```

The important bit here is constraints.  All groups at level 2 are constrained to move together along their normal.  However, as is often the case, we don't have data that covers all the panel groups so some are fixed and left behind during refinement.  Result:

Screenshot from 2019-10-28 11-02-54

Ideally, after refinement using constraints, the panel groups that weren't refined would have the new deltas applied.

I have code in cspad.cbf_meterology that does this:
https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/cspad_cbf_metrology.py#L494-L556

But I'd prefer to have this native in dials.refine.

dagewa:
This is a complete rewrite of the parameter auto-reduction code in dials.refine. The initial motivation was to fix #990, but this proved difficult with the old version of this algorithm. The new version operates at a higher level, accounts for constrained parameters, is less complex and yet provides finer-grained analysis of poorly-supported parameters. The previous version could only free parameters in a block (for example all parameters for a particular panel group). The new version will fix individual parameters that have too few reflections, while leaving other parameters in the same parameterisation unaffected. The `detector_reduce` function of the previous version is no longer required, as this was used to get around the block setting behaviour.

As the algorithm is now different, the results are also different. The tests have been updated for the new version, and a new test added to ensure that #990 is fixed. Testing against various real world datasets would be appreciated, particularly for XFEL use cases, such as metrology refinement.

I am interested in any effect on the speed of execution. This no longer uses the C++ extensions in #467, but `auto_reduction.action=fix` should not be much slower than a single step of refinement. `auto_reduction.action=remove` can be slower because this is an iterative process.


dagewa:
This also turns up in xia2 (https://github.com/xia2/xia2/commit/0f54391b70630224890c2a0e43535b86085aff57) so adding @rjgildea to take a look too.

phyy-nx:
Hi, thanks so much for looking at this.  In xfel_regression, we test this code extensively in test/command_line/tst_cspad_cbf_metrology.py.  There are two versions, a 300 and 3000 image version.  I tested this branch on the 300 image version.

In cspad.cbf_metrology, I get around #990 by manually bringing along the left-behind panels at each step of refinement.  See:
https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/cspad_cbf_metrology.py#L494-L553

Call this the 'fixer' code.

I did three tests:
1) Run on master.  Expect this to pass, keeping the detector flat.
2) Run on master, removing the fixer code.  Expect this to fail, with Z-offset panels.
3) Run on constrain-data-free-panels branch, removing the fixer code.  Expect this to pass, as the new code keeps the constraints even on panels with no reflections.

Results:
1) Pass
2) Fail. The test checks if the final model is similar to a known model, which when panels are left behind in Z, it is not.
3) Pass

Here are the dxtbx.plot_detector_models plots for the three use cases:

image

Blue and magenta (overlaid) are cases 1 and 3.  Green is case 2 where the Z axes are all jumbled up.

Timing: it ran in nearly the same amount of time 

Precision: the code did give slightly different answers, but these are very small differences (5th to 7th decimal place in mm), and the is_similar_to check at the end still passes.

Overall, thumbs way up.  Thanks!  After this is merged I'll drop the fixer code from cspad.cbf_metrology and for non-cspad detectors, we should be in much better shape.

dagewa:
Thanks for the analysis. I'm glad you brought this to my attention. It was one of those cases where the original code was overly complex, in that it was trying to figure out information that is easily derived from the full parameterisation prior to constructing it. It is far simpler and cleaner to construct, test then construct again if required. I'm merging now.

PR #1017, opened 2019-11-18
Refactor refine_bravais_settings

rjgildea:
- Use concurrent.futures instead of libtbx.easy_mp
- Remove unused normalisation option
- Float imports
- Rename some variables/methods
- Move RBS code to new bravais_settings module
- Move phil_scope (mostly) to bravais_settings module
- More useful test names
- Split up run function


algorithms/indexing/bravais_settings.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+        except RuntimeError as e:
+            if (
+                str(e) == "scitbx Error: g0 - astry*astry -astrz*astrz <= 0."
+                or str(e) == "scitbx Error: g1-bstrz*bstrz <= 0."
+            ):
+                subgroup.refined_experiments = None
+                subgroup.rmsd = None
+                subgroup.Nmatches = None
+            else:
```
Anthchirp: wondering if bailing out here with `return subgroup` is clearer than `try: except: else`

algorithms/indexing/symmetry.py
```diff
@@ -1,321 +1,22 @@
 from __future__ import absolute_import, division, print_function
 
-import copy
 import logging
-import math
 
-import dials.util
+from six.moves import StringIO
```
Anthchirp: patch incoming

Anthchirp: _should_ work. Haven't run it through with an appropriate example triggering the code path though.

command_line/refine_bravais_settings.py
```diff
@@ -218,35 +194,30 @@ def run(args=None):
...
 
-    for subgroup in Lfat:
+    for subgroup in refined_settings:
         expts = subgroup.refined_experiments
         soln = int(subgroup.setting_number)
         bs_json = "%sbravais_setting_%i.expt" % (prefix, soln)
         logger.info("Saving solution %i as %s" % (soln, bs_json))
         expts.as_file(os.path.join(params.output.directory, bs_json))
 
+    return refined_settings
```
benjaminhwilliams: Looks great.  Just one further suggestion — can we leave the file-writing and some of the final-results logging in `run`, rather than in the `refine_bravais_settings` function?  It would make it much easier to re-use the functionality of `r_b_s` in cases where you don't necessarily need to save the output to disk.

Also, I'm not sure if the program help message renders it slightly redundant, but I'd be keen on a function docstring for `r_b_s`.

```suggestion
def refine_bravais_settings(experiments, reflections, params):
    """
    FIXME:  This is really the business end of dials.r_b_s.  Can this functionality be
            distilled into a docstring?
    """
    if len(experiments.crystals()) > 1:
        if params.crystal_id is not None:
            assert params.crystal_id < len(experiments.crystals())
            experiment_ids = experiments.where(
                crystal=experiments.crystals()[params.crystal_id]
            )

            experiments = ExperimentList([experiments[i] for i in experiment_ids])
            refl_selections = [reflections["id"] == i for i in experiment_ids]
            reflections["id"] = flex.int(len(reflections), -1)
            for i, sel in enumerate(refl_selections):
                reflections["id"].set_selected(sel, i)
            reflections = reflections.select(reflections["id"] > -1)
        else:
            raise Sorry(
                "Only one crystal can be processed at a time: set crystal_id to choose "
                "experiment."
            )

    if params.refinement.reflections.outlier.algorithm in ("auto", libtbx.Auto):
        if experiments[0].goniometer is None:
            params.refinement.reflections.outlier.algorithm = "sauter_poon"
        else:
            # different default to dials.refine
            # tukey is faster and more appropriate at the indexing step
            params.refinement.reflections.outlier.algorithm = "tukey"

    cb_op_to_primitive = (
        experiments[0]
        .crystal.get_space_group()
        .info()
        .change_of_basis_op_to_primitive_setting()
    )
    if experiments[0].crystal.get_space_group().n_ltr() > 1:
        effective_group = (
            experiments[0]
            .crystal.get_space_group()
            .build_derived_reflection_intensity_group(anomalous_flag=True)
        )
        sys_absent_flags = effective_group.is_sys_absent(reflections["miller_index"])
        reflections = reflections.select(~sys_absent_flags)
    experiments[0].crystal.update(
        experiments[0].crystal.change_basis(cb_op_to_primitive)
    )
    miller_indices = reflections["miller_index"]
    miller_indices = cb_op_to_primitive.apply(miller_indices)
    reflections["miller_index"] = miller_indices

    refined_settings = refined_settings_from_refined_triclinic(
        experiments, reflections, params
    )

    return refined_settings


def run(args=None):
    usage = "dials.refine_bravais_settings indexed.expt indexed.refl [options]"

    parser = OptionParser(
        usage=usage,
        phil=phil_scope,
        read_experiments=True,
        read_reflections=True,
        check_format=False,
        epilog=help_message,
    )

    params, options = parser.parse_args(args=args, show_diff_phil=False)

    # Configure the logging
    log.config(verbosity=options.verbose, logfile=params.output.log)

    logger.info(dials_version())

    # Log the diff phil
    diff_phil = parser.diff_phil.as_str()
    if diff_phil != "":
        logger.info("The following parameters have been modified:\n")
        logger.info(diff_phil)

    experiments = flatten_experiments(params.input.experiments)
    reflections = flatten_reflections(params.input.reflections)
    if len(reflections) == 0 or len(experiments) == 0:
        parser.print_help()
        return

    assert len(reflections) == 1
    reflections = reflections[0]

    if len(experiments) == 0:
        parser.print_help()
        return

    refined_settings = refine_bravais_settings(experiments, reflections, params)

    possible_bravais_settings = {solution["bravais"] for solution in refined_settings}
    bravais_lattice_to_space_group_table(possible_bravais_settings)
    logger.info(refined_settings.as_str())

    prefix = params.output.prefix
    if prefix is None:
        prefix = ""
    summary_file = "%sbravais_summary.json" % prefix
    logger.info("Saving summary as %s" % summary_file)
    with open(os.path.join(params.output.directory, summary_file), "w") as fh:
        json.dump(refined_settings.as_dict(), fh)

    for subgroup in refined_settings:
        expts = subgroup.refined_experiments
        soln = int(subgroup.setting_number)
        bs_json = "%sbravais_setting_%i.expt" % (prefix, soln)
        logger.info("Saving solution %i as %s" % (soln, bs_json))
        expts.as_file(os.path.join(params.output.directory, bs_json))
```

benjaminhwilliams: Sorry, managed to apply that previous suggestion to a single line, rather than a block of code.  Correction incoming.

benjaminhwilliams: Hmm... It seems Github doesn't allow multi-line suggestions incorporating parts of a file that aren't included in the diff for the PR.  Hence the unexpected appearance of the review comment above.

benjaminhwilliams:
I've just got a couple of minimal little changes to add, to pacify my linter (and satisfy an urge for docstrings).

Issue #536, opened 2018-04-14
dials.image_viewer: Display -2 as "masked" occurs for non-Pilatus images


dagewa:
An example is in the backstop region of this image:
```
dials.image_viewer $(libtbx.find_in_repositories dials_regression)/image_examples/ESRF_BM14/mar225_1_001.mccd
 ```
![mask_marccd](https://user-images.githubusercontent.com/11502083/38770507-f348e9ea-400b-11e8-95fd-1d33ad5b7d5c.png)

For those with access, the images in `/dls/mx-scratch/rjgildea/fisher/013_021_G4` are worse, with many "masked" pixels across the image

rjgildea:
This is a known thing, and I'm not sure a fix is possible without a significant reworking of how the display is handled in the FlexImage class:
https://github.com/cctbx/cctbx_project/blob/master/iotbx/detectors/display.h#L236

dagewa:
I wonder if @nksauter would like to comment? Why do we need special code to display -2 as masked anyway? Shouldn't the masked pixels be defined by the mask?

rjgildea:
The code in FlexImage long-predates dxtbx and image masks (and also Pilatus detectors) - judging by labelit svn logs it dates back to 2004.

dagewa:
I haven't yet tried it, but I thought for a simple fix that this line: https://github.com/cctbx/cctbx_project/blob/master/iotbx/detectors/display.h#L199 
should instead be
`bool has_pilatus_inactive_flag = false;`
and then it is set `true` individually for DECTRIS detectors based on `vendortype` and then doesn't end up true for e.g. CCD datasets

rjgildea:
Yes, but then it won't be possible to view image masks for CCD images. The problem is that the image viewer is (ab)using the functionality in FlexImage flagging the special Pilatus -2 pixels in order to display dxtbx image masks.

graeme-winter:
I would definitely say that @nksauter would hold an opinion here as clearly this must also be a problem for pedestal subtracted CSPAD images…. 

Properly supporting the concept of a mask not derived from pixel values seems like a good idea…

rjgildea:
https://github.com/dials/dials/blob/42fcf5ff99de386d37c0390407dfc06d20a322c4/util/image_viewer/spotfinder_frame.py#L1077-L1080

dagewa:
[...]

PR #1020, opened 2019-11-19
Masks for non-pilatus-like detectors

dagewa:
Rather than using -2 to indicate masked pixels for all detectors, set explicitly-masked pixels to `MASK_VAL`. This fixes the problem with valid -2 values being shown as masked in the image viewer (#536). This branch also requires https://github.com/cctbx/cctbx_project/tree/non-pilatus-flags

### DIFF
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -639,9 +639,8 @@ class SpotFrame(XrayFrame):
         # FIXME Currently assuming that all panels are in same plane
         p_id = detector.get_panel_intersection(beam.get_s0())
         if p_id == -1:
-            p_id = (
-                0
-            )  # XXX beam doesn't intersect with any panels - is there a better solution?
+            # XXX beam doesn't intersect with any panels - is there a better solution?
+            p_id = 0
         pan = detector[p_id]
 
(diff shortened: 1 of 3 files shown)


util/image_viewer/spotfinder_frame.py
```diff
@@ -639,7 +639,9 @@ def draw_resolution_rings(self, unit_cell=None, space_group=None):
         # FIXME Currently assuming that all panels are in same plane
         p_id = detector.get_panel_intersection(beam.get_s0())
         if p_id == -1:
-            p_id = 0  # XXX beam doesn't intersect with any panels - is there a better solution?
+            p_id = (
+                0
+            )  # XXX beam doesn't intersect with any panels - is there a better solution?
```
rjgildea: Would this comment be better moved to the line above to avoid this ugly formatting?

dagewa: Probably. Black did this (I didn't touch the line on purpose)

dagewa:
@phyy-nx I'm keen to get this in to DIALS 2.1. Might you be able to check things are okay over your side? In particular, there is a small and I think safe change in https://github.com/cctbx/cctbx_project/tree/non-pilatus-flags, but I don't know if that has any effects outside dials.image_viewer usage.

dagewa:
Thanks!

PR #1027, opened 2019-11-20
Fix problem with conda dials environment file on macos

phyy-nx:
Currently .conda-envs/dials_py27_osx-64.txt uses conda-forge::hdf5-external-filter-plugins.  When I run a simple bootstrap.py --builder=dials --use-conda on my mac, conda enters an endless dependency validation loop.  This is because conda-forge::hdf5-external-filter-plugins was compiled using the default clang compilers that use libcxx v. 9 or so, but the .conda-envs/dials_py27_osx-64.txt specifies cctbx::libcxx, which is version 4.0.1.  Thus the dependencies are conflicted and the installation fails.

This patch changes the env file for dials to use conda-forge::libcxx.

Having the hdf5 plugins on conda forge built using the latest conda supplied clang is a good thing so advancing to libcxx to the latest version makes sense.  However, this adds some risk.  I’ve ran libtbx.pytest --regression in the dials folder and that passed, so the basic functionality appears to be working.  We’ll let the azure/jenkins builders chew on this pull request and then maybe merge it tomorrow if it makes sense?  Likely we’ll need someone to test using a compressed h5 file.

(if this continues to work, we’ll likely migrate cctbx/phenix off of libcxx 4.0.1 onto the conda-forge version)


Anthchirp:
Will need to verify this with a full manual build.

The cctbx::libcxx was required for another ABI incompatibility breaking DXTBX_ASSERT. We will have to check if that still occurs.

Anthchirp:
I am looking into this

Anthchirp:
Had a go at this, and think I have a working MacOS variant of the linux package selection.

I've verified the previous `DXTBX_ASSERT` issue by modifying dxtbx:
```patch
--- a/model/beam.h
+++ b/model/beam.h
@@ -234,7 +234,7 @@ namespace dxtbx { namespace model {

     /** Get the wave vector in units of inverse angstroms */
     vec3<double> get_s0() const {
-      DXTBX_ASSERT(wavelength_ != 0.0);
+      DXTBX_ASSERT(wavelength_ == 0.0);
       return -direction_ * 1.0 / wavelength_;
     }

```
and this seems to still work as expected.

I also dropped the py2app dependency which is not used by dials at all.

I didn't see any resolution issues. If it works for you then I'll squash merge.

phyy-nx:
Looks good.

Phenix uses py2app to make its .app bundle.  Ok to take it out of the dials
env file.

Anthchirp:
thanks for reporting

PR #1028, opened 2019-11-21
More unicode support (needed for minimal docker images)

phyy-nx:
(part of cctbx/cctbx_project#417)

### DIFF
--- a/libtbx_refresh.py
+++ b/libtbx_refresh.py
@@ -2,8 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 import dials.precommitbx.nagger
 import libtbx.pkg_utils
-from libtbx.utils import to_str
-import io
 
 libtbx.pkg_utils.define_entry_points(
     {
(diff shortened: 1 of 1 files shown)


ndevenish:
I think we'd prefer `from builtins import open` to importing `io.open` because it retains the standard `open` idiom and the code doesn't change at all when dropping python 2 in the future (yes, `io.open` is the same thing but is a slightly obscure way of calling). The same would go for the `to_str`, however....

 wouldn't it be better to bypass unicode entirely and `open(file, 'rb')` and compare with `b"DIALS_ENABLE_COMMAND_LINE_COMPLETION"`.

I'm also surprised that this is necessary for docker to work - are you sure it's not just the locale (or if no other resort PYTHONIOENCODING) that needs to be set up?

phyy-nx:
Thanks for your review.
1) I have no preference as to the syntax of io.opn vs builtins.open.  Is this a blocker to merging?
2) Loading the file as 'rb' means no unicode allowed in the source code.  See:
https://github.com/cctbx/dxtbx/blob/master/libtbx_refresh.py#L70
3) I spent half a day trying to figure out what was wrong with my docker image, including lots of locale work.  This just bypassed the issue.

ndevenish:
> Is this a blocker to merging?

No, we can change it later

> Loading the file as 'rb' means no unicode allowed in the source code

Weird, reading the raw byte sequences in binary mode should be entirely agnostic of the contents of the file;
```python
with open("dxtbx/libtbx_refresh.py", "rb") as f:
    assert b"format_classes" in f.read()
print("Read successfully")
```
Appears to work:
```
$ python2 test.py
Read successfully
$ python3 test.py
Read successfully
```
So I don't know why it wasn't working for you  ¯\\\_(ツ)\_/¯. It's just usually easier to deal with unicode issues by completely avoiding string handling, if possible.

>I spent half a day trying to figure out what was wrong with my docker image, including lots of locale work. This just bypassed the issue.

Right, thought it worth a suggestion because there's been several cases in docker and VMs where we've had to resort to PYTHONIOENCODING to get things to function predictably. I guess it depends on your image base.

phyy-nx:
(testing rb. I like that idea better)

phyy-nx:
Thanks, that works great.  I didn't think rb would work!  Merging.

Anthchirp:
For reference this would have been a really good pull request to squash merge.

Issue #1025, opened 2019-11-20
dials.refine _check_scan_range


dagewa:
Before this check was introduced (#778) we would occasionally get failures in scan-varying refinement for datasets where there were no reflections at the scan edges.

Now the check is there, but beamline scientists complain that DIALS does not process data that have blank images at the beginning and/or end of the scan.

I can make the check more lenient, but at some point you just start getting a dodgy refined model, which due to the smoother also begins to affect part of the scan that really do have data. Also things like the average unit cell calculated over the scan become nonsense.

Another option would be for `dials.refine` to trim the scan to match the range of strong reflections, with a border of up to 5°, say. So, for the data set with 7.25° of images with no strong spots at the start `dials.refine` could warn that scan-varying refinement would be compromised and trim the first 2.25° off.

I'm not sure if `dials.refine` _should_ be modifying the scan width like this, but in a sense the whole point of the program is to modify the experimental models, so maybe that is okay?

What do people think?

dagewa:
Example of "successful" refinement with limited data at the start of the scan:
![image](https://user-images.githubusercontent.com/11502083/69245517-dbe24980-0b9e-11ea-99d5-e9a98fba0be8.png)

Anthchirp:
specific case: https://github.com/xia2/xia2/issues/381

phyy-nx:
Trimming the scan to where the bright spots starts seems reasonable, but
wouldn't downstream integration be affected?  IE weak spots that could have
been integrated at the beginning or end of the scan would be lost?  Also,
what would be the cutoff for N strong spots?  Min num spots per parameter
or zero or something in between?

dagewa:
It is conceivable that trimming the scan would fail to integrate some extremely weak spots that would otherwise have been integrated, but refining a scan-varying model against no data is liable to produce garbage anyway (see plot above) in which case the predictions would probably miss the spots.

In this case I'm not using the minimum number of spots per parameter but just looking at the phi values for the first and last strong spot in the dataset and comparing with the scan edges.

dagewa:
Following discussion with @graeme-winter about this and ideas in #1014 we are moving towards `Scan` being the interface to the useful bit of a image sequence for some experiment. In that case:

1. On import there is one crystal-less experiment with a scan that refers to the full image range because we don't know any better.
2. We do spot-finding and indexing which potentially splits into multiple experiments depending on the number of lattices found. The scans associated with these experiments still refer to the full imported image range, because the models are global, static models.
[...]

PR #1033, opened 2019-11-22
Trim scan

dagewa:
For #1025 and xia2/xia2#381, `dials.refine scan_varying=true` will now trim scans to the observed reflection range rather than complain.

@graeme-winter this is the PR you were looking for earlier!

### DIFF
--- a/algorithms/refinement/refiner.py
+++ b/algorithms/refinement/refiner.py
@@ -170,6 +170,7 @@ class RefinerFactory(object):
             "xyzcal.px",
             "xyzobs.mm.variance",
             "flags",
+            "shoebox",
             "delpsical.weights",
         ]
         # NB xyzobs.px.value & xyzcal.px required by SauterPoon outlier rejector
--- a/algorithms/refinement/refiner.py
+++ b/algorithms/refinement/refiner.py
@@ -152,6 +152,75 @@ def _copy_experiments_for_refining(experiments):
     return out_list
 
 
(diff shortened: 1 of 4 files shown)


graeme-winter:
Looking now

0th order

dials.import / find_spots / index / refine on a troublesome not-centred data set gives:

```
Scan:
    image range:   {72,483}
    oscillation:   {-25,0.1}
    exposure time: 0
```

in refinement output => looks to do what I would expect compared with 

```
>>> from dials.array_family import flex
>>> d = flex.reflection_table.from_file('indexed.refl')
>>> i = d.select(d.get_flags(d.flags.indexed))
>>> bb = i['bbox'].parts()
>>> flex.min(bb[4])
72
>>> flex.max(bb[5])
483
```

graeme-winter:
Checking that `xia2` behaves itself in this situation - 

```
-------------------- Spotfinding SWEEP1 --------------------
11903 spots found on 420 images (max 304 / bin)
                             * * *                          
                    ** *  **** *************                
                ********************************            
          *****************************************         
        * *******************************************       
       **********************************************       
      *************************************************     
     **************************************************     
    *****************************************************   
    *****************************************************   
63                         image                         482
WARNING: Potential blank images: 1 -> 80
WARNING: Potential blank images: 461 -> 500
```

it's not coming as any surprise here that _something is wrong_ - fails to process in my macbook because µ=2° so we have a different problem here - will test on other machine

Issue #1030, opened 2019-11-21
dials.index: please do not delete scan and goniometer for still images


graeme-winter:
If we record a grid scan at multiple angles would be useful to keep this information around after indexing - right now the scan and goniometer are deliberately deleted. 

Propose we keep them iff the input experiment->imageset is an ImageSequence not ImageSet.

Current output looks like:

```
Experiment 7:
Detector:
Panel:
  name: /entry/instrument/detector
  type: SENSOR_PAD
  identifier: 
  pixel_size:{0.075,0.075}
  image_size: {4148,4362}
  trusted_range: {-1,65535}
  thickness: 0.45
  material: Si
  mu: 22.0599
  gain: 1
  pedestal: 0
  fast_axis: {0.999992,-0.00122992,-0.00377055}
  slow_axis: {-0.00125372,-0.999979,-0.00631484}
  origin: {-155.863,168.099,-151.551}
  distance: 153.195
  pixel to millimeter strategy: ParallaxCorrectedPxMmStrategy
    mu: 22.0599
    t0: 0.45


Max resolution (at corners): 1.869481
Max resolution (inscribed):  2.304973

Beam:
    wavelength: 1.77118
    sample to source direction : {-0.000611717,-0,1}
    divergence: 0
    sigma divergence: 0
    polarization normal: {0,1,0}
    polarization fraction: 0.999

Beam centre: 
    mm: (156.17,167.91)
    px: (2082.24,2238.81)

Crystal:
    Unit cell: (68.07(2), 68.07(2), 68.04(3), 109.31(3), 109.57(3), 109.553(16))
    Space group: P 1
    U matrix:  {{ 0.1011, -0.9897,  0.1014},
                {-0.7773, -0.1422, -0.6129},
                { 0.6210, -0.0168, -0.7837}}
    B matrix:  {{ 0.0147,  0.0000,  0.0000},
                { 0.0052,  0.0156,  0.0000},
                { 0.0090,  0.0090,  0.0180}}
    A = UB:    {{-0.0028, -0.0145,  0.0018},
                {-0.0177, -0.0077, -0.0110},
                { 0.0020, -0.0073, -0.0141}}
Experiment 8:
Detector:
Panel:
  name: /entry/instrument/detector
  type: SENSOR_PAD
  identifier: 
  pixel_size:{0.075,0.075}
  image_size: {4148,4362}
  trusted_range: {-1,65535}
  thickness: 0.45
  material: Si
  mu: 22.0599
  gain: 1
[...]

PR #1036, opened 2019-11-23
Convert sequences to stills on import: delete scan & gonio to make properly stills

graeme-winter:
If convert_sequences_to_stills then delete the goniometer and scan.

At the moment they are imported as a lot of sequences - see discussion in #1030

### DIFF
--- a/command_line/dials_import.py
+++ b/command_line/dials_import.py
@@ -318,6 +318,9 @@ class ManualGeometryUpdater(object):
 
         if self.params.geometry.convert_sequences_to_stills:
             imageset = ImageSetFactory.imageset_from_anyset(imageset)
+            for j in imageset.indices():
+                imageset.set_scan(None, j)
+                imageset.set_goniometer(None, j)
         if not isinstance(imageset, ImageSequence):
             if self.params.geometry.convert_stills_to_sequences:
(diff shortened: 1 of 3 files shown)


graeme-winter:
For what it's worth the tests pass in DIALS just fine, on my macbook at least

graeme-winter:
@phyy-nx I think this does something you want - if we delete the scan and gonio on import then we're providing the route into the system which guarantees you get the behaviour you expect. If you import as sequences then the gonio and scan should be saved - allowing for the information to be used in the analysis.

graeme-winter:
> Changes look relatively trivial, but I would suggest adding a brief test to cover the new behaviour. Something along the lines of running dials.import on a rotation image with `convert_sequences_to_stills=True`, and then verifying that `experiment.is_still() == True` and `experiments.all_stills() == True` and that `experiment.scan is None` and `experiment.goniometer is None`.

👍

graeme-winter:
@rjgildea given added tests you are happy with the PR as is?

phyy-nx:
Very good.  Thanks for this one.

graeme-winter:
@phyy-nx you are welcome, thanks for the feedback

Issue #932, opened 2019-09-16
Consistent timing information via logging


noemifrisina:
Labels are annoyingly inconsistent e.g. in dials.integrate_INTEGRATE.log it's labelled 'Total time taken', in dials.find_spots.log 'Time Taken' and so on... 
And yep, it's case sensitive.

dagewa:
It could perhaps be done consistently using e.g. a decorator around command line program's `run` functions?

graeme-winter:
That seems like a very reasonable suggestion. If we could work in cpu time usage as well would be excellent.

dagewa:
This might be more difficult than I thought, because typically the logger is set up inside the `run` function.

dagewa:
Current discussion is leaning towards prepending each message in the log _file_ by the elapsed time since program start. This can be done by a formatter that is used by the `logging.FileHandler` only, not the stdout to avoid cluttering the live display.

dagewa:
As noted by @Anthchirp https://stackoverflow.com/questions/25194864/python-logging-time-since-start-of-program/25196134#25196134
I'll pick this up

PR #1042, opened 2019-11-25
Add boilerplate example plus test

rjgildea:
Copy "boilerplate" example from https://github.com/dials/dials/wiki/Boilerplate-%28or-the-standard-framework-of-a-DIALS-program%29 and add a test.


doc/examples/boilerplate.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+
+def do_stuff(experiments, reflections, params):
+    """Write the behaviour of the program as functions and classes outside run()"""
+    logger.info("Hello world!")
+    # Include file output here
+    logger.info("integer_parameter: %i", params.integer_parameter)
+    logger.info("bool_parameter: %s", params.bool_parameter)
+    reflections.as_file(params.output.reflections)
```
benjaminhwilliams: I have a concern about this.  As an API user, I often want to make use of an algorithm from a DIALS module, retain the reflection table/experiment list/whatever, do further calculations and only then write an output file.  I may never wish to produce the file output that the DIALS module would normally produce when called as a stand-alone program.  So insisting on packaging up file output in functions together with the reusable algorithm code makes things difficult.

benjaminhwilliams: I appreciate that the corollary is that if one _does_ wish to produce the normal file output when re-using parts of a DIALS program via function calls, one has to put in a few `reflections.as_file(params.output.reflections)`-esque lines, but I think explicit is better than implicit...

rjgildea: I hold no strong opinions here, but I note it was you who added the output lines to the boilerplate in the wiki 🙃
https://github.com/dials/dials/wiki/Boilerplate-(or-the-standard-framework-of-a-DIALS-program)/_compare/e63e7126920cebc0069ad65362488b57917bb729

This was just a straightforward copy of the existing wiki boilerplate, with minimal changes to ensure it was actually valid code. Happy for you to change this as you see fit.

benjaminhwilliams: Yes, that was uploaded after some wrangling by committee.  I wasn't entirely satisfied with it at the time, but the whole thing was enough of an improvement on the previous boilerplate that it didn't seem worth splitting hairs.  I might push a change and pass the review baton back to you.  Perhaps we can get a couple of other pairs of eyes on it too, once we're both satisfied.  I don't feel 100% confident that I know what's best here.

benjaminhwilliams:
@rjgildea, thoughts now?

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+A docstring
+
+This can double as a helpful message which explains how the program is run.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import logging
+import sys
+from typing import List  # noqa: F401  Flake8 doesn't recognise Python 2 typing.
```
graeme-winter: Are we ready to start using typing?

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+
+import libtbx.phil
+from dials.array_family import flex
+import dials.util
+import dials.util.log
+from dials.util.options import OptionParser, flatten_reflections, flatten_experiments
```
graeme-winter: Do we want to annotate which of these are useful for what tasks?

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+    }
+    bool_parameter = False
+      .type = bool
+    integer_parameter = 0
+      .type = int
+    """
+)
+
+
+def do_stuff(experiments, reflections, params):
```
graeme-winter: Propose we state clearly here that `do_stuff` is _renamed_ otherwise we will be defining `do_stuff` methods everywhere and it will be about as useful as `run()`

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+    reflections = flatten_reflections(params.input.reflections)
+
+    # You might well wish to check here that the command-line input is appropriate.
+    if len(reflections) != 1:
+        sys.exit("Exactly one reflection file needed")
+    if len(experiments) != 1:
+        sys.exit("Exactly one 1 experiment list required")
+
+    # Do whatever this program is supposed to do.
+    do_stuff(experiments, reflections[0], params)
```
graeme-winter: `experiments[0]` surely?

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+                      which is the usable form of a parsed PHIL scope.
+
+    Raises:
+        RuntimeError:  if you pass an empty experiment list.
+    """
+    logger.info("Hello world!")
+
+    # Here's an example of an error that might be raised, as documented above.
+    if not experiments:
+        raise RuntimeError
```
graeme-winter: should be in `run()` method?

doc/examples/boilerplate.py
```diff
@@ -0,0 +1,135 @@
...
+    logger.info("The input reflection table contains %d reflections.", len(reflections))
+    logger.info(
+        "The input experiment list contains %d imagesets.", len(experiments.imagesets)
+    )
+
+    logger.info("integer_parameter: %i", params.integer_parameter)
+    logger.info("bool_parameter: %s", params.bool_parameter)
+
+
+def run(args=None, phil=phil_scope):  # type: (List[str], libtbx.phil.scope) -> None
```
graeme-winter: default to `args=` not set to `sys.argv[1:]`

doc/examples/boilerplate.py
ndevenish: I think it often does exist on our installs but only as a chained dependency. In python3 it's part of the standard library, so I think for now the better pattern is:
```
try:
    from typing import List
except ImportError:
    pass
```
Flake8 and other tools will accept this as an "imported symbol" and stop the type warnings

doc/examples/boilerplate.py
ndevenish: I suggest that we keep consistent with this - `boilerplate.py`, so `do_boilerplate`, `getLogger("dials.boilerplate")` (assuming no better way to get dispatcher entry point name?) etc. Basically, so that someone can search-and-replace to get a template rather than having to scan over everything for placeholdery words

doc/examples/boilerplate.py
ndevenish: If `OptionParser` gets passed None it defaults to sys.argv. Otherwise, you run the risk of list-caching, the same as putting a list object as a default to a function parameter (e.g. errors with `def a(some=[]): some.append(x)`)

doc/examples/boilerplate.py
benjaminhwilliams: Yes, I agree, I was looking to provide a simple example of a raise that needs documenting in the docstring, but this is probably a bad example.

doc/examples/boilerplate.py
benjaminhwilliams: Yep, the default behaviour is obfuscated by the DIALS option parser inheriting defaults from the (old) standard optparse option parser, but not documenting that clearly so nobody realises.  I'm happy to talk about ripping up the DIALS option parser, if you want...  :-P

doc/examples/boilerplate.py
benjaminhwilliams: Yes, good point, I think you're right.  I'll make a change.

doc/examples/boilerplate.py
benjaminhwilliams: Ooh yes, I was worried you might all be getting sick of my love of comments and docstrings but I'm happy to throw in more.

doc/examples/boilerplate.py
benjaminhwilliams: Great, thanks @ndevenish, I'll do that.

@graeme-winter, I think it's very helpful if we can try to use it, since it's not often clear to 
someone who is new to a particular corner of the code base which in-house special objects are being passed around.  With Nick's small addition, it's almost painless and doesn't break anything.

benjaminhwilliams:
Changes made.  Good to go?

dagewa:
I think timing could be added to the boilerplate `run` function, thus encouraging a more consistent format for presentation of this. See #932

benjaminhwilliams:
> I think timing could be added to the boilerplate `run` function, thus encouraging a more consistent format for presentation of this. See #932

Not a bad plan at all.  I'll suggest a straw man tomorrow and we can all tear it down at quiet moments during the DLS-CCP4 workshop.

benjaminhwilliams:
@dagewa, There you go — let the bikeshedding commence!

benjaminhwilliams:
OK, I think this is good to go.  @ndevenish, we're nominally waiting for your review.  Any comments?

dagewa:
@benjaminhwilliams thanks for the straw man timing example, it was useful to push the discussion forward.

benjaminhwilliams:
I don't think I'm going to hear from @ndevenish, so I'm just going to merge this.

PR #1043, opened 2019-11-26
Use unicode in report graphs

rjgildea:
Also add some subscripts and superscripts.
σ ✔
½ ✔


benjaminhwilliams:
I'm very much in favour.  Looks fine to me, since the test failure is unrelated.

Issue #1037, opened 2019-11-23
dials.symmetry fails on multi-sweep centred data set


graeme-winter:
```
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 2.0.2-g25fc2fc43-release
The following parameters have been modified:

input {
  experiments = integrated.expt
  reflections = integrated.refl
}

Detected existence of a multi-dataset reflection table 
containing 4 datasets. 
...
Incompatible unit cell: (4.80801, 12.8219, 16.5573, 73.5157, 90.0253, 79.2275)
      median unit cell: (4.80765, 12.8188, 16.5583, 106.445, 90.0189, 100.786)
```

graeme-winter:
Processed stepwise in `/dls/tmp/gw56/i19-check/raw` - integrated.(refl,expt) in there will reproduce issue. 

Reproduce issue on master.

Input fine:

```
cs03r-sc-serv-16 raw :( $ dials.show integrated.expt |grep Unit
    Unit cell: (4.8052(4), 12.8081(10), 16.5449(12), 106.458(3), 90.007(2), 100.777(2))
    Unit cell: (4.8080(4), 12.8219(10), 16.5573(12), 106.484(3), 90.025(2), 100.773(2))
    Unit cell: (4.8097(4), 12.8156(10), 16.5593(12), 106.490(3), 90.017(2), 100.804(3))
    Unit cell: (4.8073(3), 12.8224(10), 16.5604(12), 106.432(2), 90.021(2), 100.795(2))
```

suspect issue is it maps one to mC lattice then finds others don't match, as mapping back is ambiguous.

rjgildea:
Rather unhelpfully this works just fine on mac:
```
$ dials.symmetry integrated.*
$ dials.show symmetrized.expt | grep cell
    Unit cell: (25.164, 4.805, 16.545, 90.000, 106.764, 90.000)
    Unit cell: (25.192, 4.808, 16.557, 90.000, 106.794, 90.000)
    Unit cell: (25.177, 4.810, 16.559, 90.000, 106.800, 90.000)
    Unit cell: (25.191, 4.807, 16.560, 90.000, 106.741, 90.000)
```

rjgildea:
And even more unhelpfully, I can't reproduce it with my linux sources:
```
$ dials.symmetry /dls/tmp/gw56/i19-check/raw/integrated.{refl,expt}
[...]

PR #1046, opened 2019-11-27
If inputs all similar, use first reindex

graeme-winter:
Encodes a fix for #1037 - if the input unit cells are all the same (or similar) then only compute one change of basis op to a minimum cell and apply to all sweeps before analysis.

This is a horrible workaround to get past the stability issues highlighted in cctbx/cctbx_project#419

### DIFF
--- a/command_line/symmetry.py
+++ b/command_line/symmetry.py
@@ -7,7 +7,8 @@ import random
 import sys
 from dials.util import tabulate
 
-from cctbx import sgtbx
+from cctbx import sgtbx, uctbx
+from cctbx.sgtbx.bravais_types import bravais_lattice
 from cctbx.sgtbx.lattice_symmetry import metric_subgroups
 from libtbx import Auto
 import iotbx.phil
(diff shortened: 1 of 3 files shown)


command_line/symmetry.py
```diff
@@ -124,16 +124,31 @@ def map_to_minimum_cell(experiments, reflections, max_delta):
...
     Returns: The experiments and reflections mapped to the minimum cell
     """
+
+    # have a look to see if all the inputs are similar before we start - if
+    # yes use only one for this
+
+    c0 = experiments[0].crystal.get_unit_cell()
+    input_similar = all(
+        c0.is_similar_to(c)
+        for c in [expt.crystal.get_unit_cell() for expt in experiments]
```
rjgildea: There is no need for two for loops here, the following would be equivalent:
```
input_similar = all(
        c0.is_similar_to(expt.crystal.get_unit_cell())
        for expt in experiments
    )
```

Issue #846, opened 2019-07-12
image viewer saturation shown on variance image


dagewa:
The highlighting of pixels above the upper bound of the `trusted_range` should only occur for the normal image display, not for the variance image.

In this example, the image is capped at 8000 ADUs, but the variance values in the centre of the spot shown are >> 65535 (which is what the format class has set the upper trusted range bound to). The image viewer erroneously displays these with yellow overlay.
![Screenshot from 2019-07-12 12-39-03](https://user-images.githubusercontent.com/11502083/61125644-4d4c1600-a4a2-11e9-893a-d94c9d296e88.png)

PR #1049, opened 2019-12-01
Don't always show saturated pixels

dagewa:
This PR controls the display of overloaded pixels with a yellow mask so that this only occurs when the display is of `"image"` type (not `"variance"`, fixing #846, or any other type of image). It also switches off this display when > 1 image is being summed.

### DIFF
--- a/newsfragments/846.bugfix
+++ b/newsfragments/846.bugfix
@@ -1,0 +1,2 @@
+No longer show pixels that are above the trusted range upper bound as
+"saturated" on the "variance" image.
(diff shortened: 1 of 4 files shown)


dagewa:
An easy way of testing the difference between the branch and master:
```
dials.import $(dials.data get -q centroid_test_data)/*cbf trusted_range=-1,100
dials.image_viewer imported.expt
```
There are yellow saturated pixels on the "image" display, but these are not displayed if more than 1 image is summed, or if other image types such as "variance" is selected.

There is residual weird behaviour in that the pixels that are once marked yellow in "image" view become white in "variance" view, but this is cleared if you move +1 then -1 images while staying in variance view.

dagewa:
I think this change is fairly safe and I intend to merge soon (tomorrow?) for DIALS 2.1 unless anybody has any review comments.

dagewa:
For reference, I think this explains the "residual weird behaviour"
https://github.com/dials/dials/blob/327a64b4f232dbd01fdb6616de6fa325a84e1a0a/algorithms/image/filter/mean_and_variance.h#L252

Issue #1056, opened 2019-12-03
editing scans: there should be a dials program between indexing and refinement which does this


graeme-winter:
Discussion topic from this morning - 

At the moment `dials.refine` will trim experiment `scan` objects if there are no reflections. It does not handle complex cases e.g. crystals moving in / out of beam etc. and - perhaps - is not fitting in with the original remit of the program. 

Proposal floated from discussion with @dagewa this morning - we have a _new_ program which can optionally sit between `dials.index` and `dials.refine` which can be used to edit the models to allow refinement to be possible - in most cases this is not necessary but `dials.refine` can check the input and if invalid (e.g. no observations available for scan varying refinement) abort and ask that the input be tidied before attempting again. 

This will allow us to have the book keeping code for editing scans etc. to be useful and in one place, not embedded in refinement where it perhaps does not belong. 

This would imply removing the functionality from `dials.refine`

It would also mean testing this functionality would become trivial.

graeme-winter:
Activity ongoing on this in the "expeditor" branch

Issue #1029, opened 2019-11-21
dials.index: keep input experiments and extend list for new ones with crystals in 


graeme-winter:
I propose we carry all experiments through and extend the experiment list with new experiments in indexing - currently input experiments are deleted, new experiments made with crystals in. Propose to keep the input experiments then extend the list with new experiments with crystals in, so that the unindexed reflections can still refer to the crystal-less experiment. 

Means number of experiments after indexing will always be exactly equal to experiments in + lattices found. 

The graphs attached illustrate what I am talking about

![IMG_8116](https://user-images.githubusercontent.com/1241716/69353845-0b648500-0c77-11ea-9147-ee271318fd5e.jpg)

On the left we have the current behaviour - 

 - top - one experiment, no lattices indexed => 0 experiments
 - second - one experiment, one lattice => one out - usual
 - third - one experiment in, two out
 - bottom - two experiments go in, one with two lattices the other 0, you can never look at these images in the output 

Proposal would mean in every case the input is carried through faithfully and new experiments are added when lattices are found. 

This also means you can find experiments which share data as they have the same image sequence and scan objects.

graeme-winter:
Assert that the behaviour here should be:

1. Indexing extends the experiment list
2. Reflections which are indexed have the id which corresponds to that experiment
3. Reflections which are not indexed have their original id
4. No reflections leave with experiment id of -1

graeme-winter:
WIP in https://github.com/dials/dials/tree/dials-1029-30

phyy-nx:
Hi, currently the assumption is that if we have a indexed dataset, that all experiments in the experiment list have a crystal model.  I believe this proposed change to how indexed experiments are saved breaks any code like this:

```python
for expt in experiments:
  A = expt.crystal.get_A()
```

Here's an example:
https://github.com/dials/dials/blob/master/algorithms/indexing/nave_parameters.py#L49

Essentially, any code like that needs to change:
```python
for expt in experiments:
  if expt.crystal is None: continue
  A = expt.crystal.get_A()
```

graeme-winter:
Precisely

The corollary of this is if you do:

`dials.image_viewer indexed.expt`

the input experiments have not vanished. Also, since you _have_ indexed some experiments then (in an image sequence at least) the refined geometry could be used to attempt reindexing unindexed experiments or indexing further lattices.

No one is suggesting that this is a small change, but I think it should not be complex.

phyy-nx:
[...]

PR #1059, opened 2019-12-04
Expeditor - a tool to prepare data for refinement from indexing

graeme-winter:
See https://github.com/dials/dials/issues/1056 for rationale

This is a proposal for a new tool which can sit betwixt indexing and refinement for the more gnarly example data sets - for example where samples are not centred, or come and go from the beam, or you have multiple crystals arriving at the beam at different times. 

This also aims to take some responsibility for splitting crystal models before refinement where this is appropriate. 

There are also tests :-) 

Welcome feedback.

### DIFF
--- a/command_line/expeditor.py
+++ b/command_line/expeditor.py
@@ -13,7 +13,7 @@ Utility to wrangle experiments from indexing in preparation for refinement -
 will look to see that the sample is centred, that there are observations for
 the full range of every scan and rewrite the experiment lists if not.
 
-  dials.bullseye indexed.expt indexed.refl
+  dials.expeditor indexed.expt indexed.refl
 
 """
 
@@ -81,7 +81,7 @@ def reflections_fill_scan(reflections, scan, step_degrees):
 
 class Protocol(object):
     def __init__(self):
-        usage = "usage: dials.bullseye [options] indexed.expt indexed.refl"
+        usage = "usage: dials.expeditor [options] indexed.expt indexed.refl"
 
         self.parser = OptionParser(
             usage=usage,
--- a/command_line/expeditor.py
+++ b/command_line/expeditor.py
@@ -42,6 +42,7 @@ def select_scans_from_reflections(reflections, scan):
 
     z0, z1 = bbox[4], bbox[5]
     i0, i1 = scan.get_array_range()
+    o0, o1 = scan.get_oscillation()
 
     coverage = flex.int(i1 - i0, 0)
 
@@ -57,6 +58,7 @@ def select_scans_from_reflections(reflections, scan):
         l = list(g)
         s = copy.deepcopy(scan)
         s.set_image_range((l[0][1] + 1, l[-1][1] + 1))
+        s.set_oscillation((o0 + o1 * (l[0][1] + 1 - i0), o1))
         scans.append(s)
 
     return scans
--- a/command_line/expeditor.py
+++ b/command_line/expeditor.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from itertools import groupby
+from tabulate import tabulate
 import copy
 from libtbx.phil import parse
 from dials.util import show_mail_on_error
@@ -32,6 +33,35 @@ output {
 }
 """
 )
+
+
+def model_connectivity(experiments):
+    def model_connectivity_impl(experiments, model):
+        text = [""]
+        text.append("%s:" % model.capitalize())
+        models = getattr(experiments, "%ss" % model)()
+        rows = [[""] + [str(j) for j in range(len(models))]]
+        for j, e in enumerate(experiments):
+            row = ["Experiment %d" % j]
+            for m in models:
+                if getattr(e, model) is m:
+                    row.append("x")
+                else:
+                    row.append(".")
+            rows.append(row)
+        text.append(tabulate(rows, tablefmt="plain"))
+        return text
+
+    if len(experiments) == 1:
+        return ""
+
+    text = []
+    text.append("Experiment / Models")
+    text.extend(model_connectivity_impl(experiments, "detector"))
+    text.extend(model_connectivity_impl(experiments, "crystal"))
+    text.extend(model_connectivity_impl(experiments, "beam"))
+
+    return "\n".join(text)
 
 
 def select_scans_from_reflections(reflections, scan):
@@ -115,15 +145,27 @@ class Protocol(object):
         experiments = flatten_experiments(params.input.experiments)
         reflections = flatten_reflections(params.input.reflections)[0]
 
+        print(model_connectivity(experiments))
+
         z = reflections["xyzobs.px.value"].parts()[2]
 
         experiments_out = ExperimentList()
         reflections_id = flex.int(reflections.size(), -424242)
         reflections_id.set_selected(reflections["id"] == -1, -1)
 
+        crystal_scan = {}
+
         for j, e in enumerate(experiments):
+            if e.crystal in crystal_scan:
+                if e.scan is not crystal_scan[e.crystal]:
+                    print("Duplicating crystal for experiment %d" % j)
+                    e.crystal = copy.deepcopy(e.crystal)
+            else:
+                crystal_scan[e.crystal] = e.scan
+
             sel = reflections.select(reflections["id"] == j)
             i0, i1 = e.scan.get_image_range()
+
             print("Experiment %d has %d reflections" % (j, sel.size()))
             scans = select_scans_from_reflections(sel, e.scan)
 
@@ -156,6 +198,7 @@ class Protocol(object):
             i0, i1 = e.scan.get_image_range()
             print("Output experiment %d has %d reflections" % (j, sel.size()))
 
+        print(model_connectivity(experiments_out))
         experiments_out.as_file(params.output.experiments)
         reflections.as_file(params.output.reflections)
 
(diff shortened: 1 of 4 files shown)


graeme-winter:
Apropos 🚴‍♀️🏠 comments - https://en.wikipedia.org/wiki/Expeditor

benjaminhwilliams:
Looking at the changes, I think we need to merge #1042 and then (*gasp*) use it.

graeme-winter:
@benjaminhwilliams you mean retrofit the standard boiler plate example? Not offended by the suggestion.

benjaminhwilliams:
Yes.  Most of all, there's a lot going on inside that `run` method that could be outside in a function, to keep the command-line-specific parts minimal.  And in fact the whole class `Protocol` is really a function in disguise...

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+    return "\n".join(text)
+
+
+def select_scans_from_reflections(reflections, scan):
+    """Determine a list of valid scans where reflection shoeboxes are seen,
+    from within the bounds of the input scan."""
+
+    bbox = reflections["bbox"].parts()
+
+    z0, z1 = bbox[4], bbox[5]
```
rjgildea:
```suggestion
    z0, z1 = reflections["bbox"].parts()[4:]
```
might be simpler, since the `bbox` variable above isn't otherwise needed.

rjgildea: Or alternatively:
```_, _, _, _, z0, z1 = reflections["bbox"].parts()`

File: command_line/expeditor.py [hunk 2]
```diff
@@ -0,0 +1,213 @@
...
+
+    for _z0, _z1 in zip(z0, z1):
+        for j in range(_z0, _z1):
+            coverage[j] += 1
+
+    filled = (coverage > 0).iselection()
+
+    scans = []
+
+    for k, g in groupby(enumerate(filled), lambda n: n[0] - n[1]):
```
rjgildea: A comment describing this loop would be useful.

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+        l = list(g)
+        s = copy.deepcopy(scan)
+        s.set_image_range((l[0][1] + 1, l[-1][1] + 1))
```
rjgildea: The use of the variable name `l` combined with five uses of number `1` in this line makes it very hard to read. I would generally discourage the use of single latter variable names where possible.
In fact, PEP8 actively discourages the use of the variable name `l`: https://www.python.org/dev/peps/pep-0008/#names-to-avoid
> Never use the characters 'l' (lowercase letter el), 'O' (uppercase letter oh), or 'I' (uppercase letter eye) as single character variable names.

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+        s.set_oscillation((o0 + o1 * (l[0][1] + 1 - i0), o1))
```
rjgildea: Again, variable naming `o0`, `o1`, `i0` and `o1` make this very difficult to read.

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+            i0, i1 = e.scan.get_array_range()
+            sel = (reflections["id"] == j) & (z >= i0) & (z <= i1)
+            reflections_id.set_selected(sel, eid)
+
+            print(
+                "Output experiment %d has %d / %d reflections"
+                % (eid, sel.count(True), nref)
+            )
+
+            for k, s in enumerate(scans[1:]):
```
rjgildea: I don't see why this can't simply be 
```suggestion
            for k, s in enumerate(scans):
```
and eliminate the need for the duplicated code just above dealing with the first scan.

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+
+  dials.expeditor indexed.expt indexed.refl
+
+"""
+
+
+phil_scope = parse(
+    """
+output {
+  experiments = expedited.expt
```
rjgildea:
```suggestion
  experiments = trimmed.expt
```

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+    .type = path
+    .help = "Filename for fixed experiment list"
+
+  reflections = expedited.refl
```
rjgildea:
```suggestion
  reflections = trimmed.refl
```

command_line/expeditor.py
```diff
@@ -0,0 +1,213 @@
...
+
+        if len(params.input.reflections) != len(params.input.experiments):
+            print("Cardinality error")
+            self.parser.print_help()
+            return
+
+        experiments = flatten_experiments(params.input.experiments)
+        reflections = flatten_reflections(params.input.reflections)[0]
+
+        print(model_connectivity(experiments))
```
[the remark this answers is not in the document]
graeme-winter: Yes logging on the to-do - will be a side-effect of following @benjaminhwilliams recommendation of using the new-and-unreleased boilerplate.

graeme-winter:
> Adding an additional tool that sits between indexing and refinement seems like an unnecessary layer of complexity. If the use case is that:
> 
> * dials.index -> dials.refine : fails, then try:
> * dials.index -> dials.expeditor -> dials.refine : success
> 
> Then why not just incorporate this functionality into dials.refine so that users don't have to remember an additional tool (that most of the time doesn't do anything?)?

This work is *precisely because* you did not like `dials.refine` automatically trimming sweeps and said that this is not something which refinement should be doing. That said, it is something which I could see being generally useful (or more precisely - evolving in a generally useful direction) 

On my to-do is to factor this into analysis and activity modules so the analysis could be included at the end of `dials.index` and warn if refinement is likely to be unsuccessful because e.g. no observations for experiment X in scan region I1 - I2.

dagewa:
I would like to understand why trimming a scan to the observations is not a desirable feature. We have no other way at the moment to express the fact that a crystal may not diffract throughout the scan range. Images that do not contain diffraction are by definition, for that crystal, useless. If a crystal model _is_ provided for those images (i.e. by using the scan-static model), integrating them just wastes time and will not help data quality.

However, if we decide that modifying the scan is indeed not appropriate, then an alternative to using the scan to hold this information might be to make a new attribute within the crystal object to hold its observed range.

It would be possible to modify scan-varying refinement so that takes the observed range information from somewhere (or figures it out), refines a model within that range, then uses the model at the edges of that range to extrapolate out the model for missing images in the scan. But that is more work than giving the assumption that the crystal exists throughout the scan primacy, and therefore modifying the scan to ensure that remains the case.

jbeilstenedmands:
@dagewa Regarding your suggestion of a new attribute, i believe this functionality already exists, as I added it to help with the image range exclusion for scaling and related programs.
The scan has a valid_image_range property. This can be set using an experiment identifier as key, and a list of image ranges.
For example, in the most complex case you'd have something like this:
valid_ranges= {"0":[1,80,120,150], "1":[90, 110]}. This could represent a scan of 150 images, with the first experiment/lattice ("0") having 'valid' images from 1 to 80 and 120 to 150, and a separate lattice that was only valid from 90 to 110. There are also a few utility functions in `dials.util.exclude_images`.
The implementation could be tweaked but the ideas are there I think, and you don't need to modify the scan object in a way that would lose information.

dagewa:
@jbeilstenedmands this sounds useful and interesting, thanks. Before exploring this further here I will open a new issue, because behind all this lies a bad design decision in the way the scan-varying crystal model is stored in the crystal, and it may be worth dealing with that first.

rjgildea:
I do like the suggestions for somehow explicitly storing the information that only a part of the scan is valid for a particular application, rather than implicitly by deleting part of a scan. E.g. some version of reflection_table.flags, but for scans ranges:
```
scan.get_flags(scan.flags.valid_for_scan_varying)
```

https://github.com/dials/dials/blob/a6db22d71f7190d4b96224a91e35b58755ed84d5/array_family/boost_python/flex_reflection_table.cc#L1026-L1057

Also the idea of removing information (i.e. trimming the scans) seems to be at odds with @graeme-winter's suggestion in #1029 that we only ever add information, i.e. don't delete experiments that didn't index.

I also note that this PR does have some overlap with the existing [`dials.detect_blanks`](https://github.com/dials/dials/blob/master/command_line/detect_blanks.py), although obviously using different algorithms to detect "blank" regions of the scan. Whilst `dials.detect_blanks` does not itself directly remove any blank regions, xia2 can use the results to remove blanks regions and split up sweeps:
https://github.com/xia2/xia2/blob/a3db8c8c9ff7f72a15f4d1a5e20a2513c2ec2c9b/Modules/Indexer/DialsIndexer.py#L357-L425

graeme-winter:
@jbeilstenedmands re: `valid_image_range` - does this cope with the use case where we have the crystal out of the beam in the middle of the scan? 

The key above is that this generates multiple scans / experiments in this case which fits in well with our existing model. If folks have a better solution to this problem I'm all ears.

jbeilstenedmands:
@graeme-winter What is there so far is only for keeping track of the valid ranges, and a few utility functions like `get_selection_for_valid_image_ranges(reflection_table, experiment)` in `dials.util.exclude_images` but one could imagine implementing some methods to split an experiment based on the valid image ranges property.

graeme-winter:
Contemplating this some more and not trying to push a viewpoint - 

 - the experiment scan could be a subset of the imageset scan
 - if you want the full range, ask for the imageset scan
 - if you want to work with the data which exist, use the experiment scan

So the model above does meet this brief? 

Adding another object which stores scan ranges - but not _scan ranges_ - strikes me as adding to the already respectable pile of confusion.

Issue #706, opened 2019-02-13
dials.integrate background.algorithm=simple for integrating detectors


dagewa:
The default `background.algorithm=glm` is the best choice for photon-counting detectors, however my experience suggests that `background.algorithm=simple` is more appropriate for integrating detectors. The reason seems to be simply the inverse of the logic that dictates why the GLM approach is better for counts: pixel values on integrating detectors are better described by a normal distribution than a Poisson distribution. Assuming Poisson will tend to overestimate the background level. As a result, integrating beyond the resolution limit leads to intensities with a negative bias. This does not just affect CCDs, but also other types of integrating detector, such as certain types of detector used in electron diffraction and most likely detectors used at the XFEL, like the CS-PAD.

At the moment, the background algorithm is a user choice, but very few people know that the default may not be the most appropriate choice. Are there ways in which we could make this choice automatically? It seems like the [`detector_helper_sensors`](https://github.com/cctbx/cctbx_project/blob/1fce4707574dc7e5d5e134dd2b4d0c6f064e3490/dxtbx/model/detector_helpers.py#L109-L112) could be put to work here (though it is not a 1-1 mapping, some chips allow operation in either counting or integration modes).

phyy-nx:
For dials.stills_process the background algorithm default is simple, for
the reasons you describe. Most xfel data has been collected on these
integrated detectors.  But with a lot of pilatus serial data being
collected it's likely worth revisiting here too.  I'd be onboard with
adding an 'auto' option here and making that the default.

-Aaron

dagewa:
Good point - an inappropriate default can work both ways. `Auto` should solve most issues for detectors we know about. For detector types that could be either counting or integrating there is some chance the appropriate response model would be recorded in metadata (he says, while fighting natural pessimism)

dagewa:
@jmp1985 wondering if you have an opinion here?

jmp1985:
@dagewa I agree that having an auto which knows about the detector type would be useful. In general, if the pixels can assume negative values then the glm method may be the best solution; it is hard to argue that this could be correctly modelled by a Poisson distribution (in particular thinking of those data you had which had average background < 0!)

phyy-nx:
Sounds interesting.  What would the default cascade and pixel factors for
integrating detectors be?  Would putative integration code that uses these
factors suffer if the defaults were incorrect?

dagewa:
[...]

PR #1063, opened 2019-12-06
Add AutoBackgroundExt

dagewa:
Choose between glm and simple methods depending on whether the detector appears to be counting or not (for #706)

### DIFF
--- a/extensions/__init__.py
+++ b/extensions/__init__.py
@@ -45,7 +45,7 @@ class _Extension(object):
             if default_index < 0:
                 default_index = 0
             if names:
-                names[default_index] = "*" + names[default_index]
+                names[default_index] = "*" + str(names[default_index])
             return names
 
         exts = cls.extensions()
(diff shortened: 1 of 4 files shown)


phyy-nx:
This is great.  I wonder if the dials.stills_process override should be dropped when this is merged.  Currently, I have it set to change the background algorithm to simple instead of glm:
https://github.com/dials/dials/blob/master/command_line/stills_process.py#L212
This is because most XFEL detectors are integrating detectors.  But I know with serial crystallography at synchrotrons, the GLM algorithm might be getter.  I've just never tried it.  Anyone here have a report on that?

dagewa:
Hi @phyy-nx I haven't tried `dials.stills_process` on synchrotron counting detectors, but I do expect that the right behaviour would be to choose `glm` if the detector is counting and `simple` otherwise. The logic here should get that right in most cases, except for some exotic situations (one Timepix format is counting but has a gain of exactly 3)

phyy-nx:
👍

PR #1075, opened 2019-12-16
Fixes for dials 1029 and 1030

graeme-winter:
Keep goniometers and scans around if provided in input
Keep input experiments in output from dials.index
Keep experiment id column pointing at input experiment for unindexed reflections

PR #1076, opened 2019-12-16
Fix 1030: keep goniometers and scans around if provided in input

graeme-winter:
Includes:
 - do not delete things which are provided as input
 - test change to switch back outlier rejection to sauter_poon
N.B. to get original treatment can import images as stills which deletes goniometer and scan from the get go. 

These are cherry-picked from https://github.com/dials/dials/pull/1075


phyy-nx:
Hi, don't merge this yet, if that's ok.  I need to think about it.

graeme-winter:
Sure, though if you have a timescale for taking a look I’d appreciate it 🙂👍

phyy-nx:
I've had my think.  Ok to merge.  I can't think of a use case that is broken from my previous post:
https://github.com/dials/dials/issues/1030#issuecomment-557721604

I still would like to see a grid scan and its associated headers so I could look a the resultant imported .json file.

graeme-winter:
Thanks @phyy-nx yes we should add some "proper" grid data to the test set for this, probably via zenodo.

PR #902, opened 2019-08-29
Implement unique experiment identifiers

jbeilstenedmands:
**Overview**
This pull request implements the generation and use of unique experiment identifiers throughout the dials processing workflow. The motivation behind this is to allow matching of experiments and reflection tables without relying on the ordering of these objects (either on loading data or during processing. This feature has already proved useful in `dials.scale` and `dials.cosym`).

These identifiers (a uuid string by default) are stored in `experiment.identifier`, and the `reflection_table.experiment_identifiers()` map is used to record the numerical_id -> identifier mapping.
When a new experiment is created, a unique identifier is generated - this only happens in `dials.import` and `dials.index`. When a reflection table is created, the experiment_identifiers map is populated - this happens in spot-finding, indexing and integration. Other programs in the standard dials workflow add to the existing table, so the mapping is preserved.

So far I have only implemented and tested the uuid identifier generation for the sweeps case. For the stills case, I have added in function calls in the relevant locations in `stills_indexer` (which would currently generate a uuid); I invite @asmit3 and collaborators to add to this PR to modify this as you wish as I am unfamiliar with the ins and outs of the stills_process workflow.

People can probably ignore the changes to the scaling files if reviewing this. I shall continue to review the dials programs for the 'potential issue' (see below). However, the tests pass locally for me and wanted to get this PR out there for feedback.

**Detailed description on behaviour changes**
The behavioural changes are primarily to help programs which handle the input of multiple datasets in separate files - which will likely have duplicate numerical ids in the reflection tables.

1.  To implement safe handling of data on loading, regardless of order, there is a new function to perform the role of `flatten_experiments` and `flatten_refections`, which also reorders the reflection tables to match the experiments order using the experiment identifiers;
`refls, expts =  reflections_and_experiments_from_files(refl_file_objs, expt_file_objs)`
(note that the id columns in the tables are not renumbered at this point. This was something I introduced recently in `flatten_reflections` for #656 but agree that this functionality should occur elsewhere, see below).
So far I have only replaced the flatten functions in `dials.scale`, but this can be rolled out to other programs if people are happy about this. 

2. To ensure that multiple reflection tables can be joined together, one must ensure that the column id values do not clash. To do this, one can use the `renumber_table_id_columns(reflection_tables)` function in `dials.util.multi_dataset_handling`. There is also a `split_reflection_tables_on_ids(reflection_tables)` function for splitting multi-dataset tables on ids.

[...]

PR #1077, opened 2019-12-16
Handle selecting a reflection table with experiment identifiers

jbeilstenedmands:
This short PR introduces some new functionality for handling reflection table selections when experiment identifiers are set, to form the basis for future work.

As suggested by @dagewa in #902, the experiment identifiers should be updated when making a selection. So for the case of `new_refl = refl.select(selection_array)`, e.g. `new_refl = refl.select(refl["id"] == 0)`,  if `refl` contains an id column and identifiers are set, then the identifiers map is now set in `new_refl`for all ids in `new_refl["id"]`. This differs from the current behaviour, where all of the identifiers are copied across to the new table, which could cause issues later on.

Also implemented are two new select methods, for passing an Experiment or ExperimentList to reflections.select(), to allow
```
for exp in experiments:
    refl = reflections.select(exp)
```
or
```
for elist in experimentlists:
    refl = reflections.select(elist)
```
as opposed to the current motif
```
for i, exp in enumerate(experiments):
     refl = reflections.select(refl["id"] == i)
```
for a reflection table `reflections` containing data from multiple experiments.

### DIFF
--- a/array_family/boost_python/flex_table_suite.h
+++ b/array_family/boost_python/flex_table_suite.h
@@ -16,6 +16,7 @@
 #include <iterator>
 #include <iostream>
 #include <sstream>
+#include <unordered_set>
 #include <boost/python.hpp>
 #include <boost/python/def.hpp>
 #include <boost/python/slice.hpp>
@@ -28,6 +29,8 @@
 #include <dials/array_family/flex_table.h>
 #include <dials/array_family/scitbx_shared_and_versa.h>
 #include <dials/error.h>
+#include <dxtbx/model/experiment.h>
+#include <dxtbx/model/experiment_list.h>
 
 namespace dials { namespace af { namespace boost_python { namespace flex_table_suite {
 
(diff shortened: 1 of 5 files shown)


test/array_family/test_identifiers_handling.py
```diff
@@ -0,0 +1,105 @@
...
+    If no identifiers are set, then nothing should happen."""
+
+    # Test when identifiers & id column set.
+    refl = flex.reflection_table()
+    refl["id"] = flex.int([0, 0, 0, 1, 1, 1, 2, 2, 2])
+    refl.experiment_identifiers()[0] = "0"
+    refl.experiment_identifiers()[1] = "1"
+    refl.experiment_identifiers()[2] = "2"
+
+    new_refl = refl.select(refl["id"] == 0)
```
dagewa: Might be worth testing using `refl["id"] == 1` instead as a demonstration that identifiers are _not_ reset to start from 0 in the new table

test/array_family/test_identifiers_handling.py
```diff
@@ -0,0 +1,105 @@
...
+    refl.experiment_identifiers()[2] = "2"
+    experiments = ExperimentList(
+        [
+            Experiment(identifier="0"),
+            Experiment(identifier="2"),
+            Experiment(identifier="3"),
+        ]
+    )
+
+    sel_refl = refl.select(experiments)
```
dagewa: This line prints
```
0
2
3
```

array_family/boost_python/flex_table_suite.h
```diff
@@ -668,6 +710,80 @@ namespace dials { namespace af { namespace boost_python { namespace flex_table_s
...
+
+  template <typename T>
+  T select_using_experiments(T &self, dxtbx::model::ExperimentList expts){
+    typedef typename T::experiment_map_type::const_iterator const_iterator;
+    typedef dxtbx::model::ExperimentList::shared_type::const_iterator expt_const_iterator;
+    T result;
+    for (expt_const_iterator expt = expts.begin(); expt != expts.end(); ++expt) {
+
+      std::string identifier = expt->get_identifier();
+      std::cout << identifier << std::endl;
```
dagewa: Found the source of the print in the tests. This should be removed

jbeilstenedmands:
The failing test has highlighted an important consideration. My code assumes that the id values are ints, but you could have size_t or others i guess. Perhaps I need some templating here...

PR #1050, opened 2019-12-01
Prefix logger.warning messages with "WARNING:"

dagewa:
This is a way of prefixing warning messages and thus make them stand out from general information messages in a consistent way. Draft for comments whether we want this kind of behaviour. This is for #1034.

Issue #1034, opened 2019-11-22
Some logging tidy-up


dagewa:
`logger.warning` is used in a few places in the codebase. Use is not very consistent:

* `logger.warning("*** WARNING: no reflections in job %d ***", index)`
* `logger.warning("Restraints will be ignored for scan_varying=True")`
* `logger.warning("Warning: Experiment crystals differ. Using first experiment crystal for file-level data.")`

There are also warnings that aren't `warning`s
* `logger.info("Warning: %d reference spots were not matched to predictions" % (len(unmatched)))`

- [x] It would be nice to have a consistent style. If `logger.warning` took care of the `*** WARNING`/`Warning:` or whatever prefix, this would help.

- [ ] Capitalisation - `Warning: experiment` not `Warning: Experiment` preferred

- [ ] Finally, 
```python
logger.info("\n")
logger.info("thing")
logger.info("\n")
```
one call per message please

dagewa:
@Anthchirp I'm wondering how adding a prefix like `"WARNING:"` to `logger.warning` messages would work. Is it something that can be added to `dials.util.log.config`?

PR #1078, opened 2019-12-16
Colour warning (and above) messages on the console

dagewa:
This complements #1050

After this, warning messages are suitably highlighted on the console and in log files, so the various different styles used (see #1034) can be made consistent and the message can just contain the message contents.


Anthchirp:
Something like https://github.com/borntyping/python-colorlog is probably more appropriate, as it uses more than one colour and is cross-platform. (The solution proposed here does not work on Windows).

Or at least use the dlstbx ColorStreamHandler (from https://gist.github.com/mooware/a1ed40987b6cc9ab9c65), which is already linked in `util/log.py`. Either way the dlstbx reference should be removed.

Any accepted solution must respect the [NO_COLOR environment variable](https://no-color.org/) and check that the output actually goes to a console before enabling colour.

dagewa:
I was looking for ColorStreamHandler but could not find it. Happy to use that instead

dagewa:
Ok, I like colorlog. It seems to be available on pip and conda. I've changed this PR to use that. How do I add the package as dependency for builds (if this version accepted)?

dagewa:
Test case I've been using is `dials.index imported.expt strong.refl scan_varying=true`
Screenshot from 2019-12-16 21-20-19

util/log.py
```diff
@@ -6,10 +6,7 @@
 import sys
 import time
 
-try:
-    from dlstbx.util.colorstreamhandler import ColorStreamHandler
-except ImportError:
-    ColorStreamHandler = None
+from colorlog import ColoredFormatter
```
Anthchirp: We will need to trap the import similarly to the dlstbx version because even when we add the dependency everywhere not all development environments will pick it up automatically. Specifically we don't currently update conda environment dependencies.

Anthchirp:
I'd also like to suggest giving DEBUG level messages another specific colour, for example blue:
image

dagewa:
Done. I forgot we weren't updating conda. Thanks for the idea about blue debug messages, I like it.

dagewa:
I'm going to add news and tidy up existing `warning` messages in this branch before merge

Anthchirp:
please also add a newsfragment and when you squash merge replace the default commit message with a summary only :slightly_smiling_face:

rjgildea:
FWIW not everyone uses a black (or purple?) terminal:
Screenshot 2019-12-17 at 12 21 38

Anthchirp:
No problem, we can add a terminal client with black background to the DIALS distribution :+1:

dagewa:
@rjgildea  feel free to adjust colours after merge.

Issue #1061, opened 2019-12-06
Mask generation very slow now


ndevenish:
Running on both `module load dials/latest` and my own python3, mask generation seems to be slow now, such that when running `dials.find_spots` there is a long wait before spotfinding begins e.g. between these two lines of output:
```
Extracting strong pixels from images
 Using multiprocessing with 5 parallel job(s)
```
I ran profiling and it seems to be in the pre-spotfinding mask generation:

![image](https://user-images.githubusercontent.com/529963/70320526-dc284900-181c-11ea-8072-f738e183b8ef.png)

graeme-winter:
See https://github.com/dials/dials/pull/980 as probable source

graeme-winter:
Uuuhh I think this is silently a massive problem? Discuss?

graeme-winter:
Trying to find spots on an 8,000 image grid and spin-up time taking $ages of mankind.

dagewa:
suggested workaround, use the first 1000 images only to look for hot pixels. Until we have a fast way of doing this through the whole sweep

graeme-winter:
workaround of `use_trusted_range=false` gets the spot finding going straightaway...

graeme-winter:
@dagewa reasonable however even 1k hot pixel search is time consuming. Not sure of the right solution here but I think it is fair to say that the current solution is not right.

dagewa:
Original code simply used the first image. That was no good. Even just look at first and last image would be much better

graeme-winter:
Or a “large” random selection drawn from the data set… yes

Large as in, “I can’t be bothered to count that many"

dagewa:
Or, indeed, don't add trusted range into the mask. Isn't it effectively masked anyway, by being an "overload" shoebox?

Anthchirp:
massive regression on DIALS 2.1
for beamline spotfinding plots this adds 66 seconds/img * 250 images per data collection = 4.6 CPU hours.

dagewa:
Ouch, actually just read @Anthchirp's comment properly - 66 seconds _per image_! The 100 image workaround is still no good then

dagewa:
@rjgildea points out this is possibly because per image analysis treats every image separately and may well then be running this whole mask calculation (which touches every image) repeatedly

dagewa:
#1084 is now faster. Is this a sensible way to go?

graeme-winter:
as a stop-gap - `use_trusted_range=false` - ?

PR #1084, opened 2019-12-18
Fast masking

dagewa:
Dumb workaround for #1061

Pixels outside the trusted mask on at least 100 images
are considered bad and masked

### DIFF
--- a/util/masking.py
+++ b/util/masking.py
@@ -155,7 +155,14 @@ class MaskGenerator(object):
             if self.params.use_trusted_range:
                 trusted_mask = None
                 low, high = panel.get_trusted_range()
-                for image_index, _ in enumerate(imageset.indices()):
+
+                # Take 10 evenly-spaced images from the imageset. Pixels outside
+                # the trusted mask on all of these images are considered bad and
+                # masked. https://github.com/dials/dials/issues/1061
+                stride = max(int(len(imageset) / 10), 1)
+                image_indices = range(0, len(imageset), stride)
+
+                for image_index in image_indices:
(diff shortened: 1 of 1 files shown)


dagewa:
Ugh. Even on this branch, the masking calculation takes nearly 8 seconds for 100 cbfs

```
dials.import $(dials.data get -q thaumatin_i04)/*.cbf
```

Snippet from log with changes to time the calculation
```
   0.1: Start trusted range masks
   7.8: End trusted range masks
```

dagewa:
Changed tactic. Now takes 10 images distributed through the imageset. The `thaumatin_i04` job now takes about 0.5s.

graeme-winter:
@dagewa looking at this now - however may be worth explicitly ensuring that we have odd and even frames (or prime numbered frames?) to look for twinklers?

dagewa:
That's an interesting idea. Perhaps 5 pairs of sequential frames rather than 10 evenly spread? I don't know much about the various classes of dodgy pixels to devise optimal sampling.

Anthchirp:
Am wondering - Is the hot pixel information relevant enough to have it available for spot finding, or would it be permissible to accumulate the information during the spot finding main loop and then use it from indexing onwards?
(This may have been asked before, don't recall)

An implicit side-effect of this strategy is that it would only take the spot-found region of the dataset into account.

dagewa:
dials.generate_mask also runs this code, where spot finding information is not available. However it matters far less if that program is slow.

graeme-winter:
I'm trying to fix up the capricious pixel discovery code today which would probably be a better place to look at mask generation from anyways.

PR #1086, opened 2020-01-09
Use experiment identifiers from import

jbeilstenedmands:
This is an updated version of a previous PR (#902).

The idea is the same, generate a unique (uuid) string identifier on import, do the same when discovering lattices upon indexing, and propagate through the processing (and checking each time so that files without identifiers still work for back compatibility). Most of the file changes are updating the flatten_reflections/flatten_experiments calls, to handle any order of input data from the command line.

At the moment, most programs don't really use the identifier to any particular benefit, that can come as part of future work as required.

I have also generated a uuid in the stills indexer as a demonstration, and checked that stills processing still completes fine (although there are possibly more complex workflows that i am not able to properly check?). Without the changes to `stills_indexer.py`, no identifiers are generated (at any point in stills processing), and the processing completes the same. I would like to ask for feedback from @asmit3 or @phyy-nx as to which option is preferred:
1) Generate a uuid now, as per this PR, and let you guys change the format of this to timestamp at a later point if interested.
2) Leave `stills_indexer.py` unchanged, so that no identifiers are generated for stills at present (although this may cause issues if using the regular indexer for stills which is sometimes done?)
3) Let you add to this PR to generate a timestamp identifier in `stills_indexer.py`.

### DIFF
--- a/algorithms/indexing/assign_indices.py
+++ b/algorithms/indexing/assign_indices.py
@@ -60,6 +60,10 @@ class AssignIndicesGlobal(AssignIndicesStrategy):
                 sel_cryst = crystal_ids == i_cryst
                 for i_expt in experiments.where(crystal=cryst, imageset=imgset):
                     expt_ids.set_selected(sel_cryst, i_expt)
+                    if experiments[i_expt].identifier:
+                        reflections.experiment_identifiers()[i_expt] = experiments[
+                            i_expt
+                        ].identifier
 
             reflections["miller_index"].set_selected(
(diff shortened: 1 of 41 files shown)


asmit3:
Hi @jbeilstenedmands , we will take a look at this on Monday pacific time and get back. Please hold on till then. Thanks

asmit3:
Hi @jbeilstenedmands , @phyy-nx and I took a look at this PR and it looks good to be merged.

jbeilstenedmands:
@asmit3 thanks for taking a look at this.

PR #1090, opened 2020-01-13
Accommodate absorption in a diamond anvil cell

benjaminhwilliams:
Diamond anvil cells are commonly used to allow measurement structures under applied pressure.  Both the incident and diffracted X-ray beam are attenuated by absorption in the anvils between which the sample is sandwiched.  The path length of the incident beam through the diamond, and hence the amount of absorption, depends on the orientation of the cell to the incident beam.  The path length of the diffracted beam depends on the orientation of the cell and the diffracted angle.  Hence a non-uniform absorption factor applies to reflection intensities, which affects the fidelity of data reduction using the existing DIALS scaling tools.

By modelling the anvils as flat plates, of equal thickness and with parallel sides, the overall absorption factor can be calculated.  It is a function of the incident and diffracted beam directions, the thickness of each anvil and their normal (itself a function of the mounting orientation and goniometer rotation) and the linear attenuation coefficient of the anvil material (itself a function of wavelength and the material's density).

We can thereby correct the integrated intensities for the anvil absorption and then apply the post-integration components of the data reduction pipeline as normal, just as though the sample had been measured in air.

For diamond anvil cell data, this tool should be called on the output of `dials.integrate` before any subsequent processing.  The default anvil parameters reflect the cell design used on I19-2 at Diamond Light Source.


benjaminhwilliams:
There are obviously no tests included here.  This comes back to earlier discussions about the best way to incorporate regression data for small molecule work.  My preference would be to merge this, test it in the school of hard knocks (I19-2) and make adding tests a separate task.

rjgildea:
For this pull request you shouldn't need a whole dataset added to dials-data - just the integrated experiments/reflections should be enough to exercise this command. To test `correct_intensities_for_dac_attenuation` you could probably even generate some fake experiments/reflections on the fly for the purposes of exercising the function.

command_line/rescale_diamond_anvil_cell.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    % filename_root
+)
+
+# Get the tabulated NIST mass attenuation coefficient data for carbon.
+carbon_attenuation_data = attenuation_coefficient.get_table("C")
+
+
+def correct_intensities_for_dac_attenuation(
+    elist,  # type: ExperimentList
```
rjgildea: Is there any particular reason for using the variable names `elist` and `rtable` rather than `experiments` and `reflections` which appear to be much more commonly used throughout the codebase by a variety of authors? E.g.
https://github.com/dials/dials/blob/master/command_line/refine.py#L251
https://github.com/dials/dials/blob/master/command_line/index.py#L122
https://github.com/dials/dials/blob/master/command_line/scale.py#L151

benjaminhwilliams: I was just trying to avoid shadowing the parameter names with the values passed to them in lines https://github.com/dials/dials/blob/f9862c6b068ca9fb58848a98f8d28bdf04f03a33/command_line/rescale_diamond_anvil_cell.py#L286-L288.  Old habits die hard, I guess.

benjaminhwilliams: > For this pull request you shouldn't need a whole dataset added to dials-data - just the integrated experiments/reflections should be enough to exercise this command. To test `correct_intensities_for_dac_attenuation` you could probably even generate some fake experiments/reflections on the fly for the purposes of exercising the function.

OK, I'll grab you tomorrow to chat about it.  I'm not sure of the best way of measuring success/failure of such a test.

graeme-winter:
Agree with comments from @rjgildea on tests - they can run on very small data sets, anything high pressure will be fine, just need the integrated reflection files. Looking through the changes now.

command_line/rescale_diamond_anvil_cell.py
```diff
@@ -0,0 +1,302 @@
+# coding: utf-8
```
graeme-winter: Here for convenience - why is this called "rescale diamond anvil cell" rather than e.g. "correct diamond anvil" - purely a bikeshed thing but probably good to get right from the start.

command_line/rescale_diamond_anvil_cell.py
```diff
@@ -0,0 +1,302 @@
...
+        dac_norm:  A 3-vector representing the normal to the anvil surfaces in the
+                   laboratory frame when the goniometer is at zero datum, i.e. the axes
+                   are all at 0°.  The vector is assumed to be normalised.
+        thickness:  The thickness of each diamond anvil (assumed equal).
+        density:  The density of the anvil material in g·cm⁻³
+                  (units chosen for consistency with the NIST tables of X-ray mass
+                  attenuation coefficients, https://dx.doi.org/10.18434/T4D01F).
+                  Defaults to a typical value for synthetic diamond.
+    """
+    # Iterate over all experiments in the experiment list.
```
graeme-winter: As a motif I find:

```
# do thing
thing.do()
```

redundant 😉 - appreciate useful when mapping out how code will be laid out but not sure useful in final version

graeme-winter:
Not a precondition on the merge but a proposal for the future - pretty sure everything done in here could easily have been achieved with `flex` calculations rather than `scipy/numpy` stuff - which would make it slightly more in line with the rest of the dials code base, though admittedly slightly harder for the "developer in the street"

command_line/rescale_diamond_anvil_cell.py
```diff
@@ -0,0 +1,302 @@
...
+        input_errors.append(
+            "Please provide at least one valid reflection table (.refl) file."
+        )
+    if input_errors:
+        sys.exit("\n".join([parser.format_help()] + input_errors))
+
+    # These functions are commonly used to collate the input.
+    experiments = flatten_experiments(params.input.experiments)
+    reflections_list = flatten_reflections(params.input.reflections)
+    reflections = reflections_list[0]
```
jbeilstenedmands: Note, if you wanted to follow my comment and split the input reflections_list into a list of single-dataset tables, just use `parse_multiple_datasets` from `dials.util.multi_dataset_handling`

benjaminhwilliams:
> Not a precondition on the merge but a proposal for the future - pretty sure everything done in here could easily have been achieved with `flex` calculations rather than `scipy/numpy` stuff - which would make it slightly more in line with the rest of the dials code base, though admittedly slightly harder for the "developer in the street"

I wasn't sure if there was an efficient way in `flex` to perform some of the arithmetic here, such as the dot product of each of an array of vectors with each of another array of vectors.  This may simply be `flex`-ignorance on my part but sometimes life's just too short when the alternative is easy, quick and staring me in the face.

benjaminhwilliams:
> Looks good overall. Wasn't really able to follow the maths just from reading but trust you've got this sorted :)
Yes, I started trying to write it out in the PR description and then remembered that maths in GitHub isn't easy.  Perhaps it would be better in the wiki or something.  I'll finish writing it up so people can refer to it in future.

> One thing regarding multiple reflection tables, I notice you join them all together at the start, then go through selecting each one to apply the corrections individually.
> I think we've shared our differing views on this issue before, but in my opinion its better to split the input reflections into a list of single-dataset tables, then have a function which runs on a single experiment and single-dataset reflection table, as the algorithm is basically doing in this case (or have a function working on an ExperimentList and list of reflection tables the same length). The tables can then be joined together at the end for output. This was probably makes it easier for the algorithm code to be more 'functional' in my opinion.
I may be coming round to your way of thinking on this.  I'll see about refactoring it slightly.

command_line/rescale_diamond_anvil_cell.py
benjaminhwilliams: Yes that's probably exactly how it came to be there.  And I tend to err on the side of leaving comments in rather than preferring comment-free code.  I know the house style is rather the opposite... :-P  This one though, we could probably lose.

command_line/rescale_diamond_anvil_cell.py
benjaminhwilliams: Yep, it's a good point.  Opinions, @jbeilstenedmands, @rjgildea?

rjgildea: "correct" is potentially ambiguous in that it can be an adjective or a verb, whereas at least "rescale" is unambiguous in that sense...

phyy-nx:
Hi, just to help out a bit :)

a = flex.vec3_double(10, (1,0,0))
b = flex.vec3_double(10, (0,1,0))
a.dot(b)
list(a.dot(b))
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]

Also IPython's ? feature is helpful here.

I haven't read the code, but changing between flex and numpy (using
as_numpy_array() for example) does do a copy, so I try to avoid it as much
as possible.  If I start in flex or numpy I try to stay there.
-Aaron

benjaminhwilliams:
> Hi, just to help out a bit :)

Ah, thanks Aaron.  And if I want to apply each of an array of rotation operators to each of an array of vectors, is there a `flex`y way to do that?

phyy-nx:
Yep.  I'd do it like this:

from scitbx import matrix
a = flex.vec3_double(10, (1,0,0))
r = matrix.col((0,1,0)).axis_and_angle_as_r3_rotation_matrix(90, deg=True)
b = (flex.mat3_double(10, r)*a)
print(b[0])

(2.220446049250313e-16, 0.0, -1.0)

The slight numerical instability there is interesting of course.  I haven't
tried compiling with fast math off and no optimisations, but I wonder if
that would clean it up.

command_line/rescale_diamond_anvil_cell.py
benjaminhwilliams: I've settled on `dials.anvil_correction`, I hope that appeases the bikehed committee?

command_line/rescale_diamond_anvil_cell.py
benjaminhwilliams: Hi @jbeilstenedmands, I've reformatted this to use tools from `dials.util.multi_dataset_handling` as per your suggestion.  Could you look over it and verify that I'm holding it correctly?

benjaminhwilliams:
> Looks good overall. Wasn't really able to follow the maths just from reading but trust you've got this sorted :)

The changes to the docs in this branch contain details of the arithmetic, in case you want to look that over.

command_line/anvil_correction.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+    if input_errors:
+        sys.exit("\n".join([parser.format_help()] + input_errors))
+
+    # These functions are commonly used to collate the input.
+    experiments = flatten_experiments(params.input.experiments)
+    reflections_list = flatten_reflections(params.input.reflections)
+    reflections_list = parse_multiple_datasets(reflections_list)
+    reflections_list = sort_tables_to_experiments_order(reflections_list, experiments)
+    # Work around parse_multiple_datasets dropping unindexed reflections.
```
jbeilstenedmands: I think this workaround should be before the call to parse_multiple_datasets?

benjaminhwilliams: Oh dear, thanks, that was a slip.  Well done for spotting that.

benjaminhwilliams:
@rjgildea, this now has tests.  Does it have your approval?

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+    # Get the axis of the scan rotation.
+    rotation_axis = experiment.goniometer.get_rotation_axis_datum()
+    # For each reflection, get the angle of the scan rotation.
+    angles = refls_sel["xyzobs.mm.value"].parts()[2]
+    # Construct a rotation vector (parallel with the rotation axis and with
+    # magnitude equal to the rotation angle) for each reflection.
+    # The shape of this array is (N, 3), where N is the number of reflections.
+    rotvecs = np.outer(angles, rotation_axis)
+    # Remove redundant things that scale with N.
+    del angles
```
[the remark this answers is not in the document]
Anthchirp: if this is an issue then that's an argument for breaking up functions to limit the lifetime of these objects

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+try:
+    from typing import List, Sequence, SupportsFloat
+except ImportError:
+    pass
+else:
+    Vector = Sequence[SupportsFloat]
+
+
+filename_root = os.path.splitext(os.path.basename(__file__))[0]
+logger = logging.getLogger("dials.%s" % filename_root)
```
Anthchirp:
```suggestion
logger = logging.getLogger("dials.command_line.anvil_correction")
```
The `command_line` bit is missing. Doesn't make any difference, but is convention.

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+            .help = '''A 3-vector representing the normal to the anvil surfaces in the
+            laboratory frame when the goniometer is at zero datum, i.e. the
+            axes are all at zero degrees.  The vector may be given
+            un-normalised.'''
+    }
+
+    output {
+        experiments = None
+            .type = path
+            .optional = True
```
Anthchirp: is this the same `optional` thing we discussed elsewhere? intuitively correct, but really inappropriate?

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+            .help = '''The output experiment list file name.
+            If None, don't output an experiment list file.'''
+        reflections = corrected.refl
+            .type = path
+            .help = "The output reflection table file."
+        log = dials.%s.log
+            .type = path
+    }
+    """
+    % filename_root
```
Anthchirp: just use the fixed name in the string?

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+    # Work around parse_multiple_datasets dropping unindexed reflections.
+    unindexed = flex.reflection_table()
+    for r_table in reflections_list:
+        unindexed.extend(r_table.select(r_table["id"] == -1))
+    # Get a single reflection table per experiment object.
+    reflections_list = parse_multiple_datasets(reflections_list)
+    reflections_list = sort_tables_to_experiments_order(reflections_list, experiments)
+
+    # Record the density of diamond in g·cm⁻³ (for consistency with NIST tables,
+    # https://dx.doi.org/10.18434/T4D01F).
```
Anthchirp:
```suggestion
    # https://doi.org/10.18434/T4D01F).
```
new resolver address

command_line/anvil_correction.py
benjaminhwilliams: Sorry, I'm belatedly picking up on this.  I don't remember this discussion...  What's the problem?

rjgildea: https://github.com/dials/dials/pull/1157#discussion_r380094969

rjgildea: I.e. `.optional` has absolutely no effect here, so it would be better to remove it to avoid potential confusion.

benjaminhwilliams: Ah yes, thanks for the reminder.

command_line/anvil_correction.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # Get the normalised incident and diffracted beam vectors ̠̂s₀ and ̠̂s₁.
+    # Naturally, there will be one vale of ̠̂s₁ for each reflection.
+    s0_norm = experiment.beam.get_unit_s0()  # An array of shape (3).
+    s1_norm = refls_sel["s1"] * wavelength  # Shape (N, 3).
+
+    # Get the scalar product of the diamond anvil cell orientation with each of
+    # ̠̂s₀ & ̠̂s₁ for each reflection.
+    incident_cosine = np.dot(dac_orientation, s0_norm)  # An array of shape (N).
+    diffracted_cosine = np.einsum("ij,ij->i", dac_orientation, s1_norm)  # Shape (N)
```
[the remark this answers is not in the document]
benjaminhwilliams: Unfortunately, the dot product of each of an array of _N_ 3-vectors with its counterpart from another array of _N_ 3-vectors has no user-friendly NumPy function name.  `numpy.dot`, `numpy.tensordot`, `numpy.inner`, `np.multiply`, etc. don't offer a simple shorthand for `np.sum(foo * bar, axis=1)`.  Of course, I could use that expression explicitly but `einsum` is quicker:
```Python
In[1]: import numpy as np
In[2]: foo = np.arange(18).reshape(6, 3)
In[3]: bar = 2 * foo
In[4]: foo
Out[4]: 
array([[ 0,  1,  2],
       [ 3,  4,  5],
       [ 6,  7,  8],
       [ 9, 10, 11],
       [12, 13, 14],
       [15, 16, 17]])
In[5]: bar
Out[5]: 
array([[ 0,  2,  4],
       [ 6,  8, 10],
       [12, 14, 16],
       [18, 20, 22],
       [24, 26, 28],
       [30, 32, 34]])
In[6]: %timeit np.sum(foo * bar, axis=1)
4.68 µs ± 86.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
In[7]: %timeit np.einsum('ij,ij->i', foo, bar)
2.28 µs ± 89.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
I've added a comment though to make it a bit clearer.

benjaminhwilliams:
> humbly suggest moving the .dxf and .svg files into the dials.github.io repository (same as other images)

Done.

benjaminhwilliams:
> I note that this code makes heavy use of numpy/scipy routines. I don't know whether this is something we want to encourage/discourage, particularly where there are alternative ways to accomplish the same using flex arrays. I appreciate that you probably already _know_ how to accomplish certain things with numpy arrays, whereas you may not know how to do the same with flex arrays.

I leave this to future discussion.  I appreciate the desire to keep things uniform, and I know @ndevenish has opinions about this, but I reckon that's another problem for another time.

benjaminhwilliams:
It looks like the only failures in the Travis builds are Python 3 build tests that seem to be completely unaffected by this PR.  Otherwise I'm ready to merge...

Anthchirp:
Some sort of smtbx build failure, no clear idea why.
Travis on long-running branches can be a problem as it doesn't necessarily have the correct cctbx repository checkout for the time you branched off master.

If you rebased your branch on master this will likely be fixed.

PR #1105, opened 2020-01-23
Make merging statistics robust to statistics errors from too few reflections

benjaminhwilliams:
Anywhere where we use `iotbx.merging_statistics.dataset_statistics`, we are at risk of encountering errors such as `iotbx.merging_statistics.StatisticsError` when there are very few observations.  This is common in small molecule work and work-arounds to this have already been necessary in xia2 (see https://github.com/xia2/xia2/commit/974960cb0f9cccc700e369a349d13df7d0c40ee2) since the module in question was never designed with small molecule data in mind.

This PR introduces a decorated version of `iotbx.merging_statistics.dataset_statistics`, which tries the following until the first success:
1. to calculate the merging statistics with the standard default behaviour of `iotbx.merging_statistics.dataset_statistics`,
2. to calculate the merging statistics with the `"counting_sorted"` binning method, which aims to have roughly equal numbers of reflections in each bin,
3. to retain the `"counting_sorted"` binning method and progressively reduce the number of bins until there are fewer than six bins at which point,
4. it re-raises the `StatisticsError`.

This is clearly not a true fix, just an ugly work-around.  Improvement suggestions welcome.

There are a few resorted imports too.

### DIFF
--- a/algorithms/scaling/scaling_library.py
+++ b/algorithms/scaling/scaling_library.py
@@ -15,18 +15,18 @@ import uuid
 import pkg_resources
 from copy import deepcopy
 
-import iotbx.merging_statistics
 from cctbx import miller, crystal, uctbx
 from dxtbx.model import Experiment
 from dials.array_family import flex
-from dials.util.options import OptionParser
-from dials.util import Sorry
 from dials.algorithms.scaling.Ih_table import IhTable
 from dials.algorithms.scaling.model.model import KBScalingModel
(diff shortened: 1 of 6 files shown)


benjaminhwilliams:
This PR is superseded by https://github.com/cctbx/cctbx_project/pull/435 and corresponding changes in https://github.com/dials/dials/ on the branch `counting_sorted`.

PR #1107, opened 2020-01-23
Simplify rmsd convergence tests

dagewa:
The list comprehension and `all()` were probably legacy from dials.refine, which has a 3D RMSD (X, Y and Z) and thus tests each dimension. These uses in scaling appear to have a 1D RMSD only, so this can be simplified, making the functions easier to read.


dagewa:
hmm, actually there's python 3.6 test errors here... I'll have a look

Issue #237, opened 2016-10-18
Feature request: would be cute to compute a DQE map of the detector to include in the output of dials.report integrated.pickle


graeme-winter:
May give a lot of insight, particularly to small molecule users.

dagewa:
Does this just mean adding a plot a bit like this one https://github.com/dials/dials/blob/89e7d1e36d6019c07045d4b44d658e288349069c/command_line/report.py#L798 but using the `QE` column of the integrated reflections? If so, that should be trivial to add

PR #1108, opened 2020-01-24
Add QE map plot

dagewa:
Add a plot of calculated quantum efficiency for #237

Looks like this:
newplot


graeme-winter:
Worked for me thank you - interesting, a two-theta offset small molecule run still looks isotopic - 

newplot (4)

In hindsight this should be expected 🙂 since the closest point on the detector is the centre (about) => happy with this. Review coming.

PR #1116, opened 2020-01-27
Use first 360 degrees of data for dials.symmetry. Fixes #1000

### LINKED ISSUE #1000: dials.symmetry: perform analysis on first 360° of data only
Passing in a data set consisting of 1,440° of data for the analysis is redundant - after the first turn there is no new information being added. Should be trivial as a selection on the `z` value of each reflection?

Could we perform calculations only on first turn and apply to all data?

jbeilstenedmands:
Use first 360 degrees by default (if not explicitly using the exclude_images option).
Add tests for the helper function.

### DIFF
--- a/command_line/symmetry.py
+++ b/command_line/symmetry.py
@@ -225,13 +225,13 @@ def eliminate_sys_absent(experiments, reflections):
     return reflections
 
 
-def get_subset_for_symmetry(params, experiments, reflection_tables):
+def get_subset_for_symmetry(experiments, reflection_tables, exclude_images=None):
     """Select an image range for symmetry analysis, or just select
     the first 360 degrees of data."""
     refls_for_sym = []
-    if params.exclude_images:
+    if exclude_images:
         experiments = exclude_image_ranges_from_scans(
-            reflection_tables, experiments, params.exclude_images
+            reflection_tables, experiments, exclude_images
         )
         for refl, exp in zip(reflection_tables, experiments):
             sel = get_selection_for_valid_image_ranges(refl, exp)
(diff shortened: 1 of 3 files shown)


command_line/symmetry.py
```diff
@@ -224,6 +225,39 @@ def eliminate_sys_absent(experiments, reflections):
...
+    if params.exclude_images:
+        experiments = exclude_image_ranges_from_scans(
+            reflection_tables, experiments, params.exclude_images
+        )
+        for refl, exp in zip(reflection_tables, experiments):
+            sel = get_selection_for_valid_image_ranges(refl, exp)
+            refls_for_sym.append(refl.select(sel))
+    else:
+        # use first 360 degrees if <360 deg i.e. first measured data.
+        for expt, refl in zip(experiments, reflection_tables):
```
rjgildea: I think the following would achieve the same results in a lot fewer lines of code - at least with these changes the tests still pass:
```
        for expt, refl in zip(experiments, reflection_tables):
            scan_end = int(math.ceil(360/abs(expt.scan.get_oscillation()[1])))
            if scan_end < len(expt.scan):
                refl = refl.select(refl["xyzobs.px.value"].parts()[2] <= scan_end)
            refls_for_sym.append(refl)
```

command_line/symmetry.py
```diff
@@ -224,6 +225,39 @@ def eliminate_sys_absent(experiments, reflections):
     return reflections
 
 
+def get_subset_for_symmetry(params, experiments, reflection_tables):
```
rjgildea: Given this function only uses the `exclude_images` attribute of `params`, why not just directly pass in `exclude_images` as a paramer instead of `params`?

command_line/symmetry.py
jbeilstenedmands: Yes I think I agree, the px values always go from 0 upwards regardless of scan direction, so this always selects the first measured 360 degrees? Good spot.

Issue #1132, opened 2020-02-04
Table of dataset number vs image template in reports


dagewa:
Ahead of the **Scaling model plots** section, where the plots correspond to individual datasets, it would be useful to have a table of dataset number versus image template to make it easy to find and got back to integration results for particular datasets. It might also need crystal number for multiple lattice datasets.

dagewa:
As pointed out by @rjgildea this would apply anywhere that does plots for multiple datasets, not just scaling.html

PR #1133, opened 2020-02-04
Print image template

dagewa:
Make it easier to match experiments to imagesets in a multi-imageset
context (such as after cosym/scaling). Related to #1132

### DIFF
--- a/command_line/show.py
+++ b/command_line/show.py
@@ -258,6 +258,9 @@ def show_experiments(experiments, show_scan_varying=False):
             text.append("Format class: %s" % format_class.__name__)
         if expt.identifier != "":
             text.append("Experiment identifier: %s" % expt.identifier)
+        template = expt.imageset.get_template()
+        if template:
+            text.append("Image template: %s" % template)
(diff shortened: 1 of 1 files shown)


command_line/show.py
```diff
@@ -258,6 +258,9 @@ def show_experiments(experiments, show_scan_varying=False):
             text.append("Format class: %s" % format_class.__name__)
         if expt.identifier != "":
             text.append("Experiment identifier: %s" % expt.identifier)
+        template = expt.imageset.get_template()
```
rjgildea: I may be wrong, but ISTR that `get_template()` only exisits for `ImageSequence`, not `ImageSet`?

dagewa: Yes, it seems you are right

graeme-winter: Even so the addition would be useful, could `try` / `except` or `hasattr` this one?

dagewa: Yep, done in 5ba4f498d8 :+1:

Issue #1128, opened 2020-02-03
Integrating a single image (numbered 0?) fails 


graeme-winter:
```
dials.import th_8_2_0000.cbf ; dials.find_spots imported.expt ; dials.index imported.expt strong.refl ; dials.integrate indexed*
```

=>

```
********************************************************************************
Warning: 9 reference spots were not matched to predictions
********************************************************************************

Using 388 / 391 reflections for sigma calculation
Calculating E.S.D Beam Divergence.
Calculating E.S.D Reflecting Range (mosaicity).
Traceback (most recent call last):
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/integrate.py", line 665, in <module>
    script.run()
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/integrate.py", line 334, in run
    experiments = ProfileModelFactory.create(params, experiments, reference)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/factory.py", line 56, in create
    expr.scan,
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/gaussian_rs/model.py", line 228, in create
    params, reflections, crystal, beam, detector, goniometer, scan, profile
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/gaussian_rs/model.py", line 397, in create_from_reflections
    centroid_definition=params.gaussian_rs.centroid_definition,
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/gaussian_rs/calculator.py", line 579, in __init__
    algorithm=algorithm,
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/gaussian_rs/calculator.py", line 491, in __init__
    crystal, beam, detector, goniometer, scan, reflections
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/profile_model/gaussian_rs/calculator.py", line 330, in __init__
    "Something went wrong. Zero pixels selected for estimation of profile parameters."
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: Something went wrong. Zero pixels selected for estimation of profile parameters.
```

graeme-winter:
Yes, it does not work when image is numbered 0 🙄 - the same image numbered 1 works fine.

graeme-winter:
Turns out the mask values are all invalid? 🤔

PR #1121, opened 2020-01-29
Fix shoebox memory usage check

benjaminhwilliams:
Sometimes `dials.algorithms.integration.processor._average_shoebox_size` can be called on a reflection table that has no shoeboxes.  In fact, this is the usual case when `dials.algorithms.integration.processor.Manager.compute_processors` is called during a `dials.integrate` process because the predicted reflections have bboxes but don't yet have shoeboxes.

When the multiplicity is so high that the required shoebox memory exceeds the memory limit, `dials.integrate` should raise a nice `RuntimeError` with a meaningful message.  This code path was broken and is fixed by this PR.

fixes #1111

Co-Authored-By: Markus Gerstel <markus.gerstel@diamond.ac.uk>

PR #1115, opened 2020-01-27
Stream reflectiontable msgpack directly to file

Anthchirp:
The previous implementation held the reflection table up to 3 times in
memory during the output phase. This implementation passes the python
filehandle into C++ space and then writes the msgpack output directly
to the file. As a result memory consumption is significantly reduced.

Fixes #1112

PR #1136, opened 2020-02-05
DIALS 2.1.3

Anthchirp:
* Make resolution estimation more stable in presence of ice and powder rings
* Fix spot finding and integration of files with index 0 (#1128, cctbx/dxtbx#133)
* Reduce memory usage when writing .refl files (#1115)
* `dials.integrate`: Fix broken memory check in cases of high multiplicity (#1121)
* `dials.symmetry`: Prevent failures when dealing with small numbers of reflections (#1130, cctbx/cctbx_project#435)


benjaminhwilliams:
It looks like the tests failing in these checks are just because we need to merge dials/cctbx#45.  I can't see any other failures there.

Anthchirp:
Yes, Jenkins build is happy, which is what counts

PR #1144, opened 2020-02-06
Ensure restraints are not considered for fixed parameterisations

dagewa:
Otherwise the gradients check fails.


biochem-fan:
Thanks! I confirmed this fixes my test case (I tested both `auto_reduction.action=fix` and `remove`).

PR #1145, opened 2020-02-06
Update search_beam_centre to use experiments

### LINKED ISSUE #1126: dials.search_beam_position uses a single detector and beam model
Even when there are multiple imagesets and these have different detectors and beams

### LINKED ISSUE #1127: dials.search_beam_position macrocycles not working?
Looks very much the same each time:
```
Starting macro cycle 1
Running DPS using 373 reflections
Found 5 solutions with max unit cell 20.00 Angstroms.
Old beam centre: 28.67, 28.67 mm (1024.0, 1024.0 px)
New beam centre: 28.96, 28.45 mm (1034.2, 1016.1 px)
Shift: -0.29, 0.22 mm (-10.2, 7.9 px)

Starting macro cycle 2
Running DPS using 373 reflections
Found 5 solutions with max unit cell 20.00 Angstroms.
Old beam centre: 28.67, 28.67 mm (1024.0, 1024.0 px)
New beam centre: 28.96, 28.45 mm (1034.2, 1016.1 px)
Shift: -0.29, 0.22 mm (-10.2, 7.9 px)

Starting macro cycle 3
Running DPS using 373 reflections
Found 5 solutions with max unit cell 20.00 Angstroms.
Old beam centre: 28.67, 28.67 mm (1024.0, 1024.0 px)
New beam centre: 28.96, 28.45 mm (1034.2, 1016.1 px)
Shift: -0.29, 0.22 mm (-10.2, 7.9 px)
```

rjgildea:
- Replace imagesets with experiments. Since imagesets aren't relevant at to search_beam_position and mainly a legacy of its pre-experiment datablock-past, use experiments to pass around/access the experimental models directly.
- Extend test_search_small_molecule to check that the same detector (i.e. same fast/slow axes) is returned for each experiment. Previously all experiments were output with the same detector model, which was a bug.
- Use concurrent.futures instead of libtbx.easy_mp
- In tests call run method directly instead of via procrunner
- Fix n_macro_cycles option: ensure that the updated geometry is used for subsequent macro-cycles. Add test to assert that the shift in the final cycle is small.
- Move hardcoded dps phil params to lowest level

Fixes #1126, fixes #1127

### DIFF
--- a/command_line/search_beam_position.py
+++ b/command_line/search_beam_position.py
@@ -435,7 +435,7 @@ def discover_better_experimental_model(
         amax_list = []
         for future in concurrent.futures.as_completed(futures):
             result = future.result()
-            if result["solutions"]:
+            if result.get("solutions"):
                 solution_lists.append(result["solutions"])
                 amax_list.append(result["amax"])
 
--- a/command_line/search_beam_position.py
+++ b/command_line/search_beam_position.py
@@ -3,14 +3,15 @@ from __future__ import absolute_import, division, print_function
 # LIBTBX_PRE_DISPATCHER_INCLUDE_SH export PHENIX_GUI_ENVIRONMENT=1
 
 import cmath
+import concurrent.futures
 import copy
 import logging
 import math
 import random
 import sys
 
+import libtbx.introspection
 from libtbx import adopt_init_args
-from libtbx import easy_mp
 from libtbx.test_utils import approx_equal
 from libtbx.utils import plural_s
 from scitbx import matrix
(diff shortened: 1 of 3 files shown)


command_line/search_beam_position.py
```diff
@@ -397,135 +370,111 @@ def run_dps(args):
...
-    amax_list = [r["amax"] for r in results if r]
+    with concurrent.futures.ProcessPoolExecutor(max_workers=nproc) as pool:
+        futures = []
+        for expt, refl in zip(experiments, refl_lists):
+            futures.append(pool.submit(run_dps, expt, refl, max_cell))
+        solution_lists = []
+        amax_list = []
+        for future in concurrent.futures.as_completed(futures):
+            result = future.result()
+            if result["solutions"]:
```
dagewa: I'm finding it is crashing here with
```Traceback (most recent call last):
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/search_beam_position.py", line 553, in <module>
    run(sys.argv[1:])
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/search_beam_position.py", line 544, in run
    plot_search_scope=params.plot_search_scope,
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/search_beam_position.py", line 438, in discover_better_experimental_model
    if result["solutions"]:
KeyError: 'solutions
```

rjgildea:
Do the tests pass? Does it only crash on a specific example? All the tests pass for me on both Python 2 and Python 3.

PR #1148, opened 2020-02-07
Switch stills indexer based on imageset type

### LINKED ISSUE #1079: dials.index: on sequences use stills.indexer=sequences
At the moment when I am indexing `ImageSequences` of stills I use e.g.:

```
dials.index imported.expt strong.refl \
    max_lattices=3 unit_cell=47,71,102,90,90,90 \
    space_group=P222 \
    stills.indexer=sequences joint_indexing=0 \
    indexing.method=fft1d detector.fix=all beam.fix=all \
    outlier.algorithm=tukey
```

`stills.indexer=sequences` here should be `Auto` based on input of `ImageSet` vs. `ImageSequence` rather than `ImageSequence.is_still()`

See also https://github.com/dials/dials/issues/1030

graeme-winter:
If the input experiment has image set which is an instance of ImageSequence, then use sequences indexer; also move check for inconsistent input to end (i.e. reduce duplication)

Fixes #1079

Modest refactor; tests pass.

### DIFF
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -19,7 +19,7 @@ from dials.algorithms.indexing.compare_orientation_matrices import (
 from dials.algorithms.indexing.symmetry import SymmetryHandler
 from dials.algorithms.indexing.max_cell import find_max_cell
 from dials.algorithms.refinement import DialsRefineConfigError, DialsRefineRuntimeError
-from dxtbx.model import ExperimentList
+from dxtbx.model import ExperimentList, ImageSequence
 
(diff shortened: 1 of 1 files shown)


phyy-nx:
Makes me a bit nervous, especially considering the email thread on dials_support about ImageSequence.  I think scan.is_still is a better check, just for safety, but I can be talked out of this position.

graeme-winter:
Understand nerves - however if you look at the latest on that thread you may see that we have a real need for this 😉 since our use case really is rather different to the stills process behaviour. 

I suspect that this would be a good topic for an actual conversation, if that would be welcome?

Issue #1153, opened 2020-02-14
dials.spot_counts_per_image does weird things with images starting from 0


graeme-winter:
```
cs03r-sc-serv-16 imp_1000 :) $ dials.import ../gambit3911_00*cbf
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 2.dev.1378-gb6bd8a464
The following parameters have been modified:

input {
  experiments = <image files>
}

--------------------------------------------------------------------------------
  format: <class 'dxtbx.format.FormatCBFMiniPilatusDLS6MSN119.FormatCBFMiniPilatusDLS6MSN119'>
  num images: 1000
  sequences:
    still:    1
    sweep:    0
  num stills: 0
--------------------------------------------------------------------------------
Writing experiments to imported.expt
```

=> 

```
cs03r-sc-serv-16 imp_1000 :) $ dials.spot_counts_per_image imported.expt strong.refl 
```

=> 

```
---8<----
|     989 |        0 |               0 |                 0 |
|     990 |        0 |               0 |                 0 |
|     991 |        0 |               0 |                 0 |
|     992 |        0 |               0 |                 0 |
|     993 |        0 |               0 |                 0 |
|     994 |        0 |               0 |                 0 |
|     995 |        0 |               0 |                 0 |
|     996 |        0 |               0 |                 0 |
|     997 |        0 |               0 |                 0 |
|     998 |        0 |               0 |                 0 |
|     999 |        0 |               0 |                 0 |
|    1000 |        0 |               0 |                 0 |
|    1001 |       10 |              10 |             15230 |
|    1002 |        5 |               5 |             16506 |
|    1003 |       12 |              12 |             22998 |
|    1004 |        7 |               7 |             20789 |
|    1005 |        9 |               9 |              2945 |
|    1006 |        7 |               7 |             22698 |
|    1007 |        7 |               6 |             22860 |
|    1008 |        9 |               9 |              2059 |
|    1009 |        6 |               6 |             22511 |
---8<---
```

i.e. images up to 1000 are all 0's then it starts working...﻿

graeme-winter:
Oh god, it's worse than that
```
cs03r-sc-serv-16 dials-1153 :) $ dials.import gambit3911_0000*
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 2.dev.1378-gb6bd8a464
The following parameters have been modified:

input {
[...]

PR #1154, opened 2020-02-14
Fix counting output

graeme-winter:
For #1153

In the ascii-art graph need to recognise weirdness if we are counting from 0 - the output indices should be shifted (am implicitly assuming we are at worst having image 0 not image -1 ... etc.)

For the per-image analysis need to pass around the image numbers as well as
there it had been assumed that we would be running from 1... always


graeme-winter:
This is not yet ready for a review - there are many worms in this can 😕

PR #1155, opened 2020-02-16
Improve performance of image viewer tile generation.

phyy-nx:
Use of the locals() function adds an apparent overhead, at
least when tested under Python 3.6. Removing it cuts 0.7 seconds
from each "Next" image viewed in a Jungfrau16M *.h5 dataset.

(from @nksauter)


ndevenish:
Wow.

PR #1157, opened 2020-02-17
Temporarily override maximum trusted value in spot finding

graeme-winter:
Added phil parameter maximum_trusted_value. Temporarily assign this for spot finding, reassign original value at end for output.

For xia2/xia2#231

Set expert level 2 as should be used very carefully

Verified that this resets output value and includes overloaded spots in output.


command_line/find_spots.py
```diff
@@ -126,10 +132,24 @@ def run(self, args=None):
...
                 experiments
             )  # add identifier e.g. if coming straight from images
+
         if len(experiments) == 0:
             self.parser.print_help()
             return
 
+        # If maximum_trusted_value assigned, use this temporarily for the
+        # spot finding
+        if params.maximum_trusted_value > 0:
```
benjaminhwilliams:
```suggestion
        if params.maximum_trusted_value:
```

command_line/find_spots.py
```diff
@@ -177,6 +197,14 @@ def run(self, args=None):
...
         # Save the experiments
         if params.output.experiments:
+
+            # Reset the trusted ranges
+            if params.maximum_trusted_value > 0:
+                for detector in experiments.detectors():
+                    for panel in detector:
+                        trusted = input_trusted_ranges[id(panel)] = trusted
+                        panel.set_trusted_range(trusted)
+
```
benjaminhwilliams: I think you've left a bit of copypasta in (`trusted = input_trusted_ranges[id(panel)] = trusted`).  Also, I think we probably want to reset this regardless of the value of `params.output.experiments` in case we want to do anything afterwards with the experiment list before the process finishes (for example it might have a bearing on the later `params.per_image_statistics` block.

```suggestion

        # Reset the trusted ranges
        if params.maximum_trusted_value:
            for detector in experiments.detectors():
                for panel in detector:
                    trusted = input_trusted_ranges[id(panel)]
                    panel.set_trusted_range(trusted)

        # Save the experiments
        if params.output.experiments:

```

command_line/find_spots.py
```diff
@@ -69,6 +69,11 @@
       .help = "The log filename"
   }
 
+  maximum_trusted_value = 0
+    .type = float
+    .help = "Override maximum trusted value for spot finding only"
+    .expert_level = 2
+
```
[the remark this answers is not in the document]
graeme-winter: Good suggestion, implemented

command_line/find_spots.py
graeme-winter: Good suggestion, implemented

command_line/find_spots.py
graeme-winter: Good suggestion, implemented

command_line/find_spots.py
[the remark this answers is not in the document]
rjgildea: As far as I understand the `.optional` attribute is irrelevant here and has no effect - it only affects `choice` types and when scopes can be multiple (and when it _is_ relevant, the default is `.optional = True` anyway). I would suggest it is better not to add `.optional` here, as it might confuse future developers into thinking that it is required or having it somehow affects the behaviour.

command_line/find_spots.py
benjaminhwilliams: I note that you removed the copypasta but didn't implement the logic change.  Any reason for that?

benjaminhwilliams: Nor did you remove the second redundant `> 0` as suggested here.

command_line/find_spots.py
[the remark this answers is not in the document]
benjaminhwilliams: Ah, well I count myself among the confused then.

PR #1158, opened 2020-02-17
Clean up dials.array_family.flex

Anthchirp:
* Explicitly import symbols from dials flex_ext instead of importing/exporting everything and the kitchen sink.
* Avoids [unsupported use of `locals()`](https://docs.python.org/3/library/functions.html#locals)
* Removes unused symbols
  * `dials.array_family.flex.BinIndexer`²
  * `dials.array_family.flex.Reflection`²
  * `dials.array_family.flex.ShoeboxExtractor`
  * `dials.array_family.flex.centroid`²
  * `dials.array_family.flex.experiment_id_map`²
  * `dials.array_family.flex.get_real_type`
  * `dials.array_family.flex.intensity`²
  * `dials.array_family.flex.prediction`¹
  * `dials.array_family.flex.unit_cell`²

The ones marked with ¹ appear to be unused in Python land, and could potentially be removed from C++ land altogether. Discuss.

Edit: The ones now marked with ² may _appear_ to be unused, but are in fact used.


Anthchirp:
discussed: tidy up. If we need to put it back then we know where to find it

Anthchirp:
Rebuilt dials from scratch and tests still pass.

Anthchirp:
Think this is ready to go.

PR #1160, opened 2020-02-18
Stabilize search_beam_position

### LINKED ISSUE #1152: unstable test test_search_beam_position.test_search_multiple
Sometimes it passes, sometimes it:
```
run_in_tmpdir = local('/scratch/jenkins_agent/workspace/dials_tests/tests/pytest/popen-gw2/test_search_multiple0')
dials_regression = '/dls/science/groups/scisoft/DIALS/repositories/git-reference/dials_regression'

    def test_search_multiple(run_in_tmpdir, dials_regression):
        """Perform a beam-centre search and check that the output is sane.
    
        Do the following:
        1. Run dials.search_beam_centre on two datablocks and two pickled
        reflection tables, as output by dials.find_spots;
          a) Check that the program exits correctly;
          b) Check that it produces the expected output datablock.
        2. Check that the beam centre search has resulted in the expected shift
        in detector origin.
        """
    
        data_dir = os.path.join(dials_regression, "indexing_test_data", "trypsin")
        pickle_path1 = os.path.join(data_dir, "strong_P1_X6_1_0-1.pickle")
        pickle_path2 = os.path.join(data_dir, "strong_P1_X6_2_0-1.pickle")
        experiments_path1 = os.path.join(data_dir, "datablock_P1_X6_1.json")
        experiments_path2 = os.path.join(data_dir, "datablock_P1_X6_2.json")
    
        args = [experiments_path1, experiments_path2, pickle_path1, pickle_path2]
        search_beam_position.run(args)
        assert os.path.exists("optimised.expt")
    
        experiments = load.experiment_list(experiments_path1, check_format=False)
        optimised_experiments = load.experiment_list("optimised.expt", check_format=False)
        detector_1 = experiments[0].detector
        detector_2 = optimised_experiments[0].detector
        shift = scitbx.matrix.col(detector_1[0].get_origin()) - scitbx.matrix.col(
            detector_2[0].get_origin()
        )
>       assert shift.elems == pytest.approx((0.037, 0.061, 0.0), abs=1e-1)
E       assert (3.7479447822...48082165, 0.0) == approx((0.037 ...0 +- 1.0e-01))
E         -(3.7479447822514658, 2.964508548082165, 0.0)
E         +approx((0.037 +- 1.0e-01, 0.061 +- 1.0e-01, 0.0 +- 1.0e-01))

../../../../build_dials/modules/dials/test/command_line/test_search_beam_position.py:86: AssertionError
```

Anthchirp:
This doesn't make the algorithm _stable_, but it does make it _deterministic_.
The test failures in #1152 indicate that the algorithm is in fact very much not stable at all.

Can someone with more simplex knowledge have a look and see if there is a more appropriate solution for this?

Otherwise - this does fix the tests and fixes #1152, so it's still an improvement over the current situation.

### DIFF
--- a/command_line/search_beam_position.py
+++ b/command_line/search_beam_position.py
@@ -421,7 +421,8 @@ def discover_better_experimental_model(
             futures.append(pool.submit(run_dps, expt, refl, max_cell))
         solution_lists = []
         amax_list = []
-        for future in concurrent.futures.as_completed(futures):
+        concurrent.futures.wait(futures, return_when=concurrent.futures.ALL_COMPLETED)
+        for future in futures:
             result = future.result()
             if result.get("solutions"):
                 solution_lists.append(result["solutions"])
--- a/command_line/search_beam_position.py
+++ b/command_line/search_beam_position.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import cmath
 import concurrent.futures
 import copy
+import itertools
 import logging
 import math
 import random
@@ -354,7 +355,7 @@ def run_dps(experiment, spots_mm, max_cell):
     # There must be at least 3 solutions to make a set, otherwise return empty result
     if len(solutions) < 3:
         return {}
-    return dict(solutions=flex.vec3_double([s.dvec for s in solutions]), amax=DPS.amax)
+    return dict(solutions=flex.vec3_double(s.dvec for s in solutions), amax=DPS.amax)
 
 
 def discover_better_experimental_model(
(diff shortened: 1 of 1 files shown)


command_line/search_beam_position.py
```diff
@@ -153,43 +154,43 @@ def igrid(x):
...
-                evaluator=selfOO,
+    # Do a simplex minimization
+    class simplex_minimizer(object):
+        def __init__(self, wide_search_offset):
+            self.n = 2
+            self.wide_search_offset = wide_search_offset
+            self.starting_simplex = [  # guaranteed to be random
+                flex.double([0.11505456638977896, 0.6090665392794814]),
+                flex.double([0.13339096418598828, 0.24058961996534878]),
+                flex.double([0.3271390558111398, 0.8591374909485977]),
```
dagewa: I'm trying to understand why this change helps, if `flex.random_double` was indeed seeded correctly then how were we getting differences here on different test instances?

Anthchirp: tested this, and this was a red herring. I've put the random numbers back.

rjgildea:
You can just (incorrectly) override the beam centre on import before running the beam centre search, so no need for the dataset to have the beam centre incorrectly recorded in the image headers.

Anthchirp:
merging pull request, as this doesn't introduce any new issues.
opened #1152 for issue discussion

PR #1162, opened 2020-02-19
dials.array_family cleanup by namespace gatherer

Anthchirp:
Following a suggestion by @ndevenish:
* move `flex.py` to `flex_ext.py`
* make `flex.py` only import symbols from different places and not contain any actual code

Made into a pull request for the pull request mainly to see if we get LGTM feedback


ndevenish:
👍 I find this much easier to understand/read what is going on, with no messy/confusing namespace manipulations to avoid overwriting builtins; no imports at end of ~2kline file.

Anthchirp:
for the record: I've tested various warning suppression annotations, and for namespace gatherers the most appropriate annotation is
```
from X import (  # noqa: F401; lgtm
    ...
)
```
image

PR #1165, opened 2020-02-19
Add dials.dose_analysis for running pychef in dials

jbeilstenedmands:
This PR introduces a new command line tool, `dials.dose_analysis`, which runs the CHEF dose statistic calculations on an mtz file or on scaled .expt/.refl files. Produces a `dials.dose_analysis.html` report with the dose plots (same as in xia2.report).

I wasn't sure about the name, but thought it more clear than dials.chef, and this leaves it open to be expanded in future for additional analysis. Other name suggestions welcome.

For multi-sweep dials data files, each sweep is assumed to be a measurement of a new crystal and assigned a starting dose of 0. If they are measurements of the same crystal, this can be set using the option shared_crystal=True.
I believe that the default option (all start at 0) is more useful, as it gives an indication of the average level of radiation damage for images measured at that dose across all crystals.

e.g. Multicrystal TehA dataset, scan ranges varying from 20 to 50 images. 
mtz input
newplot (3)

scaled.expt, scaled.refl input
newplot (4)

scaled.expt, scaled.refl input, shared_crystal=True
newplot (5)

### DIFF
--- a/command_line/dose_analysis.py
+++ b/command_line/dose_analysis.py
@@ -22,6 +22,14 @@ number is used as a proxy for dose (i.e. dose is accumulated across sweeps),
 which may not be suitable for multi-sweep mtz files, unless all sweeps are
 measured on the same crystal.
 
+Example usage:
+
+dials.dose_analysis scaled.expt scaled.refl
+
+dials.dose_analysis mtzfile=scaled.mtz
+
+dials.dose_analysis scaled.expt scaled.refl shared_crystal=True
+
 """
 from __future__ import absolute_import, division, print_function
 import json
--- a/command_line/dose_analysis.py
+++ b/command_line/dose_analysis.py
@@ -58,6 +58,26 @@ logger = logging.getLogger("dials.command_line.dose_analysis")
 
 phil_scope = phil.parse(
     """\
+dose {
+    experiments
+    {
+        dose_per_image = 1
+        .type = ints(value_min=0)
+        .help = "The 'dose' accumulated per image. If more than one value is"
+                "given, this indicates the dose per image for each experiment,"
+                "and hence must match the number of experiments."
+        starting_doses = None
+        .type = ints(value_min=0)
+        .help = "The dose values at the start of each sweep. Must match the"
+                "number of experiments. If none given, it is assumed that each"
+                "sweep starts with zero accumulated dose."
+        shared_crystal = False
+        .type = bool
+        .help = "Option to indicate that all sweeps correspond to measurements"
+                "on the same crystal. Therefore the starting doses are"
+                "automatically adjusted to account for previously accumulated dose."
+    }
+}
 output {
     log = "dials.dose_analysis.log"
         .type = str
--- a/command_line/dose_analysis.py
+++ b/command_line/dose_analysis.py
@@ -5,14 +5,22 @@ The program dials.dose_analysis calculates dose dependent data quality statistic
 The algorithms and statistics are described in "How best to use photons",
 G. Winter et al. 2019 https://doi.org/10.1107/S2059798319003528
 
-The program makes a noteable assumption which may be unsuitable for particular
-experiments.
-For running on dials datafiles (expt, refl), the image number is
-taken as a proxy for dose (calculated individually on each sweep for multi-
-sweep datasets).
-For mtz file input, if there is no 'dose' column, then the batch number is
-used as a proxy for dose, which may not be suitable for multi-sweep mtz files,
-unless all sweeps are measured on the same crystal.
+Input data can be provided in mtz format or scaled .expt and .refl files.
+
+For dials datafiles, by default, the image number is taken as a proxy for dose.
+For multi-sweep datasets, each experiment is assumed to be a measurement of a
+new crystal, and is assigned a starting dose of zero. If the multi-sweep
+dataset is repeated measurements of one crystal, then the option
+shared_crystal=True can be given, which will accumulate dose across the sweeps.
+Alternatively, a list of starting_doses can be given. The default assumption is
+that each image corresponds to an equivalent dose. This can be overwritten with
+the dose_per_image option, where a list of values can be provided to set a
+different dose per image for each sweep.
+
+For mtz file input, if there is no 'dose' column, then by default the batch
+number is used as a proxy for dose (i.e. dose is accumulated across sweeps),
+which may not be suitable for multi-sweep mtz files, unless all sweeps are
+measured on the same crystal.
 
 """
 from __future__ import absolute_import, division, print_function
--- a/command_line/dose_analysis.py
+++ b/command_line/dose_analysis.py
@@ -196,7 +196,7 @@ class PychefRunner(object):
             experiments, start_doses, doses_per_image
         ):
             refls = reflection_table.select(expt)
-            imgno = refls["xyzobs.px.value"].parts()[2]
+            imgno = flex.ceil(refls["xyzobs.px.value"].parts()[2])
             dose = (imgno * dose_per_img) + starting_dose
             doses.extend(dose)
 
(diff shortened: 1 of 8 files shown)


test/command_line/test_dose_analysis.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_dose_analysis_dials_data(dials_data, tmpdir):
+    """Test dials.dose_analysis on scaled data."""
+    location = dials_data("l_cysteine_4_sweeps_scaled")
+    refls = location.join("scaled_20_25.refl").strpath
+    expts = location.join("scaled_20_25.expt").strpath
+
+    command = ["dials.dose_analysis", refls, expts, "min_completeness=0.4", "-v"]
+    result = procrunner.run(command, working_directory=tmpdir)
```
rjgildea: Can the tests not call the `dials.command_line.dose_analysis.run()` function directly rather than via `procrunner`?

jbeilstenedmands: I prefer to use procrunner so that we're testing the program as a user would on the command line, and to avoid potential errors in setting up the correct input for the tests.

command_line/dose_analysis.py
```diff
@@ -0,0 +1,306 @@
...
+    parser = OptionParser(
+        usage=usage,
+        phil=phil,
+        epilog=__doc__,
+        read_experiments=True,
+        read_reflections=True,
+        check_format=False,
+    )
+
+    params, _ = parser.parse_args(args=args, show_diff_phil=False)
```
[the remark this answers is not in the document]
rjgildea: I.e.:
```
    params, options, args = parser.parse_args(
        args=args, show_diff_phil=False, return_unhandled=True
    )

    if experiments and reflections:
        ...
    
    elif args and os.path.isfile(args[0]):
        mtz_object = mtz.object(file_name=args[0])
        ...
```

command_line/dose_analysis.py
```diff
@@ -0,0 +1,306 @@
...
+                raise KeyError("Input data must be scaled.")
+            script = PychefRunner.from_dials_datafiles(
+                params, experiments, reflections[0],
+            )
+        elif params.input.mtzfile:
+            try:
+                mtz_object = mtz.object(file_name=params.input.mtzfile)
+            except RuntimeError as e:
+                # If an error is encountered trying to read the mtzfile
+                raise ValueError(e)
```
rjgildea: Is there any specfic reason to re-raise as a different type of exception?

pychef/__init__.py
```diff
@@ -33,6 +32,25 @@
     dose_step = None
       .type = float(value_min=0)
   }
+  experiments
```
rjgildea: These parameters are only applicable to the command line program, and if experiments/reflections are given as input, not for mtz files. Perhaps these parameters should live in the command line phil scope? As suggested below these parameters could be passed explicitly to the `interpret_images_to_doses_options` function.

test/command_line/test_dose_analysis.py
```diff
@@ -0,0 +1,111 @@
...
+def test_dose_analysis_input_handling(dials_data, tmpdir):
+    """Test that errors are handled if more than one refl file, no refl/expt
+    file or unscaled data."""
+    location = dials_data("l_cysteine_4_sweeps_scaled")
+    refls = location.join("scaled_20_25.refl").strpath
+    expts = location.join("scaled_20_25.expt").strpath
+
+    # Too many refl files
+    command = ["dials.dose_analysis", refls, expts, refls]
+    result = procrunner.run(command, working_directory=tmpdir)
```
rjgildea: Calling `dials.command_line.dose_analysis.run()` directly would mean you could make use of `with pytest.raises(SystemExit)` here instead, with the added bonus that you can test that it throws a sys.exit rather than any old error/traceback:
```
from dials.command_line.dose_analysis import run
...
with pytest.raises(SystemExit):
    run([refls, expts, refls])
```

test/command_line/test_dose_analysis.py
rjgildea: As discussed elsewhere, calling `dials.command_line.dose_analysis.run()` directly would still exercise all of the command line parsing code.

command_line/dose_analysis.py
```diff
@@ -0,0 +1,306 @@
...
+
+        if dose is None:
+            dose = batches_to_dose(batches.data(), params.dose)
+        else:
+            dose = dose.data()
+
+        return cls(intensities, dose, params)
+
+    @classmethod
+    def from_dials_datafiles(cls, params, experiments, reflection_table):
```
rjgildea: I would probably prefer "data_files" rather than "datafiles" here.
```suggestion
    def from_dials_data_files(cls, params, experiments, reflection_table):
```

command_line/dose_analysis.py
```diff
@@ -0,0 +1,306 @@
...
+
+        for expt, starting_dose, dose_per_img in zip(
+            experiments, start_doses, doses_per_image
+        ):
+            refls = reflection_table.select(expt)
+            imgno = flex.ceil(refls["xyzobs.px.value"].parts()[2])
+            dose = (imgno * dose_per_img) + starting_dose
+            doses.extend(dose)
+
+        doses = doses.iround()
```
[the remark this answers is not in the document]
rjgildea: Alternatively if `dose_per_img` can only be integer, then `doses` can be a `flex.int` from the beginning.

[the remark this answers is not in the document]
jbeilstenedmands: Thanks for pointing this out, I was going to allow floats so that actual (calculated) doses could be given, but then found that pychef wants integer doses, so best to constrain to ints for now.

test/command_line/test_dose_analysis.py
jbeilstenedmands: I'm convinced on this one now, anecdotally test time went from about 17s to about 6s.

pychef/__init__.py
```diff
@@ -64,6 +82,45 @@
...
+            % len(experiments)
+        )
+    else:
+        doses_per_image = dpi
+
+    if params.dose.experiments.shared_crystal:
+        start_doses = [0]
+        accumulated_dose = 0
+        # adjust starting doses to account for a shared crystal.
+        for expt, dose_per_img in zip(experiments, doses_per_image):
```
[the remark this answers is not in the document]
jbeilstenedmands: I think best to leave the experiments matching the input order (of data and dose_per_image) command to avoid confusion. This can be manually set anyway by giving starting doses.

jbeilstenedmands:
Thanks @rjgildea , adopted the majority of suggestions, and I'm convinced now on the benefits of calling the run method directly, rather than going via procrunner. Any further thoughts from anyone on an improved name? If not I'll merge in a day or two.

Anthchirp:
as per previous discussion we use imperatives in command names, (`dials.integrate`, `dials.find_spots`, ...), and the general use case of pychef does not use dose in a physical sense at all. So my suggestion would be `dials.(pychef|$verb(_$noun)?)` with `verb in {analyze, plot}` and `noun in {damage}`.

graeme-winter:
+1 for `dials.analyze_damage`

rjgildea:
I think that `dials.damage_analysis` would read better than `dials.analy[s,z]e_damage`, and avoid the analyse vs. analyze arguments. I don't think we necessarily need to stick to "all-dials-commands-must-be-in-the-imperative" just for the sake of it - there are several prominent commands that don't follow this already, `dials.image_viewer` and `dials.reciprocal_lattice_viewer` primes example of this. Also programs like `dials.symmetry` and `dials.cosym` don't fit this "imperative" naming scheme.

dagewa:
I like consistency to a point, but hobgoblins etc. The users really don't care about our command naming schemes.

Anthchirp:
Discussion outcome: `dials.damage_analysis`

PR #1167, opened 2020-02-21
Replace **0.5 and **2 with flex.sqrt and flex.pow2

rjgildea:
Using flex.sqrt and flex.pow2 is faster:

```
$ cat tst.py 
import time
from scitbx.array_family import flex

a = flex.random_double(100000)

t0 = time.time()
for i in range(1000):
    a ** 0.5
t1 = time.time()

for i in range(1000):
    flex.sqrt(a)
t2 = time.time()

for i in range(1000):
    a ** 2
t3 = time.time()

for i in range(1000):
    flex.pow2(a)
t4 = time.time()

print(t1 - t0)
print(t2 - t1)
print(t3 - t2)
print(t4 - t3)

$ dials.python tst.py 
3.79648709297
0.110934972763
3.71216607094
0.0598487854004
```

Mostly touches scaling code, although also some changes elsewhere.

### DIFF
--- a/algorithms/integration/filtering.py
+++ b/algorithms/integration/filtering.py
@@ -196,7 +196,7 @@ class IceRingFilter:
         :return: True/False in powder ring
         """
         result = flex.bool(len(d), False)
-        d2 = 1.0 / d ** 2
+        d2 = 1.0 / flex.pow2(d)
         for ice_ring in self.ice_rings:
             result = result | (d2 >= ice_ring[0]) & (d2 <= ice_ring[1])
         return result
--- a/algorithms/integration/kapton_2019_correction.py
+++ b/algorithms/integration/kapton_2019_correction.py
@@ -622,7 +622,7 @@ class multi_kapton_correction(object):
                     # has already been corrected!
                 else:
                     refl_sele["intensity.sum.value"] *= k_corr
-                    refl_sele["intensity.sum.variance"] *= (k_corr) ** 2
+                    refl_sele["intensity.sum.variance"] *= flex.pow2(k_corr)
                 return refl_sele
 
             if len(refl_zero) > 0 and self.params.smart_sigmas:
--- a/algorithms/integration/kapton_correction.py
+++ b/algorithms/integration/kapton_correction.py
@@ -637,7 +637,7 @@ class multi_kapton_correction(object):
                     # has already been corrected!
                 else:
                     refl_sele["intensity.sum.value"] *= k_corr
-                    refl_sele["intensity.sum.variance"] *= (k_corr) ** 2
+                    refl_sele["intensity.sum.variance"] *= flex.pow2(k_corr)
                 return refl_sele
 
             if len(refl_zero) > 0 and self.params.smart_sigmas:
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -71,7 +71,7 @@ def prepare_merged_reflection_table(experiments, reflection_table, d_min=None):
     )
     merged_reflections = flex.reflection_table()
     merged_reflections["intensity"] = merged.data()
-    merged_reflections["variance"] = merged.sigmas() ** 2
+    merged_reflections["variance"] = flex.pow2(merged.sigmas())
     merged_reflections["miller_index"] = merged.indices()
     return merged_reflections
 
@@ -183,7 +183,7 @@ def merge_and_truncate(params, experiments, reflections):
     if params.assess_space_group:
         merged_reflections = flex.reflection_table()
         merged_reflections["intensity"] = merged.data()
-        merged_reflections["variance"] = merged.sigmas() ** 2
+        merged_reflections["variance"] = flex.pow2(merged.sigmas())
         merged_reflections["miller_index"] = merged.indices()
         logger.info("Running systematic absences check")
         run_systematic_absences_checks(experiments, merged_reflections)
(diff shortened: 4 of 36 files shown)


Anthchirp:
tests pass (modulo failing `dials.import` test for unrelated reasons)

graeme-winter:
This is deliberately incomplete? e.g.

```
./array_family/flex.py:        i_obs.set_sigmas(variances ** 0.5)
./array_family/flex.py:                        d.append((i, j, dx ** 2 + dy ** 2 + dz ** 2))
./array_family/flex.py:            (x1 - x2) ** 2 + (y1 - y2) ** 2 + (z1 - z2) ** 2
./array_family/flex.py:                        d.append((i, j, dx ** 2 + dy ** 2 + dz ** 2))
./array_family/flex.py:            (x1 - x2) ** 2 + (y1 - y2) ** 2 + (z1 - z2) ** 2
```

and 

```
./algorithms/profile_model/gaussian_rs/calculator.py:                kernel.append(math.exp(-((i - mid) ** 2) / (2 * sigma ** 2)))
```

I also wonder if `flex.pow(array, 3)` is (by analogy) faster than `array ** 3` - 


```
./command_line/report.py:            ds3_min = 1 / d_min ** 3
./command_line/report.py:            ds3_max = 1 / d_max ** 3
./command_line/report.py:            d_centre = 1 / ds3_centre ** (1 / 3)
```

graeme-winter:
Some idea of the wall-clock time differences would also be useful e.g. does this actually make a difference we can measure? For documentation not criticism.

Anthchirp:
cf. https://github.com/dials/dials/pull/1167#pullrequestreview-362919720 on your first point regarding array_family.flex

graeme-winter:
> cf. [#1167 (review)](https://github.com/dials/dials/pull/1167#pullrequestreview-362919720) on your first point regarding array_family.flex

👍 did not make the connection

graeme-winter:
> This is deliberately incomplete? e.g.
> 
> ```
> ./array_family/flex.py:        i_obs.set_sigmas(variances ** 0.5)
> ./array_family/flex.py:                        d.append((i, j, dx ** 2 + dy ** 2 + dz ** 2))
> ./array_family/flex.py:            (x1 - x2) ** 2 + (y1 - y2) ** 2 + (z1 - z2) ** 2
> ./array_family/flex.py:                        d.append((i, j, dx ** 2 + dy ** 2 + dz ** 2))
> ./array_family/flex.py:            (x1 - x2) ** 2 + (y1 - y2) ** 2 + (z1 - z2) ** 2
> ```
> 
> and
> 
> ```
> ./algorithms/profile_model/gaussian_rs/calculator.py:                kernel.append(math.exp(-((i - mid) ** 2) / (2 * sigma ** 2)))
> ```
> 
> I also wonder if `flex.pow(array, 3)` is (by analogy) faster than `array ** 3` -
> 
> ```
> ./command_line/report.py:            ds3_min = 1 / d_min ** 3
> ./command_line/report.py:            ds3_max = 1 / d_max ** 3
> ./command_line/report.py:            d_centre = 1 / ds3_centre ** (1 / 3)
> ```

Seems that there are quite a few remaining instances...? If we are aiming for consistency (though I don't know how many cases the input is a flex array) 

```
Grey-Area dials :) [flex_sqrt_pow2] $ grep -R "\*\* [0-9]" .|wc -l
     230
```

Anthchirp:
for (single) floating point numbers `x ** 2` is faster than `math.pow(x, 2)` according to
```
import timeit
print(timeit.timeit('x ** 0.5', setup='x = 123.45'))
print(timeit.timeit('math.sqrt(x)', setup='x = 123.45;import math'))
print(timeit.timeit('x ** 2', setup='x = 123.45'))
print(timeit.timeit('math.pow(x,2)', setup='x = 123.45;import math'))
```

`./algorithms/profile_model/gaussian_rs/calculator.py` is definitely not running on a flex array

rjgildea:
I only converted those where I could be sure the input was a `flex.double`. I'm pretty sure most, if not all, of those remaining are single values, where obviously `flex.sqrt` or `flex.pow2` would fail.

rjgildea:
@graeme-winter I'm pretty sure that `flex.pow(array, 3)` exactly equivalent to doing `array ** 3` (although I can't manage to follow the code to verify that this is the case). I suspect that `flex.sqrt` and `flex.pow2` are running the C++ code `std::sqrt` and `s * s` underneath, where as `flex.pow` or `array ** x` are calling `std::pow`. Using
```
t0 = time.time()
for i in range(1000):
    a ** 3
t1 = time.time()

for i in range(1000):
    flex.pow(a, 3)
t2 = time.time()

print(t1 - t0)
print(t2 - t1)
```
gives no significant difference:
```
1.55465221405
1.54827284813
```

rjgildea:
> Some idea of the wall-clock time differences would also be useful e.g. does this actually make a difference we can measure? For documentation not criticism.

Wallclock times given above in top comment.

graeme-winter:
> > Some idea of the wall-clock time differences would also be useful e.g. does this actually make a difference we can measure? For documentation not criticism.
> 
> Wallclock times given above in top comment.

I meant for the DIALS tasks 🙄 not just the `sqrt` etc. 

Thanks re: `n**3` etc.

rjgildea:
> I meant for the DIALS tasks 🙄 not just the `sqrt` etc.

Actual time savings for dials command line programs are likely to be small, and probably within the variability of time taken for running the same command. Something on the order of ~0.2s time saving for doing `flex.sqrt` on an array of 10 million values, in which case presumably other calculations would be even slower with that many reflections. But using `flex.sqrt` and `flex.pow2` is demonstrably quicker, and definitely not slower.

PR #1174, opened 2020-02-28
Support imagesets of len > 1 (e.g. grid scans) in dials.stills_process

phyy-nx:
`dials.stills_process` needs something atomic to split on to send to the ranks.  In the case of a grid scan of stills, `ExperimentListFactory` returns 1 experiment with an ImageSequence of length N.  This pull request changes that behavior so that in the `do_import` function of `dials.stills_process`, the single experiment is converted to a list of experiments of length N, each with an imageset of length 1.  Then dials.stills_process can split these off to the ranks.

Still fast because load_models=False on import.

### DIFF
--- a/command_line/stills_process.py
+++ b/command_line/stills_process.py
@@ -16,6 +16,7 @@ from dials.util import log
 from dials.array_family import flex
 from dxtbx.model.experiment_list import ExperimentListFactory
 from dxtbx.model.experiment_list import ExperimentList
+from dxtbx.model.experiment_list import Experiment
 from libtbx.utils import Abort, Sorry
 from collections import OrderedDict
 from libtbx.phil import parse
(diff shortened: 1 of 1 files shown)


command_line/stills_process.py
```diff
@@ -240,17 +241,27 @@ def do_import(filename, load_models=True):
...
+        for i in range(len(imageset)):
+            # Preserve original models if they were available (in the case of an image file
+            # they will not be, but in the case of a previously processed experiment list,
+            # then they may be available
+            expt = Experiment(
+                imageset=imageset[i : i + 1],
+                detector=experiment.detector,
+                beam=experiment.beam,
+                scan=experiment.scan,
+                goniometer=experiment.goniometer,
```
graeme-winter: Since you are expecting  `scan = None` and `goniometer = None` why not enforce that here? Thinking that merging https://github.com/dials/dials/pull/1148 will have an impact on this choice.

command_line/stills_process.py
```diff
@@ -240,17 +241,27 @@ def do_import(filename, load_models=True):
...
-        imageset.set_goniometer(None)
-        experiment.imageset = imageset
-        experiment.scan = None
-        experiment.goniometer = None
+        for i in range(len(imageset)):
+            # Preserve original models if they were available (in the case of an image file
+            # they will not be, but in the case of a previously processed experiment list,
+            # then they may be available
+            expt = Experiment(
+                imageset=imageset[i : i + 1],
```
graeme-winter: This will preserve whether the input is an `ImageSet` or `ImageSequence` so represents a behaviour change?

command_line/stills_process.py
phyy-nx: I was trying to go by the idea that not deleting scans/gonios was desired.  I tested this code with #1148 and the stills indexer and stills integrator were both picked, so I'm happy enough.

command_line/stills_process.py
phyy-nx: The `imageset_from_anyset` call converts an `ImageSequence` to an `ImageSet`, which (especially with #1148) will mean the data are treated as stills downstream.

graeme-winter:
👍 works for me - squash merging

Anthchirp:
Could I please have a 1 or 2 sentence summary what effect this PR has on a user?

Anthchirp:
suggestion:
> dials.stills_process now supports imagesets of length > 1 (e.g. grid scans)

sufficient?

phyy-nx:
Yup.

PR #1178, opened 2020-03-04
Account for March 2020 bugfix in rstbx hemisphere sampler

phyy-nx:
Details:
1. Add `march2020_bugfix = True` as parameter to fft1d indexer and search_beam_position.
2. For tests that fail now with the new behavior, set `march2020_bugfix = False` to restore the old behavior.  These are 1 test in indexing and 4 tests in search_beam_position.

An alternative to 2 is fixing the tests to account for this new behavior but I thought others could look first and opine.

### DIFF
--- a/test/command_line/test_search_beam_position.py
+++ b/test/command_line/test_search_beam_position.py
@@ -31,7 +31,7 @@ def test_search_i04_weak_data_image_range(mocker, run_in_tmpdir, dials_regressio
         "image_range=1,10",
         "image_range=251,260",
         "image_range=531,540",
-        "n_macro_cycles=2",
+        "n_macro_cycles=4",
     ]
     from rstbx.indexing_api import dps_extended
 
@@ -85,7 +85,7 @@ def test_search_multiple(run_in_tmpdir, dials_regression):
     shift = scitbx.matrix.col(detector_1[0].get_origin()) - scitbx.matrix.col(
         detector_2[0].get_origin()
     )
-    assert shift.elems == pytest.approx((0.037, 0.061, 0.0), abs=1e-1)
+    assert shift.elems == pytest.approx((-0.518, 0.192, 0.0), abs=1e-1)
 
 
 def test_index_after_search(dials_data, run_in_tmpdir):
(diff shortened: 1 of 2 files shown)


phyy-nx:
See cctbx/cctbx_project#444

dagewa:
Good job on spotting the bug. But what is the purpose of keeping the old, broken behaviour as an option?

phyy-nx:
Well, often I've had to reproduce old results exactly, either for data
provenance reasons or because we want to compare the old vs. the new.
Having a phil parameter that let's me do this has been very helpful.

graeme-winter:
Apropos test fixes: we should modify the tests to account for the new, bugless behaviour.

phyy-nx:
Force pushed a version that only adjusts the tests following new version of cctbx/cctbx_project#444.

Again, 5 tests modified.  If the original authors want to take a look, could be useful.  Thanks.

phyy-nx:
Any thoughts on this one?  Thanks :)

test/command_line/test_search_beam_position.py
```diff
@@ -31,7 +31,7 @@ def test_search_i04_weak_data_image_range(mocker, run_in_tmpdir, dials_regressio
         "image_range=1,10",
         "image_range=251,260",
         "image_range=531,540",
-        "n_macro_cycles=2",
+        "n_macro_cycles=4",
```
graeme-winter: Just for interest, why change in `n_macro_cycles`?

test/command_line/test_search_beam_position.py
```diff
@@ -146,9 +146,9 @@ def test_index_after_search(dials_data, run_in_tmpdir):
 
     # check we can actually index the resulting optimized experiments
     expected_unit_cell = uctbx.unit_cell(
-        (57.780, 57.800, 150.017, 89.991, 89.990, 90.007)
+        (59.663, 64.291, 155.061, 91.376, 89.964, 91.212)
     )
-    expected_rmsds = (0.06, 0.05, 0.001)
+    expected_rmsds = (0.3, 0.5, 0.005)
```
graeme-winter: Really such a large change in expected RMSD? 🤔

test/command_line/test_search_beam_position.py
phyy-nx: It appears for this weak dataset that the new indexing code takes a few more cycles to converge.

test/command_line/test_search_beam_position.py
phyy-nx: Yah, I was surprised too but I'm not sure if there's anything to do about it.

phyy-nx:
@graeme-winter ok to merge then?  Thanks!

graeme-winter:
@phyy-nx  nope, please go ahead - good to have the bug fixed 🙂

PR #1181, opened 2020-03-06
Respect valid_image_ranges in cosym and dials.symmetry

rjgildea:


command_line/cosym.py
```diff
@@ -90,6 +91,11 @@ def __init__(self, experiments, reflections, params=None):
             params = phil_scope.extract()
         self.params = params
 
+        self._reflections = []
+        for refl, expt in zip(reflections, experiments):
+            sel = get_selection_for_valid_image_ranges(refl, expt)
+            self._reflections.append(refl.select(sel))
+
         self._experiments, self._reflections = self._filter_min_reflections(
             experiments, reflections
```
jbeilstenedmands: Should the input to this function now be`self._reflections`?

PR #1182, opened 2020-03-07
Single panel special cases

### LINKED ISSUE #663: Remove special cases for single panel detectors in image viewer
The image viewer has special cases for single and multipanel detectors, however the time for this has long past.  This issue is here to track this problem, because fixing it requires several steps.

First, here's an example of why this matters.  Use dials.import on a single panel image, then run https://github.com/dials/dials/blob/single_panel_special_cases/util/image_viewer/slip_viewer/rotate_datablock.py which will rotate the detector by adjusting the origin, fast and slow vectors.  Use dials.image_viewer on the datablock and the viewer will not display a rotated image, even though the detector methods return the rotated coordinates correctly.  However, do this for a multipanel image, and the rotated image displays fine.  Current use case: Rayonix MX340 detector on its side at LCLS.  Since the data is rotated, polarization correction will be applied incorrectly.

Note, this isn't limited to datablocks.  A rotated full cbf exhibits the same issue.  However, check out the single_panel_special_cases branch and the images, either from a datablock or a rotated cbf, will display correctly.

Steps to fix:

1. The image viewer is currently duplicated in dials/util/image_viewer/slip_viewer because of wxPython 4 work.  However, I can't fix this issue until this duplication is resolved, as I want the fix in the cctbx copy of the code, but I also need to use dials.image_viewer.  Can we fold @ndevenish's changes back into rstbx and remove the duplicated code?
2. Commit 517d7af5b87bdfcd8340a5b894153c340146e81b in the single_panel_special_cases branch identifies where the special cases are.  In some of these the special case can be dropped, but in other cases the work around for multi-panel data is hacked in and should be resolved.
3. With this work, get_flex_image can be deprecated as get_flex_image_multipanel will be used exclusively

phyy-nx:
Remove all (or as many as we could find) single panel special cases from the image viewer.  Fixes #663 

This is a draft pull request, so testing by others would be quite helpful.  We have tested using a single panel image and a multipanel image from dials_regression.  Masking, resolution rings, reflection display (shoeboxes), and a rotated detector all display properly for single and multi-panel images.

Co-authored-by: Asmit Bhowmick <asmit3@gmail.com>


phyy-nx:
Did some more testing today and I think this is ready for review.

Tests:
- Single panel image: dials_regression/image_examples/ALS_831/q315r_lyso_001.img
- Multi panel image: dials_regression/image_examples/LCLS_cspad_nexus/idx-20130301060858401.cbf

For both images, I ran the image viewer with these cases:
- Image alone
- Image imported
- Image rotated (this is the one that didn't work for the single panel case but is now fixed. Again, used dials/util/image_viewer/slip_viewer/rotate_detector.py to generate this)
- Spotfinder spots
- Indexed experiments/reflections

All the tools in 'Actions' seem to work as do the check boxes.

Comments appreciated.  Thanks.

util/image_viewer/slip_viewer/rotate_detector.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import division, print_function
+import sys
+from dxtbx.model.experiment_list import ExperimentListFactory
+from scitbx.matrix import col
+
+"""
+Example script for rotating a detector in an experiment.json
+"""
+
+experiments = ExperimentListFactory.from_json_file(sys.argv[1])
```
graeme-winter: `ExperimentList.from_file()` maybe better?

util/image_viewer/slip_viewer/rotate_detector.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import division, print_function
+import sys
+from dxtbx.model.experiment_list import ExperimentListFactory
+from scitbx.matrix import col
+
+"""
```
graeme-winter: For info - what is the purpose of this example script?

util/image_viewer/slip_viewer/rotate_detector.py
```diff
@@ -0,0 +1,29 @@
...
+detector = experiments[0].detector
+print(detector)
+
+h = detector.hierarchy()
+
+fast = col(h.get_fast_axis())
+slow = col(h.get_slow_axis())
+origin = col(h.get_origin())
+
+normal = fast.cross(slow)
```
graeme-winter: If purpose is education, propose use
```
  get_normal(...)
      get_normal( (VirtualPanelFrame)arg1) -> tuple :
```

here

util/image_viewer/slip_viewer/rotate_detector.py
phyy-nx: Right, this is just for testing.  You run this script to rotate a detector model then load the resultant experiment file in the image viewer.

I have another issue I'll submit soon that will use this script too.  After these are finished I'm cool with this example script going away or landing on a tutorial page or something.

phyy-nx:
Bump :)

util/image_viewer/slip_viewer/ring_frame.py
```diff
@@ -159,16 +159,11 @@ def OnAutoFit(self, event):
...
-                h = detector.hierarchy()
-                if len(h) > 0:
-                    beam_pixel_fast, beam_pixel_slow = detector[0].millimeter_to_pixel(
-                        detector.hierarchy().get_beam_centre(beam.get_s0())
-                    )
-                else:
-                    beam_pixel_fast, beam_pixel_slow = detector[0].millimeter_to_pixel(
-                        detector[0].get_beam_centre(beam.get_s0())
-                    )
+            h = detector.hierarchy()
```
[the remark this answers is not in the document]
phyy-nx: I think even then hierarchy() works.
```
from dxtbx.model import Detector
d = Detector()
p = d.add_panel()
p.get_origin
p.set_frame((1,0,0), (0, -1, 0), (-50, 50, -100))
print (d.hierarchy())
...
  origin: {0,0,0}
...
```

util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -497,20 +458,10 @@ def UseLevel(self, n):
...
-        detector = self.raw_image.get_detector()
-        if sys.platform.startswith("linux"):
-            if len(detector) > 1:
-                return 0.0, 0.0
-            else:
-                return (
-                    self.center_x_lon - self.extent[0],
-                    self.center_y_lat - self.extent[3],
-                )
+        if sys.platform.lower().find("linux") >= 0:
```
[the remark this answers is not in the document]
phyy-nx: Don't know, and my quick dive through git blame isn't helping.

phyy-nx:
Any other feedback?  Otherwise I'll merge tomorrow.

graeme-winter:
Given I have not had a chance to test but others have, I don't want to hold anything up @phyy-nx  so go for it 🙂

phyy-nx:
Thanks all

PR #1183, opened 2020-03-09
Implement shared scaling models and dose_decay scaling model.

jbeilstenedmands:
I don't expect anyone to have a look over the details of this as it's a fair amount of code, but here is a summary of what this is, and happy to receive feedback on the general ideas. I reckon it would be good to get this merged before the branch for 2.2

This work introduces a new scaling model, called dose-decay, which uses an exponentially decaying correction that is linearly dependent on dose :
i.e. decay correction proportional to _exp_(c D / d<sup>&beta;</sup>),
where D is `dose' i.e. image number, c is a proportionality constant which is the free parameter to be refined. This is in addition to a smoothly varying scale correction and overall B-factor per sweep.

It is unclear what the right value of &beta; is;
suggestions are that &beta;=1: https://doi.org/10.1107/S0907444910007262
but some recent experimental studies suggest a value of close to 2: https://doi.org/10.1107/S2052252519008777
so this is set as 2 (`resolution_dependence=quadratic`) for now.

This is ongoing research, targeted towards multi-crystal processing. By default, the decay model component is shared across crystals, i.e. the assumption is that all crystals receive the same dose history (not unreasonable for small crystals immersed in the beam in my opinion). Therefore this PR also implements sharing of model components, which can be used in future e.g. to share absorption surfaces for small molecule work.

Testing on a few datasets so far seems to suggest that this does not give any worse a result than allowing free refinement of smoothly-varying B-factors and may be beneficial for some datasets/analyses/future developments.

### DIFF
--- a/algorithms/scaling/active_parameter_managers.py
+++ b/algorithms/scaling/active_parameter_managers.py
@@ -8,6 +8,7 @@ import logging
 from collections import OrderedDict
 
 from dials.array_family import flex
+from scitbx import sparse
 import six
 
 logger = logging.getLogger("dials")
@@ -27,6 +28,7 @@ class active_parameter_manager(object):
     """
 
     def __init__(self, components, selection_list):
+        # self.target = target
         self.x = flex.double([])
         self.components = OrderedDict()
         self.derivatives = None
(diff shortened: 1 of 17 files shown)


dagewa:
I continue to not like the use of image number as a proxy for dose, on the principle of call a :spades:  a :spades:, but :man_shrugging:

jbeilstenedmands:
Fair point, however if we got to the point where we could provide reliable dose estimates then this could be used.

Issue #1169, opened 2020-02-23
Strange Wilson plot from DIALS


dagewa:
This is from the standard [tutorial data](https://dials.github.io/documentation/tutorials/processing_in_detail_betalactamase.html). Processing with DIALS including `dials.scale` produces a Wilson plot in which the highest resolution reflections actually increase in intensity rather than fall to zero and stay there.
![newplot(2)](https://user-images.githubusercontent.com/11502083/75116816-faefec00-5663-11ea-9cf7-7e544cc32d1c.png)

dagewa:
Further investigation points the finger at the Wilson plot code rather than the intensities themselves. Here is a CCP4-generated Wilson plot for the same DIALS-scaled intensities
![dials-scale-ctruncate](https://user-images.githubusercontent.com/11502083/75120685-2b944d80-5685-11ea-8f8a-2dd7265befa7.png)

dagewa:
As pointed out by @rjgildea 
```
miller_array.setup_binner_d_star_sq_step(auto_binning=True)
wilson_plot = miller_array.wilson_plot(use_binning=True)
pyplot.scatter(wilson_plot.binner.bin_centers(2), flex.log10(flex.double(wilson_plot.data[1:-1])))
```
gives
<img width="1193" alt="Screenshot 2020-02-23 at 22 37 56" src="https://user-images.githubusercontent.com/11502083/75913807-83804080-5e4b-11ea-873e-91639fb385a7.png">

PR #1194, opened 2020-03-16
Calculate french-wilson plot using miller array methods. For #1169

jbeilstenedmands:
Uses cctbx miller array methods to calculate the French-Wilson plot, and mmtbx code to generate the expected distribution, avoiding anisotropic scaling.

Seems to give reasonable looking plots (dials-report.html scaling results -> analysis by resolution)
Beta-lactamase:
newplot (11)
Thermolysin:
newplot (12)
Thaumatin:
newplot (13)

### DIFF
--- a/report/plots.py
+++ b/report/plots.py
@@ -12,6 +12,7 @@ from cctbx import uctbx
 from scitbx.array_family import flex
 from scitbx.math import distributions
 from mmtbx.scaling.absolute_scaling import scattering_information, expected_intensity
+from mmtbx.scaling.matthews import matthews_rupp
 from scipy.optimize import least_squares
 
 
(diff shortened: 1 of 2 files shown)


rjgildea:
> For `n_residues` you should be able to estimate on a case-by-case basis by taking the unit cell volume and assuming 50% solvent, however as it would only be appropriate for the protein case and only affects a scale factor anyway, I don't know if this is important enough to bother.

You can get an estimate of n_residues thusly:
```
from mmtbx.scaling import matthews
matthews.matthews_rupp(intensities).n_residues
```
Used e.g. here:
https://github.com/dials/dials/blob/5302b03a9f6e5fff5d7b403115f1346ca6b17567/algorithms/symmetry/__init__.py#L252-L253

jbeilstenedmands:
Thanks for the suggestion for matthews_rupp, have added this now. @dagewa yes `scipy.optimize.least_squares` was very convenient for this simple case.

Issue #659, opened 2018-11-23
In dials.integrate track down and prevent (ideally) cause of exit code = -9


graeme-winter:
Believed to be memory allocation in second step of integration i.e. once the reference profiles are constructed, when gets to integration it falls over. This is probably the most widely reported real bug.

graeme-winter:
Multiple possible solutions for this - using threaded integration in preference to multiprocessing is one, better prediction of memory usage would be a second. Really needs a test case to throw the problem however.

dagewa:
Possible solution 1 is in the works #833

dagewa:
This is still one of the most commonly-reported bugs, and in my eyes the most serious.

We struggle to process "normal" data on a "normal" computer. For example, I have a dataset from a CCP4 workshop that is nothing particularly special: P3 cell with a=84.5 Å, c=107.0 Å, P6M detector, 360° at 0.1°/image and a moderately large mosaicity of 0.5°. I can't process this beyond 2.0 Å on my laptop with 16 GB RAM, even using just a single process! Ronan ran it on one of the CCP4 servers with nproc=1 and noted it required 31 GB of memory and took 7.5 hours to complete.

XDS has no such problem. We are keeping all integrated data in RAM for no good reason. Once a spot is integrated, it is integrated. Dump it somewhere. Or at least dump its shoebox

graeme-winter:
Excellent to see some diagnostic here, keeping all shoeboxes around is unnecessary and a very poor idea

Do you have pointers where the issues are centred?

dagewa:
I don't have any such pointers, just coming at it from a user perspective at the moment. Tried to process data, it failed. I did notice it failed during integration though, so after profile modelling. It integrated a couple of hundred images then died. I no longer have the logs for that run here though.

graeme-winter:
If the offending images are somewhere public would help a lot! 👍

dagewa:
I can't do for this case. I think a 3600 image fine-sliced dataset of thaumatin would do though. We can set the mosaicity to 0.5° manually

graeme-winter:
Reproduced:

/dls/i04/data/2020/cm26459-1/20200220/TestInsulin/ins_10/ins_10_1_master.h5

```
# disable virtual memory
ulimit -v 0

dials.import ../insulin-vanilla/ins_10_1_master.h5
dials.find_spots imported.expt nproc=8
dials.index imported.expt strong.refl
dials.refine indexed.expt indexed.refl
[...]

PR #1195, opened 2020-03-18
Override sigma b multiplier

graeme-winter:
Default is 2.0, which makes for large bounding boxes for background calculations - leaving this in place for still data. This pull request changes the default for rotation data to 1.0, which should substantially reduce the memory footprint but may have an impact on data quality. 

N.B. there is is no equivalent for sigma_m.

Will significantly help fixing https://github.com/dials/dials/issues/659

### DIFF
--- a/algorithms/integration/integrator.py
+++ b/algorithms/integration/integrator.py
@@ -438,7 +438,7 @@ def _initialize_rotation(experiments, params, reflections):
     # Compute some reflection properties
     reflections.compute_zeta_multi(experiments)
     reflections.compute_d(experiments)
-    reflections.compute_bbox(experiments)
+    reflections.compute_bbox(experiments, sigma_b_multiplier=1.0)
 
     # Filter the reflections by zeta
     mask = flex.abs(reflections["zeta"]) < params.filter.min_zeta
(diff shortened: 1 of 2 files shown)


graeme-winter:
First cut this seems to correspond to a factor of 4 reduction in memory use, which is helpful.

If this works out I will make this into a parameter, though this seems to be non-trivial (it passes through many many onion layers)

graeme-winter:
Tests pass for me 🙂

dagewa:
Can we run a few datasets to look at the effect on data quality?

graeme-winter:
Of course

The default behaviour (i.e. no parameter) is that nothing changes - if you change the parameter the outcome will change obviously. Are you asking for this to be part of the review process to assess the impact of the change if you change the parameter value?

dagewa:
No sorry, I hadn't twigged that defaults haven't changed

graeme-winter:
The default was previously hard coded somewhere in the bottom of `dials.integrate` - I have just moved this up (though the default parameter still exists, as I don't want to change API) 

Still would appreciate someone eyeballing the changes

graeme-winter:
@phyy-nx please could you confirm that this makes no difference to your processing. In principle the additional parameter could help in come cases, but that is maybe out of context

PR #1068, opened 2019-12-11
a DIALS bootstrap script

Anthchirp:
The DIALS bootstrap will only create Python3+Conda environments, that are updateable post-bootstrap. Python 2.7 and non-conda environments are not supported. Rollout is planned for February 2020, at which point xia2/DIALS developers are expected to move over to a Python 3 environment in preparation for Python 2.7 retirement.

PR #1198, opened 2020-03-18
a DIALS installer script

Anthchirp:
work towards a DIALS-specific installer, same process as #1068.
Only Python 3 conda installation is supported (system python 2.7 on the target host is fine), sensible defaults, faster, LZMA compression only, etc.
Hope to have this done and tested in time for the DIALS 3.0 release.
Longer term this will form the basis for experimental lite installers, that do not copy conda packages.

### DIFF
--- a/installer/create_installer.py
+++ b/installer/create_installer.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python
-
 """
 Script to set up an installer directory tree and copy over most of the
 necessary files.  We used to just keep the entire (Phenix) installer in a
@@ -11,15 +9,12 @@ probably need to perform additional modifications to the installer tree
 before it can be tarred.
 """
 
-from __future__ import absolute_import, division, print_function
-
 import imp
 import os
 import shutil
 import stat
 import subprocess
 import sys
-import tarfile
 import time
 import zipfile
 from optparse import OptionParser
@@ -28,7 +23,7 @@ import libtbx.auto_build.rpath
 
 # XXX HACK
 libtbx_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
-if not libtbx_path in sys.path:
+if libtbx_path not in sys.path:
     sys.path.append(libtbx_path)
 
 INSTALL_SH = """\
--- a/installer/create_installer.py
+++ b/installer/create_installer.py
@@ -190,7 +190,6 @@ class SetupInstaller(object):
         for i in ["bin", "lib"]:
             makedirs(os.path.join(self.dest_dir, i))
         self.copy_info()
-        self.write_environment_files()
         self.copy_libtbx()
         self.copy_doc()
         self.copy_modules()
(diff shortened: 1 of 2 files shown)


Anthchirp:
@rjgildea made the point that we shouldn't distribute sources in the installer. If the user wants a development environment then they should install a development environment.

Anthchirp:
another data point on this is that at least on MacOS some development installation paths survive the installer process (cf. KY's mail on dials-support, 2020-06-11 09:21)

Anthchirp:
Copying in an email from a discussion on the dials-support list:

<details><summary>

> In the past we had 5(?) different ways of setting up a DIALS installation, which all more or less resulted in the same outcome:

</summary>

> You could
>  1. run bootstrap.py
>  2. run bootstrap.py with conda
>  3. download an installer.tar.xz file
>  4. download an installer.tar.xz file with conda
>  5. download a source installer
> 
> 1, 3, and 5 were limited to a Python 2.7 installation.
> 2 and 4 could in theory provide you with a Python 2 and a Python 3 environment (however we never distributed or recommended this for Python 2).
> 
> 1 and 2 would give you something which is close to (but not exactly) a 'development' installation: You get all the source code, you get git repositories (if you have git installed, otherwise you get a snapshot), however you don't get a compiler toolchain and you need to have one installed to set up the environment.
> 
> 3 and 4 would give you something which is close to (but not exactly) a 'user' offline installation: You get a binary build for your platform, you do not get git repositories, and you do not need to have a compiler toolchain set up. However you also get all the source code, some build dependencies, and if you have a compiler toolchain you can use the installation to make changes to sources. If you do that you will have to rebuild everything because of path changes.
> 
> 5 would give you something which is close to (but not exactly) a 'development' offline installation: You get all the source code, however you do not get git repositories. You will need a compiler toolchain and need to build everything from scratch.
> 
> 
> From taking a look at https://dials.diamond.ac.uk/diamond_builds/release-2.2.5/ the download sizes are for 3) around 200 MB, for 4) around 250 MB and for 5) around 305 MB.
> 
> As you can see we don't have a true 'user' installation - and no true 'development' installation either. The idea mentioned in https://github.com/dials/dials/pull/1198 is that we start introducing this split properly.
> 
> A user installation does not actually need git repositories, compilers, or C/C++ source code. The runtime environment can be a lot leaner and the installer a lot smaller. We don't need the runtime tools to run tests or to build documentation (bye compiler toolchain, alabaster, biopython, pytest, scons, sphinx, ..). Looking into the future, if/when we get to the point where we have a dials conda-forge package this will be a direct descendant of this user installation and follow the same principles.
> 
> A development installation on the other hand should provide all these things, and since we moved to conda compilers the dials environment now contains nearly everything you need to build stuff, including the compiler toolchain. However, we do not provide an installer for such a true development installation.
> 
> 
> The use case of an offline installation is in my view orthogonal to this split.
> I propose following this split between user and developer and only supply installers for the 'user' side, and cover all use cases on the 'developer' side using bootstrap.
> Our (user) installers therefore will be binary-only. It will makes sense to still keep them self-contained, ie. not rely on 3rd party downloads. This will automatically enable their use in offline environments.
> If you want to build, develop, or extend an installation then the installers are not for you, and you should go through the bootstrap process instead. Currently our bootstrap script only allows online installations, as it will download not only the source repositories but also all package dependencies during the installation steps, before creating a non-relocatable environment - which led to the error that you observed. However we can add support for an offline installation model here. I'm thinking along the lines of eg.
> 
>   python bootstrap.py update base --download-only
> 
> on a machine with internet connection, at which point you can transfer the working directory to an airgapped machine and then set up the development environment in place with
> 
>   python bootstrap.py base build --offline
> 
> This should be relatively easy to accomplish. And while we would distribute the bootstrap.py (as we already do for all releases), we would not pack up the working directory and distribute an offline development /installer/ ourselves.

</details>

Anthchirp:
I will not be able to make this happen. However, with the recent release of [dxtbx on conda-forge](https://anaconda.org/conda-forge/dxtbx) we are now closer than ever to getting a DIALS release that can be installed without a bespoke environment and a 3-10,000 line bootstrap/installer script.

Exciting times.

PR #1202, opened 2020-03-23
Rename scaling.html --> dials.scale.html

dagewa:
for consistency with e.g. dials.cosym.html and the logfile. Unfortunately
dials-report.html is still not consistent with this pattern.


graeme-winter:
100% 👍

jbeilstenedmands:
Thanks for sorting this

dagewa:
Builds and tests passed, but the green tick hasn't appeared

PR #1207, opened 2020-03-24
No trusted range for spotfinding

dagewa:
Is this the right way to deprecate something, just emit a `logger.warning`?

### DIFF
--- a/newsfragments/1156.removal
+++ b/newsfragments/1156.removal
@@ -1,0 +1,1 @@
+``dials.find_spots``: Deprecate ``spotfinder.filter.use_trusted_range=`` parameter
--- a/util/masking.py
+++ b/util/masking.py
@@ -218,7 +218,7 @@ class MaskGenerator(object):
                 warnings.warn(
                     "Checking for hot pixels using the trusted_range is"
                     " deprecated. https://github.com/dials/dials/issues/1156",
-                    DeprecationWarning,
+                    FutureWarning,
                 )
                 trusted_mask = None
                 low, high = panel.get_trusted_range()
--- a/util/masking.py
+++ b/util/masking.py
@@ -214,6 +214,9 @@ class MaskGenerator(object):
             # the trusted range. This identifies bad pixels, but does not include
             # pixels that are overloaded on some images.
             if self.params.use_trusted_range:
+                raise DeprecationWarning(
+                    "Checking for hot pixels using the trusted_range is deprecated. https://github.com/dials/dials/issues/1156"
+                )
                 trusted_mask = None
                 low, high = panel.get_trusted_range()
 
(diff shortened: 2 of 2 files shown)


Anthchirp:
No, you want to 
```python
import warnings
warnings.warn("Text", DeprecationWarning)
```

dagewa:
done, but this is not visible

graeme-winter:
@dagewa the warning will be visible if used in a test environment

graeme-winter:
Does get me wondering though

If we have a PHIL parameter which has been deprecated should we be able to add a deprecate flag & warn in the PHIL parsing if it gets set from it's initial value? 🤔 May ticket this.

rjgildea:
As mentioned [here](https://github.com/cctbx/cctbx_project/issues/452#issuecomment-604313748) there is already a `.deprecated` attribute in phil. This emits a warning with `warnings.warn` if the parameter is used (although a user wouldn't actually see the warning by default).

dagewa:
In this particular case, the deprecation message is more for users than developers. I would like it if such messages ended in stdout and the log, automatically. I thought `logging.captureWarnings(True)` should do it, but it appeared not to.

rjgildea:
This does appear to output a warning to users:
```
$ cat deprecated.py 
import libtbx.phil

phil_scope = libtbx.phil.parse("""\
a = None
    .deprecated = True
""")
params = phil_scope.fetch(libtbx.phil.parse("a=1"))
params.show()
$ libtbx.python deprecated.py 
/Users/rjgildea/software/cctbx/modules/cctbx_project/libtbx/phil/__init__.py:1012: PhilDeprecationWarning: a is deprecated - not recommended for use.
  self.full_path(), PhilDeprecationWarning)
```

See also:
https://github.com/cctbx/cctbx_project/blob/b0210c194b5819365f4a968240312585568b7241/libtbx/phil/__init__.py#L20-L22
and
https://github.com/cctbx/cctbx_project/blob/b0210c194b5819365f4a968240312585568b7241/libtbx/phil/__init__.py#L1011-L1012

Anthchirp:
Apparently the appropriate warning category if you want to tell users about deprecated features is
[`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) as per [PEP565](https://www.python.org/dev/peps/pep-0565/#additional-use-case-for-futurewarning).

So the most careful deprecation path would be
`PendingDeprecationWarning` :arrow_right: `DeprecationWarning` :arrow_right: `FutureWarning` :arrow_right: feature removal

`PendingDeprecationWarnings` could be used to backport deprecation warnings, ie. if a feature was marked as deprecated or removed _after_ a release was made then that feature could be marked as will-be-deprecated-in-the-future in the next point release. Users would be none-the-wiser, but developers building on it would see harmless test output that can be bulk-ignored if so desired.

I'll add a commit here to change this PR to `FutureWarning`.

ndevenish:
This seems to work and is probably something we want in a X.Y.0 release so going to merge

PR #1208, opened 2020-03-24
Feature for masking: mask out whole panels

phyy-nx:
If untrusted.panel is specified but no geometric masks are specified (like rectangle, circle, etc.), then mask out the whole panel.

Suggested by @nksauter.

### DIFF
--- a/test/command_line/test_generate_mask.py
+++ b/test/command_line/test_generate_mask.py
@@ -171,3 +171,23 @@ def test_generate_mask_trusted_range(dials_data, tmpdir):
 
     # Overloads should not be included in the mask
     assert (mask1[0] == mask2[0]).all_eq(True)
+
+
+def test_generate_whole_panel_mask(input_experiment_list, tmpdir):
+    params = phil_scope.fetch(
+        phil.parse(
+            """
+untrusted {
+  panel = 0
+}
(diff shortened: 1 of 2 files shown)


dagewa:
This seems a reasonable interface change to me. Could a test be added please?

phyy-nx:
Test added. Also note a6fa94b, which was required to get `untrusted.panel=0` working.

test/command_line/test_generate_mask.py
```diff
@@ -171,3 +171,23 @@ def test_generate_mask_trusted_range(dials_data, tmpdir):
...
     # Overloads should not be included in the mask
     assert (mask1[0] == mask2[0]).all_eq(True)
+
+
+def test_generate_whole_panel_mask(input_experiment_list, tmpdir):
+    params = phil_scope.fetch(
+        phil.parse(
+            """
+untrusted {
+  panel = 0
```
rjgildea: Is this a multi-panel test case? If so it might be better to test with a non-zero panel id?

phyy-nx: The test is single panel, but in this case 0 specifically was tripping me up (see a6fa94b) so that's why it's tested.

PR #1209, opened 2020-03-25
Move sys.exit out of potential library functions from export.py

### LINKED ISSUE #1205: dials.scale exits on error without traceback
This is spun out from https://github.com/xia2/xia2/issues/430

`export_mtz` from `command_line/export_mtz.py` is treated as library code by `dials.scale`:
https://github.com/dials/dials/blob/5058b42c5b6932b2bca2a62c28b961eb5a5926cf/command_line/scale.py#L148

However a `ValueError` raised here is caught and the program simply exits, which is confusing
https://github.com/dials/dials/blob/5058b42c5b6932b2bca2a62c28b961eb5a5926cf/command_line/export.py#L263-L266
Maybe `export_mtz` should not be swallowing this `ValueError`, or/and maybe`dials.scale` should be using `dials.util.export_mtz` instead.

dagewa:
Various functions in `export.py` such as
https://github.com/dials/dials/blob/adc74ed395693983697312452a5ccc5df15f6746/command_line/export.py#L241
could be used as library functions (indeed this one is, by `dials.scale`). However, they catch exceptions and do other error testing and immediately exit in the function bodies. The changes here raise the exceptions there instead, to be caught in the __main__ block.

Fixes #1205

### DIFF
--- a/command_line/export.py
+++ b/command_line/export.py
@@ -219,12 +219,12 @@ phil_scope = parse(
 
 def _check_input(experiments, reflections, params=None, check_intensities=False):
     if not experiments:
-        raise TypeError("export requires an experiment list")
+        raise ValueError("export requires an experiment list")
     if len(reflections) != 1:
         raise ValueError("export requires 1 reflection table")
     if check_intensities:
         if not params:
-            raise TypeError("No parameters given")
+            raise ValueError("No parameters given")
         if "profile" not in params.intensity and "sum" not in params.intensity:
             raise ValueError(
                 "Only intensity options containing sum or profile can be exported in this format"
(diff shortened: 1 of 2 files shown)


dagewa:
Can still change that? I was trying to follow the guidance, viz.
- [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError): Raised when an operation or function is applied to an object of inappropriate type
- [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError): Raised when an operation or function receives an argument that has the right type but an inappropriate value

Anthchirp:
The tests are not actually checking the _type_ though, they go for truthiness. So if an empty experiment list is false-y (I'd argue it should be) then the exception will be wrong.

dagewa:
What if `None` is passed for these parameters? Would that be a `TypeError` or a `ValueError`? I was assuming the former. I take your point that empty versions of the sequence-like containers should be raising `ValueError`

PR #1210, opened 2020-03-25
Spot threshold as overlay

### LINKED ISSUE #1200: Live spot finding view in dials.image_viewer
The dials image viewer is very useful for adjusting spot finding parameters - it would be _even more useful_ if one could view the image and the pixels which are identified as "spot" at the same time i.e. something like

Screenshot 2020-03-20 at 10 01 27

computed real-time...

dagewa:
This adds a new "Threshold pixels" overlay, which indicates which pixels would be marked by the spot-finding algorithm. Fixes #1200


command_line/image_viewer.py
```diff
@@ -44,6 +44,8 @@
   .type = bool
 show_all_pix = True
   .type = bool
+show_thresh_pix = False
```
rjgildea: Explicit parameter name better than saving 3 characters:
```suggestion
show_threshold_pix = False
```

util/image_viewer/spotfinder_frame.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, *args, **kwds):
         self._mask_frame = None
 
         self.show_all_pix_timer = time_log("show_all_pix")
+        self.show_thresh_pix_timer = time_log("show_thresh_pix")
```
rjgildea: I think all these `show_*_timer`s should be removed - I'm not sure their output goes anywhere, so they are just adding clutter to the code.

util/image_viewer/spotfinder_frame.py
```diff
@@ -1161,6 +1165,48 @@ def update_settings(self, layout=True):
...
 
+        if self.settings.show_thresh_pix:
+            image = self.pyslip.tiles.raw_image
+            kabsch_debug_list = self._calculate_dispersion_debug(image)
+            cv = [kabsch.index_of_dispersion() for kabsch in kabsch_debug_list]
+            final_mask = [kabsch.final_mask() for kabsch in kabsch_debug_list]
+            final_mask = [mask.as_1d() for mask in final_mask]
+            value = []
+            for pnl, mask in enumerate(final_mask):
+                mask.reshape(cv[pnl].accessor())
```
rjgildea: I'm not sure why the mask is converted to 1d and then reshaped, but it doesn't appear to be needed (I tested the change below and it seems to work):
```suggestion
            final_mask = [kabsch.final_mask() for kabsch in kabsch_debug_list]
            value = []
            for pnl, mask in enumerate(final_mask):
```

util/image_viewer/spotfinder_frame.py
```diff
@@ -1675,6 +1721,7 @@ def __init__(self, *args, **kwargs):
         self.settings.show_ctr_mass = self.params.show_ctr_mass
         self.settings.show_max_pix = self.params.show_max_pix
         self.settings.show_all_pix = self.params.show_all_pix
+        self.settings.show_thresh_pix = self.params.show_thresh_pix
```
rjgildea:
```suggestion
        self.settings.show_threshold_pix = self.params.show_threshold_pix
```

util/image_viewer/spotfinder_frame.py
dagewa: If they do nothing then I agree they should be removed. I think that should be outside this PR though so it doesn't get lost in the squash merge commit message

util/image_viewer/spotfinder_frame.py
dagewa: probably copypasta

dagewa:
> However, I notice that the mask doesn't update with changing the parameters

Weirdly, it appears to update _sometimes_. At least while on `image` mode, fiddling with `gain` does update the overlay. I agree it doesn't appear to update while on `threshold` mode

dagewa:
@graeme-winter Fixing the stack issue was as simple as moving the `sum_images` call to after the calculation of the threshold overlay!

I have not yet figured out how to ensure the overlay updates when `display == threshold`

rjgildea:
With 18d32a0 the overlay should now be updated correctly with dispersion == threshold. However, this does result in two (presumably identical) calls to `_calculate_dispersion_debug` every time you change a spotfinding parameter - I wonder if this could be cached somehow, as it would only need to be changed when either the image changes or if a limited number of parameters have changed...

dagewa:
It would be better if all of the `Debug` stuff was cached. See also https://github.com/dials/dials/pull/948#issuecomment-535450372

Issue #670, opened 2018-12-05
Avoiding loss of information after dials.two_theta_refine


rjgildea:
For multi-crystal work, a proposed post-integration workflow in DIALS might be:

1) Analysis of indexing ambiguities - dials.cosym
2) Determination of symmetry - dials.symmetry
3) Refinement of "best overall" unit cell - dials.two_theta_refine
4) Scaling - dials.scale

Passing the resulting refined_cell.json from dials.two_theta_refine to dials.scale ensures that the correct "best overall" unit cell is written into the output mtz files for downstream analysis, and that the correct (single) unit cell parameters are used for applying any resolution cutoff in scaling (so that the same miller indices are included/excluded for each sweep).

However, this results in the loss of information of the original individual unit cells and orientation, which would be useful for e.g. unit cell clustering analysis, and analysis of the distribution of the crystal orientation. dials.two_theta_refine overwrites the input crystal models with a single crystal model and orientation (presumably the orientation of the first experiment?) which is shared across all experiments. This seems like a bit of a hack, and could also lead to problems if the user tried to use the resulting file for further refinement or integration, as the predictions would be nonsensical.

I welcome suggestions for how we can improve the current situation.

rjgildea:
Could we instead attach an additional crystal model attribute to the experiment list? E.g. experiment_list.best or experiment_list.overall?

dagewa:
The fact that `dials.two_theta_refine` overwrites orientation sounds like a bug. This should be addressed anyway, whatever we do with the experiment list.

In the presented workflow `dials.two_theta_refine` is essentially doing a form of global post-refinement. Perhaps the dxtbx `Crystal` should have a new attribute, like `post_refinement_unit_cell`, or `best_unit_cell`, and `dials.two_theta_refine` should only alter that leaving each crystal's static model intact. That would avoid attaching additional attributes at the global experiment list level, which might get problematic if you later extended an experiment list with another (for whatever reason).

graeme-winter:
Extending the crystal to have an explicit "best" unit cell seems like a neat solution - at the moment do we keep the cell constants in a unit cell or a B matrix? If we were to keep the two separate conceptually then we could update only the cell in global cell refinement and continue to use the B for predictions etc.

Then when you update B you would update the cell, but updating the cell would not apply the same modifications. So getting the cell constants should always give best values. 

Need to review how these are modelled. 

[...]

PR #1214, opened 2020-03-27
Use recalculated_unit_cell in two_theta_refine

rjgildea:
Use `recalculated_unit_cell` (if set) in dials.scale and export_mtz.

Depends on cctbx/dxtbx#153


rjgildea:
The test introduced in 0e9ef4b currently fails, as the `recalculated_unit_cell` is lost when `.change_basis` is called on the crystal in export_mtz:
https://github.com/dials/dials/blob/78f1e749952279caff1e77ccd0035f1b78f9cea8/util/export_mtz.py#L531
Obviously we could quite easily apply the change of basis to the `recalculated_unit_cell` in the `Crystal.change_basis` method, although I'm not sure whether we can do anything sensible to the recalculated unit cell parameter/volume sd's? @dagewa any thoughts here? Should we just discard the sd's in `Crystal.change_basis`?

dagewa:
I've not yet tested, but this looks like the right approach. At a low level, `dials.two_theta_refine` just updates the usual cell via the **B** matrix parameterisation. At the end we make the high level decision to copy the unit cells into the original experiments as recalculated cells.

dagewa:
> Should we just discard the sd's in `Crystal.change_basis`?

I would say yes, because some change of basis operations might result in slightly different estimated σ values after refinement if that change had been made before refinement rather than after. I haven't actually tested that, but I would not be surprised if that's the case.

Being defensive, we should also wipe out the errors any time the recalculated cell setter is called.

dagewa:
I think this fixes #670

algorithms/scaling/scaling_library.py
```diff
@@ -381,7 +381,11 @@ def determine_best_unit_cell(experiments):
     experiments."""
     uc_params = [flex.double() for i in range(6)]
     for exp in experiments:
-        for i, p in enumerate(exp.crystal.get_unit_cell().parameters()):
+        if exp.crystal.get_recalculated_unit_cell() is not None:
+            unit_cell = exp.crystal.get_recalculated_unit_cell()
+        else:
+            unit_cell = exp.crystal.get_unit_cell()
```
Anthchirp, line 387: maybe
```suggestion
        unit_cell = exp.crystal.get_recalculated_unit_cell() or exp.crystal.get_unit_cell()
```
but you iirc you don't like that syntax.

util/export_mtz.py
```diff
@@ -558,9 +558,12 @@ def export_mtz(integrated_data, experiment_list, params):
         else:
             experiment.data["ROT"] = z
 
-    mtz_writer.add_crystal(
-        params.mtz.crystal_name, unit_cell=experiment_list[0].crystal.get_unit_cell()
-    )  # Note: add unit cell here as may have changed basis since creating mtz.
+    if experiment_list[0].crystal.get_recalculated_unit_cell() is not None:
+        unit_cell = experiment_list[0].crystal.get_recalculated_unit_cell()
+    else:
+        unit_cell = experiment_list[0].crystal.get_unit_cell()
```
Anthchirp, line 564: same

algorithms/scaling/scaling_library.py
rjgildea: I think in this case it is probably justified as it is a simple enough assignment and makes the code much shorter.

PR #1221, opened 2020-04-06
dials.integrate: change memory detection logic

Anthchirp:
##### excerpt from a DIALS support call

We do have a check in place before we go an try allocating the memory we need, and if we find out that we'd need more memory than we can have we reduce the number of parallel processes to fit into memory, trying to avoid an out-of-memory crash. In your case this logic determined that dials.integrate should use 22 cores, and then it still crashed.

I believe this is because our memory check apparently does the following:
*	We calculate the amount of memory required by one process
*	We take the available system memory and multiply it by a factor of 0.75 (parameter `integration.block.max_memory_usage`)
*	We divide the first second number by the first number, and use the rounded down value of that division to limit the number of processes.
This is probably fine, but it builds on the assumption that everything else running on your computer fits inside `system memory * (1 - integration.block.max_memory_usage) + swap size`.
This is because "available system memory" literally means "theoretically available physical memory in this machine". It does not mean "memory available to allocate". Incidentally it also excludes swap memory.

For your immediate case it is easiest to just reduce the number of processes with the xia2 `nproc` option below 22, depending on the memory situation on your machine.

For us DIALS folks I think we need to do a bit better with the memory estimation. I propose the following logic:
* Find the amount of memory available for allocation ([`psutil.virtual_memory().available`](https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory)) and the amount of memory available in swap ([`psutil.swap_memory().free`](https://psutil.readthedocs.io/en/latest/#psutil.swap_memory))
* Subtract the second from the first to obtain the amount of memory available for immediate allocation, multiply it by `integration.block.max_memory_usage`
* Divide this by the amount of memory required by one process, round down and limit the number of processes to this number.
* If this number falls below 1 then
  - Check whether the amount of memory available for immediate allocation including swap is greater than the amount of memory required by one process
  - If this is the case, then continue with 1 process
  - Otherwise crash out with the usual "Not enough memory ..." error

Further:
*	raise the default of `integration.block.max_memory_usage` from 0.75 to 0.90 or maybe even higher. Reason for this is that the 'unknown' caused by the rest of the system has mostly gone away.
*	write all the obtained numbers into the log. That would be really helpful for debugging...
*	and for a bonus: limit amount of 'memory available for immediate allocation' to `resource.getrlimit(resource.RLIMIT_RSS)[0] - psutil.Process().memory_info().rss`, because psutil ignores ulimits.


Anthchirp:
Very interesting behaviour with this patch.
I get
```
Integrating reflections

 Split 91499 reflections overlapping job boundaries

Memory required per process: 13.4 GB
Available memory: 13.5 GB, swap: 4.3 GB, memory limit: 12.2 GB
No memory ulimit set
Reducing number of processes from 4 to 1 due to memory constraints.
Running this process will rely on swap usage!
Processing reflections in the following blocks of images:

 block_size: 152 frames

(...)
 Using multiprocessing with 1 parallel job(s)

 Beginning integration job 0

 Frames: 0 -> 150

(...)

 Memory usage:
  Available system memory (excluding swap): 17.8 GB
  Available swap memory: 4.3 GB
  Available system memory (including swap): 13.5 GB
  Limit shoebox memory (including swap): 16.0 GB
  Limit shoebox memory (excluding swap): 12.2 GB
  Required shoebox memory: 13.4066 GB

Running this process will rely on swap usage!

(...)
 Beginning integration job 1

 Frames: 75 -> 225

 Number of reflections
  Partial:     2337
  Full:        27900
  In ice ring: 0
  Integrate:   30237
  Total:       30237

 The following histogram shows the number of reflections predicted
 to have all or part of their intensity on each frame.

 76  [14   ]: 
(...)
 225 [1038 ]: ****

  File "/scratch/wra62962/files/dials/build/../modules/dials/command_line/integrate.py", line 665, in <module>
    script.run()
  File "/scratch/wra62962/files/dials/build/../modules/dials/command_line/integrate.py", line 353, in run
    reflections = integrator.integrate()
  File "/scratch/wra62962/files/dials/modules/dials/algorithms/integration/integrator.py", line 1205, in integrate
    self.reflections, _, time_info = processor.process()
  File "/scratch/wra62962/files/dials/modules/dials/algorithms/integration/processor.py", line 306, in process
    self.manager.accumulate(task())
  File "/scratch/wra62962/files/dials/modules/dials/algorithms/integration/processor.py", line 498, in __call__
    required_memory / 1e9,
MemoryError: Please report this error to dials-support@lists.sourceforge.net: 
        There was a problem allocating memory for shoeboxes.  This could be caused
        by a highly mosaic crystal model.  Possible solutions include increasing the
        percentage of memory allowed for shoeboxes or decreasing the block size.
        The average shoebox size is 38 x 39 pixels x 36 images - is your crystal
        really this mosaic?
            Available system memory (excluding swap): 1.7 GB
            Available system memory (including swap): 5.3 GB
            Shoebox memory limit: 4.8 GB
            Required shoebox memory: 12.3 GB
```

Looks like we are doing subsequent memory checks without freeing?

Anthchirp:
The failing check happens between integration block 1 and integration block 2.
When I replace the `raise MemoryError` with a simple `print` the integration continues, so this is definitely a false positive and not (just) bad memory handling.
I don't see the point of re-running these checks before every integration block? Either there is enough memory to run all of integration or there isn't. If memory usage on the machine changes throughout the run then that's too bad, but I don't see why we would be tracking that.

These subsequent checks are possibly the reason why we divide the limit between the number of processes.
https://github.com/dials/dials/blob/dd473848474b509c53380a4f106a516b139cef18/algorithms/integration/processor.py#L875-L878

graeme-winter:
This works for me! 

```
|     0 |  88264 |  0.62755 |  0.54335 |    0.21041 |
Updating predictions for indexed reflections

Final refined crystal model:
Crystal:
    Unit cell: 234.145(2), 234.145(2), 234.145(2), 89.99999999999999, 90.0, 89.99999999999999
    Space group: I 2 3
    U matrix:  {{ 0.5844,  0.6010, -0.5452},
                { 0.8053, -0.3470,  0.4806},
                { 0.0997, -0.7200, -0.6868}}
    B matrix:  {{ 0.0043,  0.0000,  0.0000},
                {-0.0000,  0.0043,  0.0000},
                {-0.0000, -0.0000,  0.0043}}
    A = UB:    {{ 0.0025,  0.0026, -0.0023},
                { 0.0034, -0.0015,  0.0021},
                { 0.0004, -0.0031, -0.0029}}
    A sampled at 1801 scan points
Saving refined experiments to refined.expt
Saving reflections with updated predictions to refined.refl
```

3x3x3 unit cell seems to work fine here though was very slow - however the memory usage calculations are spot on

graeme-winter:
```
dials.import /Volumes/Blue/Data/i04-ins/upload/insu_d200_1_master.h5 
dials.find_spots imported.expt nproc=8
dials.index imported.expt strong.refl space_group=I23
dials.reindex indexed.* change_of_basis_op=3a,3b,3c space_group=I23
dials.refine reindexed.*
dials.integrate refined* nproc=8
```

... used 3 cores but worked

graeme-winter:
Screenshot 2020-04-21 at 11 19 53

pretty crowded, too 🙂

graeme-winter:
```
Grey-Area memory-1221-triple-ins :) $ grep -A8 "Timing information for integration" dials.integrate.*
dials.integrate.log:4428.1: Timing information for integration
dials.integrate.log-        +-------------------+------------------+
dials.integrate.log-        | Read time         | 1030.33 seconds  |
dials.integrate.log-        | Extract time      | 709.72 seconds   |
dials.integrate.log-        | Pre-process time  | 16.87 seconds    |
dials.integrate.log-        | Process time      | 17163.21 seconds |
dials.integrate.log-        | Post-process time | 0.00 seconds     |
dials.integrate.log-        | Total time        | 19356.57 seconds |
dials.integrate.log-        +-------------------+------------------+
--
dials.integrate.save:        Timing information for integration
dials.integrate.save-        +-------------------+------------------+
dials.integrate.save-        | Read time         | 821.59 seconds   |
dials.integrate.save-        | Extract time      | 545.21 seconds   |
dials.integrate.save-        | Pre-process time  | 16.69 seconds    |
dials.integrate.save-        | Process time      | 13570.99 seconds |
dials.integrate.save-        | Post-process time | 0.00 seconds     |
dials.integrate.save-        | Total time        | 15220.12 seconds |
dials.integrate.save-        +-------------------+------------------+
```

Former is master, latter is your branch - seems to be related to using 5 processes (master) vs. 3 (branch) 🤔

algorithms/integration/processor.py
```diff
@@ -459,44 +459,69 @@ def __call__(self):
...
         ), "maximum memory usage must be > 0"
         assert (
             self.params.block.max_memory_usage <= 1.0
         ), "maximum memory usage must be <= 1"
-        limit_memory = total_memory * self.params.block.max_memory_usage
-        if sbox_memory > limit_memory:
+
+        # Compute expected memory usage and warn if not enough
+        required_memory = processor.compute_max_memory_usage()
+        available_memory = psutil.virtual_memory().available
```
graeme-winter, line 471: I note that VM on fruiters is a bit special - it is not a static amount of swap, more something where it grows according to it's own whims - 

```
 Memory usage:
  Available system memory (excluding swap): 9.2 GB
  Available swap memory: 1.6 GB
  Available system memory (including swap): 10.8 GB
  Limit shoebox memory (including swap): 1.2 GB
  Limit shoebox memory (excluding swap): 1.0 GB
  Required shoebox memory: 0.0 GB
```

is pretty misleading

algorithms/integration/processor.py
```diff
@@ -792,43 +817,111 @@ def compute_processors(self):
...
-                total_memory is not None and total_memory > 0
-            ), "psutil call appears to have given unexpected output"
-            limit_memory = total_memory * self.params.block.max_memory_usage
-            njobs = int(math.floor(limit_memory / max_memory))
-            if njobs < 1:
-
-                xsize, ysize, zsize = _average_bbox_size(self.reflections)
-                raise RuntimeError(
-                    """
+            available_memory = psutil.virtual_memory().available
```
graeme-winter, line 829: Seeing the same code block repeated - wonder if we should have some dials.util.memory wrapper which computes available wired and virtual memory? Seems like probably useful elsewhere.

algorithms/integration/integrator.py
```diff
@@ -96,9 +96,9 @@ def generate_phil_scope():
                   "number of blocks may be set to 1. If force is True then the"
                   "block size is always calculated."
 
-        max_memory_usage = 0.75
+        max_memory_usage = 0.90
```
graeme-winter, line 99: Configuration parameter perhaps useful - If I am happy to have chrome swapped out I could push this to e.g. 150%

algorithms/integration/processor.py
```diff
@@ -792,43 +817,111 @@ def compute_processors(self):
...
+                return
+            if njobs >= 1:
+                # There is enough memory to run, but not as many processes as requested
+                logger.info(
+                    "Reducing number of processes from %d to %d due to memory constraints",
+                    self.params.mp.nproc,
+                    int(njobs),
+                )
+                self.params.mp.nproc = int(njobs)
+                self.params.block.max_memory_usage /= self.params.mp.nproc  # why?
```
graeme-winter, line 887: Since you have made the effort to look at all this - `# why?` we should probably resolve as a part of this review?

algorithms/integration/processor.py
Anthchirp: Yeah, dynamic swap allocation is a thing. Then again we don't really want to work in swap anyway. An alternative would be to give a warning, but then continue and potentially run into `-9` if the turing-tape isn't actually infinite.

graeme-winter: Our prior on Turing tapes is that they are finite, so the change set is probably good. Am wondering then whether we should consider the swap at all? I guess if one process requires more memory than we have currently free in wired + swap then we are in trouble, but beyond that does it contain much signal?

At the same time, the signal-to-noise of the log output is pretty low, and rarely read, so on balance does no harm

algorithms/integration/processor.py
Anthchirp: The second check (see above 'why duplicate?') happens at the start of each task. But each task is now only allowed `1/n` of the total memory, which is what I assume this is for.
The only way this makes any sense is when the task is run on a different machine, but in that case it doesn't make any sense :thinking:

In my opinion we should scratch the per-task check. Check once at the beginning of the process and either there is enough memory or there is not. We don't catch memory disappearing while we're running inside a task, so it's not like this additional check is of much use.

algorithms/integration/integrator.py
Anthchirp: It is really inconvenient to map that desired behaviour into this number though.

Say you have 16 GB of RAM, 2 GB of swap, other stuff takes up 10 GB and chrome another 4 GB, leaving 2 GB physical RAM free. Some OS bits take up 0.5 GB swap.

So to allow to use 6 GB you may think you'd set this to 300%, but it's percent of available (2 GB physical + 1.5GB swap = 3.5GB), so 171%.

At least I'd find that mental arithmetic inconvenient, YMMV :wink:

graeme-winter: Yeh fair enough, however set(reasons for not making a parameter here) = ? Other than having to push a parameter through multiple onion layers?

algorithms/integration/processor.py
graeme-winter: +1 to scratching per-task check

algorithms/integration/integrator.py
Anthchirp: Vo reason. Just thinking an absolute "use x GB of memory for integration" with default `>=0.9 * available` could be more accessible than a relative value of something potentially obscure

graeme-winter: <- has missed something then

we have an alternative option to use?

algorithms/integration/processor.py
Anthchirp: Done.

algorithms/integration/integrator.py
Anthchirp: Right now there isn't one, but we could make one happen.

algorithms/integration/processor.py
Anthchirp: code duplication resolved by removing task check

algorithms/integration/processor.py
Anthchirp: I used swap vs ram mainly in deciding how many processes to allow, and the sum in deciding when to give up, so
10 GB RAM, 10 GB swap, 1 GB/process => maximum number of processes: 10
10 GB RAM, 10 GB swap, 10 GB/process => maximum number of processes: 1
10 GB RAM, 10 GB swap, 20 GB/process => maximum number of processes: 1
10 GB RAM, 10 GB swap, 20.1 GB/process => fail with error message

PR #1235, opened 2020-04-15
Move to conda compilers

### LINKED ISSUE #1201: Use conda compilers by default
Move away from platform-dependent builds and use the official conda compilers instead.
```
wget https://raw.githubusercontent.com/dials/dials/master/installer/bootstrap.py
python bootstrap.py update base
conda_base/bin/conda install -p conda_base c-compiler cxx-compiler
. conda_base/etc/profile.d/conda.sh
conda activate ./conda_base
python bootstrap.py build --config-flags="--compiler=conda"
```
gives a first build, but we also need to make sure that the settings 'stick' with a later `make reconf`.

Anthchirp:
Change new development environments to use native conda-forge compilers rather than system compilers.

On existing environments DIALS command line completion will stop working.
Apart from that this PR does not make any changes to existing development environments. 

After creating a new bootstrap base & build the following things change:
* compilers are added to the environment
* `setpaths.sh` and related scripts are made obsolete and will no longer work. They will point to a new `dials` script that is used to set up the environment.
* this script sets up the conda environment properly and adds the `libtbx/bin` directory on top
* as a consequence all normal dispatcher commands become available with their usual names.
  * No need for `libtbx.` prefixes any more.
  * The `conda` command will work without needing to add a `-p` option
* `make`/`make reconf` etc. will now use conda-forge compilers
* builds will now use C++11 (`-std=c++11`). On merging will amend the cctbx-bootstrap to also set this flag for the DIALS builder.

open issues:
- [x] need to check if fortran compilers are actually needed
- [x] this may break binary installs until #1198 is fixed
- [x] need to test on MacOS
- [x] need to run full regression test suite
- [x] need to update documentation & add a newsfragment

fixes #1201 
closes cctbx/cctbx_project#453, no changes in cctbx are required


phyy-nx:
Hi, I haven't read this carefully but I believe it changes the defaults of the dials bootstrap only.  Therefore you are preserving the following use case, correct?

(note using the cctbx bootstrap)

- wget https://raw.githubusercontent.com/cctbx/cctbx_project/master/libtbx/autobuild/bootstrap.py
- python bootstrap.py --builder=dials hot update
- (activate my own conda environment so I can use different compilers)
- python bootstrap.py --builder=dials build --use-conda $CONDA_PREFIX

Anthchirp:
Yes, that'll see no changes (apart from the lack of bash command line completion)

Anthchirp:
On MacOS builds this depends on cctbx/cctbx_project#409

phyy-nx:
> On MacOS builds this depends on [cctbx/cctbx_project#409](https://github.com/cctbx/cctbx_project/pull/409), otherwise eg. `import scitbx.array_family.flex` with crash the interpreter.

Can you explain why?  In my reading of cctbx/cctbx_project#409 I don't see what would affect using the conda compilers.

Anthchirp:
On crashing you get a two liner about numpy imports failing, but no trace. So I suspect it's a compiler+boost vs. numpy issue, but I'm not sure.

phyy-nx:
Is it because you need boost 1.7?  That is part of cctbx/cctbx_project#409, as it modifies the branch used when checking out boost.

Anthchirp:
That's what I'm thinking. Some sort of issue with the conda-forge compiler and the old version of boost maybe, that manifests in the interplay with numpy.

Anthchirp:
The test `libtbx.python modules/cctbx_project/scitbx/lbfgs/tst_lbfgs_fem.py` fails on MacOS but is fine on Linux. May need some investigation to see what is going on.

phyy-nx:
As an aside, I would think in this regime of using the conda compilers,
that you would not download the boost sources during update, but would
instead install a standard boost package from conda, as specified in the
environment file, yes?

Anthchirp:
As far as I know that is still unsupported. While we use the conda compilers we are still limited to using scons as build mechanism rather than conda build, so wouldn't be able to find the conda-forge boost package.

If we can get to work that'd be very welcome, but it's probably one for a future pull request.

Anthchirp:
You can now try this out by running the following commands in an empty directory:
```
curl https://raw.githubusercontent.com/dials/dials/conda-compiler/installer/bootstrap.py > bootstrap.py
python bootstrap.py --b dials@conda-compiler --b cctbx_project@dials-conda-compiler --b boost@1.70
```

Anthchirp:
Discussed this PR in DIALS-E meeting. We will review this after the 3.0 release branch has been made, as we feel it is too high a risk to merge this before the release. By then the cctbx-409 changes will have settled as well, so we will only be changing one big thing at a time.

Anthchirp:
Updated instructions to test following the cctbx-409 merge:
```bash
curl https://raw.githubusercontent.com/dials/dials/conda-compiler/installer/bootstrap.py > bootstrap.py
python bootstrap.py --b dials@conda-compiler
```

This is planned to be merged following the 3.0 release branch, around mid-May

Anthchirp:
@luisodls pointed out that the reciprocal lattice viewer is broken.

I have made a possible fix in https://github.com/cctbx/cctbx_project/pull/484 and added a test that will flag the missing gltbx library in 90f02b8fd7c62be0fbda1e917ee2dea64ec1a47d

Anthchirp:
Fix verified for linux and (thanks to @rjgildea) mac.

PR #1237, opened 2020-04-15
Decoy measurements for stills processing

phyy-nx:
Option to integrate blank shoeboxes in dials.stills_process, as a control for background modeling.

Uses body-centering as default to predict areas without Bragg diffraction, one of seven modulus=2 sublattice transformation(s) used to produce distinct coset results.  See Sauter and Zwart, Acta D (2009) 65:553.


phyy-nx:
No comments so merging :)

PR #1242, opened 2020-04-24
remove dead code

Anthchirp:
I ran https://gist.github.com/Anthchirp/e926d72ef4f9cfaf0ae578d0973bc197 over dials and found some code that nobody refers to - neither from the dials codebase, nor from cctbx_project, dxtbx, xia2, dials_scratch, etc...

Unless anyone is specifically interested in these bits I'll remove them shortly.


algorithms/refinement/parameterisation/detector_parameters.py
```diff
@@ -546,363 +546,11 @@ def compose(self):
         ]
 
     def get_state(self, multi_state_elt=0):
-
         # There is only one detector, but the req. panel must be specified
         panel = (self._model)[multi_state_elt]
         return matrix.sqr(panel.get_d_matrix())
 
 
-class PyDetectorParameterisationMultiPanel(DetectorParameterisationMultiPanel):
```
dagewa, line 555: I kept this around as sort of documentation, because it is much easier to understand what is going on in the Python `compose` method than the base class method, which calls out to C++
https://github.com/dials/dials/blob/afd9f7c613d373caaac62db26d56191c24d413a3/algorithms/refinement/parameterisation/detector_parameters.py#L523-L537

However, I can see that leaving bits of dead code around as "documentation" is far from ideal. What would be better? Perhaps we should keep the code for use in a test that specifically just checks that the results from the Python version match those from the one that uses the C++ function?

Anthchirp: having it live on in a `test_*.py` file sounds like a good idea in that case

Anthchirp:
I just reran the script, this time looking for unused functions.

algorithms/indexing/stills_indexer.py
```diff
@@ -790,10 +789,6 @@ def refine(self, experiments, reflections):
 """Mixin class definitions that override the dials indexing class methods specific to stills"""
 
 
-class StillsIndexerKnownOrientation(IndexerKnownOrientation, StillsIndexer):
```
phyy-nx, line 793: Did your script to find dead code include other repos?  Please keep this class.
https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/xtc_process.py#L1349-L1353

util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -516,16 +505,6 @@ def lon_lat_to_picture_fast_slow(self, longitude, latitude):
             (size1 / 2.0) - (latitude - self.center_y_lat),
         )
 
-    def picture_fast_slow_to_lon_lat(self, pic_fast_pixel, pic_slow_pixel):
```
phyy-nx, line 519: Please keep this one.

nksauter:
Markus,

Some of this code is indeed valuable to people, and should be preserved.
As a general practice I do not think we should rush into deleting old code
just because it is currently unlinked.   The specific high-value sections
I've noticed are:

1) IndexerKnownOrientation is a "pass" but it really should be implemented
properly, so as to index a new X-ray shot from a sample that is already
studied.
2) _print_normal_matrix() is clearly marked as a debugging tool.  It is
used by researchers, not end users, if and when the code needs to be
examined up close.
3) The compose() function serves the role of documentation, of an algorithm
that was once implemented in Python but is now pushed to C++.  However, the
Python version is far easier to understand, and so it is important to
preserve it as a reference.
4) In tile_generation.py, the picture_fast_slow_to_lon_lat() may be
important--I don't know.  My concern here is that there is a very
long-standing bug in the image viewer, in the code that provides pixel
intensity under the mouse.  The problem is that the coordinates are
calculated incorrectly, so the correct intensity is only displayed if the
mouse is in the upper-right quadrant of the pixel.  At least this is the
case for Jungfrau-NeXus images that I've recently examined.  It is possible
that the   picture_fast_slow_to_lon_lat() may be part of the bug fix.
Let's keep it around until we know for sure.

Nicholas K. Sauter, Ph. D.
Senior Scientist, Molecular Biophysics & Integrated Bioimaging Division
Lawrence Berkeley National Laboratory
1 Cyclotron Rd., Bldg. 33R0345
Berkeley, CA 94720
(510) 486-5713

phyy-nx:
@nksauter you may have meant this, but just FYI, the `pass` under StillsIndexerKnownOrientation is because the double derivation from IndexerKnownOrientation and StillsIndexer provides a full implementation of an indexer.

nksauter:
@phyy-nx I think your understanding is correct

graeme-winter:
@nksauter @phyy-nx two observations, neither meant to imply any criticism

@Anthchirp and others this week are looking into the dials integration which does include a good deal of dead wood / brush / call it what you will and have been trying to clean up some of the code to make understanding and fixing the algorithms - purely to provide some back story as to why these questions are being asked. 

If we have useful "documentation" functions or similar, or classes with derived behaviour with no obvious implementation, I would think adding tests would be extremely valuable to ensure that they don't go stale - my understanding (which could be wrong) is that these things above which were suggested for removal were untested. Already also this thread contains some useful signal - the `StillsIndexerKnownOrientation ` works by composition - tests would show that this works, and a documentation comment as suggested by @phyy-nx would remove any potential confusion regarding it's implementation. 

I know from experience e.g. in `xia2` that there can be a lot of dead wood to clear at times, so I think this housekeeping is worthwhile, even if the end game is we keep the code and update the tests and comments.

nksauter:
My comments stand regardless of adding tests.

Nicholas K. Sauter, Ph. D.
Senior Scientist, Molecular Biophysics & Integrated Bioimaging Division
Lawrence Berkeley National Laboratory
1 Cyclotron Rd., Bldg. 33R0345
Berkeley, CA 94720
(510) 486-5713

algorithms/indexing/stills_indexer.py
Anthchirp: :+1:

util/image_viewer/slip_viewer/tile_generation.py
Anthchirp: :+1:

Anthchirp:
David has kindly used the previously unused python implementation in a proper test, so it is no longer dead code. I have restored all other files mentioned.

Thank you all.

PR #1248, opened 2020-05-01
Fix propagation of best_unit_cell and d_min cutoffs in dials.scale/export_mtz

rjgildea:
A number of issues meant that the reflections output by dials.scale in the unmerged mtz file wouldn't necessarily be consistent with the resolution cutoff requested by the user. This PR includes a number of fixes to dials.scale and dials.export to address this (and tests to verify the new behaviour):

dials.export:
- optionally define best_unit_cell on command line
- otherwise get from recalculated_cell/median cell as per scaling
- use this best_unit_cell when calculating resolution cutoff

dials.scale:
- pass best_unit_cell from scaling when calling export_mtz
- respect command line best_unit_cell when calculating merging stats in scaling
- respect command line best_unit_cell when cutting resolution in scaling

### DIFF
--- a/util/export_mtz.py
+++ b/util/export_mtz.py
@@ -7,6 +7,7 @@ import time
 from collections import OrderedDict, Counter
 
 import dials.util.ext
+from dials.algorithms.scaling.scaling_library import determine_best_unit_cell
 from dials.array_family import flex
 from dials.util.version import dials_version
 from dials.util.filter_reflections import filter_reflection_table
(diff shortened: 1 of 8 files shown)


jbeilstenedmands:
Looks good to me from a quick glance, thanks for fixing.

util/export_mtz.py
```diff
@@ -558,11 +564,7 @@ def export_mtz(integrated_data, experiment_list, params):
         else:
             experiment.data["ROT"] = z
 
-    unit_cell = (
-        experiment_list[0].crystal.get_recalculated_unit_cell()
-        or experiment_list[0].crystal.get_unit_cell()
-    )
-    mtz_writer.add_crystal(params.mtz.crystal_name, unit_cell=unit_cell)
+    mtz_writer.add_crystal(params.mtz.crystal_name, unit_cell=best_unit_cell)
```
dagewa, line 567: Glad we no longer treat the first crystal as special :+1:

PR #1249, opened 2020-05-04
Map to minimum cell via joint reference settings

rjgildea:
A hopefully more numerically stable attempt at mapping to consistent minimum unit cells:

1. map all experiments to the reference setting
2. calculate the median unit cell for all the reference settings
3. calculate the change of basis op to minimum cell based on that median cell


rjgildea:
Depends on https://github.com/cctbx/dxtbx/pull/166

Issue #279, opened 2017-01-09
In dials.reciprocal_lattice_viewer if data are indexed please can we align on a*, b*, c* as option


graeme-winter:
i.e. have button a* / b* / c* which will allow you to align lattice accordingly; will then set b*; c* as close as possible to alignment with cardinal axes on image (i.e. x, y)

dagewa:
I'm going to say that now we have the option to view the reciprocal cell, manual alignment is not too tedious. This feature would still be a nice-to-have, but it may be better to add it to a `dials.rlv` replacement, possibly based on https://github.com/toastisme/dials_browser_rlv

Please feel free to re-open if you disagree.

Issue #214, opened 2016-09-20
Do we need phi= setting in reciprocal_lattice_viewer?


dagewa:
![rlv_funny](https://cloud.githubusercontent.com/assets/11502083/18666795/94e8c5f8-7f25-11e6-8e81-60afc76f216b.png)

See screenshot, i04 BAG training data, with Z range set to [0,5].

I expected to see the beam direction pointing in/out of the image, not across it. Eventually I realised that the datablock starts at 82 degrees, whilst the relps are always displayed at phi=0.

Do we need a phi= setter so that different positions in the scan can be explored in the rlv? Should this default to the start of the scan range, i.e. 82 degrees in this case?

I think this would enhance the utility of rlv as a teaching tool.

graeme-winter:
I.e should show w.r.t datum yes? If you have multiple sweeps I'd think that that is the desired behaviour. That said the positioning of the beam is misleading anyway since it in effect rotates through reciprocal space anyway as you rotate the sample. Hence again showing everything at datum. Beam at angle widget would give you option to reconstruct experiment at given frame number.

dagewa:
> Beam at angle widget would give you option to reconstruct experiment at given frame number

An alternative way of looking at this is that the beam vector display is generally unhelpful. In a 3D view the beam direction is swept out by the scan so showing a particular direction does not give much insight (but may lead to confusion). Also, I'm not sure if the beam direction display works properly when toggling "Show in crystal frame". Rather than trying to show an experiment at a given frame number, we could remove the beam vector display completely and insist that we are trying to show the scan, not its state at a particular frame?

jbeilstenedmands:
Decided: wontfix

PR #1259, opened 2020-05-12
Give option to show reciprocal lattice in crystal frame

graeme-winter:
Contributes to https://github.com/dials/dials/issues/279

Includes "Crystal frame" toggle, which shows the reciprocal lattice points and reciprocal lattice vectors in the crystal frame i.e. `B*h` rather than `U*B*h`

Illustrations:

Before:
Screenshot 2020-05-12 at 09 18 20

After:
Screenshot 2020-05-12 at 09 18 42

You have to zoom in closely to see differences between reciprocal lattices, but they _are_ there 🙂

### DIFF
--- a/array_family/flex_ext.py
+++ b/array_family/flex_ext.py
@@ -1151,7 +1151,9 @@ Found %s"""
                 self["xyzobs.mm.value"].set_selected(sel, centroid_position)
                 self["xyzobs.mm.variance"].set_selected(sel, centroid_variance)
 
-    def map_centroids_to_reciprocal_space(self, experiments, calculated=False):
+    def map_centroids_to_reciprocal_space(
+        self, experiments, calculated=False, crystal_frame=False
+    ):
         """Map mm/radian spot centroids to reciprocal space.
 
         Used to convert spot centroids provided in mm/radian units to reciprocal space
@@ -1160,6 +1162,8 @@ Found %s"""
 
         Args:
           experiments (dxtbx.model.ExperimentList): A list of experiments.
+          calculated (Bool): use calculated positions.
+          crystal_frame (Bool): return x, y, z positions as divided by U matrix
         """
 
         self["s1"] = cctbx.array_family.flex.vec3_double(len(self))
@@ -1171,6 +1175,11 @@ Found %s"""
                 sel_expt = self["imageset_id"] == i
             else:
                 sel_expt = self["id"] == i
+
+            u_mat = matrix.sqr((1, 0, 0, 0, 1, 0, 0, 0, 1))
+
+            if expt.crystal and crystal_frame:
+                u_mat = matrix.sqr(expt.crystal.get_U())
 
             for i_panel in range(len(expt.detector)):
                 sel = sel_expt & (panel_numbers == i_panel)
@@ -1198,7 +1207,9 @@ Found %s"""
                         .rotate_around_origin(rotation_axis, -rot_angle),
                     )
                     self["rlp"].set_selected(
-                        sel, tuple(fixed_rotation.inverse()) * self["rlp"].select(sel)
+                        sel,
+                        tuple((fixed_rotation * u_mat).inverse())
+                        * self["rlp"].select(sel),
                     )
                 else:
                     self["rlp"].set_selected(sel, S)
--- a/array_family/flex_ext.py
+++ b/array_family/flex_ext.py
@@ -1176,11 +1176,6 @@ Found %s"""
             else:
                 sel_expt = self["id"] == i
 
-            u_mat = matrix.sqr((1, 0, 0, 0, 1, 0, 0, 0, 1))
-
-            if expt.crystal and crystal_frame:
-                u_mat = matrix.sqr(expt.crystal.get_U())
-
             for i_panel in range(len(expt.detector)):
                 sel = sel_expt & (panel_numbers == i_panel)
                 if calculated:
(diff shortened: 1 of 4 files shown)


dagewa:
This looks useful, but I wonder if there is a better label than "Show in crystal frame"? A toggle or radio button between setting = UB and setting = B would be more explanatory (at least to me)

graeme-winter:
@dagewa I would think to ${average user} "Show in crystal frame" would make more sense than "setting=UB / B" but I welcome other viewpoints.

graeme-winter:
BTW I recognise that the "setting = B / UB" is much more precise and explicit a definition, no worries. I just think it will not be clear to our end user what this means 🙂

dagewa:
I do think this is a more expert option, the users of which should probably know what a setting matrix is. I'm also imagining adding another option in future to include **RUB** and then have an angle control. This would allow visualisation of the orientation of the crystal wrt to the beam at any image and thus address #214

graeme-winter:
I am not wedded to the name, but I would like to get the feature in. I personally think this is not an expert option - I could see _in situ_ general users using it...

graeme-winter:
This is particularly cute when you are looking at > 1 lattice on a sweep as they can be overlapped rather nicely - example with 2 - 

Screenshot 2020-05-12 at 14 40 27

becomes

Screenshot 2020-05-12 at 14 40 40

newsfragments/1259.feature
```diff
@@ -0,0 +1 @@
+Add an option to show lattice(s) in the crystal rather than laboratory frame.
```
Anthchirp, line 1:
```suggestion
dials.reciprocal_lattice_viewer: Add an option to show lattice(s) in the crystal rather than laboratory frame.
```

confusing otherwise

util/reciprocal_lattice/__init__.py
```diff
@@ -87,6 +90,16 @@ def load_models(self, experiments, reflections):
...
+            vecs = [
+                [
+                    matrix.col(l)
+                    for l in (100 * matrix.sqr(c.get_B()))
+                    .transpose()
+                    .as_list_of_lists()
+                ]
+                for c in crystals
+            ]
+            self.viewer.set_reciprocal_crystal_vectors(vecs)
```
dagewa, line 102: The similar block immediately above, which defines `vecs` the first time, should be removed after this is added

util/reciprocal_lattice/viewer.py
```diff
@@ -504,14 +522,24 @@ def DrawGL(self):
...
+                    color = self.palette[j]
+                    self.draw_cell(axes, color)
+            elif self.recip_latt_vectors:
+                for i, axes in enumerate(self.recip_latt_vectors):
+                    if self.settings.experiment_ids:
+                        if i not in self.settings.experiment_ids:
+                            continue
+                    j = (i + 1) % self.palette.size()
+                    color = self.palette[j]
+                    self.draw_cell(axes, color)
```
dagewa, line 542: These conditional branches are basically duplicates. It would be cleaner to use the conditional up front to decide which list to loop over, then do the loop outside the conditional

array_family/flex_ext.py
```diff
@@ -1172,6 +1176,11 @@ def map_centroids_to_reciprocal_space(self, experiments, calculated=False):
             else:
                 sel_expt = self["id"] == i
 
+            u_mat = matrix.sqr((1, 0, 0, 0, 1, 0, 0, 0, 1))
+
+            if expt.crystal and crystal_frame:
+                u_mat = matrix.sqr(expt.crystal.get_U())
+
```
dagewa, line 1183: I think it would be clearer to have
```
overall_rotation = fixed_rotation
if expt.crystal and crystal_frame:
    overall_rotation *= matrix.sqr(expt.crystal.get_U())
```
then use `overall_rotation` below

util/reciprocal_lattice/__init__.py
graeme-winter: Not true, I believe 

```
        if crystals:
            # the points are scaled by 100 so must do that here too
            vecs = [
                [
                    matrix.col(l)
                    for l in (100 * matrix.sqr(c.get_A()))
                    .transpose()
                    .as_list_of_lists()
                ]
                for c in crystals
            ]
            self.viewer.set_reciprocal_lattice_vectors(vecs)
            vecs = [
                [
                    matrix.col(l)
                    for l in (100 * matrix.sqr(c.get_B()))
                    .transpose()
                    .as_list_of_lists()
                ]
                for c in crystals
            ]
            self.viewer.set_reciprocal_crystal_vectors(vecs)
```

the first transforms the `A` matrices, sets them as one property, the second does the same for `B` - I _could_ refactor the code significantly to carry the U matrices to where the cell drawing is done, but this would be a major overhaul.

util/reciprocal_lattice/viewer.py
graeme-winter: Refactored loop in fcf4d8944f785ad80d65faf2ce10cd97827e0b0a - not convinced it is actually clearer - after 736c374856398a8d8a6b60c6a91e96483ca7bddd maybe it is.

array_family/flex_ext.py
graeme-winter: Refactored following this suggestion in 71137e5595aac1d7864b1eb8234e2b5077b70ff4

util/reciprocal_lattice/__init__.py
dagewa: Right, I see. At a quick glance I did not see that a different method was called after the listcomp :roll_eyes:

dagewa:
The tooltip doesn't appear to be working for me, on Linux :-(

graeme-winter:
> The tooltip doesn't appear to be working for me, on Linux :-(

Works on RHEL7 for me, and macOS:
Screenshot 2020-05-13 at 08 38 13

IMG_8290

Please excuse the Moire patterns - I could not do a screen grab without making the tooltip vanish 🙄

dagewa:
I believe it works! Just not on my laptop :shrug:

graeme-winter:
image

graeme-winter:
Please can I merge this? Was looking for exactly this feature a second ago. 

Or are the absence of tooltips on some systems a blocker?

dagewa:
Fine by me

graeme-winter:
N.B. also needed a985e1b49930f957a4d6685e83d2aa4b327485ff

PR #1263, opened 2020-05-14
Make some tools more robust to few reflections

benjaminhwilliams:
Some of the DIALS tools furthest downstream, such as `dials.scale`, `dials.symmetry` and `dials.merge`, rely on receiving a sufficient number of unique reflections to perform their calculations.  Existing checks that the input is valid are not explicit and are insufficient to catch all failure modes.  Changes here attempt to rationalise some behaviour in cases of very few input reflections, as from a very narrow wedge of a rotation data set.

Details of changes:
 * If `dials.algorithms.scaling.ScalingAlgorithm.remove_bad_data` results in no remaining data, raise an exception.
 * Fix the intended behaviour of `remove_bad_data` to correctly handle experiment UID strings.
 * In `dials.algorithms.scaling.Ih_table.IhTableBlock.setup_binner`, be less prescriptive in the choice of binner parameters in the case of few reflections.
 * If `modules.dials.algorithms.merging.merge.merge_and_truncate` does not receive enough reflections to calculate Wilson statistics for log output, catch the resulting exception and log an error-level message instead.
 * In the existing check for a similar failure to calculate and log the merging statistics, increase the logging level of the resuting message to error, for consistency.
 * In `dials.algorithms.symmetry.symmetry_base._normalise`, if an exception is encountered when trying to perform the normalisation, rather than raise a cryptic exception, simply log a warning and default to un-normalised intensities.
 * Correct a typo in `modules.dials.command_line.merge.phil_scope`.

### DIFF
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -217,14 +217,25 @@ def merge_and_truncate(params, experiments, reflections):
 
     if params.reporting.wilson_stats:
         if not intensities.space_group().is_centric():
-            wilson_scaling = data_statistics.wilson_scaling(
-                miller_array=intensities, n_residues=params.n_residues
-            )  # XXX default n_residues?
-            # Divert output through logger - do with StringIO rather than
-            # info_handle else get way too much whitespace in output.
-            out = StringIO()
-            wilson_scaling.show(out=out)
-            logger.info(out.getvalue())
+            try:
+                wilson_scaling = data_statistics.wilson_scaling(
+                    miller_array=intensities, n_residues=params.n_residues
+                )  # XXX default n_residues?
+            except (IndexError, RuntimeError) as e:
+                logger.error(
+                    "\n"
+                    "Error encountered during Wilson statistics calculation:\n"
+                    "Perhaps there are too few unique reflections.\n"
+                    "%s",
+                    e,
+                    exc_info=True,
+                )
+            else:
+                # Divert output through logger - do with StringIO rather than
+                # info_handle else get way too much whitespace in output.
+                out = StringIO()
+                wilson_scaling.show(out=out)
+                logger.info(out.getvalue())
 
     # Apply wilson B to give absolute scale?
 
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -248,9 +248,9 @@ def merge_and_truncate(params, experiments, reflections):
                 params.merging.use_internal_variance,
             )
         except DialsMergingStatisticsError as e:
-            logger.info(e)
+            logger.error(e, exc_info=True)
         else:
-            if params.merging.anomalous:
+            if params.merging.anomalous and anom_stats:
                 logger.info(make_merging_statistics_summary(anom_stats))
             else:
                 logger.info(make_merging_statistics_summary(stats))
--- a/algorithms/scaling/Ih_table.py
+++ b/algorithms/scaling/Ih_table.py
@@ -676,7 +676,8 @@ Not all rows of h_index_matrix appear to be filled in IhTableBlock setup."""
         relative_tolerance = 1e-6
         d_star_sq_max += span * relative_tolerance
         d_star_sq_min -= span * relative_tolerance
-        step = (d_star_sq_max - d_star_sq_min) / n_resolution_bins
+        # Avoid a zero-size step that would otherwise anger the d_star_sq_step binner.
+        step = max((d_star_sq_max - d_star_sq_min) / n_resolution_bins, 0.004)
 
         self.binner = ma.setup_binner_d_star_sq_step(
             auto_binning=False,
--- a/algorithms/scaling/algorithm.py
+++ b/algorithms/scaling/algorithm.py
@@ -209,6 +209,8 @@ class ScalingAlgorithm(Subject):
         start_time = time.time()
         self.scale()
         self.remove_bad_data()
+        if not self.experiments:
+            raise ValueError("All data sets have been rejected as bad.")
         self.scaled_miller_array = scaled_data_as_miller_array(
             self.reflections,
             self.experiments,
@@ -264,11 +266,9 @@ class ScalingAlgorithm(Subject):
         if removed_ids:
             logger.info("deleting removed datasets from memory: %s", removed_ids)
             expids = list(self.experiments.identifiers())
-            locs_in_list = []
-            for id_ in removed_ids:
-                locs_in_list.append(expids.index(id_))
+            locs_in_list = [expids.index(expid) for expid in removed_ids]
             self.experiments, self.reflections = select_datasets_on_ids(
-                self.experiments, self.reflections, exclude_datasets=removed_ids
+                self.experiments, self.reflections, exclude_datasets=locs_in_list
             )
         # also remove negative scales (or scales below 0.001)
         n = 0
--- a/algorithms/scaling/algorithm.py
+++ b/algorithms/scaling/algorithm.py
@@ -291,7 +291,7 @@ class ScalingAlgorithm(Subject):
                 self.params.output.use_internal_variance,
             )
         except DialsMergingStatisticsError as e:
-            logger.info(e)
+            logger.warning(e, exc_info=True)
 
     def finish(self):
         """Save the experiments json and scaled pickle file."""
--- a/algorithms/scaling/scaling_library.py
+++ b/algorithms/scaling/scaling_library.py
@@ -399,7 +399,8 @@ def merging_stats_from_scaled_array(
 
     if scaled_miller_array.is_unique_set_under_symmetry():
         raise DialsMergingStatisticsError(
-            "Dataset contains no equivalent reflections, merging statistics cannot be calculated."
+            "Dataset contains no equivalent reflections, merging statistics "
+            "cannot be calculated."
         )
     try:
         result = iotbx.merging_statistics.dataset_statistics(
@@ -411,28 +412,43 @@ def merging_stats_from_scaled_array(
             use_internal_variance=use_internal_variance,
             cc_one_half_significance_level=0.01,
         )
-        if anomalous:
-            intensities_anom = scaled_miller_array.as_anomalous_array()
-            intensities_anom = intensities_anom.map_to_asu().customized_copy(
-                info=scaled_miller_array.info()
-            )
-            anom_result = iotbx.merging_statistics.dataset_statistics(
-                i_obs=intensities_anom,
-                n_bins=n_bins,
-                anomalous=True,
-                sigma_filtering=None,
-                cc_one_half_significance_level=0.01,
-                eliminate_sys_absent=False,
-                use_internal_variance=use_internal_variance,
-            )
-        else:
-            anom_result = False
     except (RuntimeError, Sorry) as e:
         raise DialsMergingStatisticsError(
             "Error encountered during merging statistics calculation:\n%s" % e
         )
-    else:
-        return result, anom_result
+
+    anom_result = None
+
+    if anomalous:
+        intensities_anom = scaled_miller_array.as_anomalous_array()
+        intensities_anom = intensities_anom.map_to_asu().customized_copy(
+            info=scaled_miller_array.info()
+        )
+        if intensities_anom.is_unique_set_under_symmetry():
+            logger.warning(
+                "Anomalous dataset contains no equivalent reflections, anomalous "
+                "merging statistics cannot be calculated."
+            )
+        else:
+            try:
+                anom_result = iotbx.merging_statistics.dataset_statistics(
+                    i_obs=intensities_anom,
+                    n_bins=n_bins,
+                    anomalous=True,
+                    sigma_filtering=None,
+                    cc_one_half_significance_level=0.01,
+                    eliminate_sys_absent=False,
+                    use_internal_variance=use_internal_variance,
+                )
+            except (RuntimeError, Sorry) as e:
+                logger.warning(
+                    "Error encountered during anomalous merging statistics "
+                    "calculation:\n%s",
+                    e,
+                    exc_info=True,
+                )
+
+    return result, anom_result
 
 
 def intensity_array_from_cif_file(cif_file):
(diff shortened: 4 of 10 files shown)


algorithms/scaling/Ih_table.py
```diff
@@ -678,12 +678,16 @@ def setup_binner(self, unit_cell, space_group, n_resolution_bins):
...
+        try:
+            self.binner = ma.setup_binner_d_star_sq_step(
+                auto_binning=False,
+                d_max=uctbx.d_star_sq_as_d(d_star_sq_max),
+                d_min=uctbx.d_star_sq_as_d(d_star_sq_min),
+                d_star_sq_step=step,
+            )
+        # Catch the case of too few unique reflections to perform the binning.
+        except AssertionError:
+            self.binner = ma.setup_binner_d_star_sq_step()
```
rjgildea, line 690: Can you explain what is the difference between this call and the one above (that has failed if we reach this point), i.e. why does this work when the one above doesn't? And what is this more guaranteed to not fail?

algorithms/merging/merge.py
```diff
@@ -216,15 +216,22 @@ def merge_and_truncate(params, experiments, reflections):
...
+                # Divert output through logger - do with StringIO rather than
+                # info_handle else get way too much whitespace in output.
+                out = StringIO()
+                wilson_scaling.show(out=out)
+                logger.info(out.getvalue())
+        except RuntimeError:
+            logger.error(
+                "\n"
+                "Error encountered during Wilson statistics calculation:\n"
+                "Perhaps there are too few unique reflections.\n"
```
rjgildea, line 233: Should we also log the actual error encountered here?

algorithms/merging/merge.py
```diff
@@ -237,7 +244,7 @@ def merge_and_truncate(params, experiments, reflections):
                 params.merging.use_internal_variance,
             )
         except DialsMergingStatisticsError as e:
-            logger.info(e)
+            logger.error(e)
```
Anthchirp, line 247:
```suggestion
            logger.error(e, exc_info=True)
```
pass on trace information

algorithms/scaling/algorithm.py
```diff
@@ -264,11 +266,9 @@ def remove_bad_data(self):
         if removed_ids:
             logger.info("deleting removed datasets from memory: %s", removed_ids)
             expids = list(self.experiments.identifiers())
-            locs_in_list = []
-            for id_ in removed_ids:
-                locs_in_list.append(expids.index(id_))
+            locs_in_list = [expids.index(expid) for expid in expids]
             self.experiments, self.reflections = select_datasets_on_ids(
-                self.experiments, self.reflections, exclude_datasets=removed_ids
+                self.experiments, self.reflections, exclude_datasets=locs_in_list
```
rjgildea, line 271: `removed_ids` no longer appears to be used? Is this definitely correct? Is this code not already tested?

algorithms/merging/merge.py
```diff
@@ -216,15 +216,22 @@ def merge_and_truncate(params, experiments, reflections):
...
+            if not intensities.space_group().is_centric():
+                wilson_scaling = data_statistics.wilson_scaling(
+                    miller_array=intensities, n_residues=params.n_residues
+                )  # XXX default n_residues?
+                # Divert output through logger - do with StringIO rather than
+                # info_handle else get way too much whitespace in output.
+                out = StringIO()
+                wilson_scaling.show(out=out)
+                logger.info(out.getvalue())
+        except RuntimeError:
```
Anthchirp, line 229: RuntimeError is very broad. Do you know whether the error is raised in the `.wilson_scaling()` or the `.show()` method, and is it possible to not catch errors in the other?

algorithms/merging/merge.py
Anthchirp, line 233:
```suggestion
                "Perhaps there are too few unique reflections.\n",
                exc_info=True
```
pass on trace information

algorithms/symmetry/__init__.py
```diff
@@ -156,16 +156,27 @@ def _normalise(self, method):
...
+                    normalised_intensities = normalised_intensities.concatenate(
+                        normalise(intensities)
+                    )
+        except RuntimeError:
+            logger.warning(
+                "A problem occurred when trying to normalise the intensities by the "
+                "%s method.\n"
+                "There may be too few unique reflections. Resorting to using "
+                "un-normalised intensities instead.",
+                method,
```
rjgildea, line 176: Again, would it be useful to also report the specific error that was encountered when normalising the intensities?

Anthchirp, line 176:
```suggestion
                method,
                exc_info=True,
```

algorithms/symmetry/__init__.py
```diff
@@ -156,16 +156,27 @@ def _normalise(self, method):
...
+                logger.info("\n" + "-" * 80 + "\n")
+                logger.info("Normalising intensities for dataset %i\n" % (i + 1))
+                intensities = self.intensities.select(self.dataset_ids == i)
+                if i == 0:
+                    normalised_intensities = normalise(intensities)
+                else:
+                    normalised_intensities = normalised_intensities.concatenate(
+                        normalise(intensities)
+                    )
+        except RuntimeError:
```
Anthchirp, line 170: same as above - I assume we expect the `RuntimeError` to happen in `normalise()` rather than `.select` or `.concatenate` (or `flex.max`)

benjaminhwilliams:
@rjgildea, regarding keeping some degree of commit history rather than a full squash-merge, I agree.  Once this PR gets approval, I'll force-push a picked/squashed version with different changes kept separate and ask for approval again.

algorithms/scaling/Ih_table.py
benjaminhwilliams: The key thing is the default value for the parameter `auto_binning` is `True`.  If `auto_binning` is `False` and all reflections have equal _d_, such that `d_star_sq_min == d_star_sq_max` and `step == 0`, we fall foul of the `assert` at https://github.com/cctbx/cctbx_project/blob/565ab39cd9fbc222191526fe17ec5ca5cb692471/cctbx/miller/__init__.py#L1719.

I suppose an alternative would simply be to set `step` to have a minimum value of something greater than zero (which is what `auto_binning` is doing).  As to whether we should even get this far in considering such a pathological data set, or whether we should catch the issue further upstream, that requires more in-depth knowledge of what's going on in `dials.scale` than I possess.

algorithms/merging/merge.py
benjaminhwilliams: Thanks, I didn't know you could do that.

algorithms/scaling/Ih_table.py
jbeilstenedmands: As far as I remember there's nothing fancy that should be going on here, simply binning some reflection table data using standard methods, indeed I think the main, max, step approach was used to get around potential issues with the autobinning methods when there aren't many reflections present. Also I'm surprised that this code is called within dials.scale if the data is all reflections with the same d, this ought to be caught at a higher level I would think.

rjgildea: Having looked at this code more carefully, I don't think that wrapping this call to `ma.setup_binner_d_star_sq_step` in a try/except is the correct approach. If the error is because we are passing nonsense input (i.e. `d_star_sq_step=0`) then we should figure out why this is the case, and either ensure we don't pass nonsense values to `ma.setup_binner_d_star_sq_step` or raise a more appropriate error (perhaps the code should never even get to this point as @jbeilstenedmands suggests).

algorithms/scaling/algorithm.py
benjaminhwilliams: Ugh, sorry, that's a typo.  Thanks for spotting.

This code doesn't seem to be adequately tested.  It fails for non-integer experiment IDs, such as the new-style UID strings.

algorithms/merging/merge.py
benjaminhwilliams: Yes, the RuntimeError is raised in `.wilson_scaling`, which I take to be off limits, since it's in `mmtbx`.  Specifically it's a `RuntimeError: mean() argument is an empty array`, occurring at https://github.com/cctbx/cctbx_project/blob/8f1ca5d1c172015c24030314f9599d4b5934680a/mmtbx/scaling/absolute_scaling.py#L798.  Obviously, that should be a `ValueError` or almost anything other than a `RuntimeError`, but I don't have any control over that.

But you're right that I should be keeping the try/except tighter around the problem call.

algorithms/scaling/algorithm.py
jbeilstenedmands: I can imagine this was a genuine bug introduced by the UID changes, I would say this now looks correct.

algorithms/symmetry/__init__.py
benjaminhwilliams: Quite right.  Thank you.

algorithms/scaling/Ih_table.py
benjaminhwilliams: Well the question of what this ought to be doing upstream is beyond my current level of understanding of the scaling algorithms, so I've just taken a crude approach to ensure `step` is always greater than zero.

rjgildea: OK, it's hard to comment on how/why it got to this point without the context of the dials.scale job you were running.

benjaminhwilliams:
I'm afraid I've given up on the idea of unpicking every place where problems occur and building unit tests for each.  Instead I've just built some simple integration tests.  Let me know if you'd like me to revisit the unit tests, especially since it might avoid things like the previously undetected bug in `dials.algorithms.scaling.algorithm.ScalingAlgorithm.remove_bad_data`.

algorithms/scaling/algorithm.py
```diff
@@ -291,7 +291,7 @@ def calculate_merging_stats(self):
                 self.params.output.use_internal_variance,
             )
         except DialsMergingStatisticsError as e:
-            logger.info(e)
+            logger.warning(e)
```
Anthchirp, line 294:
```suggestion
            logger.warning(e, exc_info=True)
```

algorithms/merging/merge.py
```diff
@@ -217,14 +217,23 @@ def merge_and_truncate(params, experiments, reflections):
...
-            logger.info(out.getvalue())
+            try:
+                wilson_scaling = data_statistics.wilson_scaling(
+                    miller_array=intensities, n_residues=params.n_residues
+                )  # XXX default n_residues?
+            except (IndexError, RuntimeError):
+                logger.error(
+                    "\n"
+                    "Error encountered during Wilson statistics calculation:\n"
+                    "Perhaps there are too few unique reflections.\n",
```
Anthchirp, line 228: may want to include `str(e)` in message?

algorithms/scaling/scaling_library.py
```diff
@@ -411,28 +412,41 @@ def merging_stats_from_scaled_array(
...
+                    anomalous=True,
+                    sigma_filtering=None,
+                    cc_one_half_significance_level=0.01,
+                    eliminate_sys_absent=False,
+                    use_internal_variance=use_internal_variance,
+                )
+            except (RuntimeError, Sorry) as e:
+                logger.warning(
+                    "Error encountered during anomalous merging statistics "
+                    "calculation:\n%s" % e
```
Anthchirp, line 446:
```suggestion
                    "calculation:\n%s" % e, exc_info=True
```

algorithms/scaling/test_scale.py
```diff
@@ -765,3 +762,53 @@ def test_scale_cross_validate(dials_data, tmpdir, mode, parameter, parameter_val
...
+    ).extract()
+    # Does what it says on the tin.  Run scaling.
+    scaled_expt, scaled_refl = scale.run_scaling(scale_params, experiments, reflections)
+
+    # Get and use the default parameters for dials.merge.
+    merge_params = merge.phil_scope.fetch(source=phil.parse("")).extract()
+    # Run dials.merge on the scaling output.
+    merge.merge_data_to_mtz(merge_params, scaled_expt, [scaled_refl])
+
+    # Get amd use the default parameters for dials.report, suppressing HTML output.
```
Anthchirp:
```suggestion
    # Get and use the default parameters for dials.report, suppressing HTML output.
```

benjaminhwilliams: Snap.

algorithms/scaling/test_scale.py
```diff
@@ -765,3 +762,53 @@ def test_scale_cross_validate(dials_data, tmpdir, mode, parameter, parameter_val
     command = ["dials.scale", refl, expt] + extra_args
     result = procrunner.run(command, working_directory=tmpdir)
     assert not result.returncode and not result.stderr
+
+
+def test_few_reflections(dials_data):
```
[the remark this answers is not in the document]
benjaminhwilliams: Sorry, a hangover from the work in progress, before I discovered there needn't be any files created.  I've corrected the docstring.

algorithms/merging/merge.py
benjaminhwilliams: Done.

graeme-winter:
Ran through a standard data set with this change set, made no difference to a `xia2.small_molecule` run - literally the text output is identical modulo directory. 

The reflections coming out are identical 🙂

```
Grey-Area dials :( [few-reflections] $ diff /tmp/few/DataFiles/AUTOMATIC_DEFAULT_scaled_unmerged.sca /tmp/nat/DataFiles/AUTOMATIC_DEFAULT_scaled_unmerged.sca
```

gives nothing => from that perspective safe

benjaminhwilliams:
Great, thanks.  Would you be happy for me to merge?

benjaminhwilliams:
In the meantime, I'll work on restructuring the commits to be a little more minimal, then we can break with convention and merge without squashing, as recommended by @rjgildea, since there is value in being able to revert logically distinct components of this PR.

benjaminhwilliams:
@rjgildea indicated offline that he wanted to take a look before I merged...

PR #1264, opened 2020-05-14
Remove cache from expt/refl phil parsers

rjgildea:
The cache meant that if an experiment list/reflection table was read in once, and then modified, that if it was then re-read in later by another phil parser in the same process that the modified object was returned, rather than the original, unmodified, object which could lead to potentially surprising behaviour.

See https://github.com/xia2/xia2/issues/454

### DIFF
--- a/test/test_phil.py
+++ b/test/test_phil.py
@@ -39,12 +39,3 @@ def test(ExperimentListFactory, dials_data):
     assert isinstance(params.input.reflections2.data, flex.reflection_table)
     # Check we had the correct calls made
     assert ExperimentListFactory.from_json_file.call_args[0] == (experiments_path,)
-
-    # We now need to clear the phil converter caches
-    # as they are polluted by MagicMock() objects
-    for conv in (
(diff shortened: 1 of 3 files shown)


Anthchirp:
Interesting.
`git blame` flagged this commit/revert:
https://github.com/dials/dials/commit/3d65105af9658f41d0b010d6be05e77e1cef87ae
Apparently I tried to remove this 2 weeks ago and ended up reverting it back in because the cache is used to indirectly pass something from one caller to the next using magic values.

rjgildea:
Gah, it's not that simple, because the option parser appends the filename `"<image files>"` to the `ExperimentListConverters.cache`:
https://github.com/dials/dials/blob/master/util/options.py#L286
which is then relied on to retrieve the filename when the diff_phil is printed:
https://github.com/dials/dials/blob/8f91bc5e383268bebef799e35574c5b5b456c753/util/phil.py#L48-L49

rjgildea:
All dials tests pass for me on mac with these changes.

PR #1279, opened 2020-06-02
export_mtz: don't change basis to reference setting

### LINKED ISSUE #1278: dials.export makes a mess of MTZ file when space group is I2
[integrated.zip](https://github.com/dials/dials/files/4716566/integrated.zip)

Data in this zip file: 

```
dials.symmetry integrated.* 
dials.scale symmetrized.*
dials.export scaled.refl scaled.expt 
xia2.to_shelx scaled.mtz tst CHNO
shelxt tst
```

shelxt _does not solve_ the structure

Using

```
dials.symmetry best_monoclinic_beta=0 integrated.* 
dials.scale symmetrized.*
dials.export scaled.refl scaled.expt
xia2.to_shelx scaled.mtz tst CHNO
shelxt tst
```

it works very nicely indeed. Working in P1 i.e. scaling the original integrated data also works well too. The export of the ins file seems sensible:

```
CELL 0.51659 9.6953  8.7601  19.3432  90.000 98.036 90.000
ZERR 4 0. 0. 0. 0. 0. 0.
LATT -2
SYMM -x,y,-z
```

for the case which does not work in I2

```
CELL 0.51659 20.3893  8.7601  9.6953  90.000 110.053 90.000
ZERR 4 0. 0. 0. 0. 0. 0.
LATT -7
SYMM -x,y,-z
```

for the case which does work in C2 - these are (according to the documentation) correct 

I don't know if this is a bug or a feature, or something which needs documenting or what, but something != right here.

rjgildea:
This can lead to inconsistencies between the batches, miller indices and the overall mtz space group and unit cell.

Fixes #1278

### DIFF
--- a/newsfragments/1279.bugfix
+++ b/newsfragments/1279.bugfix
@@ -1,0 +1,1 @@
+Bug fix for dials.export format=mtz when given input in non-reference setting (e.g. I2, or primitive setting of a centred cell). Previously the batches, miller indices and overall mtz space group and unit cell were inconsistent.
--- a/test/command_line/test_export.py
+++ b/test/command_line/test_export.py
@@ -8,9 +8,11 @@ import pytest
 from cctbx import uctbx
 from dials.array_family import flex
 from dials.util.multi_dataset_handling import assign_unique_identifiers
+from dxtbx.model import ExperimentList
 from dxtbx.serialize.load import _decode_dict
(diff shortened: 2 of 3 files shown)


rjgildea:
This depends on https://github.com/dials/data/pull/190 and https://github.com/dials/data-files/pull/17 for the new test.

Issue #1281, opened 2020-06-03
dials.reindex: trap non-rational reindex matrix gracefully


graeme-winter:
Related to #1280 

```
Traceback (most recent call last):
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/reindex.py", line 358, in <module>
    run(sys.argv[1:])
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/reindex.py", line 319, in run
    experiments, change_of_basis_op, space_group=space_group
  File "/Users/graeme/svn/cctbx/build/../modules/dials/command_line/reindex.py", line 153, in reindex_experiments
    cryst_reindexed = cryst_reindexed.change_basis(cb_op)
RuntimeError: cctbx Internal Error: /Users/graeme/svn/cctbx/modules/cctbx_project/cctbx/sgtbx/rot_mx.cpp(19): Unsuitable value for rational rotation matrix.
```

should be trapped gracefully, "users should not see a stack trace" rule

graeme-winter:
```
Grey-Area work :( $ dials.reindex change_of_basis_op=a+2*b+c,-b+c,a scaled.{expt,refl}
The following parameters have been modified:

change_of_basis_op = "a+2*b+c,-b+c,a"
input {
  experiments = scaled.expt
  reflections = scaled.refl
}

Unsuitable change of basis: a+2*b+c,-b+c,a
```

graeme-winter:
Turns out this broke a test - reverting change for revisiting

PR #1282, opened 2020-06-03
Trap bad cb_op in dials.reindex

graeme-winter:
Fixes https://github.com/dials/dials/issues/1281

I tried pre-validating but getting the circumstances right is nigh on impossible, so EAFP just trap the RuntimeError as it happens and clean up the message for the end user.


graeme-winter:
Looking for a test data set now to add in

graeme-winter:
Waiting on https://github.com/dials/data/pull/190 - will add test given that

rjgildea:
This one should be relatively straightforward to add a test for, as you just need to test an invalid cb_op reindex and check that it raises a sys.exit rather than RuntimeError 🙂

graeme-winter:
@rjgildea yes your insulin example which you have waiting in the wings will be perfect as that was what I was looking at when the error occurred...

ndevenish:
I've written a test for this based on the discussion; It's probably worth looking over to check this is what you intended.

The previous issue mentioned that your patch caused a failed test, but I'm not sure which one?

command_line/reindex.py
```diff
@@ -311,19 +312,17 @@ def run(args):
...
             space_group = space_group.group()
-        experiments = reindex_experiments(
-            experiments, change_of_basis_op, space_group=space_group
-        )
+        try:
+            experiments = reindex_experiments(
+                experiments, change_of_basis_op, space_group=space_group
+            )
+        except RuntimeError as e:
+            sys.exit("%s" % str(e).split(":")[-1].strip())
```
ndevenish, line 324: It seems like prefixing "Error" would be slightly more enlightening - or is this a pattern we are avoiding?
```suggestion
            sys.exit("Error: %s" % str(e).split(":")[-1].strip())
```

test/command_line/test_reindex.py
```diff
@@ -238,3 +239,19 @@ def test_reindex_experiments():
         )
         # Check that the scan-varying A matrices have been copied as well
         assert cryst.num_scan_points == n_scan_points
+
+
+def test_reindex_cob_op_exit(dials_data):
```
rjgildea, line 244:
```suggestion
def test_reindex_cb_op_exit(dials_data):
```
`cb_op` is the conventional shorthand for `change_of_basis_op` in cctbx-land

command_line/reindex.py
```diff
@@ -311,19 +312,17 @@ def run(args):
...
         if space_group is not None:
             space_group = space_group.group()
-        experiments = reindex_experiments(
-            experiments, change_of_basis_op, space_group=space_group
-        )
+        try:
+            experiments = reindex_experiments(
+                experiments, change_of_basis_op, space_group=space_group
+            )
+        except RuntimeError as e:
```
rjgildea, line 323: Do we want to catch _any_ `RuntimeError`, or just `"Unsuitable value for rational rotation matrix"` errors? If the latter, we could explicitly check the error message and re-raise the error if it is something different.

ndevenish: Hmm, this is a good point, this would cause potentially unhelpful "Error" messages in other exception cases. Maybe we should throw/catch a special exception at the source rather than a RuntimeError? Alternatively, raise the `sys.exit()` where it happens. That case, however, makes it harder to use from other processes that might _catch_ the error.

Some combination of both?

ndevenish: It turns out we do have form doing this https://github.com/dials/dials/blob/c5e652f9b01cfd666db9a51171d24dae72d76534/algorithms/indexing/symmetry.py#L106-L108
so given we can't change the exception origins (in rstbx C++) I'm reasonably happy catching the error explicitly.

ndevenish:
Okay, now it only catches this error, and gives a slightly more useful message:
```
 $ dials.reindex change_of_basis_op=a+2*b+c,-b+c,a  /Users/nickd/dials/data/dials_data/insulin_processed/scaled.*
...
Error: Unsuitable value for rational rotation matrix. Is your change_of_basis_op valid?
```

PR #1299, opened 2020-06-11
Dependency updates: msgpack

Anthchirp:
Use conda-forge msgpack C sources rather than a separate download.
Update msgpack and msgpack-c versions as far as possible.


Anthchirp:
I believe the msgpack-python 0.6.2->1.0 API changes are not relevant to dxtbx/dials/xia2.

The only place I could see them mattering is in dlstbx where they are used in conjunction with kafka and TOM. May need looking at before this can be merged.

Anthchirp:
Travis builds should fix themselves within 24 hours

PR #1302, opened 2020-06-12
Improvements to image stacking

graeme-winter:
Means the images are on a sensible scale when stacked so look OK

### DIFF
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -812,7 +812,7 @@ class SpotFrame(XrayFrame):
             )
 
     def stack_images(self):
-        mode = self.params.stack_type
+        mode = self.params.stack_mode
         if self.params.stack_images > 1:
             image = self.pyslip.tiles.raw_image
             image_data = image.get_image_data()
@@ -1898,12 +1898,12 @@ class SpotSettingsPanel(wx.Panel):
         grid = wx.FlexGridSizer(cols=2, rows=1, vgap=0, hgap=0)
         txt1 = wx.StaticText(self, -1, "Stack type:")
         grid.Add(txt1, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
-        self.stack_types = ["max", "mean", "sum"]
-        self.stack_type_ctrl = wx.Choice(self, -1, choices=self.stack_types)
-        self.stack_type_ctrl.SetSelection(
-            self.stack_types.index(self.params.stack_type)
+        self.stack_modes = ["max", "mean", "sum"]
+        self.stack_mode_ctrl = wx.Choice(self, -1, choices=self.stack_modes)
+        self.stack_mode_ctrl.SetSelection(
+            self.stack_modes.index(self.params.stack_mode)
         )
-        grid.Add(self.stack_type_ctrl, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
+        grid.Add(self.stack_mode_ctrl, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
         s.Add(grid)
 
         # Image type choice
@@ -2055,7 +2055,7 @@ class SpotSettingsPanel(wx.Panel):
 
         self.Bind(wx.EVT_CHOICE, self.OnUpdateZoomLevel, self.zoom_ctrl)
         self.Bind(wx.EVT_CHOICE, self.OnUpdateImage, self.image_type_ctrl)
-        self.Bind(wx.EVT_CHOICE, self.OnUpdateImage, self.stack_type_ctrl)
+        self.Bind(wx.EVT_CHOICE, self.OnUpdateImage, self.stack_mode_ctrl)
         self.Bind(wx.EVT_CHOICE, self.OnUpdate, self.color_ctrl)
         self.Bind(wx.EVT_CHECKBOX, self.OnUpdate, self.resolution_rings_ctrl)
         self.Bind(wx.EVT_CHECKBOX, self.OnUpdate, self.ice_rings_ctrl)
@@ -2085,8 +2085,8 @@ class SpotSettingsPanel(wx.Panel):
             self.settings.image_type = self.image_types[
                 self.image_type_ctrl.GetSelection()
             ]
-            self.params.stack_type = self.stack_types[
-                self.stack_type_ctrl.GetSelection()
+            self.params.stack_mode = self.stack_modes[
+                self.stack_mode_ctrl.GetSelection()
             ]
             self.settings.show_resolution_rings = self.resolution_rings_ctrl.GetValue()
             self.settings.show_ice_rings = self.ice_rings_ctrl.GetValue()
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -830,6 +830,11 @@ class SpotFrame(XrayFrame):
                 for j, rd in enumerate(image_data):
                     rd += image_data_i[j]
 
+            # /= sum_images to put on consistent scale with single image
+            # so that -1 etc. handled correctly
+
+            image_data = tuple(i / self.params.sum_images for i in image_data)
+
             # Don't show summed images with overloads
             self.pyslip.tiles.set_image_data(image_data, show_saturated=False)
 
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -840,7 +840,6 @@ class SpotFrame(XrayFrame):
                 for j, rd in enumerate(image_data):
                     data = image_data_i[j]
                     if mode == "max":
-                        rd = rd.as_double()
                         sel = data > rd
                         rd = rd.as_1d().set_selected(sel.as_1d(), data.as_1d())
                     else:
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -236,7 +236,7 @@ class SpotFrame(XrayFrame):
 
         self.stack = PhilIntCtrl(self.toolbar, -1, name="stack", size=(65, -1))
         self.stack.SetMin(1)
-        self.stack.SetValue(1)
+        self.stack.SetValue(self.params.stack_images)
         self.toolbar.AddControl(self.stack)
         self.Bind(EVT_PHIL_CONTROL, self.OnStack, self.stack)
 
@@ -307,8 +307,8 @@ class SpotFrame(XrayFrame):
 
     def OnStack(self, event):
         value = self.stack.GetPhilValue()
-        if value != self.params.sum_images:
-            self.params.sum_images = value
+        if value != self.params.stack_images:
+            self.params.stack_images = value
             self.reload_image()
 
     def GetBoxCorners(self, layer, p1, p2):
@@ -813,7 +813,7 @@ class SpotFrame(XrayFrame):
 
     def stack_images(self):
         mode = self.settings.stack_type
-        if self.params.sum_images > 1:
+        if self.params.stack_images > 1:
             image = self.pyslip.tiles.raw_image
             image_data = image.get_image_data()
             if not isinstance(image_data, tuple):
@@ -824,7 +824,7 @@ class SpotFrame(XrayFrame):
             ).index
             imageset = self.image_chooser.GetClientData(i_frame).image_set
 
-            for i in range(1, self.params.sum_images):
+            for i in range(1, self.params.stack_images):
                 if (i_frame + i) >= len(imageset):
                     break
                 image_data_i = imageset[i_frame + i]
@@ -836,10 +836,10 @@ class SpotFrame(XrayFrame):
                         sel = data > rd
                         rd = rd.as_1d().set_selected(sel.as_1d(), data.as_1d())
 
-            # /= sum_images to put on consistent scale with single image
+            # /= stack_images to put on consistent scale with single image
             # so that -1 etc. handled correctly (mean mode)
             if mode == "mean":
-                image_data = tuple(i / self.params.sum_images for i in image_data)
+                image_data = tuple(i / self.params.stack_images for i in image_data)
 
             # Don't show summed images with overloads
             self.pyslip.tiles.set_image_data(image_data, show_saturated=False)
@@ -909,7 +909,7 @@ class SpotFrame(XrayFrame):
     def _calculate_dispersion_debug(self, image):
         request = {}
         request["index"] = image.index
-        request["sum"] = self.params.sum_images
+        request["sum"] = self.params.stack_images
         request["gain_value"] = self.settings.gain
         request["nsigma_b"] = self.settings.nsigma_b
         request["nsigma_s"] = self.settings.nsigma_s
@@ -1173,7 +1173,7 @@ class SpotFrame(XrayFrame):
                 )
 
         self.stack_images()
-        # if self.params.sum_images == 1:
+        # if self.params.stack_images == 1:
         # self.show_filters()
         if self.settings.show_threshold_pix:
             image = self.pyslip.tiles.raw_image
@@ -1360,7 +1360,7 @@ class SpotFrame(XrayFrame):
                 bbox = ref_list["bbox"]
                 x0, x1, y0, y1, z0, z1 = bbox.parts()
                 # ticket #107
-                n = self.params.sum_images - 1
+                n = self.params.stack_images - 1
                 bbox_sel = ~((i_frame >= z1) | ((i_frame + n) < z0))
                 selected = ref_list.select(bbox_sel)
                 for reflection in selected.rows():
@@ -1471,7 +1471,7 @@ class SpotFrame(XrayFrame):
                         centroid = reflection["xyzobs.px.value"]
                         # ticket #107
                         if centroid[2] >= i_frame and centroid[2] <= (
-                            i_frame + self.params.sum_images
+                            i_frame + self.params.stack_images
                         ):
                             x, y = map_coords(
                                 centroid[0], centroid[1], reflection["panel"]
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -1765,7 +1765,7 @@ class SpotSettingsPanel(wx.Panel):
             self,
             value=self.settings.brightness,
             min=1,
-            max=500,
+            max=1000,
             name="brightness",
             style=wx.TE_PROCESS_ENTER,
         )
@@ -1775,7 +1775,7 @@ class SpotSettingsPanel(wx.Panel):
             self, -1, size=(150, -1), style=wx.SL_AUTOTICKS | wx.SL_LABELS
         )
         self.brightness_ctrl.SetMin(1)
-        self.brightness_ctrl.SetMax(500)
+        self.brightness_ctrl.SetMax(1000)
         self.brightness_ctrl.SetValue(self.settings.brightness)
         self.brightness_ctrl.SetTickFreq(25)
         box.Add(self.brightness_ctrl, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -812,7 +812,7 @@ class SpotFrame(XrayFrame):
             )
 
     def stack_images(self):
-        mode = self.settings.stack_type
+        mode = self.params.stack_type
         if self.params.stack_images > 1:
             image = self.pyslip.tiles.raw_image
             image_data = image.get_image_data()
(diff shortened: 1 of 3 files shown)


graeme-winter:
Before:

Screenshot 2020-06-12 at 16 05 13

adding 10 images => 

Screenshot 2020-06-12 at 16 05 00


After:

adding 10 images gives:

Screenshot 2020-06-12 at 16 05 46


which I think is much more what you would want?

dagewa:
The function name `sum_images` is not really appropriate if the result is actually to calculate a mean

graeme-winter:
@dagewa valid - but don't want to rename it to `compute_mean_over_N_images` or something? 🤔

graeme-winter:
Set max brightness to 1000 to reproduce feel of stacking 10 images as before

graeme-winter:
While I am thinking about this, wonder if we should have a show-max mode as well? 🤔  so maximum pixel over the range rather than mean value?

dagewa:
Could definitely be useful - has been used for virtual powder patterns I think

util/image_viewer/spotfinder_frame.py
```diff
@@ -811,7 +811,7 @@ def draw_resolution_rings(self, unit_cell=None, space_group=None):
                 update=False,
             )
 
-    def sum_images(self):
+    def stack_images(self):
         if self.params.sum_images > 1:
```
rjgildea, line 815: Would it make sense to also change the command line parameter to `stack_images` (possibly keeping around `sum_images` with a deprecation warning for the time being)?

ndevenish:
I worry that if summing over a range to try to find weak signal, or signal only present on a few, that this might hide it? A "max pixel" mode might be appropriate. Maybe this does deserve a setting, but maybe that's just avoiding making a decision...

graeme-winter:
> I worry that if summing over a range to try to find weak signal, or signal only present on a few, that this might hide it? A "max pixel" mode might be appropriate. Maybe this does deserve a setting, but maybe that's just avoiding making a decision...

My thinking was about simulating a wider image in the first instance, but the max is useful as you say for getting the pattern out. I think the universe has space for both...

graeme-winter:
@ndevenish added a mean / max pulldown - illustrated difference:

Max:

Screenshot 2020-06-15 at 09 12 18


Mean:

Screenshot 2020-06-15 at 09 12 41

util/image_viewer/spotfinder_frame.py
graeme-winter: ... maybe, provided the people's national bikeshed committee are satisfied with the nomenclature of stacking the images?

graeme-winter: I don't know how to make one phil parameter alias another, and deprecate one or the other, otherwise I would have done this...

graeme-winter: I could replace the name and just break the cl api

rjgildea: R.e. marking phil parameters as deprecated:
https://github.com/cctbx/cctbx_project/issues/452#issuecomment-604313748

I would either:
1. Rename `sum_images` to `stack_images` and re-add `sum_images` with `.deprecated` attribute, and ignore. That way it doesn't fail, but users get a warning.
2. Just rename and accept that very few people likely use the CLI to the image viewer.

graeme-winter: Pushed change to this effect, if you use old command line option you get 

```
Grey-Area 0 :) $ dials.image_viewer imported.expt strong.refl sum_images=10
Sorry: Unknown command line parameter definition: sum_images = 10
```

which I think is OK

rjgildea:
As far as I can tell, you can only change the number of images to stack via the command line but not via the GUI, and conversely, you can only change the stack "type" in the GUI but not via the command line. This means that if you load the image viewer with default options, then there is a "stack type" dropdown that doesn't do anything.

The `stack_type` setting should probably be exposed via the command line, and it only makes really makes sense to have the "stack type" dropdown in the GUI settings panel if you can also control the number of images there too.

rjgildea:
This seems incompatible with setting the "image type" to be "raw":

```
$ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
```
Select "raw" in the "image type" dropdown:
```
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2140, in OnUpdateImage
    self.GetParent().GetParent().reload_image()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
    self.load_image(self.images.selected, refresh=True)
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
    show_untrusted=show_untrusted,
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
    self.update_settings()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1175, in update_settings
    self.stack_images()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 836, in stack_images
    sel = data > rd
TypeError: '>' not supported between instances of 'double' and 'int'
```

rjgildea:
Another oddity, this behaviour was probably at least in part already present before this PR - if you're in stack/sum images mode, then press the "image" button, then the displayed image goes back to the individual, rather than stacked, image, until you re-select an item from the "stack type" dropdown menu. I'm not entirely sure what the expected should be for the various thresholding maps - should we be running the threshold algorithms on the stacked image?

graeme-winter:
> As far as I can tell, you can only change the number of images to stack via the command line but not via the GUI, and conversely, you can only change the stack "type" in the GUI but not via the command line. This means that if you load the image viewer with default options, then there is a "stack type" dropdown that doesn't do anything.
> 
> The `stack_type` setting should probably be exposed via the command line, and it only makes really makes sense to have the "stack type" dropdown in the GUI settings panel if you can also control the number of images there too.

There is a "Stack" box at the top of the image window? I've been using this extensively so know that works 😉

I had wondered whether the `stack_type` should be exposed via command line, but not followed up on.

graeme-winter:
> This seems incompatible with setting the "image type" to be "raw":
> 
> ```
> $ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
> ```
> 
> Select "raw" in the "image type" dropdown:
> 
> ```
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2140, in OnUpdateImage
>     self.GetParent().GetParent().reload_image()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
>     self.load_image(self.images.selected, refresh=True)
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
>     show_untrusted=show_untrusted,
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
>     self.update_settings()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1175, in update_settings
>     self.stack_images()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 836, in stack_images
>     sel = data > rd
> TypeError: '>' not supported between instances of 'double' and 'int'
> ```

Reproduced this, will look

graeme-winter:
> Another oddity, this behaviour was probably at least in part already present before this PR - if you're in stack/sum images mode, then press the "image" button, then the displayed image goes back to the individual, rather than stacked, image, until you re-select an item from the "stack type" dropdown menu. I'm not entirely sure what the expected should be for the various thresholding maps - should we be running the threshold algorithms on the stacked image?

Thresholding should be fine against summed, mean requires some discussion re: "gain" equivalent, max just plain wrong. Not sure if this should be imposed. 

Re: the oddity - the internals of the image viewer are something of a tangled mess. Should fixing this be a necessary precondition of the PR features being made available?

graeme-winter:
> > This seems incompatible with setting the "image type" to be "raw":
> > ```
> > $ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
> > ```
> > 
> > 
> > Select "raw" in the "image type" dropdown:
> > ```
> > Traceback (most recent call last):
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2140, in OnUpdateImage
> >     self.GetParent().GetParent().reload_image()
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
> >     self.load_image(self.images.selected, refresh=True)
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
> >     show_untrusted=show_untrusted,
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
> >     self.update_settings()
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1175, in update_settings
> >     self.stack_images()
> >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 836, in stack_images
> >     sel = data > rd
> > TypeError: '>' not supported between instances of 'double' and 'int'
> > ```
> 
> Reproduced this, will look

Fixed / worked around in c3d54cc1f86fb2f65e03198e157b1dd48908ad86

graeme-winter:
> > As far as I can tell, you can only change the number of images to stack via the command line but not via the GUI, and conversely, you can only change the stack "type" in the GUI but not via the command line. This means that if you load the image viewer with default options, then there is a "stack type" dropdown that doesn't do anything.
> > The `stack_type` setting should probably be exposed via the command line, and it only makes really makes sense to have the "stack type" dropdown in the GUI settings panel if you can also control the number of images there too.
> 
> There is a "Stack" box at the top of the image window? I've been using this extensively so know that works 😉
> 
> I had wondered whether the `stack_type` should be exposed via command line, but not followed up on.

Done, in a2910e0041898ac0a3dabdf407bd4ad6a30a2950

graeme-winter:
> > Another oddity, this behaviour was probably at least in part already present before this PR - if you're in stack/sum images mode, then press the "image" button, then the displayed image goes back to the individual, rather than stacked, image, until you re-select an item from the "stack type" dropdown menu. I'm not entirely sure what the expected should be for the various thresholding maps - should we be running the threshold algorithms on the stacked image?
> 
> Thresholding should be fine against summed, mean requires some discussion re: "gain" equivalent, max just plain wrong. Not sure if this should be imposed.
> 
> Re: the oddity - the internals of the image viewer are something of a tangled mess. Should fixing this be a necessary precondition of the PR features being made available?

I have looked at the code for a while and can't see where the change should go to make this oddity go away - however since it is not a _new_ oddity I am not too worried about this. I have made sum the default, so if someone stacks images and then looks through the filters the results should be by default meaningful.

rjgildea:
> There is a "Stack" box at the top of the image window? I've been using this extensively so know that works 😉

Ah ok - I had missed that because I was naively expecting related controls to be in the _same_ window at the very least. Does it really make sense for the control for number of images to stack and the control for stack type to be in different windows?

graeme-winter:
> > There is a "Stack" box at the top of the image window? I've been using this extensively so know that works 😉
> 
> Ah ok - I had missed that because I was naively expecting related controls to be in the _same_ window at the very least. Does it really make sense for the control for number of images to stack and the control for stack type to be in different windows?

Respectfully acknowledged. However also mention that stacking images is related to loading images, which is all controlled above the image... and selecting the view kinda lives with the other settings, so maybe (just maybe) the viewer is poorly designed 🤷‍♂️

util/image_viewer/spotfinder_frame.py
```diff
@@ -823,12 +824,24 @@ def sum_images(self):
...
-            for i in range(1, self.params.sum_images):
+            for i in range(1, self.params.stack_images):
                 if (i_frame + i) >= len(imageset):
                     break
                 image_data_i = imageset[i_frame + i]
                 for j, rd in enumerate(image_data):
-                    rd += image_data_i[j]
+                    data = image_data_i[j]
+                    if mode == "max":
+                        if type(rd) != type(data):
```
rjgildea, line 834: I think you can call `.as_double()` even on an array that is already `flex.double()`, so you could avoid having to explicitly check the type and always both sides to double.

`a = flex.random_double(10).as_double()`

graeme-winter:
> > I have looked at the code for a while and can't see where the change should go to make this oddity go away - however since it is not a _new_ oddity I am not too worried about this. I have made sum the default, so if someone stacks images and then looks through the filters the results should be by default meaningful.
> 
> I'm not convinced the thresholding maps are acting on the sum/mean/max images in any case - I suspect they are effectively switching to `stack_images=1` without updating any of the controls to reflect this.

Quite possibly: perhaps this is also in the realm of 🤷‍♂️  - since stack images existed before this I am not sure it's a new bug...

rjgildea:
> > > This seems incompatible with setting the "image type" to be "raw":
> > > ```
> > > $ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
> > > ```
> > > 
> > > 
> > > Select "raw" in the "image type" dropdown:
> > > ```
> > > Traceback (most recent call last):
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2140, in OnUpdateImage
> > >     self.GetParent().GetParent().reload_image()
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
> > >     self.load_image(self.images.selected, refresh=True)
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
> > >     show_untrusted=show_untrusted,
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
> > >     self.update_settings()
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1175, in update_settings
> > >     self.stack_images()
> > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 836, in stack_images
> > >     sel = data > rd
> > > TypeError: '>' not supported between instances of 'double' and 'int'
> > > ```
> > 
> > 
> > Reproduced this, will look
> 
> Fixed / worked around in [c3d54cc](https://github.com/dials/dials/commit/c3d54cc1f86fb2f65e03198e157b1dd48908ad86)

This still breaks for me:
```
$ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10

stack_images = 10
input {
  experiments = <image files>
}

  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2141, in OnUpdateImage
    self.GetParent().GetParent().reload_image()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
    self.load_image(self.images.selected, refresh=True)
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
    show_untrusted=show_untrusted,
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
    self.update_settings()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1177, in update_settings
    self.stack_images()
  File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 839, in stack_images
    rd += data
Boost.Python.ArgumentError: Python argument types in
    int.__iadd__(int, double)
did not match C++ signature:
    __iadd__(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > {lvalue}, int)
    __iadd__(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > {lvalue}, scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > >)```

rjgildea:
Also "max" stack mode + "raw" image type looks very much like it is displaying the single image, and not a stacked image.

graeme-winter:
<- is wondering about abandoning this project as to fix all the problems here will probably require rewriting the image viewer.

graeme-winter:
> Also "max" stack mode + "raw" image type looks very much like it is displaying the single image, and not a stacked image.

Sometimes the raw image is ints, sometimes doubles, sometimes the image you are adding is ints, sometimes doubles, and they need to be added in-place so will need to cast images-for-adding to the same type as image-already-there - cannot stick with one "standard" type - so this will be some horrible hodge-podge 😞

graeme-winter:
> > > > This seems incompatible with setting the "image type" to be "raw":
> > > > ```
> > > > $ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
> > > > ```
> > > > 
> > > > 
> > > > Select "raw" in the "image type" dropdown:
> > > > ```
> > > > Traceback (most recent call last):
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2140, in OnUpdateImage
> > > >     self.GetParent().GetParent().reload_image()
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
> > > >     self.load_image(self.images.selected, refresh=True)
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
> > > >     show_untrusted=show_untrusted,
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
> > > >     self.update_settings()
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1175, in update_settings
> > > >     self.stack_images()
> > > >   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 836, in stack_images
> > > >     sel = data > rd
> > > > TypeError: '>' not supported between instances of 'double' and 'int'
> > > > ```
> > > 
> > > 
> > > Reproduced this, will look
> > 
> > 
> > Fixed / worked around in [c3d54cc](https://github.com/dials/dials/commit/c3d54cc1f86fb2f65e03198e157b1dd48908ad86)
> 
> This still breaks for me:
> 
> ```
> $ dials.image_viewer $(dials.data get -q x4wide)/*.cbf stack_images=10
> DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
> The following parameters have been modified:
> 
> stack_images = 10
> input {
>   experiments = <image files>
> }
> 
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 2141, in OnUpdateImage
>     self.GetParent().GetParent().reload_image()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 513, in reload_image
>     self.load_image(self.images.selected, refresh=True)
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 558, in load_image
>     show_untrusted=show_untrusted,
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/slip_viewer/frame.py", line 499, in load_image
>     self.update_settings()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 1177, in update_settings
>     self.stack_images()
>   File "/Users/rjgildea/software/cctbx/modules/dials/util/image_viewer/spotfinder_frame.py", line 839, in stack_images
>     rd += data
> Boost.Python.ArgumentError: Python argument types in
>     int.__iadd__(int, double)
> did not match C++ signature:
>     __iadd__(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > {lvalue}, int)
>     __iadd__(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > {lvalue}, scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > >)```
> ```

... try with 372120fc16eb07ee3dc3eb251ae9e9e9ff69dbc1

graeme-winter:
> > I have looked at the code for a while and can't see where the change should go to make this oddity go away - however since it is not a _new_ oddity I am not too worried about this. I have made sum the default, so if someone stacks images and then looks through the filters the results should be by default meaningful.
> 
> I'm not convinced the thresholding maps are acting on the sum/mean/max images in any case - I suspect they are effectively switching to `stack_images=1` without updating any of the controls to reflect this.

I'm pretty sure this is properly broken. Could disable these buttons if `stack != 1`? Alternative will be rewriting the way that it works.

graeme-winter:
@rjgildea disabled the buttons if stack now - so behaviour is consistent at least!

graeme-winter:
Summary:
 - using double for all image data (raw; corrected) so internally consistent
 - if stack: disable image filter buttons
 - command-line control of stack number, stack mode
 - tests with raw and corrected images work

PR #1308, opened 2020-06-22
Remove any_reflection_file() calls

Anthchirp:
especially in contexts where we know exactly what sort of file we are
opening.

The only potential use is in `util/resolutionizer.py`, where I guess something other than a `.mtz` could be opened, however the function name implies that it would only open `.mtz`s.
Can someone verify this?


Anthchirp:
Discussed in meeting. Good to merge

PR #1313, opened 2020-06-26
Add option to disable timing output

graeme-winter:
I find myself typing `| cut -c 9-` many times per day to remove timing info I don't want - so adding an option to not have it there following NO_COLOR.


Anthchirp:
should timings only be written to the log if run with `-v`?

dagewa:
I thought we had agreed that regarding verbosity some time back, but it just never got done

graeme-winter:
☝️ works for me 🙂

PR #1315, opened 2020-06-29
install IOTA from conda-forge

Anthchirp:
The iota/libtbx_refresh<->iota/setup.py adapter hack seems to have stopped working.
Since a DIALS environment doesn't need to be an IOTA development environment we should really just install a stable version from conda-forge

### DIFF
--- a/.conda-envs/linux.txt
+++ b/.conda-envs/linux.txt
@@ -22,6 +22,7 @@ conda-forge::eigen
 conda-forge::future
 conda-forge::h5py
 conda-forge::hdf5<1.11
+conda-forge::iota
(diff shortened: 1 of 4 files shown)


Anthchirp:
Existing development installations should not need to make any changes.

If you want to use your DIALS development environment to work on iota after this PR has been merged you can run
```
. dials
cd modules
git clone git@github.com:/ssrl-px/iota.git
pip install -e iota
```
and whenever changes are made to `iota/setup.py`:
```
. dials
cd modules/iota
python setup.py develop
```

Note that iota should not be configured for use with libtbx.

PR #1317, opened 2020-06-30
Run coverage on Azure

Anthchirp:
Before:
https://dev.azure.com/azure-dials/dials/_build/results?buildId=333&view=results
https://github.com/dials/dials/runs/811943466

After:
https://dev.azure.com/azure-dials/dials/_build/results?buildId=361&view=results
https://github.com/dials/dials/runs/822794676
https://codecov.io/gh/dials/dials/branch/coverage

Cost appears to be somewhere between 5 and 8 minutes runtime.
Benefit is that we get coverage testing on every branch and PR (which we don't get via Jenkins)
We will lose coverage on dials_regression tests which Jenkins did have.
On the other hand we can run on Python 3. The current Jenkins job is 2.7 only.

### DIFF
--- a/.azure-pipelines/unix-build.yml
+++ b/.azure-pipelines/unix-build.yml
@@ -174,7 +174,7 @@ steps:
     . dials
     cd modules/dials
     pytest -v -ra -n auto --basetemp="$(Pipeline.Workspace)/tests" --durations=10 \
-        --cov=dials --cov-report=xml --cov-branch \
+        --cov=$(pwd) --cov-report=xml --cov-branch \
         --timeout=5400 --regression --runslow || echo "##vso[task.complete result=Failed;]Some tests failed"
   displayName: Run tests
   workingDirectory: $(Pipeline.Workspace)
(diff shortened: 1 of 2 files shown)


ndevenish:
Explicitly, this loses about 8% of coverage due to loss of dials_regression: https://codecov.io/gh/dials/dials/compare/master...coverage

But it would simplify the CI environment and allow us to have coverage on branches, pull requests etc.

dagewa:
I'm aware that `dials.refine` tests make up a decent chunk of that 8%. I'd like to convert these to be `dials_regression`-free, but don't know when I'll get round to it

Anthchirp:
Since we'd see a lot more of those coverage comments I'm thinking we may want to disable the [reach graph](https://docs.codecov.io/docs/graphs#coverage-reach) and possibly even the `diff` block (the important figure should be shown above the block anyway):
```yaml
comment:
  layout: "diff, files"
  require_changes: true  # if true: only post the comment if coverage changes
  branches:          # branch names that can post comment
    - "master"
```

Anthchirp:
suggest we leave comments for the time being and discuss those separately after merge

PR #1319, opened 2020-07-01
Integration output improvements

graeme-winter:
- move a lot of the output from `dials.integrate` to debug
- fix a lot of off-by-one errors

In example of 2,400 image scan reduced output length by 68% [1] without reducing the amount of signal in the output. 

[1]  ->

```
Grey-Area original :) $ wc -l dials.integrate.log 
   24945 dials.integrate.log
Grey-Area reduced :) $ wc -l dials.integrate.log 
    8100 dials.integrate.log
```

### DIFF
--- a/algorithms/integration/integrator.py
+++ b/algorithms/integration/integrator.py
@@ -637,13 +637,13 @@ class ProfileModellerExecutor(Executor):
 
         # Print a histogram of reflections on frames
         if frame1 - frame0 > 1:
-            logger.info(
+            logger.debug(
                 " The following histogram shows the number of reflections predicted"
             )
-            logger.info(" to have all or part of their intensity on each frame.")
-            logger.info("")
-            logger.info(frame_hist(reflections["bbox"], prefix=" ", symbol="*"))
-            logger.info("")
+            logger.debug(" to have all or part of their intensity on each frame.")
+            logger.debug("")
+            logger.debug(frame_hist(reflections["bbox"], prefix=" ", symbol="*"))
+            logger.debug("")
 
     def process(self, frame, reflections):
         """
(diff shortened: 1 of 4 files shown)


graeme-winter:
I may rewrite the history here to 

 - fix off by ones
 - move output to debug

Anthchirp:
Since you're already playing with the logging here: There are _a lot_ of log calls to create empty lines, ie.
```
        logger.info("")
```
etc. Can these be rewritten into single log calls or possibly just eliminated?

These are relatively expensive as they have the full logging overhead for 0 signal, and they are also annoying if the logging does anything other than writing to console or a file.

For example: 
* If you use logging in a multiprocessing environment you can guarantee logging atomicity of messages. If you break up your message into 20 atomic logging calls, then that is no longer useful.
* If you send log messages to a log server then you do not care about 'nice' console text layouts. The 17 empty log messages are now just waste.

graeme-winter:
> Since you're already playing with the logging here: There are _a lot_ of log calls to create empty lines, ie.
> 
> ```
>         logger.info("")
> ```
> 
> etc. Can these be rewritten into single log calls or possibly just eliminated?
> 
> These are relatively expensive as they have the full logging overhead for 0 signal, and they are also annoying if the logging does anything other than writing to console or a file.
> 
> For example:
> 
> * If you use logging in a multiprocessing environment you can guarantee logging atomicity of messages. If you break up your message into 20 atomic logging calls, then that is no longer useful.
> * If you send log messages to a log server then you do not care about 'nice' console text layouts. The 17 empty log messages are now just waste.

Happy to do this, wondered myself, is correct to have `logger.info("\nThings\n")` ?

Anthchirp:
Would say yes, although I'd suggest avoiding leading newlines

graeme-winter:
@Anthchirp spent a while replacing `logger.info("")` with extra newlines, but thought on balance this should be done to remove sequential `logger.info()` calls to have multiline statements as single `f"words"` 🤔

graeme-winter:
OK, well I have got this far - a tool for finding multi-line loggers anyways:

```
import ast
import sys
from itertools import groupby
from operator import itemgetter

def check(filename):
    root = ast.parse(open(filename).read())
    text = open(filename).readlines()

    loggers = []
    
    for node in ast.walk(root):
        if isinstance(node, ast.Name):
            if node.id != "logger":
                continue
            start = node.lineno
            if hasattr(node, "end_lineno"):
                end = node.lineno
            else:
                end = start
            for j in range(start - 1, end):
                loggers.append(j)

    # identify blocks of logger calls
    for k, g in groupby(enumerate(loggers), lambda x: x[1] - x[0]):
        l = list(map(itemgetter(1), g))
        if len(l) > 1:
            print(f"{filename}:{l[0]}-{l[-1]}")
            for line in l:
                print(text[line].rstrip())
    
if __name__ == "__main__":
    for arg in sys.argv[1:]:
        check(arg)

```

graeme-winter:
OK, so the tests all pass: I will try to look at the logging cleanups some more tomorrow

graeme-winter:
After discussion with @Anthchirp leaving logging cleanups to more wholesale investigation

PR #1289, opened 2020-06-05
Update numpy version

Anthchirp:
Updating numpy to 1.18.4 causes a test failure in `algorithms/scaling/test_scale.py::test_scale_array`:

```
(...)
Results of error model refinement. Uncorrected and corrected variances
of normalised intensity deviations for given intensity ranges. Variances
are expected to be ~1.0 for reliable errors (sigmas).
+--------------------------+----------+------------------------+----------------------+
| Intensity range (<Ih>)   |   n_refl |   Uncorrected variance |   Corrected variance |
|--------------------------+----------+------------------------+----------------------|
| 90214.18 - 24947.56      |       36 |                 25.57  |                0.596 |
| 24947.56 - 15944.68      |       36 |                 15.483 |                1.376 |
| 15944.68 - 11895.09      |       49 |                 10.868 |                1.234 |
| 11895.09 - 6054.32       |      272 |                  6.001 |                1.105 |
| 6054.32 - 3081.50        |      238 |                  5.95  |                1.49  |
| 3081.50 - 1568.41        |      150 |                  8.099 |                3.02  |
| 1568.41 - 798.28         |      153 |                  6.028 |                2.924 |
| 798.28 - 406.31          |       67 |                  1.433 |                0.811 |
| 406.31 - 105.25          |       56 |                  1.779 |                1.016 |
+--------------------------+----------+------------------------+----------------------+


The reflection table variances have been adjusted to account for the
uncertainty in the scaling model

Total time taken: 0.3898s 

Traceback (most recent call last):
================================================================================


  File "/dev/shm/wra62962/conda_base/lib/python3.6/site-packages/numpy/core/function_base.py", line 117, in linspace
    num = operator.index(num)Summary of dataset partialities

+------------------+----------+
TypeError: 'float' object cannot be interpreted as an integer

| Partiality (p)   |   n_refl |
|------------------+----------|
During handling of the above exception, another exception occurred:

| all reflections  |     4413 |
Traceback (most recent call last):
| p > 0.99         |     2668 |
  File "/dev/shm/wra62962/build/../modules/dials/command_line/scale.py", line 265, in <module>| 0.5 < p < 0.99   |       32 |

    run()
| 0.01 < p < 0.5   |       27 |
| p < 0.01         |     1686 |
  File "/dev/shm/wra62962/build/../modules/dials/command_line/scale.py", line 239, in run
+------------------+----------+

[...]

PR #1320, opened 2020-07-01
bootstrap updates

Anthchirp:
* Refactor bootstrap script to make the control flow more obvious
* Rewrite the source checkout logic to allow more complex setups
* By default, check out the `master` branch from `dials/cctbx` rather than from `cctbx_project/cctbx` with the local branch name `stable`.
  * [This branch follows the `cctbx_project/cctbx` master branch, and automatically advances whenever the cctbx CI tests pass](https://github.com/cctbx/cctbx_project/commit/15676b42fc80fbe17922df64882c2c7af44b2932). Therefore the checked out copy will not be known unstable.
  * A local pre-commit is set up to reject local commits to the `stable` branch. All other branches will behave normally.
  * One can at any time `git checkout master` to get access to the regular `cctbx_project` `master` branch, and similarly check out any other branch from the `cctbx_project/cctbx` remote. Branches from the `dials/cctbx` remote are also available.
* ~PR also includes the update numpy 1.16 ➡️ 1.17 (half-way move for #1289)~

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,5 +1,5 @@
 variables:
-  CACHE_VERSION: 20200604
+  CACHE_VERSION: 20200701
 
 schedules:
   # nightly builds to populate caches
--- a/.conda-envs/dials_py36_linux-64.txt
+++ b/.conda-envs/dials_py36_linux-64.txt
@@ -21,7 +21,7 @@ conda-forge::jinja2
 conda-forge::jpeg
 conda-forge::matplotlib-base>=3.0.2
 conda-forge::mock
-conda-forge::numpy>=1.16,<1.18
+conda-forge::numpy>=1.16,<1.17
 conda-forge::pillow>=5.4.1
 conda-forge::pip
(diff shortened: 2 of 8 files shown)


graeme-winter:
`dials/cctbx as stable` - sounds sensible - the only side-effect I can see is that e.g. changes to `cctbx.xfel` would not propagate if there are (to pick a random example) failing tests in `mmtbx`

Evidently this would only affect those using the DIALS bootstrap, not the old `cctbx_project` one, so the main way anyone would notice this is that (in the circumstances above) nightly builds would not contain any `cctbx_project` fixes until other bugs are fixed. I guess this reflects the issue with keeping everything in one repo rather than a collection of repos so 🤷‍♂️  not much you can do about that...

graeme-winter:
To be clear: I'm happy with this as it will make development builds much more stable!

Anthchirp:
If there are no concerns I'll merge this over the weekend together with #1289.

ndevenish:
I'd prefer dependency updates be separated

PR #1057, opened 2019-12-03
Timing information in log files 

dagewa:
For #932. 

Currently looks like this:
```
0000.0:  DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
0000.0:  DIALS 2.dev.1179-g30f23efab
0000.0:  The following parameters have been modified:
        
0000.0:  input {
          experiments = indexed.expt
          reflections = indexed.refl
        }
        
0000.0:  Configuring refiner
0000.2:  
        Summary statistics for 19089 observations matched to predictions:
0000.2:  +-------------------+--------+----------+-----------+---------+--------+
        |                   |    Min |       Q1 |       Med |      Q3 |    Max |
        |-------------------+--------+----------+-----------+---------+--------|
        | Xc - Xo (mm)      | -1.504 | -0.08959 | -0.001273 | 0.06692 | 0.8589 |
        | Yc - Yo (mm)      | -4.761 | -0.07079 |  0.002289 | 0.08256 |  8.363 |
        | Phic - Phio (deg) | -12.89 |  -0.2159 |   0.02121 |   0.209 |  17.74 |
        | X weights         |  290.2 |    386.1 |     397.6 |   403.7 |  405.6 |
        | Y weights         |  296.6 |    386.2 |     396.9 |   403.3 |  405.6 |
        | Phi weights       |   11.1 |    11.97 |        12 |      12 |     12 |
        +-------------------+--------+----------+-----------+---------+--------+
        
0000.2:  Detecting centroid outliers using the MCD algorithm
0001.8:  4369 reflections have been flagged as outliers
0001.8:  14720 reflections remain in the manager
0001.8:  
        Summary statistics for 14720 observations matched to predictions:
0001.9:  +-------------------+---------+----------+------------+---------+--------+
        |                   |     Min |       Q1 |        Med |      Q3 |    Max |

```

PR #1322, opened 2020-07-03
Only show DEBUG messages if verbosity > 1

rjgildea:
Following 0e4ec1d, this allows for getting the normal log output plus the timing information only. Previously, you could only get the normal output only, or the DEBUG messages plus timing. Since some DEBUG messages potentially involve some expensive calculations, timing output may not be representative of normal runtime.

### DIFF
--- a/newsfragments/1322.misc
+++ b/newsfragments/1322.misc
@@ -1,0 +1,1 @@
+Only show debug output with -vv. -v will give standard output plus timing information.
--- a/test/util/test_log.py
+++ b/test/util/test_log.py
@@ -6,8 +6,9 @@ import dials.util.log
 
 
 def test_LoggingContext():
+
     # configure logging
-    dials.util.log.config(verbosity=1)
+    dials.util.log.config(verbosity=2)
 
     # get some loggers
     idx_logger = logging.getLogger("dials.algorithms.indexing")
--- a/util/log.py
+++ b/util/log.py
@@ -73,7 +73,7 @@ def config(verbosity=0, logfile=None):
     dials_logger = logging.getLogger("dials")
     dials_logger.addHandler(console)
 
-    if verbosity:
+    if verbosity > 1:
         loglevel = logging.DEBUG
     else:
         loglevel = logging.INFO
--- a/util/options.py
+++ b/util/options.py
@@ -683,7 +683,11 @@ class OptionParserBase(optparse.OptionParser, object):
 
         # Set a verbosity parameter
         self.add_option(
-            "-v", action="count", default=0, dest="verbose", help="Increase verbosity"
+            "-v",
+            action="count",
+            default=0,
+            dest="verbose",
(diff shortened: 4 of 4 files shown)


graeme-winter:
Works for me 🙂

rjgildea:
@Anthchirp as pointed out above, the current implementation as of 0e4ec1d means that you can only get timing information in combination with debug output, which may not be an accurate reflection of the true timing information (without debug output). If timing information isn't useful, why did was it added in the first place (https://github.com/dials/dials/pull/1057)? I probably would have preferred to add an additional `-t` command line flag, but this would have necessitated changing every call to `dials.util.log.config` - this PR is less invasive.

Anthchirp:
Well, timing with regular output wasn't so useful, which is why we removed it. I think the current state (timing with `-v`, no timing without `-v`) is fine - even if the timings change between `-v` and not `-v`. But I'm not invested in this either way.

rjgildea:
With the current state, timing information is definitely not useful, as it only reflects the timing of the debug mode, which may bare little relation with the actual timings. If the consensus is that timing information is _not_ useful, then perhaps the timing information introduced in #1057 should be removed entirely, but then that will effectively re-open #932 - although perhaps 0e4ec1d already does that as the user now only gets by default the inconsistent timing information that #932 complained about.

Anthchirp:
The timings in the "debug mode" will accurately describe the timings of the run in "debug mode". In my opinion it is a separate issue if we are doing wildly different things when running with `-v` compared to running without. I certainly would be surprised by a significant behavioural change when supplying a verbosity parameter.

dagewa:
It happens here in refinement
https://github.com/dials/dials/blob/0a33e73b1736c17e5ce729882404aa5bfc5a11d2/algorithms/refinement/refiner.py#L1024-L1028
The amount of time it takes to run these calculations is minimal, but there is no point at all to run them if the results are not going to be displayed.

dagewa:
> It happens here in refinement
> https://github.com/dials/dials/blob/0a33e73b1736c17e5ce729882404aa5bfc5a11d2/algorithms/refinement/refiner.py#L1024-L1028
> 
> 
> The amount of time it takes to run these calculations is minimal, but there is no point at all to run them if the results are not going to be displayed.

Actually this just avoids calculation when refiner is truly quiet. I don't think there is a difference in calculation between info and debug any more in refiner

Anthchirp:
> I like it. Adding timing is exactly an increase in verbosity.

:+1: In that case I suggest amending the command line help text to indicate `-v` can be used multiple times

and fixing the failing test before merging :slightly_smiling_face:

rjgildea:
> > I like it. Adding timing is exactly an increase in verbosity.
> 
> 👍 In that case I suggest amending the command line help text to indicate `-v` can be used multiple times
> 
> and fixing the failing test before merging 🙂

Done and done 👍

PR #1325, opened 2020-07-06
Users should not see stack trace

graeme-winter:


algorithms/scaling/scaler.py
```diff
@@ -212,7 +212,7 @@ def perform_error_optimisation(self, update_Ih=True):
                 self.params.weighting.error_model, Ih_table
             )
         except (ValueError, RuntimeError) as e:
-            logger.info(e, exc_info=True)
+            logger.debug(e, exc_info=True)
```
jbeilstenedmands, line 215: I think it would be best to keep the error message to info (without trace), so that the user knows that error modelling was unsuccessful for whatever reason/condition (there are a number in error modelling where ValueError is raised), so how about logger.info(e) followed by logger.debug(e, exc_info=True) ?

graeme-winter:
For future reference - was getting

```
Insufficient reflections (14 < 250) to perform error modelling.
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/scaling/scaler.py", line 212, in perform_error_optimisation
    self.params.weighting.error_model, Ih_table
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/scaling/error_model/engine.py", line 36, in run_error_model_refinement
    model = model_class(Ih_table.blocked_data_list[0], scope)
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/scaling/error_model/error_model.py", line 365, in __init__
    Ih_table, basic_params, min_partiality
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/scaling/error_model/error_model.py", line 410, in filter_unsuitable_reflections
    min_reflections_required=cls.min_reflections_required,
  File "/Users/graeme/svn/cctbx/modules/dials/algorithms/scaling/error_model/error_model.py", line 565, in filter_unsuitable_reflections
    % (n, min_reflections_required)
ValueError: Insufficient reflections (14 < 250) to perform error modelling.
```

which I think is not user friendly

graeme-winter:
Now 
```
8 outliers have been identified. 

Insufficient reflections (14 < 250) to perform error modelling.
Components to be refined in this cycle for all datasets: scale, decay
Performing a round of scaling with a Levenberg-Marquardt minimizer.
```

algorithms/scaling/scaler.py
graeme-winter: I don't hold a strong opinion & one line is fairly harmless, so done in aa00a0d10b2144142c02f654c83b0983422015f5

Issue #1255, opened 2020-05-11
Please use `tqdm.std.*` instead of `tqdm._tqdm.*`


picca:
Hello, while working on the Deban package of dials, I got this error message during the pytest execution.

```
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f3dce2c7190>

    @pytest.fixture
    def time(monkeypatch):
        "Fixture to monkeypatch the tqdm time access"
        ft = FakeTime()
>       monkeypatch.setattr(tqdm._tqdm, "time", ft)
E       AttributeError: module 'tqdm' has no attribute '_tqdm'

/usr/lib/python3/dist-packages/dials/test/util/test_progress.py:62: AttributeError
```

When I import tqdm._tqdm, I have this message

```
>>> import tqdm._tqdm
<stdin>:1: TqdmDeprecationWarning: This function will be removed in tqdm==5.0.0
Please use `tqdm.std.*` instead of `tqdm._tqdm.*`
```

ndevenish:
Hi Picca,

As per our dependency list: https://github.com/dials/dials/blob/9f591ba10e29f89b1a17285cba043cb15254d100/libtbx_config#L21

we don't support that version of `tqdm`, and are pinned because we use internal features. It might be possible to switch versions if the feature still works, but we haven't tested that version. If everything still works fine with a newer version, we're happy to accept a change to it.

Alternatively, I'd be open to dropping this code and just using tqdm directly if it's noninteractive behaviour is acceptable now (this code was so that we could _not_ have the noisy progress bar for short runs)

Anthchirp:
I believe removal of the custom core is the reasonable way forward, cf. #1328

Issue #1236, opened 2020-04-15
Python 3.7 and 3.8 support


phyy-nx:
I was testing cctbx/cctbx_project#409 with the dials builder to see how far it got and it stopped on base, since dials/.conda-envs/dials_py38_linux-64.txt is not available.  Perhaps this issue could serve as a starting point for getting python 3.7 and 3.8 working in dials?  First step would be getting the recipes ready.

Anthchirp:
I have looked at this in the past and found some known package restrictions that hold back 3.8 support:
https://github.com/dials/dials/blob/1d438bf3ae70e1180876cd248a080bc29ddb70f9/.conda-envs/linux.txt#L12-L38

ie. we'd need to update and validate matplotlib, biopython, scikit-learn, scipy, and sphinx.

I've updated numpy (80b50b0f1ab6d3703a18199721abd2d3c35140bd), this was required to fix #1193, and fixed the conda-forge 3.8 package for orderedset already, but haven't had the time to look at more of them yet. 3.8 support is not currently urgent for us, and there was no point looking at this in more detail until cctbx/cctbx_project#409 was ready.

I'll probably wait until it goes in before I revisit this.

Anthchirp:
With the essential dependency updates applied the current status is:
* Linux: 3.7 ✔️ 
* Linux: 3.8 ✔️ 
* MacOS: 3.7 ❌ 4 tests fail
  * `test_index_after_search`: `concurrent.futures.process.BrokenProcessPool`
  * `test_setting_c2_vs_i2[False-C 1 2 1-expected_unit_cell1]`: `concurrent.futures.process.BrokenProcessPool`
  * `test_search_small_molecule`: `concurrent.futures.process.BrokenProcessPool`
  * `test_show_extensions`: `libtbx.easy_run` returns `RuntimeError: non-zero return code: -6`
* MacOS: 3.8 ❌ 2 tests fail
  * `test_find_spots_server_client`: `TypeError: cannot pickle '_thread.lock' object`
  * `test_preservation_of_identifiers`: `UserWarning: resource_tracker: There appear to be 4 leaked semaphore objects to clean up at shutdown`
* ⚠️ a bunch of new deprecation warnings we can easily ignore until 3.9 is released in October (so beyond the scope of #1175)

https://dev.azure.com/azure-dials/dials/_build/results?buildId=155&view=results

Anthchirp:
[...]

PR #1328, opened 2020-07-07
Remove custom DIALS tqdm

Anthchirp:
Remove DIALS tqdm customizations and use unmodified, plain tqdm code instead. Retain the changed defaults when imported via `dials.util.progress`.

This allows wider compatibility with `tqdm` versions, resolving #1255, and allows updating the tqdm version (not part of this pull request), which fixes some Python 3.8 compatibility issues (#1236).


ndevenish:
Yes, happy with this, I was thinking about it earlier today in fact. I can’t remember if it tried to detect tty capture (to avoid messing up logs) but we can always handle that separately.

Anthchirp:
oh joy, this is _ludicrous_.

Removing our custom tqdm modifications randomly (yet reproducibly) breaks MacOS >=Catalina on Python 3.7, because matplotlib 3.2 opens a font file (but doesn't read it - to conserve memory), the handle of which is then `spawn()`ed to multiprocessing children that then try to close that file and :boom:.
```
(1000 line stack trace skipped)
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Couldn't close file
```

There are three solutions here:
1. switch to multiprocessing forkserver, which is what this PR does.
2. (apparently) go directly to Python 3.8, do not pass go, do not support 3.7.
3. update to matplotlib 3.3. Unfortunately that one has only just reached the rc1 release stage.

ndevenish:
I don't think it's unreasonable to change for the tests, seeing as that's the only time when we use matplotlib before multiprocessing. I wondered for a bit if changing to `spawn` might be more appropriate in advance of that being the default anyway, but guess it's not too bad since it's constrained to 3.7 only.

I'm still not clear why this only happens for 3.7 and why removing this code triggered the problem. I'm wondering it it's a test ordering edge case triggered by the test being removed, and that this change should be separate from this removal (even though the removal exhibited the problem)

Anthchirp:
I don't know. I also don't have access to an interactive system where I can test this. The build mac mini could not reproduce the issue at all, but that should be because it is pre-Catalina.

PR #1329, opened 2020-07-08
ΔCC½ histogram: plot normalised values

rjgildea:
This makes the plots easier to interpret wrt to the stdcutoff parameter for filtering bad data sets.

Don't multiply cc values by 100 - pretty much every else we report them in the range +/-1


jbeilstenedmands:
I'm generally on board with this change, although I think we might need to update the log output as I believe some confusion could be caused (I myself was confused, thought there was an error here but turns out not).
I tried on a dataset of four good sweeps:

```
CC 1/2 mean: 0.761
CC 1/2 excluding group 0: 0.753
CC 1/2 excluding group 1: 0.743
CC 1/2 excluding group 2: 0.746
CC 1/2 excluding group 3: 0.737
Dataset: 0, ΔCC½: 0.009
Dataset: 2, ΔCC½: 0.015
Dataset: 1, ΔCC½: 0.018
Dataset: 3, ΔCC½: 0.024
```

newplot (15)

Looking at the log output, all data are good and improve the overall CC1/2. However when I first looked at the plot, I thought there were too `bad` datasets due to negative sigma. However there is nothing wrong here, except my initial interpretation. This is simply because the delta CCs are calculated relative to CC1/2 mean, whereas the sigma plot is the relative variation within the delta CCs (you are correct that this plot is useful, as this is what is used to determine excluded datasets). For clarity, here is what was plotted previously:
newplot (14)

So perhaps it is most useful to show both plots, with a clearer description for the sigma plot that this is the relative spread within delta CCs?

ndevenish:
Okay, this one has mildly defeated me, since there seem to have been quite a few changes to the underlying code (e.g. `cc_half_algorithm.py` bugfixes, like https://github.com/dials/dials/commit/fab298780ebb4785d76fb3b157ed0c045b7d001b) since this branch was made. If we want to continue with this branch, this needs to be gone over with a fine comb and knowledge of the original intent of the restructuring.

Issue #1175, opened 2020-02-28
Python 2.7 deprecation timeline


Anthchirp:
As we are all aware Python 2.7 is now well [past its use-by date](https://pythonclock.org/).
DIALS will need to continue supporting Python 2.7 for a while as we still rely on this for a number of reasons (DLS operations, CCP4 bundling, Phenix bundling, ...)

This ticket outlines my proposal how we can support this critical need without obstructing ongoing development effort.

# Executive summary

On March 15 we will start the release branch for DIALS 2.2. DIALS 2.2 will support both Python 2.7 and 3.6. [Like DIALS 1.14](https://github.com/dials/dials/wiki/Version-history), this will be a long-term-supported (LTS) release, which means we will provide support and builds, initially until the end of the year. This may be extended if there is ongoing need for supported builds.

Meanwhile we will reduce the Python 2.7 compatibility in the master branch in a phased manner over a period of 6 months. This is to avoid a scenario where massive changes are introduced that then make it difficult to backport code to the 2.2 release branch.

# Phase 1 -- 15th of March 2020
> DIALS 2.2 branch date.

We stop officially providing or linking to any 2.7 nightly builds

We aim to keep Python 2.7 compatibility in dials, dxtbx, and xia2 `master` branches and refrain from any large-scale refactoring to not artifically make maintaining the 2.2 LTS release difficult. **Python-3-only code is allowed in new files and where retaining Python 2 support is unreasonable or impossible** (multiprocessing case). Where this causes tests to fail we will set up a _"skip-this-test-on-python-2"_ fixture.

We will reduce the MacOS testing regime at Diamond to reduce the demand on our build machine.

#### Jenkins
* Disable MacOS 2.7 builds for xia2
* Disable MacOS 2.7 DIALS regression builds
#### Travis
* no changes

<table><tr><th>repository</th><th>python 3 allowed in...</th><th>Python 2 testing coverage</th></tr>
<tr><td>dxtbx</td><td>new files and where Py2 support is unreasonable or impossible</td><td>Everything except newly excluded tests</td></tr>
<tr><td>dials</td><td>new files and where Py2 support is unreasonable or impossible</td><td>Linux: Everything except excluded, MacOS: No regression tests</td></tr>
<tr><td>xia2</td><td>new files and where Py2 support is unreasonable or impossible</td><td>Linux: Everything except excluded, MacOS: No test coverage</td></tr>
</table>

# Phase 2A -- 15th of May 2020
[...]

PR #1330, opened 2020-07-09
Add html report to resolutionizer

rjgildea:
Refactoring of resolutionizer code:
- Extract logic of estimating a resolution limit by fitting a curve to raw y_obs in resolution bins to a separate function to try and reduce code duplication.
- Use scitbx.math.curve_fitting.univariate_polynomial in place of local implementation of the same.
- Add module-level resolution_cc_half function
- Add tests for fitting functions
- Move cc_half_plots to module-level function in report/plots.py
- Remove png plots (plot=True option)
- Add simple_report jinja2 template for a single panel report with graphs
- Set max_iterations=100 in curve_fitting.univariate_polynomial_fit() to ensure fit completes in finite time
- Add json output

Note: adds dependency on `dataclasses` module - requires [backport](https://pypi.org/project/dataclasses/) on Python 3.6


rjgildea:
Example output:
```
$ dials.resolutionizer $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.{expt,refl} completeness=0.5 cc_half=0.8 rmerge=0.5 i_mean_over_sigma_mean=2 
...
Resolution cc_half:       1.24
Resolution I/sig:         1.20
Resolution Mn(I/sig):     1.24
Resolution Mn(I)/Mn(sig): 1.35
Resolution rmerge:        1.33
Resolution completeness:  1.20
```
Screenshot 2020-07-09 at 22 11 15
Screenshot 2020-07-09 at 22 11 24

algorithms/scaling/scale_and_filter.py
```diff
@@ -5,7 +5,7 @@
 import math
 from collections import OrderedDict
 
-from dials.report.plots import ResolutionPlotterMixin
+from dials.report.plots import d_star_sq_to_d_ticks
```
graeme-winter, line 8: 👍  this rename is really helpful 🙂

algorithms/statistics/cc_half_algorithm.py
```diff
@@ -590,9 +590,12 @@ def output_html_report(self):
                 ]
             )
             env = Environment(loader=loader)
-            template = env.get_template("delta_cc_half_report.html")
+            template = env.get_template("simple_report.html")
```
graeme-winter, line 593: 👍

algorithms/symmetry/__init__.py
```diff
@@ -373,9 +371,9 @@ def _resolution_filter(resolutionizer, min_i_mean_over_sigma_mean, min_cc_half):
     d_min_cc_half = 0
     if min_i_mean_over_sigma_mean is not None:
         try:
-            d_min_isigi = resolutionizer.resolution_i_mean_over_sigma_mean(
-                min_i_mean_over_sigma_mean
-            )
+            d_min_isigi = resolutionizer.resolution(
+                "i_mean_over_sigi_mean", limit=min_i_mean_over_sigma_mean
```
graeme-winter, line 375: This API documented somewhere? Assuming yes, and I have not yet hit it in the review, but I wonder if `"i_mean_over_sigi_mean"` would be better as an ersatz ENUM e.g. `resolutionizer.I_MEAN_OVER_SIGI_MEAN`?

algorithms/symmetry/__init__.py
```diff
@@ -349,7 +349,6 @@ def resolution_filter_from_array(intensities, min_i_mean_over_sigma_mean, min_cc
     """Run the resolution filter using miller array data format."""
     rparams = phil_defaults.extract().resolutionizer
     rparams.nbins = 20
-    rparams.plot = False
     resolutionizer = Resolutionizer(intensities, rparams)
```
graeme-winter, line 352: I note in some of the `command_line` code below we have `r = Resolutionizer()` - which could help readability esp. if we chose to go down the `r.ENUM` route

command_line/resolutionizer.py
```diff
@@ -1,7 +1,9 @@
 from __future__ import absolute_import, division, print_function
```
graeme-winter, line 2: General comment: I remain to be convinced that `dials.resolutionizer` is a _good_ name 🙂

dagewa: `dials.estimate_resolution` > `dials.resolution` > `dials.resolutionizer` IMHO

graeme-winter: `dials.estimate_resolution` is verbose but correct, and passes the Ronseal test...

algorithms/symmetry/__init__.py
rjgildea: @graeme-winter  is f41018f the kind of thing you had in mind?

graeme-winter: ☝️  is _precisely_ what I had in mind thank you, and will help you in the future

util/resolutionizer.py
```diff
@@ -27,16 +28,14 @@
...
-metrics = (
-    "cc_half",
-    "cc_ref",
-    "isigma",
-    "misigma",
-    "i_mean_over_sigma_mean",
-    "rmerge",
-    "completeness",
-)
+class metrics(enum.Enum):
```
graeme-winter, line 31: Just checked at this is supported in late 2.7 and since 3.4 so all good 🙂

command_line/resolutionizer.py
rjgildea: Resolved in c004a30

Anthchirp:
The Travis builds are currently failing, which are our Python 2.7 builds. Looks like this PR introduces some Python 3 only syntax to some `test_*.py` files, which means they need to be excluded from all testing. I've done this in the commit above.
Other tests can be parsed in Python 2.7 but fail later on. These need to be reviewed, and if they legitimately fail in Python 2 then the python3 fixture needs to be added to the test, ie.
`def test_function():` -> `def test_function(python3):`

See #1175 for more information and the current policy regarding the level of Python 2 support and allowed Python 3

PR #1333, opened 2020-07-10
dials.import: pass image_range to ExperimentListTemplateImporter

rjgildea:
This means that it is possible to use `image_range=` in combination with `template=` when there are missing images, as long as the missing images are outside of the defined `image_range`.

Add tests to cover the new behaviour.

Depends on https://github.com/cctbx/dxtbx/pull/191


test/command_line/test_import.py
```diff
@@ -300,6 +300,40 @@ def test_extrapolate_scan(dials_data, tmpdir):
     assert exp[0].identifier != ""
 
 
+def test_template_with_missing_image(dials_data, tmpdir):
+    # All the images except image #4
+    images = dials_data("centroid_test_data").listdir(
+        fil="centroid_*[1,2,3,5,6,7,8,9].cbf", sort=True
+    )
+    for image in images:
+        image.copy(tmpdir)
```
Anthchirp, line 309: It would be better if we could avoid copying the files, eg. by creating symlinks instead:
```python
def test(tmp_path):  # instead of tmpdir
  ...
  for image in images:
     (tmp_path / image.basename).symlink_to(image)
  ...
  procrunner.run(
            ["dials.import", f"template={tmp_path.joinpath('centroid_####.cbf')}",]
            ...
            working_directory=tmp_path,
  ...
```

This also uses the newer `tmp_path` instead of `tmpdir`, which works identically but returns a [native Python `pathlib` object](https://docs.python.org/3/library/pathlib.html).

Further this allows splitting up the tests without any further cost.

ndevenish: I'm unclear over whether using symlinks still causes problems with windows testing?

ndevenish:
Something something `1333.{bugfix|feature}`. Since we didn't claim this to work before, maybe this is a feature.

test/command_line/test_import.py
dagewa: almost certain it does. Could revert to copy on Windows only?

Anthchirp: [Admin privileges are no longer required if you enable developer mode](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) on Windows 10.
Which a developer might consider doing. This is also the default on Azure.

Would suggest catching the resulting OSError on Windows and skipping the test with a relevant message.

rjgildea: Is it really that big an issue copying a small handful of files to a temporary directory, if the alternative (symlinks) may not work on all platforms and require catching any subsequent errors? I could copy _fewer_ files if that helps.

rjgildea:
Added test currently failing because it depends on cctbx/dxtbx#191.

test/command_line/test_import.py
Anthchirp: Disk space usage is a thing on our VMs. While individually this doesn't matter much it leads down a death-by-a-thousand-cuts path. I'm happy to make the necessary changes.

PR #1335, opened 2020-07-13
Change numpy import commands

Anthchirp:
Follow the convention of importing `numpy as np`

### DIFF
--- a/test/algorithms/shoebox/test_mask_overlapping.py
+++ b/test/algorithms/shoebox/test_mask_overlapping.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import numpy as np
+
 
 def predict_reflections(sequence, crystal):
     from dials.algorithms import shoebox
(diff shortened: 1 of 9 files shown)


graeme-winter:
Thank you for this, I think it’ll help new people and I agree it’s good to follow conventions

test/algorithms/shoebox/test_mask_overlapping.py
```diff
@@ -105,7 +107,6 @@ def tst_non_overlapping(reflections, non_overlapping, image_size):
 
 def tst_overlapping(reflections, overlapping, adjacency_list, image_size):
```
graeme-winter, line 108: ISTR convention is we now call test functions `test_` not `tst_` - does this mean this test is missed by `pytest`?

Anthchirp: No, they are helper functions for one main `test()` function. This is pretty old code from pre-pytest days that was just directly converted. Nowadays we would lay it out differently.

graeme-winter: 👍

PR #1336, opened 2020-07-13
Officially support Python 3.7 and 3.8

### LINKED ISSUE #1236: Python 3.7 and 3.8 support

Anthchirp:
Closes #1236 

Disables `find_spots_server` test on MacOS+Python 3.8+ as described in https://github.com/dials/dials/issues/1236#issuecomment-655183518

### DIFF
--- a/newsfragments/1236.feature
+++ b/newsfragments/1236.feature
@@ -1,0 +1,1 @@
+DIALS now supports Python 3.7 and 3.8
--- a/test/command_line/test_find_spots_server_client.py
+++ b/test/command_line/test_find_spots_server_client.py
@@ -16,7 +16,7 @@ def start_server(server_command, working_directory):
 
 def test_find_spots_server_client(dials_data, tmpdir):
     if sys.hexversion >= 0x3080000 and sys.platform == "darwin":
-        pytest.skip("find_sports server known to be broken on MacOS with Python 3.8+")
+        pytest.skip("find_spots server known to be broken on MacOS with Python 3.8+")
 
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     s.bind(("", 0))
--- a/test/command_line/test_find_spots_server_client.py
+++ b/test/command_line/test_find_spots_server_client.py
@@ -4,6 +4,7 @@ import multiprocessing
 import procrunner
 import pytest
 import socket
+import sys
 import time
 import timeit
 from xml.dom import minidom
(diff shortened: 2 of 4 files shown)


phyy-nx:
Hi, I do know of folks still using dials.find_spot_server and
dials.find_spot_client.  It would be good not to lose testing for this.

test/command_line/test_find_spots_server_client.py
```diff
@@ -14,6 +15,9 @@ def start_server(server_command, working_directory):
 
 
 def test_find_spots_server_client(dials_data, tmpdir):
+    if sys.hexversion >= 0x3080000 and sys.platform == "darwin":
+        pytest.skip("find_sports server known to be broken on MacOS with Python 3.8+")
```
rjgildea, line 19:
```suggestion
        pytest.skip("find_spots server known to be broken on MacOS with Python 3.8+")
```

graeme-winter:
It's probably fair to say that servers are not a core part of a data processing package: I would support moving this to an external package though making it trivially importable by the end user. To be precise: to the vast majority of end users who are actually processing data this is not useful: only for facility operators is this relevant (and then I could probably count the number of installations without taking my socks off)

graeme-winter:
Thinking more about it, the number of people running find_spots_server in macOS in a production environment is probably 0 => disabling the test seems pretty reasonable in that situation.

test/command_line/test_find_spots_server_client.py
```diff
@@ -14,6 +15,9 @@ def start_server(server_command, working_directory):
 
 
 def test_find_spots_server_client(dials_data, tmpdir):
+    if sys.hexversion >= 0x3080000 and sys.platform == "darwin":
```
graeme-winter, line 18: Makes me wonder how many places we look at the version in a less reliable manner 🙄

newsfragments/1236.feature
```diff
@@ -0,0 +1 @@
+DIALS now supports Python 3.7 and 3.8
```
graeme-winter, line 1: 👍

Anthchirp:
Happy to keep the find_spots_server test. However right now it is broken. Well, technically _the test_ is fine, it's _the server_ that is broken. As I mentioned in the [linked issue](https://github.com/dials/dials/issues/1236#issuecomment-655183518
) the proper fix is to rewrite the server to properly follow python server patterns, or use one of the many python packages that are around to support these things, such as flask.

This obviously introduces extra dependencies that are very much not core-dials-related, so I like the idea of putting this into an external package.

For the purposes of this PR though I believe this is out of scope.

ndevenish:
It might be better to have a separate package, but I don't see any harm in it remaining in here. We know the test failure is a quirk of platform and can always reenable the tests when the problem is fixed.

PR #1339, opened 2020-07-14
Avoid passing phil params object to library functions in export/merging

rjgildea:
This makes library functions inconvenient to use outside of calling from a command line script. Instead, pass parameters as explicit arguments and keywords.

Rename internal `integrated_data` variable in `export_mtz` to reflect that this doesn't only deal with integrated data now.


algorithms/merging/merge.py
```diff
@@ -114,76 +131,103 @@ def make_MAD_merged_mtz_file(params, experiments, reflections, wavelengths):
...
+    d_max=None,
+    combine_partials=True,
+    partiality_threshold=0.4,
+    anomalous=True,
+    use_internal_variance=False,
+    assess_space_group=False,
+    truncate=True,
+    n_residues=200,
+    n_bins=20,
+    show_wilson_stats=False,
```
jbeilstenedmands, line 196: Looking over the parameter list, I don't think we need the two show_* parameters. This information should always be logged, or if it is not very useful then perhaps demoted to debug level. Then log output can be suppressed in the usual way if not needed when being called by other code. If you agree I'll make the change.

PR #1343, opened 2020-07-20
dials.integrate: give option to not output unintegrated reflections

### LINKED ISSUE #1342: dials.integrate: should not copy input to output
Assert: in `dials.integrate` we should not copy the input data to the output file. Like `dials.find_spots`, `dials.integrate` is "special" in that it accesses the raw data. Currently it _copies_ all the input reflections to the output as well as adding the integrated reflection output. I believe that this is _not_ desirable behaviour - in export we need to strip them out, every time I _look_ at the output I have to strip them out. 

Not only is this unnecessary and inconvenient we pay for this in storage and run time

Storage
-------

To illustrate, stripping the unintegrated reflection data with:

```
import sys

fin = sys.argv[1]
fout = fin.replace(".refl", "_strip.refl")

data = flex.reflection_table.from_file(fin)
data = data.select(data.get_flags(data.flags.integrated))
data.as_file(fout)
```

cuts a lot of data:

```
Silver-Surfer integrate :) $ du -hs 11*refl
962M	11_integrated.refl
761M	11_integrated_strip.refl
```

~ 25% of the file size which is replicated at _every_ step downstream e.g. symmetry and scaling to give a substantial aggregate improvement in disk usage. 

Time
----

Demonstration of the time saving by running `dials.symmetry` on the integrated data: the only difference here is the lack of need to read and write the unintegrated reflections:

```Silver-Surfer integrate :) $ for j in 1 2 3 ; do time dials.symmetry 11_integrated.expt 11_integrated.refl > /dev/null; done

real	1m35.731s
user	1m5.560s
sys	0m12.466s

real	1m34.454s
user	1m5.349s
sys	0m11.984s

real	1m34.085s
user	1m6.500s
sys	0m12.516s
```

vs.

```
Silver-Surfer integrate :) $ for j in 1 2 3 ; do time dials.symmetry 11_integrated.expt 11_integrated_strip.refl > /dev/null; done

real	1m28.774s
user	1m4.625s
sys	0m10.121s

real	1m21.750s
user	1m5.293s
sys	0m9.892s

real	1m20.753s
user	1m3.889s
sys	0m9.884s
```

So saving of ~ 14s on ~100s (which is substantial) as well as a saving in memory use. 

Proposal
--------

Propose the following fixes:
- by default do not include unintegrated reflections in the output reflection data
- add a PHIL parameter to copy input to output e.g. `copy_input_reflections`

I believe that this will also align the program with what a user may expect. Comments welcome, will prepare a PR for this.

graeme-winter:
Only output integrated reflections, using the filter `all=False` which takes any of profile fitted or summation integrated reflections. Fixes #1342. To restore previous behaviour use copy_input_reflections=True.

### DIFF
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -53,6 +53,7 @@ phil_scope = parse(
 
     output_unintegrated_reflections = True
       .type = bool
+      .expert_level = 2
       .help = "Include unintegrated reflections in output file"
 
     reflections = 'integrated.refl'
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -51,6 +51,10 @@ phil_scope = parse(
       .type = str
       .help = "The experiments output filename"
 
+    copy_input_reflections = False
+      .type = bool
+      .help = "Copy input reflections to output"
+
     reflections = 'integrated.refl'
       .type = str
       .help = "The integrated output filename"
@@ -553,6 +557,16 @@ def run_integration(params, experiments, reference=None):
 
     # Integrate the reflections
     reflections = integrator.integrate()
+
+    # Remove unintegrated reflections
+    if not params.output.copy_input_reflections:
+        keep = reflections.get_flags(reflections.flags.integrated, all=False)
+        logger.info(
+            "Removing %d unintegrated reflections of %d total"
+            % (keep.count(False), keep.size())
+        )
+
+        reflections = reflections.select(keep)
 
     # Append rubbish data onto the end
     if rubbish is not None and params.output.include_bad_reference:
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -51,7 +51,7 @@ phil_scope = parse(
       .type = str
       .help = "The experiments output filename"
 
-    output_unintegrated_reflections = False
+    output_unintegrated_reflections = True
       .type = bool
       .help = "Include unintegrated reflections in output file"
 
--- a/command_line/integrate.py
+++ b/command_line/integrate.py
@@ -51,9 +51,9 @@ phil_scope = parse(
       .type = str
       .help = "The experiments output filename"
 
-    copy_input_reflections = False
+    output_unintegrated_reflections = False
       .type = bool
-      .help = "Copy input reflections to output"
+      .help = "Include unintegrated reflections in output file"
 
     reflections = 'integrated.refl'
       .type = str
@@ -559,7 +559,7 @@ def run_integration(params, experiments, reference=None):
     reflections = integrator.integrate()
 
     # Remove unintegrated reflections
-    if not params.output.copy_input_reflections:
+    if not params.output.output_unintegrated_reflections:
         keep = reflections.get_flags(reflections.flags.integrated, all=False)
         logger.info(
             "Removing %d unintegrated reflections of %d total"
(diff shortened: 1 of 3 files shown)


graeme-winter:
🤦‍♂️  in committing I accidentally tried to use formatting around `all=False` which got lost as it was interpreted by the shell

command_line/integrate.py
```diff
@@ -51,6 +51,10 @@
       .type = str
       .help = "The experiments output filename"
 
+    copy_input_reflections = False
```
rjgildea, line 54: `copy_input_reflections` is incorrect/misleading here, as you are removing reflections that were predicted and extracted from the images, but flagged as not integrated (i.e. there was something "wrong" with the shoebox, e.g. overlapping the edge of a panel, or containing bad pixels etc). It _may_ also remove some input reflections (e.g. unindexed reflections), but I suspect in some cases they will mostly be newly predicted reflections that weren't in the original input.

graeme-winter: This is countable. I suspect in the cases I was looking at the majority of the reflections I am excluding here _are_ input ones for the simple reason that the reduction in file size was approximately the same as the input reflection file size. However I also recognise that in your recent experience the case could be different.

graeme-winter: Would `output_unintegrated_reflections` be a more precise option?

graeme-winter:
<- working through tests right now

graeme-winter:
OK, with changes in the branch the tests all pass. 

Running on a fairly hefty 6 sweep data set (3 wavelength MAD, 2 x 360° scans at each wavelength, Eiger 16M) get identical merging stats in the results. Disk usage for the full processing drops from 68G to 63G (which is still excessive, but 10% less so)

graeme-winter:
Wall-clock time wise (not a statistical sample) - I matched up the equivalent tasks and plotted wall-clock-time before and after - looks like:

times

i.e. on average the changes in here reduce the wall clock time of tasks, in some cases substantially.

command_line/integrate.py
graeme-winter: Believe so so made change...

graeme-winter:
Default is changed to no change in behaviour so we can evaluate effects of this e.g. in xia2.

ndevenish:
Perfectly happy to have as option for evaluation

newsfragments/1343.feature
```diff
@@ -0,0 +1,2 @@
+dials.integrate: add option to remove unintegrated data from output reflection file, output_unintegrated_reflections=False
```
ndevenish, line 1:
```suggestion
Experimental feature in ``dials.integrate``: Specify ``output_unintegrated_reflections=False`` to discard unintegrated data from output reflection file
```

test/command_line/test_integrate.py
```diff
@@ -26,6 +26,7 @@ def test2(dials_data, tmpdir):
             "modified_input.json",
             "profile.fitting=False",
             "integration.integrator=3d",
+            "output_unintegrated_reflections=True",
```
ndevenish, line 29: I'd prefer not adding this everywhere where it isn't needed (because of the default) - it's verbose, and makes it harder to change/remove/rename

command_line/integrate.py
```diff
@@ -51,6 +51,10 @@
       .type = str
       .help = "The experiments output filename"
 
+    output_unintegrated_reflections = True
+      .type = bool
+      .help = "Include unintegrated reflections in output file"
```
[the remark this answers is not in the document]
graeme-winter: 10-4

test/command_line/test_integrate.py
graeme-winter: Done

graeme-winter:
@ndevenish changes made as suggested 👍

PR #1348, opened 2020-07-22
dials.background: allow analysis of corrected data

huwjenkins:
The plot in Supplementary Figure 1 of https://doi.org/10.1101/2020.04.30.061895 produced with `dials.background` appears to show the output of get_corrected_data - i.e. (raw_data - pedestal)/gain. 

Even if this is not the case it would be useful to be able to plot this for detectors where a pedestal is applied. This PR adds that option.

### DIFF
--- a/command_line/background.py
+++ b/command_line/background.py
@@ -121,7 +121,7 @@ def run(args):
         pyplot.show()
 
 
-def background(imageset, indx, n_bins, corrected, mask_params=None):
+def background(imageset, indx, n_bins, corrected=False, mask_params=None):
     from dials.array_family import flex
     from libtbx.phil import parse
     from scitbx import matrix
--- a/command_line/background.py
+++ b/command_line/background.py
@@ -28,6 +28,9 @@ n_bins = 100
 images = None
   .type = ints
   .help = "Images on which to perform the analysis (otherwise use all images)"
+corrected = False
+  .type = bool
+  .help = "Use corrected data (i.e after applying pedestal and gain) in analysis"
 plot = False
   .type = bool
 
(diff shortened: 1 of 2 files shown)


dagewa:
Hi @huwjenkins, I just checked back through my notes. Apparently the figure was created with the command
```
dials.background collection1_1.35m_0.85s_nogain_c2_20_num7_1/pixel_correct/refined_experiments.json masking.d_max=5 images=18 masking.d_min=2.0 plot=True
```
I just ran that again now and get (please excuse the ridiculously-sized window title)
Screenshot from 2020-07-22 12-58-33

so it really was the raw data that was used. The plot is slightly different this time because I didn't actually have the refined geometry to hand, so the resolutions were not quite the same.

dagewa:
I copied the refined experiments over and reproduced the plot, just to be absolutely sure I was using the same dataset.
Figure_1
It does look like 2x2 binning (`image_size: {2048,2048}`), however I was used to much worse negative bias too from other datasets. I'm not sure what they did on these datasets to improve matters. I can try to find out.

I think your PR is fine by the way and will approve. The contributing guidelines haven't been updated, but these days we include a newsfragment in each PR. You can also add yourself to AUTHORS if you like. @ndevenish will correct me if I got the protocol wrong ;-)

command_line/background.py
```diff
@@ -114,7 +121,7 @@ def run(args):
         pyplot.show()
 
 
-def background(imageset, indx, n_bins, mask_params=None):
+def background(imageset, indx, n_bins, corrected, mask_params=None):
```
ndevenish, line 124:
```suggestion
def background(imageset, indx, n_bins, corrected=False, mask_params=None):
```
I believe this is possibly the origin of the failing test (is called from here: https://github.com/dials/dials/blob/master/util/best.py#L13), and making this a defaulted parameter was something I was going to suggest anyway

huwjenkins: Thanks! I have done that - that does appear to be why the test fails - apologies for not checking before raising PR.

ndevenish:
Looks good to me also - could you add a `newsfragments/1348.feature` file with a one-line description of the addition, and feel free to add yourself to the `AUTHORS` file in the root?

command_line/background.py
ndevenish: That's what the CI system is for :)

ndevenish:
Thanks for the contribution, this is in master now and will be released into Dials 3.1.0, which is currently scheduled for 17th August.

dagewa:
I don't understand this detector, from the gain to the negative bias. Other datasets from the same detector appear different! This bothers me.

PR #1350, opened 2020-07-23
Remove unrefined reflections in index

### LINKED ISSUE #1349: dials.index: if refinement fails in multi-lattice indexing the experiment id remains assigned
```
dials.index split_07602.* stills.indexer=sequences indexing.method=real_space_grid_search space_group=P4 unit_cell=58,58,150,90,90,90 max_lattices=8 beam.fix=all detector.fix=all
```

Data attached as split_07602.zip

[split_07602.zip](https://github.com/dials/dials/files/4960674/split_07602.zip)

Test code:

```
from dxtbx.model.experiment_list import ExperimentList

refl = flex.reflection_table.from_file("indexed.refl")
expt = ExperimentList.from_file("indexed.expt")

ids = set(refl["id"]) - set([-1])

for i in ids:
    e = expt[i]
```

verifies that every experiment id corresponds to an experiment... however following 

```
Too few reflections to parameterise Crystal8Phi1, Crystal8Phi2, Crystal8Phi3, Crystal8g_param_0, Crystal8g_param_1.
Try modifying refinement.parameterisation.auto_reduction options
```

we find that the experiment id remains assigned. Interesting that it _does not_ break refinement, only integration when it tries to split the data by experiment. 

PR incoming

graeme-winter:
If indexing works and refinement fails in multi-lattice case, can have experiment ids assigned which do not correspond to experiments in the output. This change simply removes the experiment ids where there is no refined experiment.

Fixes #1349


graeme-winter:
Can add a test once https://github.com/dials/data/pull/210 merged

Test can use 

```
from dxtbx.model.experiment_list import ExperimentList

refl = flex.reflection_table.from_file("indexed.refl")
expt = ExperimentList.from_file("indexed.expt")

assert max(refl["id"]) + 1 == len(expt)
```

to ensure all indexed reflections have experiments

graeme-winter:
Added test which depends on dials_data PR

graeme-winter:
@rjgildea test data added so will welcome your review

algorithms/indexing/indexer.py
```diff
@@ -705,7 +705,18 @@ def index(self):
...
+        # in case refinement failed, clean the reflections which may have had
+        # indices assigned to an experiment which was not saved...
+        refl = self.refined_reflections
+        expt = self.refined_experiments
+        spare = set(refl["id"]) - set(range(len(expt))) - set([-1])
+        for s in spare:
+            sel = refl["id"] == s
+            logger.info("Removing %d reflections with id %d" % (sel.count(True), s))
+            refl["id"].set_selected(sel, -1)
+            refl.unset_flags(sel, refl.flags.indexed)
```
rjgildea, line 717: This code would probably be better as part of the try/except where `self.refine()` is called, i.e. when the failed refinement was actually trapped.

algorithms/indexing/indexer.py
```diff
@@ -705,7 +705,18 @@ def index(self):
                 rotation_matrix_differences(self.refined_experiments.crystals())
             )
 
-        self._xyzcal_mm_to_px(self.experiments, self.refined_reflections)
+        # in case refinement failed, clean the reflections which may have had
+        # indices assigned to an experiment which was not saved...
+        refl = self.refined_reflections
+        expt = self.refined_experiments
```
rjgildea, line 711: I would prefer to just use the original variables as they are, without re-assigning to a shorter name for a few lines of code.

newsfragments/1350.bugfix
```diff
@@ -0,0 +1 @@
+dials.index: if multi-lattice indexing and last lattice not refined, remove experiment ids that correspond to that lattice.
```
rjgildea, line 1:
```suggestion
dials.index: In multi-lattice indexing, ensure that reflections corresponding to lattices where refinement failed are flagged as unindexed.
```

test/command_line/test_index.py
```diff
@@ -0,0 +1,31 @@
...
+    assert not result.returncode and not result.stderr
+    assert tmpdir.join("indexed.expt").check(file=1)
+    assert tmpdir.join("indexed.refl").check(file=1)
+
+    refl = flex.reflection_table.from_file(tmpdir / "indexed.refl")
+    expt = ExperimentList.from_file(tmpdir / "indexed.expt")
+
+    ids = set(refl["id"]) - set([-1])
+
+    assert max(ids) + 1 == len(expt)
```
rjgildea, line 31: If all you're doing is comparing the max refl["id"] then there is no need to mess around with sets:
```suggestion
    assert flex.max(refl["id"]) + 1 == len(expt)
```

algorithms/indexing/indexer.py
graeme-winter: Acknowledged

However the full variable names did make the code rather harder to read since the lines become _very_ long. 

I'll have a look while fixing this.

algorithms/indexing/indexer.py
graeme-winter: 👍

Agreed working on this now

graeme-winter:
> Is there any reason to put the new test in a separate file to the rest of the existing high-level indexing tests?
> 
> https://github.com/dials/dials/blob/master/algorithms/indexing/test_index.py

This is a test of the command-line program so I put the tests where I would expect the test go to... which I think is reasonable.

graeme-winter:
@rjgildea changes made according to your suggestions

graeme-winter:
> > Is there any reason to put the new test in a separate file to the rest of the existing high-level indexing tests?
> > https://github.com/dials/dials/blob/master/algorithms/indexing/test_index.py
> 
> This is a test of the command-line program so I put the tests where I would expect the test go to... which I think is reasonable.

mmk well moved over - we can revisit this more widely I guess but this is consistent with the code base as it stands.

graeme-winter:
Rerunning all tests before squash merging.

graeme-winter:
@ndevenish  will leave this one to rest for a bit on master but will want to pick across to release branch I think (not so sure about backport)

ndevenish:
> will want to pick across to release branch I think (not so sure about backport)

That is backporting

graeme-winter:
> > will want to pick across to release branch I think (not so sure about backport)
> 
> That is backporting

Sorry, I meant back to 2.2 series 🙂

Anthchirp:
This was merged with failing tests.
The tests did not lie.
Tests are now failing on master.
This should not have been merged.

PR #1357, opened 2020-07-26
Enable type checking on DIALS using mypy

Anthchirp:
Add a pre-commit and GitHub action to do type checking. Add section about mypy to contribution guidelines.

There are a number of errors (as indicated by the failing GitHub action) that would need to be addressed - either by fixing or ignoring - before this can be merged.

The existing errors also serve as an example as to what to expect.

Opinions on whether we should do this are welcome.


Anthchirp:
I watched [Dustin Ingram's Pycon 2020 talk (Static Typing in Python)](https://www.youtube.com/watch?v=ST33zDM9vOE) yesterday, and now think we should also consider running [pytype instead of mypy](https://docs.google.com/presentation/d/1GYqLeLkknjYaYX2JrMzxX8LGw_rlO-6kTk-VNPVG9gY/edit#slide=id.g58dc4d159c_0_141)

dagewa:
Does the type checker modify the source to add type hints first, or is it done just based on the state of the source it encounters?

Anthchirp:
Enabling type checking does not in itself modify the source code.

There is software out there that does runtime analysis on code (based on the test suite) to automatically add type annotations, and we can consider doing this at a later date. We do not need to do this to enable type checking, and so I think this is out of scope for now.

We still get benefits without having full annotation coverage. A trivial example: the type checker found `dials/report/test_analysis.py:4: error: Module 'mock' has no attribute 'mock'`, which is a real bug - the import statement is wrong.

Anthchirp:
I had a play with pytype. I'd say it is not suitable for pre-commits, as it is rather slow. It also finds tons of issues with the current code base. Looking through those issues though some of them seem to be genuine bugs. So it's probably worth adding it as an additional GitHub action.

To demonstrate how ridiculously powerful this is, here is one error it found:
```
[56/574] check dials.util.scale_down_image
FAILED: /scratch/wra62962/files/dials/modules/dials/.pytype/pyi/dials/util/scale_down_image.pyi 
/scratch/wra62962/files/dials/build/../conda_base/bin/python3.6 -m pytype.single --disable import-error --imports_info /scratch/wra62962/files/dials/modules/dials/.pytype/imports/dials.util.scale_down_image.imports --module-name dials.util.scale_down_image -V 3.6 -o /scratch/wra62962/files/dials/modules/dials/.pytype/pyi/dials/util/scale_down_image.pyi --analyze-annotated --nofail --quick /scratch/wra62962/files/dials/modules/dials/util/scale_down_image.py
File "/scratch/wra62962/files/dials/modules/dials/util/scale_down_image.py", line 65, in write_image_from_flex_array: Function bytes.split was called with the wrong arguments [wrong-arg-types]
         Expected: (self, sep: Optional[bytes] = ..., ...)
  Actually passed: (self, sep: str)
Called from (traceback):
  line 85, in scale_down_image
File "/scratch/wra62962/files/dials/modules/dials/util/scale_down_image.py", line 75, in write_image_from_flex_array: unsupported operand type(s) for +: 'str' and 'bytes' [unsupported-operands]
  Function __add__ on str expects str
```
To unpack this: When the function `write_image_from_flex_array`
https://github.com/dials/dials/blob/d07482e2040335c744be41155fb69cbdd5745193/util/scale_down_image.py#L53-L65
is called from `scale_down_image`:
https://github.com/dials/dials/blob/d07482e2040335c744be41155fb69cbdd5745193/util/scale_down_image.py#L78-L85
the variable `header` comes from the output of the `read_image_to_flex_array()` call:
https://github.com/dials/dials/blob/d07482e2040335c744be41155fb69cbdd5745193/util/scale_down_image.py#L36-L50
This always returns a byte string, and you can't call `.split()` with a string on a byte string.

Anthchirp:
I'm moving `pytype` discussion out into a separate ticket, as I believe this to be useful on its own and independently of `mypy`

Anthchirp:
[Dials core meeting discussion outcome](https://dials.github.io/kb/core/20200917):
* Let's _not_ enable this straight-away
* Have an introductory lecture on typing, including references to PEP articles
* Allow people to run this stand-alone and outside of a pre-commit/pull request check
* Reconsider at some point in the future

ndevenish:
I think this has been entirely superseded by #1411 and no longer under discussion at this time.

Anthchirp:
Nope, these are independent things. This is held back pending my typing intro lecture.

Anthchirp:
Relevant minutes: https://dials.github.io/kb/core/20200917#typing-mypy

Anthchirp:
We should still do this. However, judging from the number and types of errors thrown up this is likely more tractable after we have moved DIALS to a src-layout. Closing for now.

Issue #1189, opened 2020-03-10
Shoeboxes displayed wrong for rotated detector models


phyy-nx:
Generally we have panels close to 90 degrees so It's not too surprising we haven't noticed this.  Here are two examples:

![image](https://user-images.githubusercontent.com/13471434/76335856-5d76f800-62b2-11ea-91b8-d45ba11b8e44.png)

Here I used rotate_detector.py (see #1182) to rotate `dials_regression/image_examples/ALS_831/q315r_lyso_001.img` 90 degrees, then ran `dials.find_spots rotated.expt`, `dials.index rotated.expt strong.refl`, and `dials.image_viewer indexed.*`.  Interestingly, because the detector is not constrained in `dials.index` by design, the detector Z drifted ~20 degrees-ish.  You can see the pixels are marked correctly, but the centroid cross is a bit funny and the red shoebox is wrong.

Another example:
![image](https://user-images.githubusercontent.com/13471434/76337488-aa5bce00-62b4-11ea-8e39-f9ea4cd18328.png)

Here I used rotate_detector.py on `dials_regression/image_examples/LCLS_cspad_nexus/idx-20130301060858401.cbf`, but changed the angle to 45 degrees, then used `dials.find_spots rotated.expt gain=25 filter.min_spot_size=2` and `dials.image_viewer rotated.expt strong.refl`.  Now you can see the centroid cross and shoebox are both flattened.

Note, this bug is in the master now, but can only be seen on multi-panel images until #1182 is merged.

dagewa:
> Interestingly, because the detector is not constrained in dials.index by design, the detector Z drifted ~20 degrees-ish
Could this be because the detector was rotated, but the rotation axis was not?

phyy-nx:
> > Interestingly, because the detector is not constrained in dials.index by design, the detector Z drifted ~20 degrees-ish
> 
> Could this be because the detector was rotated, but the rotation axis was not?

You mean the goniometer axis?  Likely!  Not relevant to the second case of course which is a still, but it's a good point for the single image sweep.

dagewa:
This issue seems to be because the shoebox display only takes the coordinates of diametric corners and draws a rectangle between them.
https://github.com/dials/dials/blob/2d515a3340b7ce39a7f0e21897c5c7aae5962e81/util/image_viewer/spotfinder_frame.py#L1387-L1388

dagewa:
With the fix in 62dab086f, the original example:
```
[...]

PR #1368, opened 2020-08-05
Rotate shoeboxes

dagewa:
See #1189 for details


graeme-winter:
@dagewa seriously that's all the diffs?

Dang

Though I would personally prefer this in the image frame rather than lab frame, this is nevertheless an improvement thank you 🙂

newsfragments/1189.bugfix
```diff
@@ -0,0 +1 @@
+Correctly show shoeboxes for rotated panels
```
graeme-winter, line 1:
```suggestion
dials.image_viewer: rotate shoeboxes to align with rotated detector panels
```

PR #1373, opened 2020-08-11
Make Python 3.8 the default developer version

Anthchirp:
As per decision in https://dials.github.io/kb/core/20200811.

I suggest merging this after the imminent DIALS 3.1 release.


Anthchirp:
This will only affect new developer installations and the version of Python we distribute in our installers.
No changes are required for existing developer installations.

Issue #1268, opened 2020-05-19
Reflection table flags has no "scaled"


graeme-winter:
`flex.reflection_table.flags` has indexed, integrated, many other things but not scaled. We can say that it was not scaled with `user_excluded_in_scaling` or `excluded_for_scaling` but that is not super helpful. 

Proposed solution:
 - define `flags.scaled`
 - use it 😉

rjgildea:
I _assume_ that this is just the complement of `flex.reflection_table.flags.bad_for_scaling`, which is itself the combination of `UserExcludedInScaling | OutlierInScaling | ExcludedForScaling`:
https://github.com/dials/dials/blob/327a64b4f232dbd01fdb6616de6fa325a84e1a0a/array_family/reflection_table.h#L149

E.g.:
https://github.com/dials/dials/blob/4bdd392671c09e9d20794deaf90433c669e5f70b/util/filter_reflections.py#L643-L648

https://github.com/dials/dials/blob/505749539cce136fef4b0acb4ac1a3e9daa54c05/algorithms/scaling/algorithm.py#L318

graeme-winter:
Does feel a bit ... cumbersome though right? ~which reflections are good~ these reflections are not bad for this reason and not bad for that reason too.

dagewa:
What does it mean for a reflection to not be `scaled`? Every reflection output has a scaled intensity given by application of the scaling model. So a reflection is scaled if it exists in scaled.refl, right?

graeme-winter:
```
from dials.array_family import flex
import sys

data = flex.reflection_table.from_file(sys.argv[1])

flags = data.flags.names

for f in sorted(flags):
    print(f"{f:40s} -> {data.get_flags(flags[f]).count(True)}")
```

Enumerate all possible flags, count how many reflections in a table are flagged, example output:

```
background_includes_bad_pixels           -> 9794
bad_for_refinement                       -> 0
bad_for_scaling                          -> 0
bad_reference                            -> 2762
bad_shoebox                              -> 0
bad_spot                                 -> 0
centroid_outlier                         -> 2463
dont_integrate                           -> 354
excluded_for_refinement                  -> 292
excluded_for_scaling                     -> 10262
failed_during_background_modelling       -> 23
failed_during_profile_fitting            -> 6577
failed_during_summation                  -> 5560
foreground_includes_bad_pixels           -> 5560
in_powder_ring                           -> 0
includes_bad_pixels                      -> 5382
indexed                                  -> 10526
[...]

PR #1377, opened 2020-08-14
Add a 'scaled' flag to the reflection table.

jbeilstenedmands:
Introduce a 'scaled' flag, which is set at the end of scaling to indicate which are the good scaled reflections (not excluded, not outliers etc). This is the inverse of the current 'bad_for_scaling' flag, which is unclear and prone to misuse, as indicated by the confusion in the discussion in #1268.

Ideally, the scaled flag will replace all uses of 'bad_for_scaling' outside scaling code - this has not been implemented yet, as it will require updating some of the regression test files (e.g. l_cysteine_4_sweeps_scaled). If we're happy with the proposed changes here, then once merged I can update the test files and implement further changes of bad_for_scaling to scaled in places such as `dials.util.filter_reflections`

### DIFF
--- a/.conda-envs/linux.txt
+++ b/.conda-envs/linux.txt
@@ -42,7 +42,7 @@ conda-forge::scipy
 conda-forge::scons
 conda-forge::setuptools
 conda-forge::six
-conda-forge::sphinx
+conda-forge::sphinx<3.2
 conda-forge::sqlite
 conda-forge::tabulate
 conda-forge::tqdm
--- a/.conda-envs/macos.txt
+++ b/.conda-envs/macos.txt
@@ -46,7 +46,7 @@ conda-forge::scipy
 conda-forge::scons
 conda-forge::setuptools
 conda-forge::six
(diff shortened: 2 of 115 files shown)


ndevenish:
I think we're ready to merge this when the conflicts are resolved

ndevenish:
What's the relation between `scaled` and `bad_for_scaling`? It seems to set scaled as the exact opposite quite often, but I can't tell if there are scenarios where they diverge.

ndevenish:
Also, needs a newsfragment (probably `.api`?)

jbeilstenedmands:
@ndevenish you are correct, as in the PR description, I'm moving over to the scaled flag in general outside of internal scaling code, although keeping bad_for_scaling for convenience within scaling (at least for now). Newsfragment incoming.

ndevenish:
D'oh, that's what I get for reading the code and not the description.

Is the long term idea to properly deprecate the bad_for_scaling?

Issue #66, opened 2016-01-21
dials.image_viewer isoresolution lines not available for 2theta!=0


Anthchirp:
dials.image_viewer resolution on mouseover does not match the drawn resolution rings on images taken with 2theta!=0.

I noticed that the resolution rings are not distorted to ellipses as one would expect, so probably the mouseover value is more trustworthy.

example file:

```
$ dials.image_viewer /dls/i19-1/data/2016/mt11397-3/TKR948/TKR948_06_01302.cbf
```

Mouseover resolutions on the ring marked 1.14A go from 1.4 (right) and 1.5 (top/bottom) to 2.7 (left)

rjgildea:
The resolution rings are meaningless at 2theta != 0, so should be ignored.
They assume that the detector is perpendicular to the beam.

Anthchirp:
So this would be a thing for the future then to provide resolution indicators. I23 might be interested as well.

graeme-winter:
Following discussion just now hyperbolic resolution lines will be a thing for 45<|2theta|<135 degrees; give we know the principle axes (and are free to select definitions of x, y) we should be able to plot these correctly.

Anthchirp:
Ellipses now appear to work properly. Next step hyperbolas, then multi-panel detectors.

Anthchirp:
The demand for hyperbolas seems rather limited, so the work required to display those is not proportionate at this time.

biochem-fan:
I reopen this old issue because 2theta != 0 configurations are quite common with RIGAKU's in house diffractometers (Synergy with HyPix) and small molecular crystallography. It is also necessary to support non flat detectors. For example, panels in the CITIUS detector being developed at SACLA/SPring-8 are at different depths; outer rows are shallower than inner rows.

TODO:
- support different camera lengths for each panel (draw arcs instead of a circle/ellipse)
- support hyperbola

biochem-fan:
> different camera lengths for each panel

This has been already available in `calculate_isoresolution_lines` in `util/image_viewer/spotfinder_frame.py`. Because of the way panels are projected in the image viewer, I missed it. Sorry for my false alarm.

> support hyperbola

This should be also possible with `calculate_isoresolution_lines` but somehow nothing is displayed in my RIGAKU Synergy datasets. I will check.

biochem-fan:
It turned out to be a simple logic error. I will submit a pull request later today.

biochem-fan:
I found this line is problematic:

[...]

PR #1387, opened 2020-09-01
Fix up some image viewer problems

### LINKED ISSUE #1385: image_viewer: "Save Image" broken for multi-panel detectors
Image viewer:
image (2)
Saved png:
image (3)

### LINKED ISSUE #1388: image_viewer:  Saving image as PDF fails because of undeclared dependency
An attempt to save a PDF image in the image viewer using "💾 Save As..." and selecting PDF format results in
```
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/slip_viewer/frame.py", line 877, in OnSaveAs
    from reportlab.lib.units import inch
ModuleNotFoundError: No module named 'reportlab'
```

It seems this is an undeclared dependency.

benjaminhwilliams:
Things this needs:
 - [x] Replace deprecated `wx` classes.
 - [x] A proper declaration of `reportlab` as a dependency, to resolve #1388 (and close #1390)
 - [x] A fix that fixes #1385.
 - [x] A fix for the stack trace below, produced by doing a PDF export with axis annotations switched on (the text of the axis labels exports in an incorrectly defined colour).

```
Only centered placement available when drawing text on pdf
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/slip_viewer/frame.py", line 1112, in OnSaveAs
    txt.setFillColor(textcolour)
  File "/home/ekm22040/DIALS/conda_base/lib/python3.6/site-packages/reportlab/pdfgen/textobject.py", line 99, in setFillColor
    raise ValueError('Unknown color %r' % aColor)
ValueError: Unknown color wx.Colour(0, 0, 0, 255)
```
 - [ ] A fix for the following when ticking 'Show ice rings' (suggesting `ecc` > 1, whatever `ecc` is):
```
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 2173, in OnUpdate
    self.GetParent().GetParent().update_settings()
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 1244, in update_settings
    self.draw_resolution_rings(unit_cell=unit_cell, space_group=space_group)
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 744, in draw_resolution_rings
    b = math.sqrt(a * a * (1 - (ecc * ecc)))
ValueError: math domain error
```
 - [x] A fix for the following when ticking 'Threshold pixels':
```
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 2173, in OnUpdate
    self.GetParent().GetParent().update_settings()
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 1211, in update_settings
    y, x = self.pyslip.tiles.flex_image.tile_readout_to_picture(
AttributeError: 'FlexImage_d' object has no attribute 'tile_readout_to_picture'
```


benjaminhwilliams:
That merge commit resolves #1385.

benjaminhwilliams:
Thanks for the nudge, @Anthchirp, I had been putting this off because of the difficulty of testing it.

I have now tested the PDF export by:
 - checking out the current state of this branch and updating all other repositories in the DIALS build to their latest state on 'master' branches,
 - removing `build`, `conda_base` and `miniconda` directories from my DIALS build location,
 - running `python bootstrap.py base build`,
 - (making a pot of tea),
 - running `dials.image_viewer <example_experiment_list>.expt <example_reflection_table>.refl`,
 - Turning on various annotations,
 - Saving the image as a PDF.

  Everything now seems to work, though the placement of text leaves something to be desired.  Saving an image with the reciprocal basis vector labels turned on gave the following warnings:
```
Only centered placement available when drawing text on pdf
Only centered placement available when drawing text on pdf
Only centered placement available when drawing text on pdf
```
Presumably, that's one for each of 'a*', 'b*', and 'c*'.

  For comparison, here's the image viewer at the point of saving the PDF:
image
and here's the PDF itself:
[test_image.pdf](https://github.com/dials/dials/files/5260721/test_image.pdf)

benjaminhwilliams:
And thanks for resoling the merge conflicts too @Anthchirp.

benjaminhwilliams:
Of course Codecov complains a lot because the image viewer code is basically untested.  I suspect that's such a large problem though that it's beyond the scope of this PR.

Anthchirp (approved):

benjaminhwilliams:
@nksauter can you confirm that this allows the DIALS image viewer to work as you expect on a clean build?  Perhaps you could copy my testing protocol ([detailed above](https://github.com/dials/dials/pull/1387#issuecomment-696636130)) and then check that you can get PDFs that look as you would expect without errors.

benjaminhwilliams:
On a second pass, I've spotted some issues that really count as regressions:
 - On ticking 'Show ice rings' I get the following:
```
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 2173, in OnUpdate
    self.GetParent().GetParent().update_settings()
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 1244, in update_settings
    self.draw_resolution_rings(unit_cell=unit_cell, space_group=space_group)
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 744, in draw_resolution_rings
    b = math.sqrt(a * a * (1 - (ecc * ecc)))
ValueError: math domain error
```
 - On ticking 'Threshold pixels' I get this:
```
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 2173, in OnUpdate
    self.GetParent().GetParent().update_settings()
  File "/home/ekm22040/DIALS/modules/dials/util/image_viewer/spotfinder_frame.py", line 1211, in update_settings
    y, x = self.pyslip.tiles.flex_image.tile_readout_to_picture(
AttributeError: 'FlexImage_d' object has no attribute 'tile_readout_to_picture'
```

So I'll try to get fixes for those ASAP.  Please raise other problems if you spot any.

benjaminhwilliams:
As to the ice rings problem, this seems to arise from a hard-baked assumption that the projection on the detector of the cone corresponding to each ice ring is an ellipse.  In fact, any conic section is valid and the projection will only be an ellipse when the half-aperture of the cone (the 2θ<sub>B</sub> angle, where θ<sub>B</sub> is the Bragg angle of the ice ring) does not exceed the angle between the detector surface and the incident beam vector (90° – 2θ<sub>d</sub>, where 2θ<sub>d</sub> is the 'detector two-theta' angle or the angle of the detector normal to the incident beam vector).

  I happened to test this with some data from a small-molecule experiment on the chemical crystallography beamline I19 at Diamond Light Source, which has a three- and a four-circle diffractometer and where data are routinely collected with 2θ<sub>d</sub> substantially greater than 0°.  It seems that in this case the ice rings with highest Bragg angle are projected as hyperbolas and/or a parabola.

  So while this isn't a regression as a result of changes introduced here, it is a bug that deserves to be fixed, especially for the chemical crystallographers that use DIALS.

benjaminhwilliams:
For the threshold pixels display problem, perhaps @rjgildea or @dagewa could provide some insight, since this was introduced in bc1dcb289faea587136dcd13caf797cff83fa872?

dagewa:
I'll have a look

dagewa:
The difference seems to be because before the changes `self.pyslip.tiles.flex_image` was a `iotbx_detectors_ext.generic_flex_image` when it got here, but after the changes it is a `iotbx_detectors_ext.FlexImage_d`

dagewa:
I don't know why that should be, but I note that `self.pyslip.tiles.flex_image.tile_readout_to_picture` is pretty common in that file (appears 8 times), so it doesn't seem wrong to be calling it

benjaminhwilliams:
OK, I wonder why that should happen.  I'll try to find time to take a closer look.  Thanks.

util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -296,14 +296,28 @@ def set_image(
...
 
-        self.flex_image = get_flex_image_multipanel(
-            brightness=self.current_brightness / 100,
-            panels=detector,
-            show_untrusted=self.show_untrusted,
-            image_data=image_data,
-            beam=self.raw_image.get_beam(),
-            color_scheme=self.current_color_scheme,
-        )
+        if len(detector) > 1:
```
dagewa, line 299: @benjaminhwilliams it appears that it is this change that causes the threshold pixel failure. That is, if I change this to `if True` then the failure does not occur

dagewa: It seems that the return types of `get_flex_image_multipanel` and `get_flex_image` are dissimilar.

dagewa: Yes, `get_flex_image`→ `FlexImage` whereas `get_flex_image_multipanel`→ `generic_flex_image` :thinking:

benjaminhwilliams: Yes, I've discovered the same in parallel. Without more documentation, I'm struggling to understand what each type is meant to represent (there are three, I think — FlexImage, FlexImage_d and generic_flex_image, which is a class in spite of the lack of camel case).

Anthchirp:
> As to the ice rings problem, this seems to arise from a hard-baked assumption
> that the projection on the detector of the cone corresponding to each ice ring
> is an ellipse. In fact, any conic section is valid

Yes, this is well known - and also: No. No such assumption was ever hard-baked.
I still remember the issue number (#66). There was some low-level library support present to draw ellipses, so I made them work. There was absolutely no support available to draw hyperbolae, and I found that to be too difficult at the time to implement from scratch. Plus there was no demand - and so they ended up at the bottom of the pile and I closed the issue 18 months later.

benjaminhwilliams:
OK, that issue was opened and closed before I even joined the project, so I'm playing catch-up here.  I guess 'hard-baked' is too strong a term — I only meant that the code only allows for ellipses, which I didn't realise before.  Sorry if I'm going over old ground.

benjaminhwilliams:
@dagewa do you have any single-panel data that we can test this with?  We now know that it works for multiple panels but I don't have any single-panel data to hand.  Or have you tested this already?

dagewa:
I tried it with images in `dials_regression/image_examples`, with I23 for multi panel and a randomly-chosen CCD image for single panel (`APS_22ID/`)

phyy-nx:
This PR reverts #1182, essentially breaking Rayonix single panel image support at XFELs where the detector is rotated 90 degrees.  It adds back in all the `len(detector) > 1` tests that #1182 removed.  I agreed with @graeme-winter that that would be ok if #1182 was restored _very soon_, in a way that fixes both the problems it originally solved and the issues here.  As an aside, I still think it would be better to fix master directly, rather than add back in bugs that were hard to fix in the first case.

[This comment](https://github.com/dials/dials/pull/1182#issuecomment-596764997) details the tests I ran to verify #1182 is working.  I've re-stated my commands here to assist in testing:

```
dials.import `libtbx.find_in_repositories dials_regression`/image_examples/ALS_831/q315r_lyso_001.img output.experiments=singlepanel.expt
dials.import `libtbx.find_in_repositories dials_regression`/image_examples/LCLS_cspad_nexus/idx-20130301060858401.cbf output.experiments=multipanel.expt
libtbx.python `libtbx.find_in_repositories dials`/util/image_viewer/slip_viewer/rotate_detector.py singlepanel.expt
mv rotated.expt singlepanel_rotated.expt
libtbx.python `libtbx.find_in_repositories dials`/util/image_viewer/slip_viewer/rotate_detector.py multipanel.expt
mv rotated.expt multipanel_rotated.expt
```
Then use dials.image_viewer on each of the .expt files in turn to see what happens.  On master, both single and multi panel images appear rotated correctly.  On this branch, multi-panel works but single panel is broken.

benjaminhwilliams:
Thanks @phyy-nx for the information and for the detailed description of how to test for regressions of #1182.  I fear that we're trying to create incompatible patches to different corner cases, but I'm sure that a fix for both is possible, given time and effort.  The expense of these patches is starting to prevent other work getting done though, so I don't really know how to proceed.  Meanwhile, I've poured some of my frustrations with the image viewer into #1428 and would welcome your opinions on that.

PR #1390, opened 2020-09-02
remove image viewer PDF export

Anthchirp:


util/image_viewer/slip_viewer/frame.py
```diff
@@ -873,251 +873,6 @@ def OnSaveAs(self, event):
...
-                        txt.setFont(fontname, fontsize * scale)
-                        txt.setFillColor(textcolour)
-                        txt.setStrokeColor(textcolour)
-                        txt.textLine(tdata)
-                        pdf_canvas.drawText(txt)
-
-            pdf_canvas.save()
-
-        self.update_statusbar("Writing " + file_name + "..." + " Done.")
-
```
benjaminhwilliams, line 1120: Even if we remove the PDF export, we would want to keep this little status bar message, no?

benjaminhwilliams (changes requested):
Aside from the small change below, I think there is value in retaining some sort of export functionality that retains the annotations provided by the image viewer.  The PNG export basically just coverts from raw image format to a bitmap without annotations.  See https://github.com/dials/dials/issues/1388#issuecomment-685867630.

Anthchirp:
Although #1387 has been stalled, I assume this the PDF export has been fixed up to an appropriate level so it is no longer considered for removal.

PR #1391, opened 2020-09-03
Sort imports and add isort to pre-commits

ndevenish:
This commit adds isort to the pre-commit checks, and applies isort to all the files in dials.

This is using isort >5, which solves a lot of the problems that happened with older versions; for the most part, imports are only sorted within a block of imports - any non-import code acts as a delimiter and imports won't be moved above them unless specifically requested, and comments on imports will be moved. Because of this, there is some extra pre-filtering work done to find cases where this is being overly cautious, but special cases like matplotlib work without manual intervention, which I feel means is ready for general dials use.

The import order is configured to create structured blocks in the form:
- Future imports
- Standard library
- External Third Party
- cctbx_project imports (libtbx, scitbx, etc)
- dxtbx
- dials
- relative imports

My intent is to eventually move towards floating the inline imports that are safe, like we did in dxtbx - there is a WIP that floats just the standard library and `six` imports [here](https://github.com/ndevenish/dials-fork/compare/isort...ndevenish:float_std), but this pull is just the basic in-place sorting for now.


ndevenish:
Okay there was some objections in slack on having dxtbx and cctbx as separate categories, so I've combined them

PR #1393, opened 2020-09-04
Trim scan

### LINKED ISSUE #1374: Trimming Scans Breaks Refinement With Multiple Image Sets
If multiple image sets are affiliated with a single scan, the automatic [trimming](https://github.com/dials/dials/blob/d9e9629744be5cea3496fe70befe928fb4425795/algorithms/refinement/refiner.py#L155) in the current DIALS version can create new truncated scan objects. This will trigger a [sanity check](https://github.com/dials/dials/blob/d9e9629744be5cea3496fe70befe928fb4425795/algorithms/refinement/parameterisation/configure.py#L490) and cause scan varying refinement to fail. I put together a minimal [example](https://github.com/kmdalton/dials-example) to illustrate this behavior.

dagewa:
* Put scan trimming behaviour under user control
* Deprecate the poorly named `trim_scan_edges` parameter and rename it `scan_margin`
* ~~Make it possible to see `warnings.warn` messages in the console log~~

Fixes #1374


algorithms/refinement/parameterisation/configure.py
```diff
@@ -105,6 +105,14 @@
               "of the serialized model small. At the moment, this only has"
               "an effect for crystal unit cell (B matrix) errors."
 
+    trim_scan_to_observations = True
+      .type = bool
+      .expert_level = 1
+      .help = "For scan-varying refinement, scan objects will be trimmed to"
+              "the range of observed reflections if this is true. This avoids"
+              "failures in refinement for cases where the extremes of scans"
+              "contain no data, such as when the crystal moves out of the beam."
```
Anthchirp, line 114:
```suggestion
      .help = "For scan-varying refinement, trim scan objects to"
              "the range of observed reflections. This avoids"
              "failures in refinement for cases where the extremes of scans"
              "contain no data, such as when the crystal moves out of the beam."
```
This wording feels like the linguistic equivalent of 
```
if x is True:
  ...
elif x is False:
  ...
```
See also the help text immediately below:
```Set True to write out...```
which might as well just be
```Write out...```

File: algorithms/refinement/reflection_manager.py [hunk 2]
```diff
@@ -314,14 +319,23 @@ def from_parameters_reflections_experiments(
                 *params.weighting_strategy.constants, stills=do_stills
             )
 
+        # Check for deprecated parameter
+        if params.trim_scan_edges > 0.0:
+            warnings.simplefilter("default")
```
Anthchirp, line 324: this line changes the process-wide warning settings and probably should not be here
```suggestion
```

algorithms/refinement/reflection_manager.py
```diff
@@ -314,14 +319,23 @@ def from_parameters_reflections_experiments(
...
+            warnings.warn(
+                "The parameter trim_scan_edges is deprecated and will be removed shortly",
+                DeprecationWarning,
```
Anthchirp, line 327:
```suggestion
                DeprecationWarning,
                stacklevel=2,
```
I find warnings much more helpful when they indicate the line where a function was called from rather than the line where the warning itself is created.

algorithms/refinement/reflection_manager.py
```diff
@@ -72,16 +73,20 @@
...
               "first or last image of the scan will be removed before"
               "refinement, unless doing so would result in too few remaining"
               "reflections. Reflections that are truncated at the scan edges"
               "have poorly-determined centroids and can bias the refined model"
               "if they are included."
-      .type = float(value_min=0,value_max=1)
+      .type = float(value_min=0,value_max=5)
       .expert_level = 1
 
+    trim_scan_edges = 0.0
```
Anthchirp, line 86:
```suggestion
    trim_scan_edges = None
```
This makes the check for deprecation purposes more reliable, and makes
```dials.command scan_margin=5.0 trim_scan_edges=0.0``` behave similarly to
```dials.command scan_margin=5.0 trim_scan_edges=0.00000001``` instead of entirely different

File: algorithms/refinement/reflection_manager.py [hunk 5]
```diff
@@ -314,14 +319,23 @@ def from_parameters_reflections_experiments(
                 *params.weighting_strategy.constants, stills=do_stills
             )
 
+        # Check for deprecated parameter
+        if params.trim_scan_edges > 0.0:
```
Anthchirp, line 323:
```suggestion
        if params.trim_scan_edges is not None:
```
if the suggestion above to change the default was accepted

util/log.py
```diff
@@ -88,6 +88,11 @@ def config(verbosity=0, logfile=None):
     #   logging.getLogger("dxtbx").setLevel(logging.DEBUG)
     console.setLevel(loglevel)
 
+    # Enable print of messages from warnings.warn to console
+    py_warnings_logger = logging.getLogger("py.warnings")
+    py_warnings_logger.addHandler(console)
+    logging.captureWarnings(True)
+
```
Anthchirp, line 95: This is actually quite intrusive and suggests that what you want to use is `UserWarning` instead of using `DeprecationWarning` and changing the behaviour of all warnings.

Anthchirp (changes requested):

util/log.py
Anthchirp: Just for completeness: [PEP 565 suggests `FutureWarning`](https://www.python.org/dev/peps/pep-0565/) over `UserWarning`. I suppose they behave identically.

dagewa: Will remove and change the message instance over to `logger.warn` and away from `warnings` as this should be pure user land stuff

algorithms/refinement/reflection_manager.py
```diff
@@ -314,14 +318,21 @@ def from_parameters_reflections_experiments(
                 *params.weighting_strategy.constants, stills=do_stills
             )
 
+        # Check for deprecated parameter
+        if params.trim_scan_edges is not None:
+            logger.warn(
```
ndevenish, line 323: `logger.warn` is [deprecated](https://docs.python.org/3.6/library/logging.html#logging.Logger.warning)
```suggestion
            logger.warning(
```

ndevenish (changes requested):

algorithms/refinement/reflection_manager.py
ndevenish: Although I wonder if as in conversation above, `FutureWarning` would be the more correct way. Did this not show to the user?

dagewa: Oh I somehow missed that above. Yes, it does show to the user, but it is not highlighted in yellow like `logger.warning` is.

dagewa:
I think I have addressed all review comments and would like to merge this soon to close #1374

PR #1399, opened 2020-09-10
Optimise sigma calculation code in dials.integrate

jbeilstenedmands:
Replace 
```
selection = flex.size_t(range(i0, i1))
val = x.select(selection)
```
with functionally identical
```val = x[i0:i1]```

Gives significant performance improvement as the second case is within a loop with the target function calculation.
Also remove an unnecessary enumerate.

Just testing on beta-lactamase:
**master**
```  
6.1: Using 90666 / 91208 reflections for sigma calculation
        Calculating E.S.D Beam Divergence.
7.3: Calculating E.S.D Reflecting Range (mosaicity).
51.1:  sigma b: 0.039018 degrees
         sigma m: 0.056053 degrees
```
branch
```
   6.0: Using 90666 / 91208 reflections for sigma calculation
        Calculating E.S.D Beam Divergence.
   7.1: Calculating E.S.D Reflecting Range (mosaicity).
  29.5:  sigma b: 0.039018 degrees
         sigma m: 0.056053 degrees
```
Verified the output is identical using @graeme-winter's new matching tool.


dagewa:
That's a remarkable improvement for a simple change. Good spot!

PR #1406, opened 2020-09-15
Retire use of the thaumatin_i04 dataset in tests

Anthchirp:
The `thaumatin_i04` dataset is a very large dataset (3.1 GB) taking up a lot of resources in the Azure CI.
This PR rewrites the test to use a much smaller dataset that is already used elsewhere - in fact it just uses pre-processed data and does not look at a raw dataset at all.

(`thaumatin_i04` is still used in the tutorials, but that is not relevant for running the tests.)

### DIFF
--- a/test/command_line/test_search_beam_position.py
+++ b/test/command_line/test_search_beam_position.py
@@ -1,12 +1,6 @@
-from __future__ import absolute_import, division, print_function
-
 import glob
 import os
-import sys
 
-import mock
-import procrunner
-import py.path
 import pytest
 
 import scitbx
@@ -14,7 +8,6 @@ from cctbx import uctbx
 from dxtbx.model import ExperimentList
 from dxtbx.serialize import load
 
-import dials.command_line.dials_import
 from dials.algorithms.indexing.test_index import run_indexing
 from dials.command_line import search_beam_position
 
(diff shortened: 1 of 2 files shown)


test/command_line/test_search_beam_position.py
```diff
@@ -93,68 +86,50 @@ def test_index_after_search(dials_data, run_in_tmpdir):
...
+        original_origin[2],
     )
-    print(shift)
+    experiments[0].detector.hierarchy().set_local_frame(
+        experiments[0].detector.hierarchy().get_fast_axis(),
+        experiments[0].detector.hierarchy().get_slow_axis(),
+        shifted_origin,
+    )
+    assert experiments[0].detector.hierarchy().get_origin() == shifted_origin
+    experiments.as_file(run_in_tmpdir / "shifted.expt")
```
rjgildea, line 112: You can probably just use the code from `dials.modify_geometry` to do this. Something like:
```
from dials.command_line import modify_geometry
modify_geometry.run([insulin/"imported.expt", "mosflm_beam_centre=96,97"])
```
which should output a new file "modified.expt" with the updated beam centre.

Note: I think you'll have to pass along the `args` variable at this line:
https://github.com/dials/dials/blob/38aefdfe50dea24ceecf23999cf3e8dc794ae116/command_line/modify_geometry.py#L33

rjgildea (approved):
No real good reason for this test to depend on a 3 Gb data set 👍

test/command_line/test_search_beam_position.py
Anthchirp: I'm not too familiar with the geometry modification code - I'll just use what I have for now. Feel free to open a ticket for this if you think it'd be useful to follow up

PR #1409, opened 2020-09-16
dials.stills_process updates 

phyy-nx:
Improvements from recent beamtime.

- Change behavior of reference_geometry to sync the detector instead of overwriting it
- Add proccess_percent to dials.stills_process 
- Handle composite mode for spotfinding and fix up mp.composite_stride


phyy-nx:
Don't feel the need for a review on this.  Will merge when tests pass.

PR #1413, opened 2020-09-18
Add a more complete help_message for dials.merge

dagewa:
The main motivation here is that the single line `help_message` does not contain a newline, hence the output of `dials.merge` with no arguments is messy.

Make the epilogue message multi-line by adding some examples.


ndevenish:
All looks good!

PR #1416, opened 2020-09-18
Change test for #1410 to use dials-data

Anthchirp:
@jbeilstenedmands In #1410 you added a test that codecov complained about. This is because dials-regression tests are not run on Azure.

I've amended the test to run with a dials-data dataset instead. Are you happy with these changes?
(The test will not work right now, as the `centroid-test-data` dataset does not currently include `indexed.*` files, so if you're happy I'll add those first)

### DIFF
--- a/newsfragments/1416.misc
+++ b/newsfragments/1416.misc
@@ -1,0 +1,1 @@
+Rewrote test to use dials-data for increased coverage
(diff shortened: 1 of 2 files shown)


jbeilstenedmands (approved):
Thanks for this, sorry forgot about the dials-regression/azure issues when adding this.

PR #1419, opened 2020-09-19
Explicitly skip experiments with no reflections.

### LINKED ISSUE #1417: Empty reflection blocks break refinement
This issue may seem academic, but it does have consequences for users (okay maybe just _this user_). 

The problem is that in certain complicated experiment models, refinement fails with the following traceback. 

```
`Traceback` (most recent call last):                                                                      
  File "/home/kmdalton/opt/dials-v3-0-4/build/../modules/dials/command_line/refine.py", line 481, in <mod
ule>                                                                                                    
    run()                                                                                                
  File "/home/kmdalton/opt/dials-v3-0-4/build/../modules/dials/command_line/refine.py", line 396, in run
  File "/home/kmdalton/opt/dials-v3-0-4/build/../modules/dials/command_line/refine.py", line 284, in run_
dials_refine                                                                                            
    refiner, reflections, history = run_macrocycle(params, reflections, experiments)                    
  File "/home/kmdalton/opt/dials-v3-0-4/build/../modules/dials/command_line/refine.py", line 187, in run_
macrocycle                                                                                              
    params, reflections, experiments                                                                    
  File "/home/kmdalton/opt/dials-v3-0-4/modules/dials/algorithms/refinement/refiner.py", line 276, in fro
m_parameters_data_experiments                                                                            
    return cls._build_components(params, reflections, experiments)                                      
  File "/home/kmdalton/opt/dials-v3-0-4/modules/dials/algorithms/refinement/refiner.py", line 401, in _bu
ild_components                                                                                          
    autoreduce()                                                                                        
  File "/home/kmdalton/opt/dials-v3-0-4/modules/dials/algorithms/refinement/parameterisation/autoreduce.p
y", line 260, in __call__                                                                                
    self.check_and_remove()                                                                              
  File "/home/kmdalton/opt/dials-v3-0-4/modules/dials/algorithms/refinement/parameterisation/autoreduce.p
y", line 202, in check_and_remove                                                                        
    self.pred_param.compose(obs)                                                                        
  File "/home/kmdalton/opt/dials-v3-0-4/modules/dials/algorithms/refinement/parameterisation/scan_varying
_prediction_parameters.py", line 310, in compose                                                        
    for block in range(flex.min(blocks), flex.max(blocks) + 1):                                          
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: min() argument is an empty
 array
```

Which is triggered by this line:
https://github.com/dials/dials/blob/d9e9629744be5cea3496fe70befe928fb4425795/algorithms/refinement/parameterisation/scan_varying_prediction_parameters.py#L310

In my case, it resulted from having an experiment that had no strong spots. You might decide that this means I have posited an unreasonable or invalid experiment model. However, I'd argue that valid experiment models trigger this `RuntimeError` more often than you'd expect. In fact, I've now hit this same error while working on two separate projects. 

The issue can arise when you have multiple "Experiments" which share the same global parameters. In my case, I had a single crystal, beam, gonio, scan, and detector model with multiple image sets leading to separate experiments for each image set. I would still expect dials to refine the model if one of the weaker sets has all its reflections rejected as outliers. 

I've also run into this issue while working on polychromatic data wherein we have many beam objects with different wavelengths. In this model there are shared crystal, goniometer, and detector objects. Importantly, I fixed the beam directions so that all the refinement parameters were truly global. By chance, one wavelength can lose all its spots during outlier rejection and trip this `RuntimeError`. 

In my local copy of `dials/algorithms/refinement/parameterisation/scan_varying_prediction_parameters.py`, I've bypassed the error by adding the following control flow

<table>
<tr>
<th>Original</th>
<th>Modified</th>
</tr>
<tr>
<td>

```python
306 # reset current frame cache for scan-varying parameterisations
307 self._current_frame = {}
308 
309 # get state and derivatives for each block
310 for block in range(flex.min(blocks), flex.max(blocks) + 1):
    ...
```

</td>
<td>

```python
# reset current frame cache for scan-varying parameterisations
self._current_frame = {}

if len(blocks) > 0:
    block_ids = range(flex.min(blocks), flex.max(blocks) + 1)
else:
    block_ids = []

# get state and derivatives for each block
for block in block_ids:
    ...
```

</td>

</tr>
</table>

I am not submitting this as a pull request, because this is pretty deep inside dials.refine and outside my comfort zone. I don't know if my workaround is an acceptable solution. What do you all think? I hope I explained that clearly. Let me know if you are mystified by my comment, and I'll try to be more explicit.

dagewa:
Fixes #1417

### DIFF
--- a/algorithms/refinement/parameterisation/scan_varying_prediction_parameters.py
+++ b/algorithms/refinement/parameterisation/scan_varying_prediction_parameters.py
@@ -295,6 +295,10 @@ class ScanVaryingPredictionParameterisation(XYPhiPredictionParameterisation):
             sel = reflections["id"] == iexp
             isel = sel.iselection()
 
+            # skip empty experiments (https://github.com/dials/dials/issues/1417)
+            if len(isel) == 0:
+                continue
+
             blocks = reflections["block"].select(isel)
 
             # identify which parameterisations to use for this experiment
--- a/newsfragments/1417.bugfix
+++ b/newsfragments/1417.bugfix
@@ -1,0 +1,2 @@
+dials.refine: a crash is now avoided that was triggered if there are experiments
+with zero reflections and the `auto_reduction.action=remove` option was chosen.


dagewa:
@kmdalton It appears I can't request a review from you, but can you please check this branch to see if that deals with your issue?

kmdalton:
I'll look at this today. Thanks!

kmdalton:
@PrinceWalnut and I tested this change, and it looks good on our end. I'd support that this PR be merged.

PR #1420, opened 2020-09-22
Make sys.exit messages show red in terminal

ndevenish:
Logging shows errors in red now, but calls to `sys.exit` bypassed this. In this pull, this is corrected by the standard dials context wrapper, previously used to show the mail message to users.

I've called the new function `show_mail_handle_errors` instead of `show_mail_on_error` because it's starting to do more work than just mail notice handling, and I thought that the scope of what it actually does had crept up enough that the name is confusing.

I'm open to suggestions for better names.

(The most important changes are in [`util/__init__.py`](https://github.com/dials/dials/pull/1420/files#diff-0d55fa38d139143d6fcd846d32fccdbd ) )

### DIFF
--- a/util/__init__.py
+++ b/util/__init__.py
@@ -1,20 +1,28 @@
-from __future__ import absolute_import, division, print_function
-
 import contextlib
+import faulthandler
 import functools
+import os
+import signal
 import sys
 
-import six
 import tabulate as _tabulate
 import tqdm
 
 from libtbx.utils import Sorry
 
+# Coloured exit code support on windows
+try:
+    import colorama
+except ImportError:
+    pass
+else:
+    colorama.init()
+
 __all__ = [
     "debug_console",
     "debug_context_manager",
     "progress",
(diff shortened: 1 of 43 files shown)


util/__init__.py
```diff
@@ -140,34 +148,47 @@ def __exit__(self, *args, **kwargs):
...
-    if six.PY3:
-        import faulthandler
-
-        faulthandler.enable()
-        with contextlib.suppress(AttributeError, ImportError):
-            import signal
-
-            faulthandler.register(signal.SIGUSR2, all_threads=True)
+    faulthandler.enable()
+    with contextlib.suppress(AttributeError, ImportError):
```
Anthchirp, line 152:
```suggestion
    with contextlib.suppress(AttributeError):
```
`signal` should be available on all platforms - also you've floated it anyway, so this does nothing

Anthchirp (approved):
Looking good. I like the simplified control flow

PR #1424, opened 2020-09-23
Change default separate_panels=False

dagewa:
When separate_panels=True, outlier rejection is performed for each panel
separately prior to refinement. This is appropriate when we expect the
data on each panel to be drawn from different distributions (for example
prior to metrology corrections). However, it causes a large increase in
the execution time of refinement - dominating the overall job time for
a test case using data from I23.

This parameter should be set to true when doing metrology refinement,
but for normal data processing use, false is likely to be acceptable.


phyy-nx:
> This parameter should be set to true when doing metrology refinement, but for normal data processing use, false is likely to be acceptable.

This comment made me want to verify how we are using it in xfel land.  And yah, it's set to True for metrology refinement:
https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/cspad_cbf_metrology.py#L101

Whereas for ensemble refinement (which refines the detector as a whole for batches of data), it's set to False:
https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/striping.py#L141

dagewa:
That seems right. It makes sense to do the same for scans as stills too.

dagewa:
@rjgildea are you happy with this for I23 processing?

rjgildea (approved):
This seems like a sensible change. Unless you are explicitly doing metrology refinement, it seems sensible to assume that the metrology is sufficiently close to reality that all centroids come from the same distribution. It certainly makes a notable difference to the runtime in some cases.

Issue #1408, opened 2020-09-16
nproc=auto as default for find_spots and integrate


dagewa:
Setting `nproc` > 1 is standard behaviour for `dials.find_spots` and `dials.integrate`. There is no real reason for all that extra keyboard bashing at the command line in most cases, because we could accept `nproc=Auto` as default and let the program decide what is appropriate. #1019 suggests there was some attempt to do this with `dials.find_spots` (now broken). For `dials.integrate`, this gets as far as **`Modelling reflection profiles`** and then fails.

graeme-winter:
👍  sounds like an excellent idea 🙂

dagewa:
Is there any reason not to set this to the equivalent of `libtbx.show_number_of_processors`?

dagewa:
maybe something something `psutils`?

Anthchirp:
yes, the libtbx function is not the correct number to use.

Anthchirp:
Correct solution as far as I can tell is to use the following:
* [`len(os.sched_getaffinity(0))`](https://docs.python.org/3.6/library/os.html#os.sched_getaffinity)
* [`len(psutil.Process().cpu_affinity())`](https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_affinity) as primary fallback
* [`os.cpu_count()`](https://docs.python.org/3/library/os.html#os.cpu_count) as secondary fallback
* [`psutil.cpu_count()`](https://psutil.readthedocs.io/en/latest/#psutil.cpu_count) as tertiary fallback
* `1` as last resort

Anthchirp:
and step 0: if the environment variable `NSLOTS` is set to an integer >=1 then use that instead

graeme-winter:
So we need a `dials.util.how_many_cores` type helper function yes?

PR #1430, opened 2020-09-24
Add available_nproc function

dagewa:
For #1408

### DIFF
--- a/util/mp.py
+++ b/util/mp.py
@@ -24,24 +24,20 @@ def available_cores() -> int:
     except ValueError:
         pass
 
-    # Available on Linux, Python 3.3+
     try:
         return len(os.sched_getaffinity(0))
     except AttributeError:
         pass
 
-    # Availability: Linux, Windows
     try:
         return len(psutil.Process().cpu_affinity())
     except AttributeError:
         pass
 
-    # None if 'undetermined'
     nproc = os.cpu_count()
     if nproc is not None:
         return nproc
 
-    # None if 'undetermined'
     nproc = psutil.cpu_count()
     if nproc is not None:
         return nproc
--- a/util/mp.py
+++ b/util/mp.py
@@ -21,20 +21,24 @@ def available_nproc():
     if nproc >= 1:
         return nproc
 
+    # Available on Linux, Python 3.3+
     try:
         return len(os.sched_getaffinity(0))
     except AttributeError:
         pass
 
+    # Availability: Linux, Windows
     try:
         return len(psutil.Process().cpu_affinity())
     except AttributeError:
         pass
 
+    # None if 'undetermined'
     nproc = os.cpu_count()
     if nproc is not None:
         return nproc
 
+    # None if 'undetermined'
     nproc = psutil.cpu_count()
     if nproc is not None:
         return nproc
--- a/util/mp.py
+++ b/util/mp.py
@@ -1,8 +1,45 @@
 from __future__ import absolute_import, division, print_function
 
+import os
+
 import future.moves.itertools as itertools
+import psutil
 
 import libtbx.easy_mp
+
+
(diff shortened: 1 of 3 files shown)


rjgildea:
Would it be possible to add some comments describing the circumstances under which each level of fallback will get hit? Presumably these are system/Python version dependent?

I guess testing this function is not trivial, although as a minimum it might be possible to test getting the value from the environment with `monkeypatch.setenv`:
https://docs.pytest.org/en/stable/monkeypatch.html

util/mp.py
```diff
@@ -1,10 +1,51 @@
...
 import future.moves.itertools as itertools
+import psutil
 
 import libtbx.easy_mp
 
 
+def available_nproc():
+    """
+    Choose a suitable value for nproc based on the availability of processors.
+    """
```
Anthchirp, line 14:
```suggestion
def available_cores() -> int:
    """
    Determine the number of available processor cores.
    There are a number of different methods to get this information, some of which may not
    be available on a specific OS and/or version of Python. So try them in order and return
    the first successful one.
    """
```

Anthchirp (approved):
I don't think comments are particularly helpful here, as they depend on particular versions of Python, psutils, and considerations within those packages...
Possibly more useful to put in links to the relevant documentations instead?

I would suggest a different function name though.

util/mp.py
dagewa: can do - took `nproc` as being the accepted jargon in DIALS, but `cores` seems better. Given your improved docstring I'll remove the comments I added

util/mp.py
```diff
@@ -1,10 +1,51 @@
...
+
+    nproc = os.environ.get("NSLOTS", 0)
+    try:
+        nproc = int(nproc)
+    except ValueError:
+        nproc = 0
+    if nproc >= 1:
+        return nproc
```
Anthchirp, line 25:
```suggestion
        nproc = int(nproc)
        if nproc >= 1:
            return nproc
    except ValueError:
        pass
```

Python 3 does not allow string / int comparison, so only consider returning the value in the `try` block

dagewa:
When #1408 is fixed then this will get regular real world testing on the platforms we are interested in. Do we really want an additional unit test for this when the code is trivial?

rjgildea:
> When #1408 is fixed then this will get regular real world testing on the platforms we are interested in. Do we really want an additional unit test for this when the code is trivial?

After #1408 is fixed, we should probably explicitly set `nproc` in most/all of the spotfinding tests in order to play nicely with parallel tests.

phyy-nx:
Can pytests be marked to use N number of processes as part of pytest-xdist?  That would be helpful here.  We have been doing some of this in [libtbx land](https://github.com/cctbx/cctbx_project/commit/f5f88156b518a1acd6362e156cad679a93c9aebe) for OpenMP tests.  You can now mark some tests as parallel, meaning they are run in serial after all the other tests and use the number of processors specified in nproc.

Anthchirp:
Just mock out the `available_cores` function? Or just use all cores. I don't believe this has a huge penalty?

Anthchirp:
test timings: (`pytest test_mp.py  --durations=0 -vv`)
```
=== slowest test durations ===
0.00s setup    test/util/test_mp.py::test_core_detection
0.00s setup    test/util/test_mp.py::test_core_detection_override
0.00s teardown test/util/test_mp.py::test_core_detection_override
0.00s teardown test/util/test_mp.py::test_core_detection
0.00s call     test/util/test_mp.py::test_core_detection
0.00s call     test/util/test_mp.py::test_core_detection_override
```
we can afford that :slightly_smiling_face:

rjgildea (approved):
Looks good. Thanks for adding the test @Anthchirp 👍

PR #1439, opened 2020-10-02
Turn ticket references in changelog into links

Anthchirp:
The internet is apparently all about clicking on things.

Existing entries rewritten using
```python
import re

with open("CHANGELOG.rst", "r+") as fh:
    changelog = fh.read()
    fh.seek(0)
    fh.write(
        re.sub(
            "#([0-9]+)",
            lambda m: f"`#{m.group(1)} <https://github.com/dials/dials/issues/{m.group(1)}>`_",
            changelog,
        )
    )
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,16 +4,16 @@ DIALS 3.1.3 (2020-09-28)
 Bugfixes
 --------
 
-- ``dials.integrate``: fix integrator=3d_threaded crash if njobs > 1 (#1410)
-- ``dials.integrate``: Check for and show error message if shoebox data is missing (#1421)
+- ``dials.integrate``: fix integrator=3d_threaded crash if njobs > 1 (`#1410 <https://github.com/dials/dials/issues/1410>`_)
+- ``dials.integrate``: Check for and show error message if shoebox data is missing (`#1421 <https://github.com/dials/dials/issues/1421>`_)
 - ``dials.refine``: Avoid crash for experiments with zero reflections if the
-  `auto_reduction.action=remove` option was active (#1417)
+  `auto_reduction.action=remove` option was active (`#1417 <https://github.com/dials/dials/issues/1417>`_)
 
 Improved Documentation
 ----------------------
 
-- ``dials.merge``: improve help message by adding usage examples (#1413)
-- ``dials.refine``: More helpful error message when too few reflections (#1431)
+- ``dials.merge``: improve help message by adding usage examples (`#1413 <https://github.com/dials/dials/issues/1413>`_)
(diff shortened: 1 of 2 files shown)


ndevenish:
I think the idea was to rely on the github issue-rendering. Except I've checked and it apparently doesn't do that, so this is good.

My only problem is that a mostly 80-column file is now... not, and now a plain-text readable file is scattered with repeated markup. I'm not entirely sure which balance of viewing to prioritise.

I mildly prefer this form:
```restructuredtext
- ``dials.integrate``: fix integrator=3d_threaded crash if njobs > 1 `#1410`_

.. _#1410: https://github.com/dials/dials/issues/1410
```
Which preserves the readability, but also makes clicky linkys in the html document. I'm also okay with merging this and fixing the formatting later, though

ndevenish (approved):

Anthchirp:
Happy to make this change, but would go for
```restructuredtext
- ``dials.integrate``: fix integrator=3d_threaded crash if njobs > 1 (`#1410`_)

.. _#1410: https://github.com/dials/dials/issues/1410
```

> makes clicky linkys in the html document

what HTML document?

ndevenish:
> what HTML document?

When rendered e.g. viewed inside github e.g. https://github.com/dials/dials/blob/master/CHANGELOG.rst

Anthchirp:
An issue with your format is that that I believe it won't integrate nicely with towncrier.

Converting the existing file is easy:
```python
import re

with open("CHANGELOG.rst", "r+") as fh:
    changelog = fh.read()
    fh.seek(0)
    linky = set()
    sorty = sorted

    def linkify(match):
        linky.add(int(match.group(1)))
        return f"`{match.group(0)}`"

    fh.write(re.sub("#([0-9]+)", linkify, changelog))
    fh.write("\n")
    for clicky in sorty(linky):
        fh.write(f".. _#{clicky}: https://github.com/dials/dials/issues/{clicky}\n")
```
but I can't see how you would make a towncrier `issue_format` string that preserves the integrity of the top level bullet list

ndevenish:
I'd been assuming that towncrier output would need to be manually rewritten to add the links. It can be configured to actually link them in?

Anthchirp:
Yes, see this pull request for an example :slightly_smiling_face:

ndevenish:
Oh. I missed the towncrier change at the bottom, I thought this would be something manually applied.

I'd be happy to just have to apply these manually or via an extra script stage, I tend to edit the CHANGELOG anyway before commiting. If put in as-is it's likely I'd switch to hyperlink targets manually for mine, or at least format them so they don't break document width, so aren't really that bothered what happens in an automated way.

PR #1441, opened 2020-10-06
Set nproc=auto by default for spot-finding and integration

dagewa:
Draft to check tests


phyy-nx:
Hi, so `dials.stills_process` handles multiprocessing differently.  A single set of multiprocessing parameters overrides all multiprocessing for downstream API calls.  For example, with `dials.stills_process mp.nproc=12`, call to the integrator uses 1 processor, since the 11 other processors are dedicated to other images.

Does this PR maintain this behavior?  Consider especially processing with multiple lattice search on, where each shot could have multiple lattices passed to the integrated.  Under this PR, I could imagine hitting a situation where the integrator is using more processes than are available to it.

Is it possible instead for the programs `dials.integrate` and `dials.find_spots` to override the nproc parameters to Auto and maintain the defaults in the algorithm phil scopes as 1?

dagewa:
I guess it would be okay if `dials.stills_process` overrides the defaults. Nevertheless,

> Is it possible instead for the programs dials.integrate and dials.find_spots to override the nproc parameters to Auto and maintain the defaults in the algorithm phil scopes as 1?

That's a better approach, thanks

dagewa:
This could do with checking by anyone who is familiar with running `dials.find_spots` and `dials.integrate` using the various cluster submission methods to be sure it still does a sensible thing in those contexts.

phyy-nx (approved):
Thanks for the changes.  I haven't tested big-block integration in dials.find_spots/dials.integrate but the changes look good to me for stills.

graeme-winter:
In discussion: modify tests using command line find spots and integrate to use 1 core?

dagewa:
The macos test failures are a bit odd here. Anyone got any insight?

rjgildea (approved):
This is a sensible change that will save typing, and make dials "faster" by default.

I suspect that `newsfragments/1019.bugfix` and `newsfragments/1408.feature` are superceded by `newsfragments/1441.feature` and could _probably_ be removed to simplify collating future release notes.

I wonder whether we could now remove many of the references to setting `nproc` in the tutorials, e.g.:

https://github.com/dials/dials/blob/297f19248cb4267da2f68c00c1d6db5631538708/command_line/generate_tutorial_text.py#L49

```
$ find . -name "*.rst" -type f | xargs grep "nproc=" | wc -l
29
```

dagewa:
Not sure how I ended up with so much news! Maybe it's related to the amount of time spent doomscrolling on my phone lately. The alternative is that `1019.bugfix` and `1408.feature` are useful and `1441.feature` was superfluous. Happy to go either way. What do @ndevenish and @Anthchirp think? If a PR closes an issue and adds a feature are these both news?

Anthchirp:
I would delete `1408.feature`.

Issue #1440, opened 2020-10-06
best_unit_cell in MTZ output


dagewa:
`dials.scale` and `dials.export` both have a `best_unit_cell` parameter for resolution cutting. In the case of `dials.export` this also controls the overall resolution of the output MTZ, but this does not occur for either the unmerged or merged MTZs that can be produced directly by `dials.scale`. This seems inconsistent. Furthermore, `dials.merge` does _not_ have this parameter. This means that while it is possible to produce an unmerged MTZ with a particular cell in DIALS (by using `best_unit_cell` in `dials.export`) it is not currently possible to produce a merged MTZ with a particular cell.

jbeilstenedmands:
When investigating this issue, I found that the unmerged mtz file produced by dials.scale _does_ use the best unit cell given, and this (should be) covered by the test `algorithms\scaling\test_scale.py::test_scale_best_unit_cell_d_min`. Could I ask you recheck what you were seeing in your example case, and if there is perhaps a different bug then I can investigate that, but this shouldn't be the issue here.

dagewa:
Yes, sorry, you are right. It is only the _merged_ MTZ it outputs that does not use the best unit cell.

PR #1444, opened 2020-10-06
Add best_unit_cell option to dials.merge

jbeilstenedmands:
Allow the setting of a `best_unit_cell` in `dials.merge`.

If `best_unit_cell` is chosen in `dials.scale`, propagate this if creating a merged mtz file.

For #1440


command_line/merge.py
```diff
@@ -106,6 +112,15 @@ def merge_data_to_mtz(params, experiments, reflections):
     ]
     dataset_names = params.output.dataset_names
     crystal_names = params.output.crystal_names
+
+    # check if best_unit_cell is set.
+    best_unit_cell = params.best_unit_cell
+    if best_unit_cell is None:
```
rjgildea, line 118: Are there any circumstances where we need to explicitly compare this to `None`?
```suggestion
    if not best_unit_cell:
```

newsfragments/1444.feature
```diff
@@ -0,0 +1 @@
+dials.merge: Add best_unit_cell option. Allow this to be set from dials.scale also.
```
rjgildea, line 1: > Allow this to be set from dials.scale also.

Does this mean the user can control it from dials.scale, or do you just mean that dials.scale will now use the best unit cell when outputting the merged mtz file?

test/command_line/test_merge.py
```diff
@@ -98,12 +100,18 @@ def test_merge_dmin_dmax(dials_data, tmpdir):
         "project_name=ham",
         "crystal_name=jam",
         "dataset_name=spam",
+        "best_unit_cell=%s" % best_unit_cell,
```
rjgildea, line 103: Can this be an f-String?

test/command_line/test_merge.py
```diff
@@ -98,12 +100,18 @@ def test_merge_dmin_dmax(dials_data, tmpdir):
...
     ]
     result = procrunner.run(command, working_directory=tmpdir)
     assert not result.returncode and not result.stderr
 
     # check we only have reflections in range 8 - 1A
```
rjgildea, line 108: This comment is now separated from the code it refers to - can you move the comment further down?

command_line/merge.py
```diff
@@ -62,6 +63,11 @@
     .help = "All reflections with partiality values above the partiality
         threshold will be retained. This is done after any combination of
         partials if applicable."
+best_unit_cell = None
+    .type = unit_cell
+    .help = "Best unit cell value, to use when performing resolution cutting,"
+            "and as the overall unit cell in the merged mtz. If None, the median"
```
rjgildea, line 69:
```suggestion
            "and as the overall unit cell in the merged mtz. If undefined, the median"
```

rjgildea (approved):
Appreciate the tests :)

If I understand/remember correctly, when it is called from multiplex it should now automatically pick up the two theta-refined unit cell that is set on the experiments?

https://github.com/xia2/xia2/blob/420ae7df5c37e8c8e30b170ee81c951331a743ce/Modules/MultiCrystal/ScaleAndMerge.py#L363-L370

jbeilstenedmands:
@rjgildea  thanks for the suggestions, I have adopted them.
 
For multiplex, this PR shouldn't change any existing behaviour. If the best_unit_cell is not set, then `dials.merge` just uses the median cell of the experiments. I presume that after two theta refinement, all of the unit cells will be set identically and therefore will be set correctly in the output.

PR #1446, opened 2020-10-07
Combine masking logger calls

rjgildea:
* Combine up to six consecutive `logger.info()` calls into a single call
* Change to debug-level log messages

### DIFF
--- a/util/masking.py
+++ b/util/masking.py
@@ -247,9 +247,7 @@ class MaskGenerator(object):
 
             # Add a border around the image
             if self.params.border > 0:
-                logger.debug(
-                    "Generating border mask:\n" + f" border = {self.params.border}"
-                )
+                logger.debug(f"Generating border mask:\n border = {self.params.border}")
                 border = self.params.border
                 height, width = mask.all()
                 borderx = flex.bool(flex.grid(border, width), False)
--- a/util/masking.py
+++ b/util/masking.py
@@ -247,8 +247,9 @@ class MaskGenerator(object):
 
             # Add a border around the image
             if self.params.border > 0:
-                logger.info("Generating border mask:")
-                logger.info(" border = %d" % self.params.border)
+                logger.info(
+                    "Generating border mask:\n" + f" border = {self.params.border}"
+                )
                 border = self.params.border
                 height, width = mask.all()
                 borderx = flex.bool(flex.grid(border, width), False)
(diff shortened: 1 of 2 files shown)


graeme-winter:
👍

graylog2 also says thank you

rjgildea:
Tested with:
```
$ dials.generate_mask $(dials.data get -q centroid_test_data)/*0001.cbf d_min=2 d_max=10 resolution_range=4,4.5 untrusted.rectangle=100,200,100,200 untrusted.circle=100,100,10 border=10 polygon=100,100,100,200,200,200,200,100 ice_rings=True -vv
```

util/masking.py
```diff
@@ -271,20 +270,24 @@ def generate(self, imageset):
...
-                        logger.info(" panel = %d" % region.panel)
-                        logger.info(" xc = %d" % xc)
-                        logger.info(" yc = %d" % yc)
-                        logger.info(" radius = %d" % radius)
+                        logger.debug(
+                            "Generating circle mask:\n"
+                            + f" panel = {region.panel}\n"
+                            + f" xc = {xc}\n"
+                            + f" yc = {yc}\n"
+                            + f" radius = {radius}"
```
ndevenish, line 278: Assuming the split lines are deliberate, I think the `+` are redundant here e.g. 
```suggestion
                            f" panel = {region.panel}\n"
                            f" xc = {xc}\n"
                            f" yc = {yc}\n"
                            f" radius = {radius}"
```

util/masking.py
```diff
@@ -271,20 +270,24 @@ def generate(self, imageset):
...
-                        logger.info(" y0 = %d" % y0)
-                        logger.info(" x1 = %d" % x1)
-                        logger.info(" y1 = %d" % y1)
+                        logger.debug(
+                            "Generating rectangle mask:\n"
+                            + f" panel = {region.panel}\n"
+                            + f" x0 = {x0}\n"
+                            + f" y0 = {y0}\n"
+                            + f" x1 = {x1}\n"
+                            + f" y1 = {y1}"
```
ndevenish, line 289:
```suggestion
                            f" panel = {region.panel}\n"
                            f" x0 = {x0}\n"
                            f" y0 = {y0}\n"
                            f" x1 = {x1}\n"
                            f" y1 = {y1}"
```

ndevenish:
I've argued myself in circles over whether the concatenation makes sense so am going agnostic

PR #1451, opened 2020-10-14
fixes multiple stills image viewing

dermen:
I expect this wont go in as-is, but it at least solves an ongoing issue with the image viewer and stills. The use-case is

`
dials.image_viewer split*.expt split*.refl
`
or 

`
dials.image_viewer combined.expt combined.refl
`

Without these changes, running the above results in weird broken behavior of the image viewer when the expt and refl files are in the format output by dials.stills_process (all z-coords set to 0 ; bbox z1,z2 = 0,1 ; expt.scan is None, expt.goniometer is None, expt.imageset has length 1 ) .

### DIFF
--- a/util/image_viewer/spotfinder_frame.py
+++ b/util/image_viewer/spotfinder_frame.py
@@ -96,8 +96,44 @@ class SpotFrame(XrayFrame):
 
         super().__init__(*args, **kwds)
 
+        self.viewing_stills = True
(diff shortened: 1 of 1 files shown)


dermen:
Im digging a bit more, and this might only be a problem for multi-image formats ... i'll hold off with the pull request until I verify that...

PR #1453, opened 2020-10-16
Remove out-of-date and misleading shell script tutorials

dagewa:
In their current state, these tutorials are misleading as they do not work with current DIALS installations. They are also largely made obsolete by the web-page tutorials, which are run nightly. I suggest to remove them, though I appreciate another viewpoint would be to update them.


rjgildea (approved):

Anthchirp (approved):
Maybe a .misc news file?

PR #1459, opened 2020-10-21
Local indexing: take into account phi in nearest neighbour analysis

### LINKED ISSUE #1458: Improve Local Indexing
#### Issue
As per a discussion on the `dials-support` Slack channel, local index assignment often fails for longer phi scans. This can be exacerbated by crystal slippage during the scan. @jbgreisman has encountered several data sets which could be indexed with XDS but failed under DIALS even though a reasonable lattice was found. The problem most likely due to index assignment from the candidate lattice. The local indexing algorithm based on XDS, `index_assignment.method=local`, is meant to solve this, but it doesn't work well in our hands. 

#### Underlying cause
As implemented, the [AssignIndicesLocal](https://github.com/dials/dials/blob/7417272c0a119adf098fd2003277cfe62563036e/algorithms/indexing/assign_indices.py#L72) class computes reciprocal lattice point nearest neighbors based on reciprocal space distance across a whole data set. This can be problematic if the scan wraps around several times if there is some systematic drift in crystal orientation during rotation. That is, RLP_{hkl} at phi=0 might not superpose all that well with RLP_{hkl} at phi=720 if the crystal slipped a bit during the scan.  

#### Proposed solution
To make local indexing more robust to slippage/drift, I propose that we assign indices in 180 degree chunks. This could be implemented as a phil parameter `indexing.index_assignment.local.block_width = 180`.

rjgildea:
In the "local" index assignment method, the nearest neighbour analysis should only identify as neighbouring those reflections that are close in phi. If, for example, there is a rotation misset in a scan that is more than one full rotation, this previously could cause issues for the local index assignment, leading to low numbers of indexed reflections.

Fixes #1458

### DIFF
--- a/algorithms/indexing/index.h
+++ b/algorithms/indexing/index.h
@@ -197,16 +197,17 @@ namespace dials { namespace algorithms {
       // convert into a single array for input to AnnAdaptor
       // based on flex.vec_3.as_double()
       // scitbx/array_family/boost_python/flex_vec3_double.cpp
-      af::shared<double> rlps_double(reciprocal_space_points.size() * 3,
+      af::shared<double> rlps_double(reciprocal_space_points.size() * 4,
                                      af::init_functor_null<double>());
       double* r = rlps_double.begin();
       for (std::size_t i = 0; i < reciprocal_space_points.size(); i++) {
         for (std::size_t j = 0; j < 3; j++) {
           *r++ = reciprocal_space_points[i][j];
         }
+        *r++ = phi[i];
       }
 
-      AnnAdaptor ann = AnnAdaptor(rlps_double, 3, nearest_neighbours);
+      AnnAdaptor ann = AnnAdaptor(rlps_double, 4, nearest_neighbours);
       ann.query(rlps_double);
 
       const double one_over_epsilon = 1.0 / epsilon;
--- a/algorithms/indexing/test_assign_indices.py
+++ b/algorithms/indexing/test_assign_indices.py
@@ -244,3 +244,48 @@ def test_index_reflections(dials_regression):
     assert "miller_index" in reflections
     counts = reflections["id"].counts()
     assert dict(counts) == {-1: 1390, 0: 114692}
+
+
+def test_4rotation_local(dials_regression):
+    """Test the fix for https://github.com/dials/dials/issues/1458"""
+
+    data_dir = os.path.join(dials_regression, "indexing_test_data", "4rotation")
+    experiments = load.experiment_list(
+        os.path.join(data_dir, "experiments.json"), check_format=False
+    )
+    reflections = flex.reflection_table.from_file(
+        os.path.join(data_dir, "strong.pickle")
+    )
+    reflections["imageset_id"] = flex.int(len(reflections), 0)
+    reflections["id"] = flex.int(len(reflections), -1)
+
+    # Known crystal model
+    experiments[0].crystal = Crystal(
+        (30.809543619194347, -9.657021003319263, -35.831392296125614),
+        (-0.694910011020102, 26.577348370524163, -40.23283798190742),
+        (52.28735147737371, 80.23844876465284, 23.2414871375408),
+        space_group=sgtbx.space_group(),
+    )
+
+    # Prepare reflections for indexing and trim resolution range
+    reflections.centroid_px_to_mm(experiments)
+    reflections.map_centroids_to_reciprocal_space(experiments)
+    reflections = reflections.select(1 / reflections["rlp"].norms() > 4)
+
+    # Assign indices with the correct scan oscillation
+    AssignIndicesLocal()(reflections, experiments)
+    n_indexed = (reflections["miller_index"] == (0, 0, 0)).count(True)
+
+    # Modify the scan oscillation such that we are out by 1 degree per rotation
+    experiments[0].scan.set_oscillation((0, 1 - 1 / 360), deg=True)
+
+    # Reset miller indices and re-map to reciprocal space
+    reflections["miller_index"] = flex.miller_index(len(reflections), (0, 0, 0))
+    reflections["id"] = flex.int(len(reflections), -1)
+    reflections.centroid_px_to_mm(experiments)
+    reflections.map_centroids_to_reciprocal_space(experiments)
+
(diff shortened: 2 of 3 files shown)


rjgildea:
Tested by manually modifying the scan oscillation for the `4rotation` dials_regression indexing test dataset to set oscillation to 0.997 degrees:

```
$ dials.show $(libtbx.find_in_repositories dials_regression)/indexing_test_data/4rotation/experiments.json | grep Scan -A 3
Scan:
    image range:   {1,1440}
    oscillation:   {0,1}
    exposure time: 2

$ dials.show modified.expt | grep Scan -A 3
Scan:
    image range:   {1,1440}
    oscillation:   {0,0.997}
    exposure time: 2
```

Results with previous behaviour:
```
$ dials.index modified.expt $(libtbx.find_in_repositories dials_regression)/indexing_test_data/4rotation/strong.pickle index_assignment.method=local
...
RMSDs by experiment:
|     0 | 103431 |  0.75239 |  0.64086 |     1.0838 |

Refined crystal models:
model 1 (107677 reflections):
Crystal:
    Unit cell: 48.1989(15), 48.1579(16), 98.554(3), 75.8452(7), 75.9054(6), 60.0346(6)
    Space group: P 1
    U matrix:  {{ 0.6536, -0.3604, -0.6655},
                {-0.7534, -0.2265, -0.6173},
                { 0.0717,  0.9049, -0.4195}}
    B matrix:  {{ 0.0207,  0.0000,  0.0000},
                {-0.0120,  0.0240,  0.0000},
                {-0.0035, -0.0035,  0.0106}}
    A = UB:    {{ 0.0202, -0.0063, -0.0070},
                {-0.0108, -0.0032, -0.0065},
                {-0.0079,  0.0232, -0.0044}}
|          0 |      107677 |        198601 | 35.2%       |
```
newplot (11)
newplot (13)

Results with this PR:
```
$ dials.index modified.expt $(libtbx.find_in_repositories dials_regression)/indexing_test_data/4rotation/strong.pickle index_assignment.method=local
...
RMSDs by experiment:
|     0 | 143568 |  0.63947 |  0.87009 |     1.1982 |

Refined crystal models:
model 1 (299644 reflections):
Crystal:
    Unit cell: 48.2151(14), 48.2297(14), 98.546(3), 90.0396(6), 75.8949(6), 59.9967(6)
    Space group: P 1
    U matrix:  {{-0.6534, -0.3604,  0.6658},
                { 0.7541, -0.2320,  0.6145},
                {-0.0670,  0.9035,  0.4233}}
    B matrix:  {{ 0.0207,  0.0000,  0.0000},
                {-0.0120,  0.0239,  0.0000},
                {-0.0070,  0.0035,  0.0106}}
    A = UB:    {{-0.0139, -0.0063,  0.0070},
                { 0.0141, -0.0034,  0.0065},
                {-0.0152,  0.0231,  0.0045}}
|          0 |      299644 |          6634 | 97.8%       |
```
newplot (12)
newplot (14)

dagewa (approved):
This is a very useful change that will improve indexing success for wide scans where experimental geometry is not static, or even static but bad initial geometry.

Anthchirp:
Is the `4rotation` data set particularly suitable for this or is there anything in `dials_data` that would work just as well?

rjgildea:
> Is the `4rotation` data set particularly suitable for this or is there anything in `dials_data` that would work just as well?
> 
> I'm just mindful that `4rotation` is only used in one other place and is quite chunky.

Yes - the bug was only highlighted when using data sets with multiple full rotations. `dials_data` doesn't even contain a data set with a single full rotation as far as I can tell. We could probably generate some predicted reflection centroids for the purpose of this test, however the additional test is already many more lines of code than the bug fix itself!

dagewa:
Generating "observations" is a bit of a strange dance that could be made easier. The way I've done it in the past does prediction twice, to first get rays that intersect the detector and then to get a reflection table that includes flags etc. Then you can set the observations from the predictions and invent some variances (for refinement). There are various messy examples in the refinement tests, like here:
https://github.com/dials/dials/blob/241d539ac2967ac294274a7781c0ea95832335da/test/algorithms/refinement/test_refinement_regression.py#L184-L228

rjgildea:
> Generating "observations" is a bit of a strange dance that could be made easier. The way I've done it in the past does prediction twice, to first get rays that intersect the detector and then to get a reflection table that includes flags etc. Then you can set the observations from the predictions and invent some variances (for refinement).

It turns out it's much easier in this case, as all I need are some observed centroids, so I can use `flex.reflection_table.from_predictions(experiment)` and copy the calculated centroids to predicted centroids:
https://github.com/dials/dials/pull/1459/commits/d1c9d51e22f47f0e71739f60b8eefc17405c460b

@Anthchirp happy now? 🙂

Anthchirp:
> @Anthchirp happy now? 🙂

I was already happy with your last response, but - yes, certainly 🙂

algorithms/indexing/index.h
```diff
@@ -197,16 +197,17 @@ namespace dials { namespace algorithms {
...
       // scitbx/array_family/boost_python/flex_vec3_double.cpp
-      af::shared<double> rlps_double(reciprocal_space_points.size() * 3,
+      af::shared<double> rlps_double(reciprocal_space_points.size() * 4,
                                      af::init_functor_null<double>());
       double* r = rlps_double.begin();
       for (std::size_t i = 0; i < reciprocal_space_points.size(); i++) {
         for (std::size_t j = 0; j < 3; j++) {
           *r++ = reciprocal_space_points[i][j];
         }
+        *r++ = phi[i];
```
graeme-winter, line 207: My only comment here is that stuff like `*r++` always gets me scratching my head as to whether `++` happens before or after the operation etc. - and we already know that this should be written to `rlps_double[4*i+3]` - applies to the line above which is obviously unchanged as well. 

Appreciate that this is consistent with the "house style" locally and does mean an increment vs. add/multiply but I wonder how much difference that makes to the runtime (i.e. compilers are clever & can unroll the add/multiply version vs. cannot unroll the increment version)

graeme-winter (approved):
Important set of fixes thank you - wonder if your news fragment should include a comment for > 360° sweeps to aid the end user in understanding the significance of this?

PR #1461, opened 2020-10-21
Fix float comparison causing gaussianrs test failure

### LINKED ISSUE #1456: gaussianrs test failures
Tests
```
test.algorithms.integration.test_parallel_integrator.test_gaussianrs_detector_space_intensity_calculator 
test.algorithms.integration.test_parallel_integrator.test_gaussianrs_detector_space_with_deconvolution_intensity_calculator
test.algorithms.integration.test_parallel_integrator.test_gaussianrs_detector_space_with_deconvolution_intensity_calculator2
```
have been failing for a while on Jenkins with output such as
```python
    def test_gaussianrs_detector_space_intensity_calculator(data):
        algorithm = IntensityCalculatorFactory.create(
            data, detector_space=True, deconvolution=False
        )
    
        reflections = flex.reflection_table_to_list_of_reflections(data.reflections)
    
        errors = 0
        for r in reflections:
            try:
>               algorithm(r, [])
E               RuntimeError: dials Internal Error: /scratch/jenkins_agent/workspace/dials_test_linux_python3/build_dials/modules/dials/algorithms/profile_model/gaussian_rs/transform/transform.h(950): DIALS_ASSERT(fraction <= 1.0) failure.
```

Tests can be run with
```bash
$ pytest test/algorithms/integration/test_parallel_integrator.py::test_gaussianrs_detector_space_{intensity_calculator,with_deconvolution_intensity_calculator{,2}}
```

Since these tests rely on dials_regression they are not run on Azure, which is why the Azure build appears to be happy.

It looks like this was triggered by environment changes. My existing developer build works, but a newly created one fails.

ndevenish:
GCC updating in conda from 7→9 caused failure in these assertions by ~1e-13 threshold.

Since calculating this involves error functions, it's safe to assume that nowhere was relying on sub-epsilon behaviours from this function, and other comparisons here use epsilon for checking too.

Fixes #1456.

### DIFF
--- a/algorithms/profile_model/gaussian_rs/transform/transform.h
+++ b/algorithms/profile_model/gaussian_rs/transform/transform.h
@@ -947,8 +947,8 @@ namespace dials {
                       DIALS_ASSERT(kk < zfraction.accessor()[1]);
                       DIALS_ASSERT(k < zfraction.accessor()[0]);
                       double fraction = zfraction(k, kk);
-                      DIALS_ASSERT(fraction <= 1.0);
-                      DIALS_ASSERT(fraction >= 0.0);
+                      DIALS_ASSERT(fraction <= 1.0 + EPS);
+                      DIALS_ASSERT(fraction >= 0.0 - EPS);
                       double value = fraction * area * data(k, j, i);
                       profile_(kk, jj, ii) += value;
                     }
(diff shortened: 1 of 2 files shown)


Anthchirp:
Thanks 👍 

Just a note that the relevant tests are not covered by Azure, so the green tick isn't all that meaningful.

ndevenish:
Right, merge and wait for the overnights I guess.

PR #1462, opened 2020-10-22
fixes all the warnings at image viewer launch by replacing deprecated…

dermen:
Quick fix to silence the warnings


graeme-winter:
@dermen thank you for these fixes, they have been annoying me for a good long while but not bothered to do anything about it 🙄

(I can't test though as I am on Catalina so things are differently-broken right now)

ndevenish:
Yes, this is helpful!

PR #1464, opened 2020-10-23
export_mmcif: faster construction of cif loop model

rjgildea:
Avoid Python loop over every reflection.

Depends on cctbx/cctbx_project#547


rjgildea:
Avoid Python loop over every reflection.

PR #1469, opened 2020-10-27
Add Windows builds to Azure

Anthchirp:
this doesn't produce working Windows builds yet, but lays some
groundwork. I had previously prepared these changes in a branch,
but they might as well move over into the main repository.

With 3.2 branched out of the way now seems an good time to touch the bootstrap script.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -80,4 +80,4 @@ jobs:
 
   - template: .azure-pipelines/azure-pipelines-mac.yml
 
-# - template: .azure-pipelines/azure-pipelines-windows.yml
+  - template: .azure-pipelines/azure-pipelines-windows.yml
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,5 +1,5 @@
 variables:
-  CACHE_VERSION: 20200701
+  CACHE_VERSION: 20201102
 
 schedules:
   # nightly builds to populate caches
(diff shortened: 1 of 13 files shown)


Anthchirp:
So: this works.

Few caveats:
* This provides only 'light' testing on Windows
  * No regression data (for reasons that will become clear shortly)
  * 7 tests marked as xfail which need investigating
  * This only adds a single Windows build job, running on Python 3.6.
* The environmental overhead is considerable
  * It takes 5 minutes to downgrade the pre-installed Visual Studio compilers to an older version because `cbflib_adaptbx` doesn't with with current generation compilers. Possible that this can be fixed, but I'm no expert on this. The plan is to get rid of `cbflib_adaptbx` anyway, which would probably make this go away.
  * It then takes 5 minutes to set up the Python environment. In the Linux/MacOS builds we can cache this step. Here the caching works fine, too - but when the environment is loaded from the cache it causes the cache process to crash. We can sort this, but not now.
  * While the build itself _is_ cached, that is a very slow step. The reason for that is that the build directory is in excess of 4 GB. It is not clear why the build directory is so massive. We are likely doing something very wrong in our build. On Azure/Linux the build directory weighs in at 500 MB.
  * As the overall space limitation still applies, we should ideally address the size issue before even looking at regression testing.
  * Surprise: there are no Windows `scipy` builds on conda-forge. `scikit-learn` is also missing as it is downstream. So we `pip install` it instead. There is some work ongoing on this at conda-forge, but don't expect miracles any time soon. (scipy on Windows is hitting maximum build runtime limits on their CI)
* Other notes:
  * `make reconf` does not work on Windows?!
  * I'm not even convinced that `make` does anything?! This needs verifying.
  * While we do generate code coverage statistics we don't upload them. [codecov-exe should be useful here](https://github.com/codecov/codecov-exe).
* All in all this is still a very useful first step to some proper Windows build coverage.

PR #1470, opened 2020-10-28
partiality_threshold=0.4 in cosym, symmetry and ttr

rjgildea:
Set the default partiality_threshold to 0.4 in dials.cosym, dials.symmetry and dials.two_theta_refine. The previous default of 0.99 could occasionally be overzealous in rejecting reflections, especially with particularly narrow wedges. This issue was identified by errors encountered running xia2.multiplex on [James Holton's "Micro-focus Data Processing Challenge" dataset](https://bl831.als.lbl.gov/~jamesh/challenge/microfocus/).

### DIFF
--- a/command_line/cosym.py
+++ b/command_line/cosym.py
@@ -33,7 +33,7 @@ logger = logging.getLogger("dials.command_line.cosym")
 
 phil_scope = iotbx.phil.parse(
     """\
-partiality_threshold = 0.99
+partiality_threshold = 0.4
   .type = float
   .help = "Use reflections with a partiality above the threshold."
 
--- a/command_line/symmetry.py
+++ b/command_line/symmetry.py
@@ -73,7 +73,7 @@ relative_length_tolerance = 0.05
 absolute_angle_tolerance = 2
   .type = float(value_min=0)
 
-partiality_threshold = 0.99
+partiality_threshold = 0.4
   .type = float
   .help = "Use only reflections with a partiality above this threshold."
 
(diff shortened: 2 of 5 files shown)


dagewa:
For `dials.two_theta_refine` is the idea to use the same set of reflections as used in `dials.cosym` and `dials.symmetry`? For the latter two programs, this change increases the number of reflections accepted, whereas for `dials.two_theta_refine` it decreases the number, but as a result essentially the same set of reflections gets used. Is that right?

rjgildea:
> for `dials.two_theta_refine` it decreases the number

Perhaps I'm missing something, but I don't understand why this would be the case? The default `partiality_threshold` for `dials.util.filter_reflections.filter_reflection_table` is 0.99, and previously this parameter was essentially hard-coded in `dials.two_theta_refine` and not user-controllable:

https://github.com/dials/dials/blob/master/util/filter_reflections.py#L457

rjgildea:
With `module load dials/latest`:
```
$ dials.two_theta_refine $(dials.data get -q l_cysteine_4_sweeps_scaled)/scaled_20_25.{expt,refl} | egrep 'Removed|Summary'
Removed 20 reflections below partiality threshold
Summary statistics for 3585 observations matched to predictions:
```
With this PR:
```
$ dials.two_theta_refine $(dials.data get -q l_cysteine_4_sweeps_scaled)/scaled_20_25.{expt,refl} | egrep 'Removed|Summary'
Summary statistics for 3605 observations matched to predictions:
```

I.e. with this PR, fewer reflections are removed (none in this case) for being below the partiality threshold, and more reflections in total are used in refinement.

dagewa:
I see, thanks. I didn't realise anything was filtered previously in `dials.two_theta_refine` - I thought it operated on all the reflections it was given.

dagewa (approved):
It makes sense that these tools use the same threshold for post-scaling reflection selection. I did wonder whether it would make sense for `filter_reflection_table` to set a flag, but I think that can be a separate discussion.

jbeilstenedmands (approved):

PR #1471, opened 2020-10-28
For MPI, don't use single log file

phyy-nx:
Using a single log file locks up the file system.  We already had log splitting in place, so this change just prevents configuring the log file twice if MPI is enabled.


phyy-nx:
Anyone want to review this one?  Not super vital.  Otherwise I'll merge it tomorrow.

Issue #1457, opened 2020-10-21
Write out scaling/merging statistics in mmCIF format by default in xia2 and DUI


dagewa:
Following #281 we can write out mmCIF formatted reflections files from DIALS programs. As a subset of this, we include data scaling statistics in this file (see 007bea58d0a). The PDB deposition pipeline can now accept the mmCIF-formatted scaling statistics files to merge with the coordinate mmCIF file, as noted by @berrisfordjohn. The developers of CCP4 interfaces would like it if this file was written out by default. In particular, if xia2 and DUI produced these files on every run to completion, then it would make deposition easier down the line.

Should this file be output by default whenever `dials.scale` is run?

What issues are there? I can see one - how do we ensure that the statistics match with a particular MTZ used for structure determination?

graeme-winter:
I humbly note https://github.com/dials/dials/issues/1438#issuecomment-700601729 that you explicitly rejected this idea three weeks ago on the grounds that "derived data go stale"

I offer no objection to adding the additional `dials.export` step to `xia2` and you can do precisely the same in `DUI` but I would think always outputting mmCIF format reflections from `dials.scale` every time it is run is excessive, particularly as by design you may want to run it again?

jbeilstenedmands:
With regards to ensuring the matching of (unmerged?) MTZ and mmcif, these should contain the same data (or directly derived in a merged MTZ case), one should therefore be able to run the same merging code on both (using the audited dials version perhaps?) to verify that both give the same stats as given in the mmcif.

Also I know the mmcif has several auditing fields, we should be able to add detailed auditing to MTZ (in MTZ language I think this is referred to as 'history') so that it's obvious if the two were created at a different point. Although this might require allowing dials.export to export multiple formats at once, this should be easy to add.

dagewa:
Agree that adding to xia2 and DUI sounds reasonable. What I meant with `dials.scale` was to only output the merging statistics portion, but I recognise the potential headaches therein. If we don't mind exporting the whole reflections mmCIF, including the stats, then it is certainly easier to track provenance

graeme-winter:
> Agree that adding to xia2 and DUI sounds reasonable. What I meant with `dials.scale` was to only output the merging statistics portion, but I recognise the potential headaches therein. If we don't mind exporting the whole reflections mmCIF, including the stats, then it is certainly easier to track provenance

[...]

PR #1482, opened 2020-11-05
Compress exported mmcif on-the-fly

### LINKED ISSUE #1480: On-the-fly compression for dials.export of unmerged mmcif
> I guess if xia2 is going to write a compressed mmcif it would be make sense if DIALS writes the file as present, and xia2 does the compression?

Ideally, `dials.export` could compress on-the-fly depending on the file extension - e.g. if you pass `mtz.hklout=foo.mmcif.gz` 
or `mtz.hklout=foo.mmcif.bz2` then it would output a compressed file as appropriate.

_Originally posted by @rjgildea in https://github.com/dials/dials/issues/1457#issuecomment-715055863_

dagewa:
Fix #1480


Anthchirp:
Not quite happy with the automagic here.
Would it be more helpful to have an explicit `compress=gz`/`compress=bz2` option (which may be set by default)?
At the open stage we can the check that the relevant extension is in the name, and if not then append it.

My main issues are with 
* compatibility: (ie. DIALS 2.2/3.2 allow you to write plain text to a file ending in `.bz2`, no problem), and
* expectation and information (user is not told that writing to a `.bz2` file results in a compressed file, whereas writing to a `.zip` file does not)

dagewa:
Explicit compress sounds okay. I mean, we have to explicitly set the `format=mmcif` anyway, not detect based on filename (which I did find annoying already - see discussion at https://github.com/dials/dials/issues/1457#issuecomment-714320850). However, happy to make that change if there is distaste for automagic.

dagewa:
Maybe just `compress=True` and always make it gzip? @rjgildea points out that the PDB uses .gz for downloads

graeme-winter:
> Maybe just `compress=True` and always make it gzip? @rjgildea points out that the PDB uses .gz for downloads

Since this is never (I suggest) going to be used by the "woman in the street" I would say that we should be explicit in preference to implicit - `compress=gz` as default with `bz2` or whatever as options would be preferable to me, provided we also append `.gz` to the filename unless already specified

dagewa:
Ok, I've done that, but I left the default as no compression. That makes it even more explicit. You get exactly the filename you wanted, unless you want one of the supported compression modes and then you get a standard extension. Like you say, this is generally going to get set within xia2 or dui, not by the human in front of the computer.

util/export_mmcif.py
```diff
@@ -324,10 +324,12 @@ def write(self, experiments, reflections):
...
-        elif filename.endswith(".gz"):
+        if self.params.mmcif.compress is not None:
+            filename += "." + self.params.mmcif.compress
+        if self.params.mmcif.compress == "gz":
             open_fn = gzip.open
+        elif self.params.mmcif.compress == "bz2":
+            open_fn = bz2.open
         else:
             open_fn = open
         with open_fn(filename, "wt") as fh:
```
Anthchirp, line 335:
```suggestion
        with open_fn(filename, "w") as fh:
```
`t` just raises questions and is the default anyway

Anthchirp (approved):

util/export_mmcif.py
dagewa: Unfortunately this fails abruptly with the message `a bytes-like object is required, not 'str'`

dagewa: no stacktrace and I've not dug into where this happens, but suspect it is within `iotbx/cif/`

dagewa: Yes, it is here: https://github.com/cctbx/cctbx_project/blob/8737e695766c1c7395da6650e5a89777a2d3f145/iotbx/cif/model.py#L83

Anthchirp: did I ever mention that this print-to-file pattern is so, so, so special...

Leave it then :slightly_smiling_face:

Issue #1414, opened 2020-09-18
Running command without arguments for help leaves droppings


dagewa:
Running a command like `dials.index` with no arguments to print help to screen writes the file `dials.index.log` to disk.

```bash
$ cat dials.index.log 
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 3.dev.194-gc84be8f56
```
This is worse with `dials.integrate`, which also leaves the empty file `dials.integrate.phil`

dagewa:
of course what you are _supposed_ to do is run, e.g. `dials.integrate -h`

dagewa:
See also #1429 for a suggestion to change the behaviour of `dials.program -h` to add value over just typing `dials.program`

rjgildea:
It's not just `dials.index` and `dials.integrate`, I suspect it is any program that generates a log file (e.g. I tested that `dials.refine` also exhibits this behaviour).

The issue is that (as recommended in the [boilerplate command line example](https://github.com/dials/dials/blob/75742c527cb73287934adad7c10caf39a04401df/doc/examples/boilerplate.py#L135-L136)) we first configure the logging, before parsing the output:
https://github.com/dials/dials/blob/75742c527cb73287934adad7c10caf39a04401df/doc/examples/boilerplate.py#L135-L136
followed by logging the dials version and the phil diff:
https://github.com/dials/dials/blob/75742c527cb73287934adad7c10caf39a04401df/doc/examples/boilerplate.py#L138-L145
and then finally checking whether there are any input experiments/reflections, and exiting (and in some cases calling `parser.print_help()`):
https://github.com/dials/dials/blob/75742c527cb73287934adad7c10caf39a04401df/doc/examples/boilerplate.py#L147-L155
The issue is that by this point the log file has already been created - the log file is created by the call to:
`dials.util.log.config(verbosity=options.verbose, logfile=params.output.log)`

PR #1483, opened 2020-11-05
Add dano/sigdano to the output of dials.merge, and make html report

jbeilstenedmands:
One useful anomalous indicator that other programs report (XSCALE, shelxc) is dano/sigdano, i.e. the mean of the absolute differences over the mean of the difference sigmas: < | F(+) - F(-) | > / < sig(F(+) - F(-)) >. 
The staff at I23 beamline have requested this be added to DIALS as a key indicator of anomalous signal. As it relies on Fs, it seems most sensible to report on this in `dials.merge`, so in `dials.merge` the merging stats now have an extra final column, e.g.
```
            ----------Merging statistics by resolution bin----------           

 d_max  d_min   #obs  #uniq   mult.  %comp       <I>  <I/sI>    r_mrg   r_meas    r_pim   r_anom   cc1/2   cc_ano dF/s(dF)
 69.31   3.29  22601   3392    6.66  98.63     532.6    69.2    0.037    0.040    0.015    0.042   0.999*   0.297*   1.909
  3.29   2.61  22267   3319    6.71  97.53     193.3    47.8    0.050    0.054    0.021    0.060   0.998*   0.338*   1.766
  2.61   2.28  22519   3258    6.91  96.76      99.2    35.7    0.065    0.070    0.027    0.073   0.997*   0.362*   1.456
  2.28   2.07  22169   3243    6.84  95.95      67.5    27.3    0.078    0.085    0.032    0.081   0.996*   0.249*   1.223
  2.07   1.93  21502   3167    6.79  95.19      44.0    20.5    0.099    0.107    0.041    0.101   0.994*   0.204*   1.095
  1.93   1.81  21655   3187    6.79  94.37      25.0    14.1    0.136    0.147    0.056    0.133   0.991*   0.188*   0.966
  1.81   1.72  21246   3151    6.74  93.89      15.6    10.0    0.180    0.195    0.074    0.175   0.986*   0.146*   0.863
  1.72   1.65  20887   3104    6.73  93.07      10.4     7.2    0.232    0.252    0.096    0.220   0.978*   0.148*   0.761
  1.65   1.58  21639   3123    6.93  92.64       8.3     6.1    0.275    0.297    0.112    0.241   0.973*   0.055*   0.702
  1.58   1.53  20674   3082    6.71  92.00       6.3     4.7    0.327    0.354    0.136    0.304   0.955*   0.063*   0.667
  1.53   1.48  20145   3066    6.57  91.58       5.0     3.8    0.382    0.415    0.161    0.348   0.939*   0.021    0.588
  1.48   1.44  20801   3029    6.87  90.74       3.7     3.0    0.473    0.512    0.194    0.416   0.922*   0.055*   0.533
  1.44   1.40  18330   2982    6.15  90.31       3.2     2.5    0.523    0.571    0.227    0.497   0.891*  -0.002    0.494
  1.40   1.37  13881   2480    5.60  72.83       2.7     2.0    0.585    0.646    0.268    0.593   0.811*  -0.001    0.454
  1.37   1.33   9578   1749    5.48  52.71       2.4     1.8    0.638    0.704    0.292    0.650   0.748*  -0.019    0.426
  1.33   1.31   6726   1287    5.23  38.73       2.3     1.6    0.654    0.725    0.305    0.714   0.737*   0.003    0.430
  1.31   1.28   4850   1001    4.85  29.69       2.0     1.3    0.745    0.833    0.362    0.826   0.603*  -0.018    0.373
  1.28   1.26   2923    700    4.18  21.03       1.8     1.1    0.854    0.973    0.452    1.003   0.406*   0.029    0.354
  1.26   1.23   1214    407    2.98  12.37       1.5     0.8    0.895    1.067    0.567    1.156   0.284*  -0.156    0.279
  1.23   1.21    305    191    1.60   5.64       1.6     0.6    0.746    0.987    0.640    1.033   0.489*  -0.357    0.388
 69.19   1.21 315912  48918    6.46  72.87      69.3    17.1    0.064    0.070    0.027    0.070   0.999*   0.358*   0.794
```

### DIFF
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -342,8 +342,8 @@ def make_dano_plots(anomalous_data):
             "dano": {
                 "data": [],
                 "help": """\
-This plot shows the size of the anomalous differences of F relative to the uncertainties.
-(< |F(+)-F(-)|  / σ(F(+)-F(-)) >). A value of 0.8 is indicative of pure noise, and
+This plot shows the size of the anomalous differences of F relative to the uncertainties,
+(<|F(+)-F(-)|/σ(F(+)-F(-))>). A value of 0.8 is indicative of pure noise, and
 a suggested cutoff is when the value falls below 1.2, although these guides require
 reliable sigma estimates. For further information see
 https://strucbio.biologie.uni-konstanz.de/ccp4wiki/index.php?title=SHELX_C/D/E
@@ -394,13 +394,13 @@ https://strucbio.biologie.uni-konstanz.de/ccp4wiki/index.php?title=SHELX_C/D/E
     )
 
     data["dF"]["dano"]["layout"] = {
-        "title": "< |ΔF| / σ(ΔF) > vs resolution",
+        "title": "<|ΔF|/σ(ΔF)> vs resolution",
         "xaxis": {
             "title": "Resolution (Å)",
             "tickvals": d_star_sq_tickvals,
             "ticktext": d_star_sq_ticktext,
         },
-        "yaxis": {"title": "< |ΔF| / σ(ΔF) >", "rangemode": "tozero"},
+        "yaxis": {"title": "<|ΔF|/σ(ΔF)>", "rangemode": "tozero"},
     }
     return data
 
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -3,8 +3,10 @@ from __future__ import absolute_import, division, print_function
 
 import logging
 
+from jinja2 import ChoiceLoader, Environment, PackageLoader
 from six.moves import cStringIO as StringIO
 
+from cctbx import uctbx
 from mmtbx.scaling import data_statistics
 
 from dials.algorithms.scaling.Ih_table import (
@@ -21,6 +23,8 @@ from dials.algorithms.symmetry.absences.run_absences_checks import (
 )
 from dials.array_family import flex
 from dials.report.analysis import make_merging_statistics_summary, table_1_summary
+from dials.report.plots import d_star_sq_to_d_ticks
+from dials.util import tabulate
 from dials.util.export_mtz import MADMergedMTZWriter, MergedMTZWriter
 from dials.util.filter_reflections import filter_reflection_table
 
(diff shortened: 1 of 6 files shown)


algorithms/merging/merge.py
```diff
@@ -278,3 +278,42 @@ def truncate(merged_intensities):
...
+    vals.append(delta_F_mean_over_sig_delta_F_mean(anomalous_amplitudes))
+    return vals
+
+
+def delta_F_mean_over_sig_delta_F_mean(anomalous_amplitudes):
+    """Calculate < |F(+) - F(-)|> / <sigma(F(+) - F(-))> i.e. DANO/SIGDANO."""
+    diff = anomalous_amplitudes.anomalous_differences()
+    if not diff.data() or not diff.sigmas():
+        return 0.0
+    return flex.mean(flex.abs(diff.data())) / flex.mean(diff.sigmas())
```
rjgildea, line 302: You could make this a single function, in line with various methods in `miller.array`, e.g.:
https://github.com/cctbx/cctbx_project/blob/5ae04d2955fa897b34ed4ec5bd8aea427aeeeafc/cctbx/miller/__init__.py#L3045-L3075

If you're feeling brave, it might be useful to add this to cctbx so that it can be used elsewhere (e.g. `xia2.merging_statistics`), and this might avoid the gymnastics of adding it to the merging stats output.

algorithms/merging/merge.py
```diff
@@ -278,3 +278,42 @@ def truncate(merged_intensities):
...
+        if count > -1 and count < len(dFsdF):
+            fmt = "% 8.3f" % dFsdF[count]
+            if not l.endswith("*"):
+                fmt = " " + fmt
+            lines[i] += fmt
+            count += 1
+        if l.startswith(" d_max"):
+            lines[i] += " dF/s(dF)"
+            count = 0
+    return "\n".join(lines)
```
rjgildea, line 319: This seems liable to cause problems down the line if anything changes upstream with the formatting of the merging stats output. Is there not any better way to modify/override the mergings stats object to do this without having to edit a table string in this manner?

rjgildea (commented):
Does this statistic have a more commonly used name elsewhere? The current function names are a bit of a mouthful to say the least!

The way the merging stats output is modified to add the new statistic feels a little "hacky", I wonder if there is a better way to do this.

jbeilstenedmands:
I had a think about this. As this property is based on amplitudes, I'm not sure it belongs in the merging statistics output which is intensity-based. What I was doing here was also too hacky, I've changed it so that a separate table is printed out for dano/sigdano in `dials.merge`, and I have added a dials.merge.html report which plots this property.

jbeilstenedmands:
Here is an example of the plot generated in `dials.merge.html` for the MAD example dataset to 1.9A.
newplot (33)
I think this would be a useful plot for the `xia2.html` in 'Analysis by resolution' so I'll make a corresponding PR there.

newsfragments/1483.feature
```diff
@@ -0,0 +1 @@
+dials.merge: Add dF/s(dF), also known as dano/sigdano, to statistics reported.
```
rjgildea, line 1: New html report for dials.merge should probably be mentioned here too

algorithms/merging/merge.py
```diff
@@ -279,3 +283,134 @@ def truncate(merged_intensities):
...
+    rows = []
+    for i, dF in enumerate(dFsdF):
+        rows.append(
+            [
+                f"{resolution_bin_edges[i]:6.2f}",
+                f"{resolution_bin_edges[i+1]:6.2f}",
+                f"{dF:6.3f}",
+            ]
+        )
+    logger.info(tabulate(rows, header))
```
rjgildea, line 326: It might be better to return either the table, or string representation here and move the `logger.info` calls to the command line script, which would potentially make this function more useful elsewhere.

algorithms/merging/merge.py
```diff
@@ -279,3 +283,134 @@ def truncate(merged_intensities):
...
+            "title": "Resolution (Å)",
+            "tickvals": d_star_sq_tickvals,
+            "ticktext": d_star_sq_ticktext,
+        },
+        "yaxis": {"title": "Dano/SigDano", "rangemode": "tozero"},
+    }
+    return data
+
+
+def generate_html_report(mtz_file, filename):
```
rjgildea, line 386: Surely it shouldn't be necessary to re-read the mtz files that we've just written in order to re-extract the anomalous arrays. Are they not already in memory somewhere? It seems a little inefficient at best.

algorithms/merging/merge.py
```diff
@@ -279,3 +283,134 @@ def truncate(merged_intensities):
...
+        dict: A dictionary containing the plotting data.
+    """
+
+    data = {
+        "dF": {
+            "dano": {
+                "data": [],
+                "help": """\
+This plot shows the size of the anomalous differences of F relative to the uncertainties.
+Dano/SigDano = (< |F(+)-F(-)| > / < sigma(F(+)-F(-)) >)
```
rjgildea, line 346:
```suggestion
Dano/SigDano = (< |F(+)-F(-)| > / < σ(F(+)-F(-)) >)
```

rjgildea (approved):

rjgildea:
I note that if you run `dials.merge [...] truncate=False` then an empty html report is generated. Perhaps it would be better not to generate the html at all if no plots to include?

jbeilstenedmands:
Yes I wasn't sure about what to do in the case `truncate=False` and whether it is better for the program to be consistent in the number of output files it produces at the expense of an empty file in some cases.

algorithms/merging/merge.py
```diff
@@ -279,3 +283,134 @@ def truncate(merged_intensities):
...
+            {
+                "x": d_star_sq_bins,
+                "y": list(dFsdF),
+                "type": "scatter",
+                "name": "\u03BB" + f"={wave:.4f}",
+            }
+        )
+
+    data["dF"]["dano"]["layout"] = {
+        "title": "Dano/SigDano vs resolution",
```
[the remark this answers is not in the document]
rjgildea: `"< |ΔF| > / < σ(ΔF) > vs resolution"` would be another alternative

dagewa:
Ew, no, I don't like empty files. I'm reminded of some others too #1414

jbeilstenedmands:
Updated look of the output plot (a different dataset to above):
newplot (34)

PR #1488, opened 2020-11-10
Include small integer types from scitbx

graeme-winter:
Not currently published via cctbx.array_family but that does not make a real difference

### DIFF
--- a/newsfragments/1488.feature
+++ b/newsfragments/1488.feature
@@ -1,0 +1,2 @@
+Include small integer types from scitbx (see https://github.com/cctbx/cctbx_project/pull/533 for details)
+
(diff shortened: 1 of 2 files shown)


newsfragments/1488.feature
```diff
@@ -0,0 +1,2 @@
+Include small integer types from scitbx (see https://github.com/cctbx/cctbx_project/pull/533 for details)
```
Anthchirp, line 1: This is fine for a commit message, which has the context of the changes, but if you read this in isolation or in the changelog you have no idea what it refers to.
```suggestion
Added `support for small integer types <https://github.com/cctbx/cctbx_project/pull/533>`_ to DIALS flex arrays
```

Anthchirp (approved):

PR #1490, opened 2020-11-10
Ensure reproducible ordering of sym_ops in cosym

rjgildea:
Non-deterministic ordering of the sym_ops meant that the rij_matrix
could contain the same values, but in a different order. This can
lead to significant differences in the calculated functional, gradients
and curvatures, ultimately causing cosym to be non-deterministic,
despite setting random seeds.

### DIFF
--- a/algorithms/symmetry/cosym/target.py
+++ b/algorithms/symmetry/cosym/target.py
@@ -86,9 +86,7 @@ class Target(object):
 
         self._sym_ops = OrderedSet(["x,y,z"])
         self._lattice_group = lattice_group
-        self._sym_ops.update(
-            OrderedSet(op.as_xyz() for op in self._generate_twin_operators())
-        )
+        self._sym_ops.update(op.as_xyz() for op in self._generate_twin_operators())
         if dimensions is None:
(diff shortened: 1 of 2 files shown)


algorithms/symmetry/cosym/target.py
```diff
@@ -86,7 +86,9 @@ def __init__(
 
         self._sym_ops = OrderedSet(["x,y,z"])
         self._lattice_group = lattice_group
-        self._sym_ops.update({op.as_xyz() for op in self._generate_twin_operators()})
+        self._sym_ops.update(
+            OrderedSet(op.as_xyz() for op in self._generate_twin_operators())
+        )
```
ndevenish, line 91:
```suggestion
        self._sym_ops.update(op.as_xyz() for op in self._generate_twin_operators())
```
Possibly?

ndevenish: e.g. it was the going via set that was the problem and `OrderedSet.update` only expects that it's input is in the order you want to preserve

ndevenish (approved):
This is an extremely satisfying bugfix 👍

PR #1492, opened 2020-11-11
Use only cc_half by default in resolution analysis

rjgildea:
CC1/2 is now accepted as the most appropriate for estimating the resolution
cutoff (other than paired refinement). I/sigI-based cutoffs can be unreliable
due to dependence of the sigmas on error modelling in integration/scaling.
This is now consistent with xia2 default behaviour.

See also xia2/xia2#374

### DIFF
--- a/newsfragments/1492.feature
+++ b/newsfragments/1492.feature
@@ -1,1 +1,1 @@
-dials.estimate_resolution: use only cc_half by default in resolution analysis
+dials.estimate_resolution: change default to only use cc_half in resolution analysis
--- a/newsfragments/1492.feature
+++ b/newsfragments/1492.feature
@@ -1,0 +1,1 @@
+dials.estimate_resolution: use only cc_half by default in resolution analysis
--- a/test/command_line/test_estimate_resolution.py
+++ b/test/command_line/test_estimate_resolution.py
@@ -72,11 +72,7 @@ def test_multi_sequence_with_batch_range(dials_data, run_in_tmpdir, capsys):
     )
     captured = capsys.readouterr()
 
-    expected_output = (
-        "Resolution cc_half:       0.61",
-        "Resolution I/sig:         0.59",
-        "Resolution Mn(I/sig):     0.59",
-    )
+    expected_output = "Resolution cc_half:       0.61"
     for line in expected_output:
         assert line in captured.out
     assert run_in_tmpdir.join("dials.estimate_resolution.html").check(file=1)
--- a/util/resolution_analysis.py
+++ b/util/resolution_analysis.py
@@ -349,12 +349,12 @@ phil_str = """
   cc_half_fit = polynomial *tanh
     .type = choice
     .expert_level = 1
-  isigma = 0.25
+  isigma = None
     .type = float(value_min=0)
     .help = "Minimum value of the unmerged <I/sigI> in the outer resolution shell"
     .short_caption = "Outer shell unmerged <I/sigI>"
     .expert_level = 1
-  misigma = 1.0
+  misigma = None
     .type = float(value_min=0)
     .help = "Minimum value of the merged <I/sigI> in the outer resolution shell"
     .short_caption = "Outer shell merged <I/sigI>"


newsfragments/1492.feature
```diff
@@ -0,0 +1 @@
+dials.estimate_resolution: use only cc_half by default in resolution analysis
```
graeme-winter, line 1: possibly "change default to only use cc_half in resolution analysis"?

graeme-winter (approved):
Makes sense, though I wonder what will happen if $USER passes in a data set with multiplicity ~ 1? Overall though good that this enforces current best practice for consistency.

rjgildea:
> Makes sense, though I wonder what will happen if $USER passes in a data set with multiplicity ~ 1? Overall though good that this enforces current best practice for consistency.

It would only use those reflections with multiplicity > 1, obviously how many of these there are will depend on the symmetry etc. The `cc_half_significance` will also come into play here, as it will only use bins where the calculated CC1/2 is significant given the number of reflections in the bin. This could be a use case for `cc_half_method=sigma_tau`.

First scale a single narrow wedge of data:
```
$ dials.scale $(dials.data get -q multi_crystal_proteinase_k)/{experiments_1.json,reflections_1.pickle}
```

Run `dials.estimate_resolution` with defaults:
```
dials.estimate_resolution scaled.{expt,refl}
```
newplot (15)

Using `cc_half_method=sigma_tau`:
```
dials.estimate_resolution scaled.{expt,refl} cc_half_method=sigma_tau
```
newplot (16)

Using `misigma=1` (I don't know how reliable the sigmas will be after scaling with such low multiplicity):
```
dials.estimate_resolution scaled.{expt,refl} misigma=1
```
newplot (17)

dagewa (approved):
It makes sense for this to be consistent with xia2.

The issue of _what_ cut off value to use has come up again recently in CCP4 land. 0.143 has been suggested as being better than 0.3, because a CC1/2 of 0.143 corresponds to CC* of 0.5. But having a CC* of 0.5 is also an arbitrary value. The right value of the cut off is out of scope here, but I thought it might be worth noting the discussion.

jbeilstenedmands (approved):
Makes sense to me too.

To test a really low-multiplicity dataset, you can do something like this:
`$ dials.scale $(dials.data get -q l_cysteine_dials_output)/{11_integrated.expt,11_integrated.refl,23_integrated.refl,23_integrated.expt}`
`$ dials.estimate_resolution scaled.{expt,refl}`

N.B. This fails both on master and this branch with a stacktrace, but this shouldn't be a blocker for this work:
```  File "/dials/modules/dials/util/resolution_analysis.py", line 257, in resolution_cc_half
    merging_stats, metric, model, limit, sel=sel
  File "/dials/modules/dials/util/resolution_analysis.py", line 143, in resolution_fit_from_merging_stats
    return resolution_fit(d_star_sq, y_obs, model, limit, sel=sel)
  File "/Users/whi10850/dials/modules/dials/util/resolution_analysis.py", line 178, in resolution_fit
    raise RuntimeError("No reflections left for fitting")
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: No reflections left for fitting```

**Comment by rjgildea** [comment]
The issue here is that even with just 4 bins, there is only 1 reflection in each of 2 bins with a multiplicity > 1:
```
  Statistics by resolution bin:
 d_max  d_min   #obs  #uniq   mult.  %comp       <I>  <I/sI>    r_mrg   r_meas    r_pim   r_anom   cc1/2   cc_ano
  1.94   1.01     11     10    1.10   3.36    3856.6    39.3    0.004    0.005    0.004    0.007   0.000   0.000
  1.01   0.77      9      9    1.00   2.47    1965.0    24.0    0.000    0.000    0.000     0.000   0.000
  0.77   0.64     11     10    1.10   2.06    1170.4    19.1    0.130    0.184    0.130    0.251   0.000   0.000
  0.64   0.62      9      9    1.00   6.12     502.5    10.1    0.000    0.000    0.000     0.000   0.000
  1.94   0.62     40     38    1.05   2.82    1907.3    23.5    0.032    0.045    0.032    0.063   0.995   0.000
```
Even the sigma-tau CC1/2 method requires > 1 unique reflection with multiplicity > 1 in order to calculate a non-zero CC1/2 value:
https://github.com/cctbx/cctbx_project/blob/5ae04d2955fa897b34ed4ec5bd8aea427aeeeafc/cctbx/miller/__init__.py#L5127-L5158

Granted, we should probably gracefully handle the resulting error, but as you say, that is outside the context of the PR.

PR #1498, opened 2020-11-13
Update pytest-mock

Anthchirp:
Allow running with newer versions of `pytest-mock`


Anthchirp:
This needs changes in xia2 before it can be merged

PR #1502, opened 2020-11-19
Handle experiment identifiers for dials.slice_sequence block_size=x

jbeilstenedmands:
I found that trying to slice a dataset using the block_size option failed, due to non-handling of experiment identifiers, so the behaviour is fixed and a test added (this section of the code was untested).

```
DIAMRM5030:mpges whi10850$ dials.slice_sequence 7_refined.* block_size=10

block_size = 10
input {
  experiments = 7_refined.expt
  reflections = 7_refined.refl
}

  File "/dials/build/../modules/dials/command_line/slice_sequence.py", line 245, in <module>
    run()
  File "/dials/build/../conda_base/python.app/Contents/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/dials/build/../modules/dials/command_line/slice_sequence.py", line 241, in run
    script.run(args)
  File "/dials/build/../modules/dials/command_line/slice_sequence.py", line 178, in run
    sliced_experiments.append(exp)
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: dxtbx Internal Error: /dials/modules/dxtbx/model/experiment_list.h(190): DXTBX_ASSERT(index < 0) failure.
```

### DIFF
--- a/command_line/slice_sequence.py
+++ b/command_line/slice_sequence.py
@@ -167,7 +167,7 @@ class Script(object):
 
             if len(experiments) > 1:
                 raise Sorry(
-                    "For slicing into blocks please provide a single " "scan only"
+                    "For slicing into blocks please provide a single scan only"
                 )
             scan = experiments[0].scan
 
@@ -178,10 +178,8 @@ class Script(object):
             sliced = [
                 slice_experiments(experiments, [sr])[0] for sr in params.image_range
             ]
-            sliced_experiments = ExperimentList()
             generate_experiment_identifiers(sliced)
-            for exp in sliced:
-                sliced_experiments.append(exp)
+            sliced_experiments = ExperimentList(sliced)
 
             # slice reflections if present
             if slice_refs:
--- a/command_line/slice_sequence.py
+++ b/command_line/slice_sequence.py
@@ -6,7 +6,9 @@ from dxtbx.model.experiment_list import ExperimentList
 
 import dials.util
 from dials.algorithms.refinement.refinement_helpers import calculate_frame_numbers
+from dials.array_family import flex
 from dials.util import Sorry
+from dials.util.multi_dataset_handling import generate_experiment_identifiers
 from dials.util.slice import slice_experiments, slice_reflections
 
 help_message = """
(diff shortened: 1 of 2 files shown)


command_line/slice_sequence.py
```diff
@@ -158,33 +160,42 @@ def run(self, args=None):
...
-                sliced_experiments = ExperimentList()
-                for exp in sliced:
-                    sliced_experiments.append(exp)
+            sliced = [
+                slice_experiments(experiments, [sr])[0] for sr in params.image_range
+            ]
+            sliced_experiments = ExperimentList()
+            generate_experiment_identifiers(sliced)
+            for exp in sliced:
+                sliced_experiments.append(exp)
```
rjgildea, line 184: I think this should work and be simpler?
```suggestion
            generate_experiment_identifiers(sliced)
            sliced_experiments = ExperimentList(sliced)
```

command_line/slice_sequence.py
```diff
@@ -158,33 +160,42 @@ def run(self, args=None):
...
                 sliced = [
                     slice_reflections(reflections, [sr]) for sr in params.image_range
                 ]
-                sliced_reflections = sliced[0]
-                for i, rt in enumerate(sliced[1:]):
-                    rt["id"] += i + 1  # set id
+                sliced_reflections = flex.reflection_table()
+                identifiers = sliced_experiments.identifiers()
+                for i, rt in enumerate(sliced[0:]):
+                    if rt.experiment_identifiers():
```
rjgildea, line 194: Is this `if` statement necessary? Presumably if no identifiers set then the loop will exit at the first iteration...

command_line/slice_sequence.py
```diff
@@ -158,33 +160,42 @@ def run(self, args=None):
...
+                        for k in rt.experiment_identifiers().keys():
+                            del rt.experiment_identifiers()[k]
```
rjgildea, line 196: Wonder if an easier way of deleting all identifiers would be useful?

test/command_line/test_slice_sequence.py
```diff
@@ -58,14 +58,40 @@ def test_slice_sequence_with_first_images_missing(dials_regression, tmpdir):
...
+        [
+            "dials.slice_sequence",
+            "block_size=10",
+            "output.experiments=sliced.expt",
+            "output.reflections=sliced.refl",
+            expt,
+            refl,
+        ],
+        working_directory=tmpdir,
+    )
```
rjgildea, line 83: Could you call the command line script `run()` method directly and avoid the `procrunner` indirection?

test/command_line/test_slice_sequence.py
```diff
@@ -58,14 +58,40 @@ def test_slice_sequence_with_first_images_missing(dials_regression, tmpdir):
...
+
+    sliced_expts = load.experiment_list(
+        tmpdir.join("sliced.expt").strpath, check_format=False
+    )
+    print(list(sliced_expts.identifiers()))
```
rjgildea, line 88: Left over debugging print statement?

command_line/slice_sequence.py
```diff
@@ -158,33 +160,42 @@ def run(self, args=None):
...
             # slice reflections if present
             if slice_refs:
                 sliced = [
                     slice_reflections(reflections, [sr]) for sr in params.image_range
                 ]
-                sliced_reflections = sliced[0]
-                for i, rt in enumerate(sliced[1:]):
-                    rt["id"] += i + 1  # set id
+                sliced_reflections = flex.reflection_table()
+                identifiers = sliced_experiments.identifiers()
```
rjgildea, line 192: A comment around this block along the lines of "resetting experiment identifiers" might be useful for future readers

rjgildea (approved):
Needs some news :)

PR #1507, opened 2020-11-25
Remove tntbx

Anthchirp:
[As discussed in core meeting](https://dials.github.io/kb/core/20201125#remove-tntbx-dependency)


Anthchirp:
I have archived https://github.com/dials/tntbx

PR #1508, opened 2020-11-25
Add is_stills parameter to the spotfinder API to allow dials.stills_process to work with scans

### LINKED ISSUE #1180: still_process fails spotfinding on image(s) in middle of range
Using https://zenodo.org/record/55058 as a test data set:
```
$ dials.stills_process test_1/test_1_0400.cbf.bz2
....
Error spotfinding test_1_0400.cbf 3
```
However, normal dials import works:
```
$ dials.import test_1/test_1_0400.cbf.bz2 && dials.find_spots imported.expt
...
Found 19 strong pixels on image 400
...
Saved 3 reflections to strong.refl
```

phyy-nx:
Fixes #1180

### DIFF
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -403,7 +403,7 @@ class SpotFinderFactory:
     """
 
     @staticmethod
-    def from_parameters(params=None, experiments=None):
+    def from_parameters(params=None, experiments=None, is_stills=False):
         """
         Given a set of parameters, construct the spot finder
 
@@ -464,6 +464,7 @@ class SpotFinderFactory:
             max_spot_size=params.spotfinder.filter.max_spot_size,
             no_shoeboxes_2d=no_shoeboxes_2d,
             min_chunksize=params.spotfinder.mp.min_chunksize,
+            is_stills=is_stills,
         )
 
     @staticmethod
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -408,6 +408,8 @@ class SpotFinderFactory:
         Given a set of parameters, construct the spot finder
 
         :param params: The input parameters
+        :param is_stills:   [ADVANCED] Force still-handling of experiment
+                            ID remapping for dials.stills_process.
         :returns: The spot finder instance
         """
         if params is None:
--- a/command_line/stills_process.py
+++ b/command_line/stills_process.py
@@ -1021,7 +1021,9 @@ The detector is reporting a gain of %f but you have also supplied a gain of %f. 
         logger.info("*" * 80)
 
         # Find the strong spots
-        observed = flex.reflection_table.from_observations(experiments, self.params)
+        observed = flex.reflection_table.from_observations(
+            experiments, self.params, is_stills=True
+        )
 
         # Reset z coordinates for dials.image_viewer; see Issues #226 for details
         xyzobs = observed["xyzobs.px.value"]
--- a/algorithms/spot_finding/finder.py
+++ b/algorithms/spot_finding/finder.py
@@ -675,6 +675,7 @@ class SpotFinder:
         max_spot_size=20,
         no_shoeboxes_2d=False,
         min_chunksize=50,
+        is_stills=False,
     ):
         """
         Initialise the class.
@@ -703,6 +704,7 @@ class SpotFinder:
         self.mp_njobs = mp_njobs
         self.no_shoeboxes_2d = no_shoeboxes_2d
         self.min_chunksize = min_chunksize
+        self.is_stills = is_stills
 
     def __call__(self, experiments):
         warnings.warn(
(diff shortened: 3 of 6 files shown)


algorithms/spot_finding/factory.py
```diff
@@ -403,7 +403,7 @@ class SpotFinderFactory:
     """
 
     @staticmethod
-    def from_parameters(params=None, experiments=None):
+    def from_parameters(params=None, experiments=None, is_stills=False):
```
graeme-winter, line 406: `is_stills` -> `nasty_hack_is_stills_flag`

Predicate: this fix is sufficient to fix `dials.stills_process` with data which is in `ImageSequence`s not `ImageSet`s (which would be very useful)

Annotation: adding `is_stills` everywhere is a nasty hack and one which I would like to see reverted once we have resolved the underlying design issues in `dxtbx`. I am however very aware that this will not happen in a timely manner and meanwhile this fix is very well justified. What I would _not_ like to happen is have this stay in the API for the foreseeable future -> being explicit that this is a nasty (if justified) hack seems sensible.

algorithms/spot_finding/factory.py
```diff
@@ -464,6 +464,7 @@ def from_parameters(params=None, experiments=None):
             max_spot_size=params.spotfinder.filter.max_spot_size,
             no_shoeboxes_2d=no_shoeboxes_2d,
             min_chunksize=params.spotfinder.mp.min_chunksize,
+            is_stills=is_stills,
```
graeme-winter, line 467: Without further comment I note that this is parameter 19 to a constructor 🙄

To me this suggests that this code needs refactoring (as well as emphasising that the parameter addition is probably not a good thing in the wider scope)

I am not in any sense asking for this to be addressed here and now, though I think raising an issue following this code being merged would be appropriate and justified.

command_line/stills_process.py
```diff
@@ -1021,7 +1021,9 @@ def find_spots(self, experiments):
         logger.info("*" * 80)
 
         # Find the strong spots
-        observed = flex.reflection_table.from_observations(experiments, self.params)
+        observed = flex.reflection_table.from_observations(
```
graeme-winter, line 1024: Codecov notes that this is not covered by the tests: I think this is a valid concern 🙂

I wonder if we could add a very small test for this code?

graeme-winter (changes requested):
Assuming that this change set is sufficient to have stills process work on ImageSequence data (which happen to be non-rotating) I am happy to see this merged _after_ the `is_stills` flag is renamed to something which indicates that it is (i) transient (ii) not a part of the published API and (iii) a nasty temporary hack to get things working after other changes. 

Though I proposed a rather strongly worded replacement in the review comments maybe `is_still_kludge` or something similar would be reasonable?

graeme-winter:
@phyy-nx  did you have thoughts on this with particular reference to `is_still_kludge` ?

algorithms/spot_finding/factory.py
phyy-nx: Issue raised

command_line/stills_process.py
phyy-nx: Added!  The new test triggers #1180 on master but not on the `dsp_fix_grid_scan` branch.  I don't think Codecov will pick it up as it uses a external call to stills process.

phyy-nx:
> @phyy-nx did you have thoughts on this with particular reference to `is_still_kludge` ?

I'd prefer `is_stills`, especially since we (or at least I) have no intention of changing it anytime soon and I like nice variables :)  I honestly don't see a path to fixing this differently, otherwise I'd do it!

I think I've otherwise addressed the review comments?

ndevenish:
I'm okay with `is_stills` as a name, it's likely that once in this will become a baked part of the interfaces so it might as well have a good name.

I'd really like to understand the root cause here and why the inferred information from missing/zero-oscillation ImageSet/ImageSequence isn't sufficient to distinguish these cases?

phyy-nx:
From dials core meeting: Aaron to merge later today after ND adds a news fragment, then add a ticket to resolve this in the dx2 re-write

Anthchirp:
Added my test refactor, hope that's okay with you. Relevant meeting minutes are at https://dials.github.io/kb/core/20210114#dials

phyy-nx:
👍

ndevenish:
I'm just trying to pull a test case together to double check. Ideally we should have added an xfail test for this already.

phyy-nx:
test_pseudo_scan in test_stills_process (not in master) will fail on master if you cherry pick it.

ndevenish:
yep, thanks, that and the original case seem to work fine, same results as <2.1

ndevenish:
Okay, so I added news, and also added:
```diff
+        :param is_stills:   [ADVANCED] Force still-handling of experiment
+                            ID remapping for dials.stills_process.
```
to the docstrings for the functions. Hopefully this should dissuade/remind anyone that this isn't a general flag but breaks assumptions in a limited and specific way.

I think this is fine to squash-merge now (unless you have disagreements about this docstring in which case... probably better to merge without it and sort something out later).

phyy-nx:
Thanks

PR #1514, opened 2020-12-07
Cosym: choose cluster containing the most identity reindexing ops

rjgildea:
This can be important if the lattice symmetry is only very approximately related by pseudosymmetry to the true symmetry. If all potential reindexing ops are genuine indexing ambiguities, then it doesn't matter which one is chosen, however if not, then choosing the wrong one will distort the true unit cell. In such cases it is likely that the input datasets were already indexed consistently, therefore default to choosing the cluster that contains the most identity reindexing ops.

### DIFF
--- a/algorithms/symmetry/cosym/__init__.py
+++ b/algorithms/symmetry/cosym/__init__.py
@@ -175,10 +175,12 @@ class CosymAnalysis(symmetry_base, Subject):
                 sg_best = sg_primitive.change_basis(cb_op_best_primitive.inverse())
                 # best_subgroup above is the bravais type, so create thin copy here with the
                 # user-input space group instead
+                best_subsym = best_subsym.customized_copy(
+                    space_group_info=sg_best.info()
+                )
                 best_subgroup = {
-                    "best_subsym": best_subsym.customized_copy(
-                        space_group_info=sg_best.info()
-                    ),
+                    "subsym": best_subsym.change_basis(cb_op_inp_best.inverse()),
+                    "best_subsym": best_subsym,
                     "cb_op_inp_best": cb_op_inp_best,
                 }
 
@@ -297,7 +299,7 @@ class CosymAnalysis(symmetry_base, Subject):
     def _optimise(self, termination_params):
         NN = len(self.input_intensities)
         dim = self.target.dim
-        n_sym_ops = len(self.target.get_sym_ops())
+        n_sym_ops = len(self.target.sym_ops)
         coords = flex.random_double(NN * n_sym_ops * dim)
 
         import scitbx.lbfgs
@@ -357,9 +359,7 @@ class CosymAnalysis(symmetry_base, Subject):
             self._symmetry_analysis = None
             return
 
-        sym_ops = [
-            sgtbx.rt_mx(s).new_denominators(1, 12) for s in self.target.get_sym_ops()
-        ]
+        sym_ops = [sgtbx.rt_mx(s).new_denominators(1, 12) for s in self.target.sym_ops]
         self._symmetry_analysis = SymmetryAnalysis(
             self.coords, sym_ops, self.subgroups, self.cb_op_inp_min
         )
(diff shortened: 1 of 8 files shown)


algorithms/symmetry/cosym/target.py
```diff
@@ -84,15 +85,15 @@ def __init__(
                 last_id = lid
                 self._lattices.append(n)
 
-        self._sym_ops = OrderedSet(["x,y,z"])
+        self.sym_ops = OrderedSet(["x,y,z"])
```
graeme-winter, line 88: Curious why you moved this from "_private"

algorithms/symmetry/cosym/__init__.py
```diff
@@ -175,10 +175,12 @@ def _map_space_group_to_input_cell(intensities, space_group):
                 sg_best = sg_primitive.change_basis(cb_op_best_primitive.inverse())
                 # best_subgroup above is the bravais type, so create thin copy here with the
                 # user-input space group instead
+                best_subsym = best_subsym.customized_copy(
```
graeme-winter, line 178: I think this stuff could do with some kind of annotation somewhere... I am certain you have gained some understanding as a side-effect of these changes which should probably be documented inline?i.e. the _why_ of `est_subsym.change_basis(cb_op_inp_best.inverse())` etc.

I ask this as someone who may have to debug this one day :-)

algorithms/symmetry/cosym/__init__.py
```diff
@@ -370,28 +372,6 @@ def _analyse_symmetry(self):
         )
         self.params.cluster.n_clusters = len(cosets.partitions)
 
-    def _space_group_for_dataset(self, dataset_id, sym_ops):
```
graeme-winter, line 375: You're not wrong that retaining this adds confusion, as I was trying to work out how things still work with all this stuff removed...

algorithms/symmetry/cosym/__init__.py
```diff
@@ -380,14 +380,18 @@ def _reindexing_ops_for_dataset(self, dataset_id, sym_ops, cosets):
...
         for i_cluster in range(n_clusters):
-            isel = (self.cluster_labels == i_cluster).iselection()
-            dataset_ids = isel % len(self.input_intensities)
-            sel = (dataset_ids == dataset_id).iselection()
-            for s in sel:
-                sym_op_id = isel[s] // len(self.input_intensities)
-                for partition in cosets.partitions:
-                    if sym_ops[sym_op_id] in partition:
-                        if i_cluster not in reindexing_ops:
+            # Select all points within this cluster
```
graeme-winter, line 378: This is quite a large re-arrangement of code.... which looks like something which would benefit from a unit test to show that this is tidying up / refactoring not changing behaviour (and if that is a bug fix, adding a regession test)

command_line/cosym.py
```diff
@@ -177,9 +178,17 @@ def run(self):
         self.cosym_analysis.run()
 
         reindexing_ops = {}
+        sym_op_counts = {
```
graeme-winter, line 181: Is behaviour well defined in the unlikely event of a tie?

command_line/cosym.py
```diff
@@ -177,9 +178,17 @@ def run(self):
...
+            cluster_id: collections.Counter(
```
graeme-winter, line 182: Wait, what is this syntax?

Learning opportunity for YT

graeme-winter: Oh, dictionary creation from iterator...

graeme-winter: Possibly re-arranging the code slightly (e.g. a dummy variable to move everything onto a single line) would make it look less magic?

newsfragments/1514.bugfix
```diff
@@ -0,0 +1 @@
+dials.cosym: by default choose cluster containing the most identity reindexing ops. Under some circumstances, particularly in the case of pseudosymmetry, the previous behaviour could result in reindexing operators being chosen that weren't genuine indexing ambiguities, instead distorting the input unit cells.
```
graeme-winter, line 1: ... case of _approximate_ pseudosymmetry is where this has greatest impact

graeme-winter (approved):
Popped a couple of comments inline, mostly asking for small improvements to the documentation of the code as there are some `sgtbx` type rabbit holes in this... I assume that the code does what the commit messages say and overall the spirit of the change set makes a lot of sense. I would wonder if this is a case for _not_ squash merging however as the structure of the PR does have bug fix and tidying commits separately. Maybe squash down to 2 commits? So bugfix+news, cleaning+test changes

algorithms/symmetry/cosym/target.py
rjgildea: Because previously it was unnecessarily hidden behind a `self.get_sym_ops()` method (see 43f44952d78283350f47f3dc552ddc896001e470).

graeme-winter: I couldn't see the removal of the coresponding `get_sym_ops()` call in the commit -> wondering. Not a problem, just a question.

graeme-winter: Ah, because _deprecated_ and warning added -> gotcha.

algorithms/symmetry/cosym/__init__.py
rjgildea: It's mostly just renaming variables and adding comments for clarity (added as I re-looked at the code to figure out what is going on). The only real change is the addition of the `break` and moving `if i_cluster not in reindexing_ops` to the middle loop. Previously it would have continued iterating over the inner two loops after adding `i_cluster` to `reindexing_ops`, but not doing anything because of the inner `if i_cluster not in reindexing_ops` check. Now with the addition of the `break` and moving the `if` statement up a level, it will effectively skip straight to the next outer loop.

rjgildea: Looking at this again, I wonder if changing this to:
```
                if i_cluster in reindexing_ops:
                    break
```
and dedenting the rest of the code might be better.

graeme-winter: That may well be clearer. 

I am not questioning the correctness of the code, only how easy it is for a person-in-the-future to understand

algorithms/symmetry/cosym/__init__.py
rjgildea: This is just to obtain the `subsym` in the input setting rather than the "best" setting -> `cb_op_inp_best.inverse()`

algorithms/symmetry/cosym/__init__.py
rjgildea: Added a minimal test in https://github.com/dials/dials/pull/1514/commits/0ed77e801985d37175e5fe13be7fc795d1adf426. I cherry-picked this commit to master and it passed there too.

PR #1515, opened 2020-12-08
Fix some minor issues in the indexing code

benjaminhwilliams:
Catch some undefined variables and stuff like that.

* Previously `entry_point` in `getattr` call as part of `strategy_class` call in `dials.algorithms.indexing.lattice_search.LatticeSearch.__init__` would be the value of the last entry point in `pkg_resources.iter_entry_points("dials.index.lattice_search")`, not the matching entry point.
* ~Buried deep in `dials.algorithms.indexing.indexer.Indexer.index`, there's a code path with an undefined variable `refined_reflections`.  Presumably, `reflections_for_refinement` is what was intended.~
* Remove the impossible default parameter `params=None` in initialising `dials.algorithms.indexing.indexer.Indexer` and its subclasses.  `None.indexing` is always going to raise an `AttributeError`.

### DIFF
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -654,12 +654,12 @@ class Indexer(object):
                         # with this deleted experiment - indexed flag removed
                         # below
                         last = len(experiments)
-                        sel = reflections_for_refinement["id"] == last
+                        sel = refined_reflections["id"] == last
                         logger.info(
                             "Removing %d reflections with id %d"
                             % (sel.count(True), last)
                         )
-                        reflections_for_refinement["id"].set_selected(sel, -1)
+                        refined_reflections["id"].set_selected(sel, -1)
 
                         break
 
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -654,12 +654,12 @@ class Indexer(object):
                         # with this deleted experiment - indexed flag removed
                         # below
                         last = len(experiments)
-                        sel = refined_reflections["id"] == last
+                        sel = reflections_for_refinement["id"] == last
                         logger.info(
                             "Removing %d reflections with id %d"
                             % (sel.count(True), last)
                         )
-                        refined_reflections["id"].set_selected(sel, -1)
+                        reflections_for_refinement["id"].set_selected(sel, -1)
 
                         break
 
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -571,6 +571,8 @@ class Indexer(object):
                 # no more lattices found
                 break
 
+            refined_experiments = None
+            refined_reflections = None
             for i_cycle in range(self.params.refinement_protocol.n_macro_cycles):
                 if (
                     i_cycle > 0
--- a/algorithms/indexing/indexer.py
+++ b/algorithms/indexing/indexer.py
@@ -301,7 +301,7 @@ phil_scope = iotbx.phil.parse(phil_str, process_includes=True)
 
 
 class Indexer(object):
-    def __init__(self, reflections, experiments, params=None):
+    def __init__(self, reflections, experiments, params):
         self.reflections = reflections
         self.experiments = experiments
 
--- a/algorithms/indexing/lattice_search/__init__.py
+++ b/algorithms/indexing/lattice_search/__init__.py
@@ -108,7 +108,7 @@ basis_vector_search_phil_scope.adopt_scope(
 
 
 class LatticeSearch(indexer.Indexer):
-    def __init__(self, reflections, experiments, params=None):
+    def __init__(self, reflections, experiments, params):
         super(LatticeSearch, self).__init__(reflections, experiments, params)
 
         self._lattice_search_strategy = None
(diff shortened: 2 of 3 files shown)


benjaminhwilliams:
I find it a little difficult to figure out how to use these classes and what to pass them as arguments/how to mock those things, so I'm a bit stumped when it comes to writing tests.

algorithms/indexing/indexer.py
```diff
@@ -654,12 +654,12 @@ def index(self):
                         # with this deleted experiment - indexed flag removed
                         # below
                         last = len(experiments)
-                        sel = refined_reflections["id"] == last
+                        sel = reflections_for_refinement["id"] == last
```
rjgildea, line 657: The original code was correct. This line is only hit in multi-lattice indexing, and there was a refinement failure after indexing a second or subsequent lattice - so the `refined_reflections` variable exists from the previous loop iteration. In which case, we delete the last experiment and remove the experiment ids associated with this experiment from the reflection table and reset the ids.

algorithms/indexing/indexer.py
```diff
@@ -654,12 +654,12 @@ def index(self):
...
                         logger.info(
                             "Removing %d reflections with id %d"
                             % (sel.count(True), last)
                         )
-                        refined_reflections["id"].set_selected(sel, -1)
+                        reflections_for_refinement["id"].set_selected(sel, -1)
```
rjgildea, line 662: As above

rjgildea (changes requested):

algorithms/indexing/indexer.py
rjgildea: See also https://github.com/dials/dials/commit/8f30e07556ec99216f03668e9cc0de857e5a8c66

benjaminhwilliams: Ah, OK, sorry, I just couldn't work it out from the code.  I'll revert that commit.

ndevenish:
Could we maybe add
```python
refined_experiments = None
refined_reflections = None
```
before the main loop that this happens in e.g. at line 573: https://github.com/dials/dials/blob/b36efe50b5118c43a243dd5801de9929640018e4/algorithms/indexing/indexer.py#L573-L574 - that'd at least make it explicit that these variables are supposed to scope over multiple loop runs. I'm guessing the first time through the loop that `self.params.refinement_protocol.mode == "repredict_only"`?

benjaminhwilliams:
Yes please, I'd be very happy with that, if @rjgildea is.

ndevenish:
The  `strategy_class` stuff in `LatticeSearch` looks like an actual bug, to me, unless someone can persuade me otherwise. In which case, it'll also need a `.bugfix` newsfragment 😀

benjaminhwilliams:
Sure thing.  I have a newsfragment that I haven't committed yet.  I was holding off until we knew whether it ought to be `.bugfix` or `.misc`.

rjgildea:
> The `strategy_class` stuff in `LatticeSearch` looks like an actual bug, to me, unless someone can persuade me otherwise. In which case, it'll also need a `.bugfix` newsfragment 😀

It hasn't been an issue thus far because there is currently only one entry point defined for `dials.index.lattice_search` so the loop implicitly exits after the first iteration meaning that it works by happy accident. Of course it would likely be broken if any further entry points were added.

https://github.com/dials/dials/blob/b36efe50b5118c43a243dd5801de9929640018e4/libtbx_refresh.py#L23-L25

rjgildea (approved):

ndevenish:
Ah 👍 I was thinking of basis vectors searches!

algorithms/indexing/indexer.py
```diff
@@ -571,6 +571,8 @@ def index(self):
                 # no more lattices found
                 break
 
+            refined_experiments = None
+            refined_reflections = None
```
rjgildea, line 575: This change is actually in the wrong loop level, and breaks the very test that was designed to exercise the relevant try/except loop:
```
$ pytest algorithms/indexing/test_index.py::test_refinement_failure_on_max_lattices_a15 --regression
...
  File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/dials/command_line/index.py", line 261, in <module>
    run()
  File "/Users/rjgildea/software/cctbx_py3/conda_base/python.app/Contents/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/dials/command_line/index.py", line 244, in run
    indexed_experiments, indexed_reflections = index(
  File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/dials/command_line/index.py", line 162, in index
    indexed_experiments, indexed_reflections = _index_experiments(
  File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/dials/command_line/index.py", line 114, in _index_experiments
    idxr.index()
  File "/Users/rjgildea/software/cctbx/modules/dials/algorithms/indexing/indexer.py", line 659, in index
    sel = refined_reflections["id"] == last
TypeError: Please report this error to dials-support@lists.sourceforge.net: 'NoneType' object is not subscriptable
```

https://github.com/dials/dials/blob/fcfd7b452e93a82e718aba9663e5712e0244a821/algorithms/indexing/test_index.py#L567-L604

PR #1520, opened 2020-12-11
Sphinx library documentation: autodoc __init__ methods

rjgildea:


rjgildea:
E.g. despite having a lengthy documentation string in `dials.algorithsm.symmetry.cosym.target.Target.__init__`, this does not currently appear in the auto-generated sphinx documentation:

https://github.com/dials/dials/blob/29c2830be8cbac07addbee821499d374f5edbf2c/algorithms/symmetry/cosym/target.py#L19-L59

https://dials.github.io/documentation/library_reference/algorithms/dials.algorithms.symmetry.html#dials.algorithms.symmetry.cosym.target.Target

ndevenish:
What are your thoughts on this vs moving to documenting `__init__` on the class-level docstring?

ndevenish:
Actually, that's probably a discussion best kept for another time - this change is simple enough with enough benefit to justify. We'd want to do this even if we agreed to transition to the other approach.

PR #1522, opened 2020-12-11
dials.merge: Fix output of SigF, N+, N- in merged.mtz

jbeilstenedmands:
For anomalous=True (the default), when merging anomalous Fs to normal Fs,
don't use_internal_variances. This gives an incorrect (overly high) value for SigF by several factors.

Also, the miller index multiplicities were being written as N+,
N- however this was intended to be the redundancy of observations
of each symmetry group, which is fixed here.

### DIFF
--- a/1522.bugfix
+++ b/1522.bugfix
@@ -1,0 +1,1 @@
+dials.merge: Fix incorrect output of SigF, N+, N- in merged.mtz
--- a/algorithms/merging/merge.py
+++ b/algorithms/merging/merge.py
@@ -93,7 +93,7 @@ class MTZDataClass(object):
         merged_anomalous_array=None,
         amplitudes=None,
         anomalous_amplitudes=None,
-        redundancies=None,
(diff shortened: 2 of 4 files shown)


jbeilstenedmands:
Note: this does not affect the `xia2` pipeline which uses cctbx_FrenchWilson to convert Is to Fs, only manual merging with `dials.merge`

rjgildea (commented):
I think the term "multiplicity" is generally preferable to "redundancy" - can we use this instead? It would also be more consistent with existing usage within dials:
```
$ find . -name "*.py" -type f | xargs grep -i multiplicit | wc -l
      93
$ find . -name "*.py" -type f | xargs grep -i redundanc | wc -l  
       7
```
and xia2:
```
$ find . -name "*.py" -type f | xargs grep -i multiplicit | wc -l
     101
$ find . -name "*.py" -type f | xargs grep -i redundanc | wc -l 
      12
```

PR #1532, opened 2021-01-05
Fix tests after adding "number of images" to dials.show output

dagewa:
Fix `dials.show` tests following change to add `number of images` to the output, in https://github.com/cctbx/dxtbx/pull/271


dagewa:
Thanks @ndevenish, appreciated. I would have got round to it eventually, but it wouldn't have been this week, so :+1:

PR #1534, opened 2021-01-06
Remove "run_tests.py" from data files

RobertRosca:
`"run_tests.py"` is marked as a required data file for the package, but the file is now missing (seems to have been recently removed?).

So now running pip install on the repository throws an error:

```
  running install_data
  creating build/bdist.linux-x86_64/wheel/dials-0.0.1.data
  creating build/bdist.linux-x86_64/wheel/dials-0.0.1.data/data
  creating build/bdist.linux-x86_64/wheel/dials-0.0.1.data/data/dials
  copying conftest.py -> build/bdist.linux-x86_64/wheel/dials-0.0.1.data/data/dials
  copying __init__.py -> build/bdist.linux-x86_64/wheel/dials-0.0.1.data/data/dials
  copying libtbx_refresh.py -> build/bdist.linux-x86_64/wheel/dials-0.0.1.data/data/dials
  error: can't copy 'run_tests.py': doesn't exist or not a regular file
  ----------------------------------------
  ERROR: Failed building wheel for dials
Failed to build dials
ERROR: Could not build wheels for dials which use PEP 517 and cannot be installed directly
```

Removing the reference to `"run_tests.py"` fixes the error.

### DIFF
--- a/newsfragments/1534.misc
+++ b/newsfragments/1534.misc
@@ -1,0 +1,1 @@
+fix up setup.py in preparation for future packaging work
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     packages=find_packages(),
     package_dir={"dials": "../dials"},
     data_files=[
-        ("dials", ["conftest.py", "__init__.py", "libtbx_refresh.py", "run_tests.py"])
+        ("dials", ["conftest.py", "__init__.py", "libtbx_refresh.py"])
     ],
     setup_requires=setup_requirements,
     test_suite="tests",


Anthchirp:
Thank you for your contribution. You are correct, and this is an issue - however, a word of warning: DIALS isn't a regular python project and we don't support the installation via `setup.py`/pip at this time, part of the reason for this is that right now the entire package layout is not conducive to installing DIALS as a normal python package.

For now I'd recommend following the installation instructions in https://dials.github.io/documentation/installation_developer.html instead

ndevenish:
I wonder if we should put some special code that fails `setup.py` installation if someone is trying this. Like, detecting the presence of the environment variable `YES_I_KNOW_THIS_IS_BROKEN` or something.

RobertRosca:
> Thank you for your contribution. You are correct, and this is an issue - however, a word of warning: DIALS isn't a regular python project and we don't support the installation via `setup.py`/pip at this time, part of the reason for this is that right now the entire package layout is not conducive to installing DIALS as a normal python package.
> 
> For now I'd recommend following the installation instructions in https://dials.github.io/documentation/installation_developer.html instead

Aha, makes sense, thanks! I'm working with a student who's helping us at European XFEL to create [Spack](https://spack.readthedocs.io/en/latest/) packages for some of the software we use, as we're planning on using Spack to manage our software installations at EuXFEL. Asked for help with some mysterious `Failed building wheel for dials` error and I didn't look into it much past noticing the missing data files, my bad.

> I wonder if we should put some special code that fails setup.py installation if someone is trying this. Like, detecting the presence of the environment variable YES_I_KNOW_THIS_IS_BROKEN or something.

Something like that should be possible, glanced through the bootstrap script and it's not too clear to me if/when setup.py gets used. It's possible to do something like:

```python
if __name__ == "__main__":
    raise Exception("manual use of setup.py is not supported")
```

And I think that during pip install, setup.py gets called with `['-c', 'egg_info']`, so a more reliable check might be something like:

```python
if __name__ == "__main__":
    if 'egg_info' in sys.argv:
        raise Exception("manual use of setup.py is not supported")
```

Either of those would throw:

```
    File "setup.py", line 52, in <module>
      raise Exception("manual use of setup.py is not supported")
  Exception: manual use of setup.py is not supported
```

ndevenish:
My recollection was that the `setup.py` was used for configuring entrypoints, but it looks like that's done elsewhere (https://github.com/dials/dials/blob/master/libtbx_refresh.py is the configuration inside the libtbx build).

I imagine this was added as some initial explorations into trying to get it to build like a normal package. I don't know whether it's worth keeping this in at all.

We've started planning towards being more of a normal package as part of moving towards having a `conda install dials`, but it's a way off and there are other problems to solve.

Anthchirp (approved):
Well, fixing up the `setup.py` is on the cards hopefully this year, following a pattern based on [what we did with xia2](https://github.com/xia2/xia2/pull/528). Until then I'm happy to merge this either way.

PR #1538, opened 2021-01-08
Debug log message when rejecting lattices

rjgildea:
More verbose debug logs from dials.index when rejecting crystal models that are inconsistent with input symmetry


dagewa (approved):
Looks good, thanks

Issue #144, opened 2016-05-19
mp.method fails with Eiger data for spot finding


graeme-winter:
<pre>
[gw56@cs04r-sc-com99-01 parallel]$ dials.find_spots ../datablock.json njobs=10 nproc=20 mp.method=sge
DIALS 1.dev.297-gd7ebd46
The following parameters have been modified:

spotfinder {
  mp {
    method = none *sge lsf pbs
    njobs = 10
    nproc = 20
  }
}
input {
  datablock = ../datablock.json
}

Setting spotfinder.filter.min_spot_size=3
Configuring spot finder from input parameters
--------------------------------------------------------------------------------
Finding strong spots in imageset 0
--------------------------------------------------------------------------------

Finding spots in image 1 to 1800...
Extracting strong pixels from images
 Using sge with 10 parallel job(s) and 20 processes per node

gw56 has deleted job 151556
Traceback (most recent call last):
  File "/dls/science/users/gw56/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 196, in <module>
    halraiser(e)
  File "/dls/science/users/gw56/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 194, in <module>
    script.run()
  File "/dls/science/users/gw56/svn/cctbx/build/../modules/dials/command_line/find_spots.py", line 143, in run
    datablocks[0], params)
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/array_family/flex.py", line 173, in from_observations
    return find_spots(datablock)
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 415, in __call__
    table, hot_mask = self._find_in_imageset(imageset)
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 499, in _find_in_imageset
    spots_all.extend(extract_spots(imageset[j0:j1]))
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/algorithms/spot_finding/finder.py", line 301, in __call__
    callback       = process_output)
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/util/mp.py", line 225, in batch_multi_node_parallel_map
    preserve_exception_message = True)
  File "/dls/science/users/gw56/svn/cctbx/modules/dials/util/mp.py", line 134, in multi_node_parallel_map
    preserve_exception_message = preserve_exception_message)
[...]

PR #1539, opened 2021-01-08
Hack to avoid call to nullify_format_instance()

rjgildea:
The call to `nullify_format_instance()` is only necessary for HDF5 when we are using multiprocessing. Here, we are using a single processor, so it only adds unnecessary overhead. Avoiding the call can save ~0.15 seconds per image when this function is called via the dlstbx PIA service (~10% of the runtime).

See also https://github.com/dials/dials/issues/144


ndevenish:
Although as per #1539 I think this is actually never necessary any more, I don't think this risks breaking anything at all.

ndevenish (approved):

PR #1542, opened 2021-01-12
Strip out calls to nullify_format_instance

ndevenish:
We had problems in the past with the behaviour of HDF5 under forking/parallel reads e.g. #35 #144 https://github.com/dials/dials/commit/01463cc47d963d68bbd6e546c7fc6fbab3f1760d https://github.com/dials/dials/commit/2550f7c9f704b547901c064ea932141ba0dc9ead https://github.com/dials/dials/commit/b19510e7b50f5367dc49d0ae5fe80d7c9359f4ce https://github.com/dials/dials/commit/36f63d63f9c70139bd53950dc4eca05c56e2cfea and also #1539.

While testing [a fix](https://github.com/ndevenish/dxtbx/commit/20979a52945596d3dfe2cd2b5a3d9f6aed695551) to centralise this check I stripped out the behaviour of `nullify_format_instance` and found that... I couldn't reproduce any of the problems. After some [digging](https://forum.hdfgroup.org/t/hdf5-and-parallelism-with-fork-2/5225), it looks like in HDF5 [1.10.5](https://portal.hdfgroup.org/display/support/HDF5+1.10.5) this problem was fixed by:

> Added a new option to enable/disable using pread/pwrite instead of
      read/write in the sec2, log, and core VFDs.
 > This option is enabled by default when pread/pwrite are detected.

Which does reading/writing in a stateless way. So, all of these workaround hacks have become redundant.

Possibly.

All the tests I've run appear to pass - this doesn't remove the need for `nullify_format_instance` completely because there seem to be some vestigial problem with Lazy* formats https://github.com/cctbx/dxtbx/issues/245. Does anyone have any memories/documented cases where this broke before to run tests through also? Otherwise, I'd suggest that we put it in and see if any problems come up related to it.

### DIFF
--- a/algorithms/integration/image_integrator.py
+++ b/algorithms/integration/image_integrator.py
@@ -266,13 +266,6 @@ class ManagerImage(object):
         frames = self.experiments[0].scan.get_array_range()
         self.manager = ReflectionManagerPerImage(frames, self.reflections)
 
-        # Parallel reading of HDF5 from the same handle is not allowed. Python
-        # multiprocessing is a bit messed up and used fork on linux so need to
-        # close and reopen file.
-        for exp in self.experiments:
-            if exp.imageset.reader().is_single_file_reader():
-                exp.imageset.reader().nullify_format_instance()
-
         # Set the initialization time
         self.time.initialize = time() - start_time
 
--- a/algorithms/integration/parallel_integrator.py
+++ b/algorithms/integration/parallel_integrator.py
@@ -566,11 +566,6 @@ class IntegrationManager(object):
             self.blocks, self.reflections, self.params.integration.mp.njobs
         )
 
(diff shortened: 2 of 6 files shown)


phyy-nx:
Did you test it under MPI conditions?  test_stills_process.py has an mpi test in it.

phyy-nx:
conda install -c conda-forge mpi4py openmpi
mpirun hostname # test mpi is working, should see hostname for each available rank
libtbx.pytest -n 3 test_stills_process.py --regression # all three tests should run 

I did the above on master and on the nullify_pid branch and the tests pass.  I presume it's ok then.

ndevenish:
I've merged this now, with the understanding that we should back it out quickly if it is causing any apparent problems.

PR #1551, opened 2021-01-16
Include a legend on dials.background plots.

dagewa:
Draw a legend, but only when there are multiple lines, up to the size of the color cycle.


dagewa:
Superseded by #1554

PR #1554, opened 2021-01-18
Support multiple imagesets in dials.background

rjgildea:
Remove top-level `plot` parameter as likely to cause confusion with recently-added `output.plot` parameter.

```
$ dials.background $(dials.data get -q thaumatin_grid_scan)/thau_3_2_00*.cbf.bz2 $(dials.data get -q centroid_test_data)/centroid_000*.cbf plot=background.png images=1,2 size_inches=16,10
```
image


ndevenish:
Trying to reconcile the merge-error with https://github.com/dials/dials/commit/342a0543b88f7ccf30a0f5958677f06326b11884 - this seems to completely remove the `show()` behaviour, as in - there is no way to get the plot displayed interactively, as it did before. Was this intended?

rjgildea:
> Trying to reconcile the merge-error with [342a054](https://github.com/dials/dials/commit/342a0543b88f7ccf30a0f5958677f06326b11884) - this seems to completely remove the `show()` behaviour, as in - there is no way to get the plot displayed interactively, as it did before. Was this intended?

Yes - I don't think that having an interactive plot is necessary, and having two similar-but-different parameters `plot=True/False` and `output.plot=/path/to/background.png` is unnecessarily confusing.

ndevenish:
Okay. I've added `1554.removal` to note this.

command_line/background.py
```diff
@@ -67,90 +69,95 @@ def run(args=None):
 
     # Ensure we have either a data block or an experiment list
     experiments = flatten_experiments(params.input.experiments)
-    if len(experiments) != 1:
+    if 0 and len(experiments) != 1:
```
ndevenish, line 72:
```suggestion
    if not experiments:
```

rjgildea: Thanks - I'd removed this code block locally but missed it off my commit

PR #1556, opened 2021-01-18
Logic error: should be not a and not b before stopping

graeme-winter:
This fails when it should succeed as the test is logically incorrect


rjgildea (commented):
This looks like a useful and important bug fix - so it might be helpful to add a test case. It should probably be fairly straightforward to write - just take an existing test and strip out each of `""intensity.sum.value"` and `""intensity.prf.value"` in turn and verify that you can still export the reflections. This could even be done in memory directly calling the `export_mtz` so you don't need to write back out the modified reflection tables.

graeme-winter:
> This looks like a useful and important bug fix - so it might be helpful to add a test case. It should probably be fairly straightforward to write - just take an existing test and strip out each of `""intensity.sum.value"` and `""intensity.prf.value"` in turn and verify that you can still export the reflections. This could even be done in memory directly calling the `export_mtz` so you don't need to write back out the modified reflection tables.

Minded to agree, good point

test/command_line/test_export.py
```diff
@@ -405,3 +405,20 @@ def test_json_shortened(dials_data, tmpdir):
...
+        removed = tmpdir / f"removed_{remove}.refl"
+        data = flex.reflection_table.from_file(refl)
+        del data[f"intensity.{remove}.value"]
+        del data[f"intensity.{remove}.variance"]
+        data.as_file(removed)
+
+        procrunner.run(
+            ["dials.export", expt, removed],
+            working_directory=tmpdir.strpath,
+        )
```
rjgildea, line 424: Would suggest verifying that `dials.export` actually output an mtz file as expected, rather than simply exiting silently (not tested this change):
```suggestion


def test_export_sum_or_profile_only(dials_data, tmpdir):
    expt = dials_data("insulin_processed") / "integrated.expt"
    refl = dials_data("insulin_processed") / "integrated.refl"

    for remove in "prf", "sum":
        removed = tmpdir / f"removed_{remove}.refl"
        data = flex.reflection_table.from_file(refl)
        del data[f"intensity.{remove}.value"]
        del data[f"intensity.{remove}.variance"]
        data.as_file(removed)

        procrunner.run(
            ["dials.export", expt, removed, f"mtz.hklout=removed_{remove}.mtz"],
            working_directory=tmpdir.strpath,
        )
        assert (tmpdir / "removed_{remove}.mtz").check(file=1)
```
@Anthchirp would probably suggest using the `tmp_path` (returns a `pathlib.Path` object) instead of `tmpdir` (returns a `py.path.local` object) for new tests :)

rjgildea (approved):
Thanks for adding a test 👍

test/command_line/test_export.py
Anthchirp: correct, this is because `py.path` is deprecated and `pathlib` is supported out of the box, and therefore it's not really worth learning or propagating the `py.path`-specific features such as `.check(file=1)` - with pathlib it's `.is_file()`

ndevenish:
Will rewrite to pathlib in a mo.

This extra mtz check appears to be good, as this test actually fails now:
```
DIALS 3.dev.341-ge208ee8be

mtz {
  hklout = "removed_prf.mtz"
}
input {
  experiments = /Users/nickd/dials/data/dials_data/insulin_processed/integrated.expt
  reflections = /private/var/folders/54/rx9zwmy96sqbczfvw51plf900000gn/T/pytest-of-nickd/pytest-7/test_export_sum_or_profile_onl0/removed_prf.refl
}

Data appears to be unscaled, setting intensity = profile+sum
Cannot export intensity kind 'profile'; missing column(s): intensity.prf.value

...

>           assert (tmpdir / "removed_{remove}.mtz").check(file=1)
E           AssertionError: assert False
```

ndevenish:
Okay, I've updated to `tmp_path` and actually check the `dials.export` return code... so it's still failing but in the right place now.

ndevenish:
Error now:
```
DIALS 3.dev.343-ge11295d77

mtz {
  hklout = "removed_prf.mtz"
}
input {
  experiments = /Users/nickd/dials/data/dials_data/insulin_processed/integrated.expt
  reflections = /private/var/folders/54/rx9zwmy96sqbczfvw51plf900000gn/T/pytest-of-nickd/pytest-15/test_export_sum_or_profile_onl0/removed_prf.refl
}

Cannot export intensity kind 'profile'; missing column(s): intensity.prf.value
Data appears to be unscaled, setting intensity = profile+sum


>           assert not result.returncode and not result.stderr
E           assert (not 1)

test/command_line/test_export.py:424: AssertionError
```

graeme-winter:
Should we merge this now?

Issue #1449, opened 2020-10-09
get_raw_data called twice for image 0 in spotfinding?


ndevenish:
Maybe three times for trusted_range evaluation.... Found when profiling the `find_spots_server`, because for grid scans this is regenerated every time.

Is called in `MaskGenerator.generate` before getting it to run the threshold algorithm:
https://github.com/dials/dials/blob/5005f0390c241e02592f33719cacfa3c34008102/util/masking.py#L209

as part of `_find_spots_in_imageset`:
https://github.com/dials/dials/blob/5005f0390c241e02592f33719cacfa3c34008102/algorithms/spot_finding/finder.py#L822

The only place this seems to be used, is to get the panel size for generating the mask size - I think this is encoded in the panel information, no?
https://github.com/dials/dials/blob/5005f0390c241e02592f33719cacfa3c34008102/util/masking.py#L246

PR #1569, opened 2021-01-27
Mask generation: remove unnecessary get_raw_data()

rjgildea:
The mask generation code doesn't actually use the image data, but just uses the image to get the panel sizes - which are available in the panel object! This extra call to `imageset.get_raw_data()` for the first image of an imageset causes significant overhead e.g. to the per-image analysis service used at DLS.

Convert single-method class to a function.

### DIFF
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import pickle
 import time
@@ -8,10 +9,10 @@ from dxtbx.imageset import ImageSequence
 from iotbx.phil import parse
 
 import dials.extensions
+import dials.util.masking
 from dials.algorithms.background.simple import Linear2dModeller
 from dials.algorithms.spot_finding.finder import SpotFinder
 from dials.array_family import flex
-from dials.util.masking import MaskGenerator
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


ndevenish:
Fixes https://github.com/dials/dials/issues/1449

graeme-winter (approved):
Clever solution to the problem, looks sane, but pls can has news fragement aimed at developers of other packages - masking is the kind of area I would imagine being used elsewhere & API change could affect them.

ndevenish:
That's a good point - we should probably put in a `MaskGenerator` shim that prints a deprecationwarning, decide on a release and put in a `.removal`

graeme-winter:
> That's a good point - we should probably put in a `MaskGenerator` shim that prints a deprecationwarning, decide on a release and put in a `.removal`

Probably a good idea

In fairness though, we can probaby be more assertive e.g. `MaskGenerator` throws a `RuntimeError` or something to _really let people know_ that it is gone. Keep for 1 release cycle.

ndevenish:
That'll have the same effect as just removing - breaking the thing they are trying to do and forcing them to fix it right then and there. I don't think there is any harm in leaving in a deprecated interface for a few months.

ndevenish:
Suggest this commit:
- `DeprecationWarning` until 3.4 (~2021-03-15)
- Then `UserWarning` until 3.5 (~2021-05-24)

ndevenish (approved):

PR #1575, opened 2021-02-01
Windows build

dagewa:
I'm working through Windows build errors on this branch (see https://hackmd.io/510HFFicQ7-6yYSUXqHq-A). Still have more to fix, but I wanted to make this visible early (particularly to @Anthchirp), in case there are better approaches. First issue, need this otherwise `import ssl` fails down the line.


ndevenish:
This SSL issue was with the windows store version of python, right?

dagewa:
No, it's in the subprocess calls to the executable defined as `conda_python`

dagewa:
It turns out that adding conda's DLL directory to PATH during the build was enough to get through with the set up [described](https://hackmd.io/510HFFicQ7-6yYSUXqHq-A). There are still many test failures, but I'd rather handle them separately, so marking this as ready now.

Anthchirp (approved):
@dagewa could you have a look if this works for you?

dagewa:
@Anthchirp thanks, can confirm the build works for me.

PR #1582, opened 2021-02-08
re-enable Azure cache

Anthchirp:
We had those disabled for nearly 3 months in
1f1dbef
due to an Azure cache integrity bug.

It _looks_ like this is no longer an issue.


Anthchirp:
Still broken

PR #1584, opened 2021-02-09
Indeterminate normal equations

dagewa:
When the normal equations have zero elements on the right hand side, instead of
```
There is at least one normal equation with a right hand side of zero,
meaning that the parameters are not all independent, and there is no unique
solution.  Mathematically, some kind of row reduction needs to be performed
before this can be solved.
```
report
```
The normal equations have an indeterminate solution. The problematic parameters are Detector1Dist, Detector1Shift1, Detector1Shift2, Detector1Tau1, Detector1Tau2, Detector1Tau3.
```


dagewa:
the actual change here is trivial -  checking that the newsfragment looks okay

ndevenish (approved):

Issue #1587, opened 2021-02-10
Spotfinding overhead computing resolution map


rjgildea:
During spotfinding with a `d_min` or `d_max` filter set, there is an expensive overhead computing a map of the resolution at each pixel:

https://github.com/dials/dials/blob/9711b078f8837f9b20064529caf0f6fa0f9ef10f/util/masking.py#L184-L187

For an EIGER 16M constructing the resolution map can take an additional ~2.8 seconds compared to ~1.6 seconds total for spotfinding if no resolution limits have been specified.

During standard spotfinding (i.e. on a full imageset) this is only done once, so the overhead is not significant compared to the time taken to spotfind an entire dataset. The resolution map is also cached for given beam and panel models to alleviate this to some extent in the case of per-image analysis (see #1177).

However, in the context of per-image analysis at Diamond, this still results in a significant overhead. For example, we may perform PIA on 250 image files selected from a rotation scan, and these are spread randomly across 80 PIA service instances, with each instance seeing ~3-4 images. This means that the additional overhead will affect the processing time of around 1/4 to 1/3 of all images.

phyy-nx:
FWIW, in SFX data, where the wavelength changes for each image, the d_min/d_max masking changes slightly for each image.  Because of the mask generation overhead that you mention here, I usually use dials.generate_mask on a single image with resolution limits a bit beyond what I want, then use that mask for spotfinding.  It skips the per-pixel calculation of `s1` and runs much faster.  I then limit the spots used in indexing using indexing d_min/d_max limits.

ndevenish:
I think this issue exists outside of DIALS, in our live analysis services, and we can address this there.

PR #1590, opened 2021-02-10
Faster generation of pixel-resolution map

rjgildea:
Faster construction of the pixel-resolution map by calculating `s1` vector explicitly (via the *d* matrix) rather than calling `panel.get_resolution_at_pixel()` (which invokes a call to `std::acos`).

Additional savings (these are more significant) by disabling parallax correction for ResolutionMaskGenerator. The accuracy of the parallax correction is probably not necesary for constructing a resolution mask, so set the px_mm_strategy to `SimplePxMmStrategy` before calculating, leading to a significant time saving in calculating the pixel-resolution map.

In total, the time taken to construct the pixel-resolution map is reduced from ~2.8 seconds to ~0.9 seconds for an EIGER 16M image.

See #1587


util/masking.h
```diff
@@ -38,15 +38,12 @@ namespace dials { namespace util {
         : resolution_(
           af::c_grid<2>(panel.get_image_size()[1], panel.get_image_size()[0])) {
       vec3<double> s0 = beam.get_s0();
+      double wavelength_r = 1.0 / beam.get_wavelength();
```
graeme-winter, line 41: 1 / wavelength is wavenumber?

util/masking.h
```diff
@@ -38,15 +38,12 @@ namespace dials { namespace util {
...
       for (std::size_t j = 0; j < resolution_.accessor()[0]; ++j) {
         for (std::size_t i = 0; i < resolution_.accessor()[1]; ++i) {
           vec2<double> px(i + 0.5, j + 0.5);
-          try {
-            resolution_(j, i) = panel.get_resolution_at_pixel(s0, px);
-          } catch (dxtbx::error) {
-            // Known failure: resolution at beam center is undefined
-            resolution_(j, i) = 0.0;
-          }
+          vec3<double> s1 = panel.get_pixel_lab_coord(px).normalize() * wavelength_r;
```
graeme-winter, line 45: Wonder if any benefit in manually computing lab frame coordinate rather than calling `get_pixel_lab_coord` - would mean compiler could see what you are doing, may unroll loops in clever way etc.

util/masking.py
```diff
@@ -184,7 +188,13 @@ def generate_ice_ring_resolution_ranges(beam, panel, params):
 @lru_equality_cache(maxsize=3)
 def _get_resolution_masker(beam, panel):
     logger.debug("resolution masker cache miss")
-    return ResolutionMaskGenerator(beam, panel)
+    t0 = time.perf_counter()
```
graeme-winter, line 191: Do we want to leave the timing in?

newsfragments/1590.misc
```diff
@@ -0,0 +1 @@
+Faster pixel-resolution map calculation for generation of resolution maps
```
graeme-winter, line 1: ... by ignoring the parallax correction in the sensor?

util/masking.py
```diff
@@ -184,7 +188,13 @@ def generate_ice_ring_resolution_ranges(beam, panel, params):
...
+    panel = copy.deepcopy(panel)
+    panel.set_px_mm_strategy(SimplePxMmStrategy())
```
graeme-winter, line 193: Should we leave the option of computing with the native strategy? e.g. for cases where you are making a mask manually so don't mind a 2s delay (for metal rings could make a difference to output too)

graeme-winter (commented):
Flagged a couple of questions, would like some thoughts on this e.g. for making the option to keep the "slow" method

util/masking.py
rjgildea: I've modified the implementation so now this can be controlled by setting the new masking parameter `parallax_correction=False` (the default is `parallax_correction=None`, i.e. use whatever `PxMmStrategy` is set on the panel. This depends on https://github.com/cctbx/dxtbx/pull/319.

util/masking.py
rjgildea: I think it is possibly useful information to include (and is only debug log level), however I have removed the "resolution masker cache miss" logging statement above, so the net number of log statements is the same as before.

util/masking.h
rjgildea: Possibly, however I think this change is a demonstrable enough improvement that it is worth including as is. We have discussed elsewhere other possible ways for computing the resolution map which may be better still.

rjgildea:
This PR now depends on https://github.com/cctbx/dxtbx/pull/319.

newsfragments/1590.feature
```diff
@@ -0,0 +1 @@
+New masking parameter ``parallax_correction=``. Set to ``False`` to speed up generation of resolution masks by disabling parallax correction.
```
graeme-winter, line 1: Not sure, but I think a little more context here would be useful so that the general user does not think this will usefully speed up 1,800 image find spots job?

util/masking.h
```diff
@@ -38,15 +38,12 @@ namespace dials { namespace util {
...
         for (std::size_t i = 0; i < resolution_.accessor()[1]; ++i) {
           vec2<double> px(i + 0.5, j + 0.5);
-          try {
-            resolution_(j, i) = panel.get_resolution_at_pixel(s0, px);
-          } catch (dxtbx::error) {
-            // Known failure: resolution at beam center is undefined
-            resolution_(j, i) = 0.0;
-          }
+          vec3<double> s1 = panel.get_pixel_lab_coord(px).normalize() * wavenumber;
+          resolution_(j, i) = 1/(s1 - s0).length();
```
graeme-winter, line 46: Low risk, but should we cope gracefully here with `s1 == s0` - unlikely in experimental conditions, though maybe not so unlikely in the 1st cut if someone records a beam centre in integral pixels? Or half integral.

Would be easy to add a test case above for this.

util/masking.py
```diff
@@ -35,6 +37,12 @@
             "accepted by the filtering algorithm."
     .type = float(value_min=0)
 
+  parallax_correction = None
```
graeme-winter, line 40: Why `None` not `True`?

graeme-winter (approved):
Good change set, appreciate having the boolean option for this and the pain it caused to implement that way 🙂

I flag a couple of questions, but I think the `s1 == s0` one is the only one which really needs attention - the others are pretty cosmetic.

util/masking.py
rjgildea: I went with this name for consistency with https://github.com/dials/dxtbx/blob/2caec59df9e63219892e152d47dbdef2afc845f4/model/detector.py#L108-L112. I didn't want to set the default to `True` in case that implied that it would enable `parallax_correction` even if the detector model didn't already specify it. Only setting it to `False` will have any effect. Perhaps `disable_parallax_correction=False` would be clearer in this context? Welcome suggestions here...

graeme-winter: Without wishing to bikeshed, I agree that `disable_parallax_correction=False` is more honest (had occurred to me as well)

PR #1594, opened 2021-02-12
Fixes for importing still images with an image_range

### LINKED ISSUE #1592: dials.import: import on subset of grid scan fails
```

cs04r-sc-com99-25 vmxi-thau-grid-example-76xx :( $ dials.import mask=pixels.mask image_range=7600,7700 $(cat ../vmxi-thau-grid-example/imported_data.txt)
DIALS 3.3.2-g209f02519-release

input {
  experiments = <image files>
}
geometry {
  scan {
    image_range = 7600 7700
  }
}
lookup {
  mask = "pixels.mask"
}


Applying input geometry in the following order:
  1. Manual geometry

  File "/dls_sw/apps/dials/dials-v3-3-2/build/../modules/dials/command_line/dials_import.py", line 933, in <module>
    run()
  File "/dls_sw/apps/dials/dials-v3-3-2/conda_base/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/dls_sw/apps/dials/dials-v3-3-2/build/../modules/dials/command_line/dials_import.py", line 929, in run
    script.run(args)
  File "/dls_sw/apps/dials/dials-v3-3-2/build/../modules/dials/command_line/dials_import.py", line 790, in run
    experiments = metadata_updater(imageset_importer())
  File "/dls_sw/apps/dials/dials-v3-3-2/build/../modules/dials/command_line/dials_import.py", line 555, in __call__
    subset = imageset[j : j + 1]
  File "/dls_sw/apps/dials/dials-v3-3-2/modules/dxtbx/imageset.py", line 279, in __getitem__
    return self.partial_set(start - offset, stop - offset)
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: dxtbx Internal Error: /scratch/jenkins_agent/workspace/dials_release_build_linux/build_dials/modules/dxtbx/imageset.h(932): DXTBX_ASSERT(index < indices_.size()) failure.
```

Feels like this should work...

graeme-winter:
Fixes for importing still images with an image_range i.e.

```
dials.import path/to/grid.h5 image_range=100,200
```

Resolves #1592, includes test to prevent regression.


dagewa:
I'd like to have a look to see what happens with scans, re `subset = imageset[j - offset : j - offset + 1]`. Should be something similar, right?

test/command_line/test_import.py
```diff
@@ -474,6 +471,36 @@ def test_import_still_sequence_as_experiments_subset(dials_data, tmpdir):
...
+    )
+
+    imported_exp = load.experiment_list(tmpdir.join(out).strpath)
+    assert len(imported_exp) == 3
+    for exp in imported_exp:
+        assert exp.identifier != ""
+
+    iset = set(exp.imageset for exp in imported_exp)
+    assert len(iset) == 1
+    assert len(imported_exp[0].imageset) == 3
```
rjgildea, line 497: My one concern with this test is that it only tests that we only import 3 images, not which 3 images are imported. I.e. we don't explicitly check that we have imported images 3-5 rather than 1-3 (just thinking of ways in which the code could break, yet the tests still pass).

Perhaps checking the summed counts for each image in the imported imageset would suffice (assuming that each image has different total counts)?

command_line/dials_import.py
```diff
@@ -571,9 +571,19 @@ def __call__(self, imageset_list):
                 if imageset.get_scan().is_still():
                     # make lots of experiments all pointing at one
                     # image set
-                    start, end = imageset.get_scan().get_array_range()
+
+                    # check if user has overridden the input - if yes, recall
+                    # that these are in people numbers (1...) and are inclusive
+                    if self.params.geometry.scan.image_range is not None:
+                        ustart, uend = self.params.geometry.scan.image_range
```
rjgildea, line 578:
```suggestion
                        user_start, user_end = self.params.geometry.scan.image_range
```
I assume "u" is short for "user" here? Might be worth spelling this out long hand to save future me the trouble of trying to figure out "what on earth does 'u' stand for?".

rjgildea (approved):

dagewa:
> I'd like to have a look to see what happens with scans, re `subset = imageset[j - offset : j - offset + 1]`. Should be something similar, right?

Ok, I chased this through `ExperimentListTemplateImporter`  --> `DataBlockTemplateImporter` --> `ImageSetFactory.make_sequence` and I'm lost and don't really know why it didn't just work already. The first first and last images should get set in the `Scan` object, and we should be able to use this to get the image range even if this is a still series

graeme-winter:
> > I'd like to have a look to see what happens with scans, re `subset = imageset[j - offset : j - offset + 1]`. Should be something similar, right?
> 
> Ok, I chased this through `ExperimentListTemplateImporter` --> `DataBlockTemplateImporter` --> `ImageSetFactory.make_sequence` and I'm lost and don't really know why it didn't just work already. The first first and last images are set in the `Scan` object, and we use that to access the images we want, even if it is a stills series. I think I agree with the assessment that this adds technical debt, because there is something more fundamental broken.

It's different, as a still sequence is N experiments with N scans, whereas rotation data are 1 experiment with 1 scan

command_line/dials_import.py
```diff
@@ -571,9 +571,19 @@ def __call__(self, imageset_list):
...
                     # make lots of experiments all pointing at one
                     # image set
-                    start, end = imageset.get_scan().get_array_range()
+
+                    # check if user has overridden the input - if yes, recall
+                    # that these are in people numbers (1...) and are inclusive
+                    if self.params.geometry.scan.image_range is not None:
+                        user_start, user_end = self.params.geometry.scan.image_range
+                        offset = imageset.get_scan().get_array_range()[0]
+                        start, end = ustart - 1, uend
```
rjgildea, line 580:
```suggestion
                        start, end = user_start - 1, user_end
```
Sorry, the above suggested change was incomplete (I should have modified it fully such that applying the suggested change would work).

test/command_line/test_import.py
```diff
@@ -474,6 +471,36 @@ def test_import_still_sequence_as_experiments_subset(dials_data, tmpdir):
     assert all(exp.goniometer is not None for exp in imported_exp)
 
 
+def test_import_still_sequence_as_expts_subset_by_range(dials_data, tmpdir):
+    image_files = dials_data("centroid_test_data").listdir("centroid*.cbf", sort=True)
+
+    out = "experiments_as_still.expt"
+
+    _ = procrunner.run(
```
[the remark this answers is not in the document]
graeme-winter: Done

test/command_line/test_import.py
```diff
@@ -474,6 +471,36 @@ def test_import_still_sequence_as_experiments_subset(dials_data, tmpdir):
...
+        [
+            "dials.import",
+            "scan.oscillation=10,0",
+            "image_range=3,5",
+            "output.experiments=%s" % out,
+        ]
+        + [f.strpath for f in image_files],
+        working_directory=tmpdir.strpath,
```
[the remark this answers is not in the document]
graeme-winter: 🤷‍♂️

test/command_line/test_import.py
graeme-winter: This point is a good point

command_line/dials_import.py
```diff
@@ -571,9 +571,19 @@ def __call__(self, imageset_list):
                 if imageset.get_scan().is_still():
                     # make lots of experiments all pointing at one
                     # image set
-                    start, end = imageset.get_scan().get_array_range()
+
+                    # check if user has overridden the input - if yes, recall
+                    # that these are in people numbers (1...) and are inclusive
+                    if self.params.geometry.scan.image_range is not None:
```
ndevenish, line 577:
```suggestion
                    if self.params.geometry.scan.image_range:
```
Are there any non-None valid falseish settings for this parameter?

test/command_line/test_import.py
```diff
@@ -474,6 +471,38 @@ def test_import_still_sequence_as_experiments_subset(dials_data, tmpdir):
...
+            "image_range=3,5",
+            "output.experiments=%s" % out,
+        ]
+        + [f.strpath for f in image_files],
+        working_directory=tmp_path,
+    )
+
+    assert result.returncode == 0
+
+    imported_exp = load.experiment_list(str(out))
```
ndevenish, line 492:
```suggestion
    imported_exp = load.experiment_list(out)
```
`load.experiment_list` [explicitly](https://github.com/cctbx/dxtbx/blob/daf2542a15134ddb54c03a4af43a80335a5d735a/serialize/load.py#L119-L122) supports the `pathlib` protocols

test/command_line/test_import.py
```diff
@@ -474,6 +471,38 @@ def test_import_still_sequence_as_experiments_subset(dials_data, tmpdir):
...
+    out = tmp_path / "experiments_as_still.expt"
+
+    result = procrunner.run(
+        [
+            "dials.import",
+            "scan.oscillation=10,0",
+            "image_range=3,5",
+            "output.experiments=%s" % out,
+        ]
+        + [f.strpath for f in image_files],
```
ndevenish, line 486:
```suggestion
            f"output.experiments={out}",
            *image_files,
        ],
```
I find it really hard to notice that the comprehension is part of the build. I checked, and procrunner/subprocess appears to deal fine with both `py.path` and `pathlib` objects.

ndevenish (approved):
I'm not entirely sure that I can follow the start/end/offset shifting. Would feel happier with Richard's suggestion of confirming we got the right range and that we haven't e.g. just introduced _another_ off-by-one error somewhere

test/command_line/test_import.py
graeme-winter:
```
assert list(iset)[0].get_image_identifier(0) == image_files[2].strpath
```

seemed simpler

test/command_line/test_import.py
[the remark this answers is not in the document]
graeme-winter: @ndevenish added changes

graeme-winter:
> I'm not entirely sure that I can follow the start/end/offset shifting. Would feel happier with Richard's suggestion of confirming we got the right range and that we haven't e.g. just introduced _another_ off-by-one error somewhere

Added an assertion that the first image in the image set is the image we _think_ it should be.

It is also (FWIW) the same image as importing the numbered images as elsewhere in the tests.

ndevenish:
I will merge this when the tests pass 👍

PR #1598, opened 2021-02-16
dials.image_viewer basis vector scale

dagewa:
The basis vectors were previously scaled by a hard-coded 10 before display. This is not always appropriate, particularly for small molecules. If an axis was scaled too long it would not be displayed at all:
Screenshot from 2021-02-16 16-12-01

In such cases, the PR allows to set e.g. `basis_vector_scale=5` when starting `dials.image_viewer`
Screenshot from 2021-02-16 16-12-32


util/image_viewer/spotfinder_frame.py
```diff
@@ -1652,7 +1652,14 @@ def map_coords(x, y, p):
...
                     beam_x, beam_y = map_coords(beam_x, beam_y, panel)
-                    for i, h in enumerate(((10, 0, 0), (0, 10, 0), (0, 0, 10))):
+                    self.params.basis_vector_scale
+                    for i, h in enumerate(
+                        (
+                            (self.params.basis_vector_scale, 0, 0),
+                            (0, self.params.basis_vector_scale, 0),
+                            (0, 0, self.params.basis_vector_scale),
+                        )
+                    ):
```
rjgildea, line 1662:
```suggestion
                    for i, h in enumerate(((1, 0, 0), (0, 1, 0), (0, 0, 1))):
                        r = A * matrix.col(h) * self.params.basis_vector_scale
```
would probably achieve the same in fewer lines of code (and probably be more readable).

PR #1600, opened 2021-02-19
Make symlink optional to fix tests on Windows

dagewa:
This allows all tests in `test_import.py` to pass on Windows even without Administrator access, or Developer Mode.


Anthchirp (approved):

test/command_line/test_import.py
```diff
@@ -370,13 +371,22 @@ def test_extrapolate_scan(dials_data, tmpdir):
...
     for image in images:
-        (tmp_path / image.basename).symlink_to(image)
-    return tmp_path.joinpath("centroid_####.cbf")
+        try:
+            (tmp_path / image.basename).symlink_to(image)
+        except OSError:
+            shutil.copy(image, tmp_path)
+    yield tmp_path.joinpath("centroid_####.cbf")
+    for image in images:
+        (tmp_path / image.basename).unlink()
```
dagewa, line 389: This gave
```
E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\fcx32934\\AppData\\Local\\Temp\\pytest-of-fcx32934\\pytest-67\\popen-gw9\\test_template_with_missing_ima1\\centroid_0005.cbf'
```
in my Windows test run.

dagewa: Looks like an intermittent error. Something like Windows file locking at play, perhaps

Anthchirp: fix incoming

Anthchirp: I suspect this is us not correctly dealing with file handles (ie. not using context handlers _everywhere_), leading to a race against the garbage collection.

PR #1608, opened 2021-02-23
Windows fixes

dagewa:
Some small changes that reduce the number of test failures on Windows.


algorithms/symmetry/cosym/target.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         cb_op_to_primitive = data.change_of_basis_op_to_primitive_setting()
         data = data.change_basis(cb_op_to_primitive).map_to_asu()
 
-        order = lattice_ids.argsort()
+        order = lattice_ids.argsort().astype(np.uint64) # Convert to uint64 avoids crashes on Windows
```
rjgildea, line 71: Could you expand this comment here? The commit message comment was clearer, but I had to read the commit message to better understand the code change.

dagewa, line 71:
```suggestion
        # Convert to uint64 avoids crashes on Windows when later constructing flex.size_t
        # (related: https://github.com/cctbx/cctbx_project/issues/591)
        order = lattice_ids.argsort().astype(np.uint64)
```

PR #1612, opened 2021-02-25
deepcopy -> copy, fixes #1611

### LINKED ISSUE #1611: dials.search_beam_position very slow for fine sliced data (0.01°)
```
/dls/i03/data/2021/cm28170-1/processing/xia2-blah-29-4/DEFAULT/SAD/SWEEP1/index/2_SWEEP1_strong.*
```

as input - which does not have a _huge_ number of reflections - has so far taken 20 minutes to run -> suspect there is something in the process which depends on the oscillation width

graeme-winter:
Do not need a deep copy of the experiment, and deep copies of experiments can be _very_ expensive. Shallow copy and reassignment of detector is plenty.


graeme-winter:
@rjgildea is the only thing which changes in this routine the detector origin? Or should we copy the beam as well?

graeme-winter:
> @rjgildea is the only thing which changes in this routine the detector origin? Or should we copy the beam as well?

Looks like beam is unaffected -> 

```
Silver-Surfer dials-1611 :) $ dials.show 2_SWEEP1_strong.expt | grep -A6 'Beam:'Beam:
    wavelength: 0.976246
    sample to source direction : {0,0,1}
    divergence: 0
    sigma divergence: 0
    polarization normal: {0,1,0}
    polarization fraction: 0.999
Silver-Surfer dials-1611 :) $ dials.show optimised.expt | grep -A6 'Beam:'
Beam:
    wavelength: 0.976246
    sample to source direction : {0,0,1}
    divergence: 0
    sigma divergence: 0
    polarization normal: {0,1,0}
    polarization fraction: 0.999
```

So the change set in here is probably fine, though would welcome critique

graeme-winter:
As a datum - copying _that_ experiment list via `deepcopy` was around 1.6s a go 

```
cs03r-sc-serv-36 slow-copy :) $ dials.python xerox.py 
Copying experiment list 10 times took 16.0s
cs03r-sc-serv-36 slow-copy :) $ cat xerox.py 
import copy
import time
experiments = ExperimentList.from_file("72000_fine.expt", check_format=False)
t0 = time.time()
for j in range(10):
    duplicate = copy.deepcopy(experiments[0])
print(f"Copying experiment list 10 times took {(time.time() - t0):.1f}s")
```

shallow copy is virtually free...

PR #1614, opened 2021-02-26
Label rlp

dagewa:
To aid orientation and reflection matching between the reciprocal lattice and image viewers, this change introduces optional labelling for the reciprocal lattice point nearest the view centre.
Screenshot from 2021-02-26 16-55-22


toastisme (approved):
Tested using SACLA-MPCCD-Phase3-21528-5images.h5 after running dials.import and dials.find_spots. Behaviour looks correct and visualises already existing code (nearest neighbours), so more tests aren't needed. Definitely a helpful change. Looks like build errors are just from the master/main change.

Anthchirp:
Yes, merging `main` into here should fix that.

PR #1617, opened 2021-03-04
Reindex output not as pickle

### LINKED ISSUE #1616: dials.image_viewer does not work on reindexed experiments
```
Grey-Area timepix :( $ dials.reindex indexed.* change_of_basis_op=h,k,l

change_of_basis_op = "h,k,l"
input {
  experiments = indexed.expt
  reflections = indexed.refl
}

Saving reindexed experimental models to reindexed.expt
Saving reindexed reflections to reindexed.refl
Grey-Area timepix :) $ dials.image_viewer reindexed.*

input {
  experiments = <image files>
  experiments = reindexed.expt
}

Error: experiment has no detector
```

graeme-winter:
Use reflections.as_file - fixes #1616

Previously e.g. using image viewer or show would fail on reindexed data.


newsfragments/1617.bugfix
```diff
@@ -0,0 +1 @@
+dials.reindex: output data as messagepack not pickle, so the image viewer etc. works
```
rjgildea, line 1:
```suggestion
dials.reindex: output reflections using MessagePack, rather than pickle, format for compatibility with downstream programs
```

rjgildea (approved):
This was definitely an oversight, so good to catch it now. I wonder if we have other similar instances lurking in the source code...

newsfragments/1617.bugfix
graeme-winter: Good use of suggestion :)

PR #1621, opened 2021-03-04
cosym fix np.random.seed

rjgildea:
Fix occassional test failure by also setting np.random.seed


dagewa (approved):
Tests pass for me on Windows :+1:

algorithms/symmetry/cosym/_generate_test_data.py
```diff
@@ -71,12 +75,10 @@ def generate_test_data(
     twin_fractions=None,
     map_to_minimum=True,
 ):
-
-    import random
-
     if seed is not None:
         flex.set_random_seed(seed)
         random.seed(seed)
+        np.random.seed(seed)
```
graeme-winter, line 81: 👍

graeme-winter (approved):
Just what was needed - this is (again) one of those tests which seems RNG dependent - thanks for the fix. To state obvious, it worked for me (but then it would, right?)

rjgildea:
> this is (again) one of those tests which seems RNG dependent

Yes - by definition the cosym procedure is dependent on a random number generator as it is used to generate random starting coordinates for the procedure:

https://github.com/dials/dials/blob/c9cdceba7cc98754877e1a7b33820bf5db5c572b/algorithms/symmetry/cosym/__init__.py#L295

In the command line script all the potentially relevant random seeds (random, flex.random, np.random) are set. Setting `np.random.seed` also in the tests was an oversight in the recent flex -> numpy conversion.

PR #1623, opened 2021-03-05
Avoid infinite loop in intensity combination if average < 0.

### LINKED ISSUE #1622: dials.scale: determine_Imids infinite loop
Under some circumstances it appears that the `_determine_Imids()` method in `dials.scale` can enter an infinite loop, eventually consuming all available memory before being killed:

https://github.com/dials/dials/blob/a8710a8eb1345eefcf7715b825f5feface612c90/algorithms/scaling/combine_intensities.py#L142-L150

This has been observed in autoprocessing at DLS.

jbeilstenedmands:
Fixes #1622

Tested on the dataset which revealed this issue, now completes fine.


algorithms/scaling/combine_intensities.py
```diff
@@ -144,7 +144,7 @@ def _determine_Imids(self, raw_intensities):
             avg = flex.mean(raw_intensities)
             Imid = flex.max(raw_intensities) / 10.0
             Imid_list = [0, 1, avg, Imid]
-            while Imid > avg:
+            while (Imid > avg) and (Imid > 10):
```
rjgildea, line 147: What is the significance of the value `10` here? A comment explaining the reasoning might be useful.

newsfragments/1623.bugfix
```diff
@@ -0,0 +1 @@
+``dials.scale``: Fix potential for memory crash in intensity combination in rare cases
```
rjgildea, line 1: "... by avoiding infinite loop"

rjgildea (approved):
Confirmed this fixes the offending dials.scale job. I'd be interested in knowing the significance of the value `10`, and a comment in the source code would probably valuable.

newsfragments/1623.bugfix
ndevenish: I've got the newsfragment

algorithms/scaling/combine_intensities.py
jbeilstenedmands: I'll add a comment to the text and update the newsfragment once the CI job has finished.
Basically, what we're doing here is testing a value midpoint value across the span of the intensities (midpoint meaning for crossing from summation to profile intensities). So if the max intensity is 50K, for the midpoints we try mean, 5K, 500, 50 and 5. So the value of 10 means the last value to test is in the 1>10 range, which seems a sensible place to stop (but we only get down to this point if the mean is below 10).

ndevenish: I wonder if `1` would be a bit less magical-looking and still as useful?

ndevenish: Also maybe `while I > max(avg, 10):` would be clearer - setting an explicit limit? It's kinda got quite drastically different output behaviour for `avg=0+eps` and `avg=0-eps`

ndevenish:
Pulling this in now as-is as minimal change for release. I encourage (nay urge!) cleanup/deduplication of the calls.

PR #1624, opened 2021-03-05
Refactor Scaling Imid-determining code in combine_intensities

ndevenish:
Originally caused #1622, I merged the "working" code to get it in as discussion was taking time.

The changes here are:
- Deduplicate repeat declarations
- Set a limit on average calculation to avoid cases where avg<0

This doesn't change any behaviour, but will result in extremely long Imid-series generated when the limit is used. A larger value might be more sensible here. Discuss.


jbeilstenedmands (changes requested):
Thank you for refactoring out the duplicated code, this is an improvement.
However, removing the `Imid > 10` condition just reintroduces the bug. In terms of the 'magic number', let's not lose sight of what the numbers actually mean in this case. We're dealing with raw intensity values i.e. pixel counts, so it becomes meaningless to talk about midpoint values much below 1. Having the threshold of 10 means the final value tested is in the 1-10 range, which seems sensible.

ndevenish:
The `Imin > 10` part is absorbed into the `max(0, flex.mean(raw_intensities))` - the average is now guaranteed to be above/equal to zero so you don't get the original bug - an infinite loop that consumed all memory.

However, right, as it's pixel counts, presumably 10 pixels is a threshold underneath which the values aren't helpful, so a somewhat meaningful ballpark

jbeilstenedmands (approved):

PR #1580, opened 2021-02-04
Reject Wilson outliers in resolution analysis

rjgildea:
Presence of spurious observations with much larger than expected intensities, e.g. from ice rings at high resolution, may adversely affect the calculations used in resolution analysis. Therefore, attempt to identify such outliers by:
- performing quasi-normalisation of intensities to obtain E^2 values
- rejecting intensities with E^2 > emax^2, where default emax=4

See also:
- Read, R. J. (1999). Acta Cryst. D55, 1759-1764
- Evans, P. (2006). Acta Cryst. D62, 72-82
- http://legacy.ccp4.ac.uk/html/aimless.html#outlier

**Example containing ice rings**

`emax=0` (i.e. switch off outlier rejection, the previous behaviour):
```
Resolution cc_half:       2.31
```
cchalf_emax_0

`emax=4`:
```
Removing 520 Wilson outliers with E^2 >= 16.0
Resolution cc_half:       2.42
```
cchalf_emax_4
intensities_emax_4

`emax=8`:
```
Removing 24 Wilson outliers with E^2 >= 64.0
Resolution cc_half:       2.40
```
cchalf_emax_8
intensities_emax_8

`emax=10`:
```
Removing 13 Wilson outliers with E^2 >= 100.0
Resolution cc_half:       2.41
```
cchalf_emax_10
intensities_emax_10

**Example containing no ice rings (x4wide)**
```
$ dials.estimate_resolution $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.{expt,refl}
...
Removing 1 Wilson outliers with E^2 >= 16.0
Resolution cc_half:       1.20
```
intensities_x4wide

PR #1627, opened 2021-03-09
Reject Wilson outliers in scaling

jbeilstenedmands:
This change adds additional outlier rejection based on normalised intensity values during scaling.

This is performed before each round of standard outlier rejection, the normal intensities being recalculated each time due to changes in scale factor or intensity/variance estimates.

Also included is the separating out of common quasi_normalisation code into a separate file.


algorithms/symmetry/__init__.py
```diff
@@ -215,35 +216,7 @@ def quasi_normalisation(intensities):
...
-            d_max=uctbx.d_star_sq_as_d(d_star_sq_max),  # cctbx/cctbx_project#588
-            d_star_sq_step=d_star_sq_step,
-            auto_binning=False,
-        )
-        normalisations = work.intensity_quasi_normalisations()
-        return intensities.customized_copy(
-            data=(intensities.data() / normalisations.data()),
-            sigmas=(intensities.sigmas() / normalisations.data()),
-        )
+        return quasi_normalisation(intensities)
```
rjgildea, line 219: Is there any point in keeping this static method? We might as well replace its use with direct calls to `dials.util.normalisation.quasi_normalisation`.

algorithms/scaling/outlier_rejection.py
```diff
@@ -71,6 +72,27 @@ def reject_outliers(reflection_table, experiment, method="standard", zmax=6.0):
...
+    normalised_intensities = quasi_normalisation(intensities)
+
+    sel = normalised_intensities.data() > (emax ** 2)
+    logger.info(
+        f"{sel.count(True)} outliers identified from normalised intensity analysis (E\xb2 > {(emax ** 2)})"
+    )
+    outlier_indices = Ih_table.Ih_table_blocks[0].Ih_table["loc_indices"].select(sel)
+    datasets = Ih_table.Ih_table_blocks[0].Ih_table["dataset_id"].select(sel)
+    if Ih_table.n_datasets == 1:
+        return [outlier_indices]
```
rjgildea, line 89: Is there a real reason for separately handle the case `n_datasets == 1`? I can't imagine it results in a quantifiable performance benefit, and adds code complexity.

algorithms/scaling/scaler.py
```diff
@@ -790,14 +793,23 @@ def round_of_outlier_rejection(self):
         """Perform a round of outlier rejection, set a new outliers array."""
         assert self.global_Ih_table is not None
         if self.params.scaling_options.outlier_rejection:
+            E2_outlier_indices = determine_Esq_outlier_index_arrays(
```
rjgildea, line 796: Could we be consistent in usage of `E2_*` vs `Esq`? Possibly a slight preference for the former, but no strong opinion.

algorithms/scaling/scaling_options.py
```diff
@@ -194,6 +194,10 @@
       .help = "Cutoff z-score value for identifying outliers based on their
                normalised deviation within the group of equivalent reflections"
       .expert_level = 1
+    emax = 4
```
rjgildea, line 197: Is this default appropriate for scaling? I note that AIMLESS has a default cutoff of EMAX=10 for acentric reflections:
http://legacy.ccp4.ac.uk/html/aimless.html

rjgildea (approved):
I haven't tested this, but assuming they work, the changes look sensible, although I do wonder whether the default `emax=4` is too extreme (c.f. `EMAX=10` for AIMLESS). Have you done much comparison on scaling outcomes for datasets with/without obvious outliers (e.g. ice rings)?

algorithms/scaling/outlier_rejection.py
jbeilstenedmands: I've now moved this line up by one, and in the case of `n_datasets = 1` it now saves doing two selections, one of which is on the whole dataset, so I'd like to keep it this way as it will be called multiple times per job.

algorithms/scaling/scaler.py
jbeilstenedmands: Thanks, I've previously used `Esq` in the scaling code so I've now changed to that term that here for consistency

algorithms/scaling/scaling_options.py
jbeilstenedmands: Updated now to 10 based on precedent set by AIMLESS 👍

rjgildea: I did use a default of `emax=4` (based on trial and error) in the Wilson outlier rejection I added to resolution analysis (https://github.com/dials/dials/pull/1580), but I think it is probably approriate for scaling to have a more conservative cutoff for outlier rejection than resolution analysis. There are more details on the choice in AIMLESS in section 5.3 of [Evans, P. (2006). Acta Cryst. D62, 72-82](https://journals.iucr.org/d/issues/2006/01/00/ba5084/):

> Spurious observations arising from ice or salt spots are often very large and may be rejected if they have an intensity much larger than would be expected (Read, 1999). This test is performed on the normalized amplitudes E, normalized as a function of resolution such that 〈E2〉 = 1. An E of > 4 is very unlikely, but because of the errors in normalization, particularly at low resolution where the mean intensity is changing rapidly with resolution and there are relatively few reflections, or with anisotropic data, it is better to reject only observations with E > 8–10.

rjgildea:
This does also make me wonder if we should add Wilson outlier rejection in symmetry analysis (we do already calculate normalised intensities here) 🤔

algorithms/scaling/scaler.py
```diff
@@ -790,14 +793,23 @@ def round_of_outlier_rejection(self):
...
             outlier_indices = determine_outlier_index_arrays(
                 self.global_Ih_table,
                 self.params.scaling_options.outlier_rejection,
                 self.params.scaling_options.outlier_zmax,
             )[0]
             self.outliers = flex.bool(self.n_suitable_refl, False)
             self.outliers.set_selected(outlier_indices, True)
+            self.outliers.set_selected(Esq_outlier_indices, True)
             if self._free_Ih_table:
+                Esq_outlier_indices = determine_Esq_outlier_index_arrays(
```
rjgildea, line 808: These particular outlier indices don't look to be used?

algorithms/scaling/scaler.py
```diff
@@ -790,14 +793,23 @@ def round_of_outlier_rejection(self):
         """Perform a round of outlier rejection, set a new outliers array."""
         assert self.global_Ih_table is not None
         if self.params.scaling_options.outlier_rejection:
+            Esq_outlier_indices = determine_Esq_outlier_index_arrays(
+                self.global_Ih_table, self.experiment, self.params.scaling_options.emax
+            )[0]
```
rjgildea, line 798: Moved below to be conditional on `self.params.scaling_options.emax` being non-zero:
```suggestion
            Esq_outlier_indices = determine_Esq_outlier_index_arrays(
                self.global_Ih_table, self.experiment, self.params.scaling_options.emax
            )[0]
```

algorithms/scaling/outlier_rejection.py
```diff
@@ -71,6 +72,29 @@ def reject_outliers(reflection_table, experiment, method="standard", zmax=6.0):
     return reflection_table
 
 
+def determine_Esq_outlier_index_arrays(Ih_table, experiment, emax=4.0):
+    # first calculate normalised intensities and set in the Ih_table.
+    if not emax:
+        return [flex.size_t() for _ in range(Ih_table.n_datasets)]
```
rjgildea, line 78: If you make calling this function conditional on non-zero `emax` then you can probably remove this extra check here.

algorithms/scaling/scaler.py
```diff
@@ -790,14 +793,23 @@ def round_of_outlier_rejection(self):
...
             outlier_indices = determine_outlier_index_arrays(
                 self.global_Ih_table,
                 self.params.scaling_options.outlier_rejection,
                 self.params.scaling_options.outlier_zmax,
             )[0]
             self.outliers = flex.bool(self.n_suitable_refl, False)
             self.outliers.set_selected(outlier_indices, True)
+            self.outliers.set_selected(Esq_outlier_indices, True)
```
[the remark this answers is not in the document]
rjgildea: Same comments apply below to other instances of calling `determine_outlier_index_arrays()`.

algorithms/scaling/scaler.py
jbeilstenedmands: Good spot, there was a line of code missing in here.

jbeilstenedmands:
Thanks for the reviews, i've adopted the suggestions. I've tested on some of the standard datasets such as beta-lactamase and x4-wide, and with `emax=10` no E2 outliers are found. I had a look at the dataset that lead to us implementing the Wilson outlier rejection for `dials.estimate_resolution`, in that case 27 outliers are found (dataset size 440K refls), however these were already being picked up by the standard outlier rejection so no change in results here. I imagine this change will be most beneficial for low multiplicity datasets with ice rings, where it was more difficult to identify these kind of outliers from groups with only 2 or 3 reflections. I haven't got any datasets like this to hand, but if anyone knows of some then I'd be happy to test out.

PR #1633, opened 2021-03-16
Stop Azure when unrenamed newsfragments present

Anthchirp:
This prevents running _2n_ builds in a pull request which are then
immediately superseded because the GitHub Action drops in another _2n_
build jobs.


Anthchirp:
And when it is renamed:

ndevenish:
At the moment the newsfragment job doesn't run for drafts. Will this cause no tests to be run for draft releases if the contributor has included a blank newsfragment?

Anthchirp:
Correct. Can fix this by 
* calling it a feature,
* telling Azure to always run checks if the PR is a draft, even if an unrenamed fragment is present (silly),
* moving the Newsfragment job to Azure, always running it but not enforcing a file being present if the PR is a draft, **or**
* always running the Newsfragment job, even if the PR is a draft

ndevenish:
Could also split up the checking and renaming parts of the newsfragment action. Don't really want to overcomplicate things too much.

Anthchirp:
Yes, that's number 3 and also my preferred solution

ndevenish:
read as (moving the Newsfragment job to Azure) and (always running it but not enforcing a file being present if the PR is a draft)

doesn't need to be fixed to put this in, anyhow

ndevenish (approved):

PR #1639, opened 2021-03-29
Cosym target: faster numpy operations

rjgildea:
Significant speedup of computation of cosym target functional, gradients and curvatures.

* Use `np.square(x)` instead of `np.power(x, 2)`
* Avoid explicit loops over number of dimensions

* Functional now ~10x quicker
* Gradients ~20x quicker
* Curvatures ~4x quicker

For a real dials.cosym example, 44 datasets in P3121, this change cut about 25s off the total runtime:
```
real	1m18.899s
user	1m4.166s
sys	0m2.899s

real	0m52.221s
user	0m39.257s
sys	0m2.973s
```


rjgildea:
@dwpaley it would be interesting to hear what performance impact this has on your use case.

algorithms/symmetry/cosym/target.py
```diff
@@ -371,32 +366,13 @@ def compute_gradients(self, x: np.ndarray) -> np.ndarray:
           f: The value of the target function at coordinates `x`.
           grad: The gradients of the target function with respect to the parameters.
         """
-        grad = np.empty(x.shape)
+        x = x.reshape((self.dim, x.size // self.dim))
         if self.wij_matrix is not None:
             wrij_matrix = np.multiply(self.wij_matrix, self.rij_matrix)
+            grad = -2 * (x @ wrij_matrix - x @ np.multiply(self.wij_matrix, x.T @ x))
```
benjaminhwilliams, line 372: Would this work?  Reducing the number of matrix multpilications simplifies the calculation.
```suggestion
            grad = -2 * x @ (wrij_matrix - np.multiply(self.wij_matrix, x.T @ x))
```

For comparison, a dumb example:
```
In [1]: import numpy as np

In [2]: x = np.random.randint(10, size=(2, 5))

In [3]: y = np.random.randint(10, size=(5, 2))

In [4]: z = np.random.randint(10, size=(5, 2))

In [5]: x, y, z
Out[5]: 
(array([[3, 0, 4, 0, 8],
        [1, 2, 8, 1, 0]]),
 array([[3, 9],
        [9, 7],
        [6, 9],
        [6, 8],
        [2, 2]]),
 array([[1, 6],
        [4, 3],
        [8, 6],
        [8, 3],
        [1, 0]]))

In [6]: %timeit x @ y - x @ z
2.22 µs ± 29.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit x @ (y - z)
1.33 µs ± 22.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

algorithms/symmetry/cosym/target.py
```diff
@@ -371,32 +366,13 @@ def compute_gradients(self, x: np.ndarray) -> np.ndarray:
...
-            tmp = np.outer(coords[i], coords[i])
-            if self.wij_matrix is not None:
-                tmp = np.multiply(self.wij_matrix, tmp)
-            for j in range(self.dim):
-                tmp_array[j * NN : (j + 1) * NN] = np.matmul(tmp, coords[j])
-            grad -= tmp_array
-        grad *= -2
-
-        return grad
+            grad = -2 * (x @ self.rij_matrix - x @ (x.T @ x))
```
benjaminhwilliams, line 374: See above.
```suggestion
            grad = -2 * x @ (self.rij_matrix - x.T @ x)
```

benjaminhwilliams (changes requested):
I haven't seen the algebra, so I can't check the consistency of the code with the desired arithmetic, but I presume that's not at question here.  Changes look good and improve clarity, with a nice performance improvement.  I have one small suggestion (affects two places).  I haven't tested my suggestion though, so YMMV.  Otherwise, LGTM.

algorithms/symmetry/cosym/target.py
rjgildea: Thanks, good spot. Based on limited testing it doesn't appear to give a _significant_ performance benefit in the wider context, but it seems a sensible and useful improvement nonetheless 👍

benjaminhwilliams (approved):

benjaminhwilliams (approved):

algorithms/symmetry/cosym/target.py
```diff
@@ -370,31 +370,30 @@ def compute_gradients(self, x: np.ndarray) -> np.ndarray:
...
         NN = x.size // self.dim
-        for i in range(self.dim):
-            coords.append(x[i * NN : (i + 1) * NN])
+        coords = x.reshape((self.dim, NN))
 
         # term 1
+        grad = np.empty(coords.shape)
         for i in range(self.dim):
-            grad[i * NN : (i + 1) * NN] = np.matmul(wrij_matrix, coords[i])
+            grad[i] = np.matmul(wrij_matrix, coords[i])
```
elena-pascal: `matmul` can also take stacks of matrices and would interpret it as such if `coords` dimensions is > 2. I think 
`grad = np.matmul(wrij_matrix, coords)` _should_ also work

Maybe for future considerations

Issue #1638, opened 2021-03-29
Profile fitting issues for reflections in panel gaps.


jbeilstenedmands:
Following on from #1297, we now integrate reflections that have part of their intensity in the panel gaps, however I have found that the profile fitting seems to be not working well in this case. This has become apparent in some datasets I've been processing, where I am seeing many profile fitted intensities being given zero or near zero intensities, but being marked as successfully profile fitted with the `integrated_prf` flag.

I tried to simulate this using some ideas from the `test_profile_fitting.py` code. To simulate a reflection moving into a panel gap, I simulated a `13x13x13` pixel box, with a gaussian centred at `6,6,6` with width `2,2,2`, progressively masking a region of the pixel box to simulate an edge (code example below). Plotted below is the ratio of the fitted to true intensity for a number of rows of the box that are masked, repeated 1000 times over random seed values for the poisson noise.

![Intensity_variation](https://user-images.githubusercontent.com/30625594/112855209-82e5c800-90a6-11eb-90e9-1f417d59540b.png)

And here are a few histograms of the same data.
One row masked (vast majority within 1% so good):
![histogram_1](https://user-images.githubusercontent.com/30625594/112855640-ecfe6d00-90a6-11eb-9678-5ab5a146dc5d.png)

Three rows masked (starting to see larger variability):
![histogram_3](https://user-images.githubusercontent.com/30625594/112855703-f982c580-90a6-11eb-8384-d65354e8e7c2.png)

Six rows masked (significant variability):
![histogram_6](https://user-images.githubusercontent.com/30625594/112855744-02739700-90a7-11eb-8ffc-9ddb4abec9f8.png)

Eleven rows masked, highly unreliable, here you can see that some fits are given zero intensity:
![histogram_11](https://user-images.githubusercontent.com/30625594/112855772-0bfcff00-90a7-11eb-8375-1231c8ad4806.png)

Picture of a real reflection where this occurs:
![Screenshot_panelgap](https://user-images.githubusercontent.com/30625594/112857292-9bef7880-90a8-11eb-912a-57defa2f86c2.png)

FWIW I'm not suggesting reverting #1297 as I think we should be integrating cases where there are only a few masked pixels and the tests from that PR show that there is only small bias for a low number of masked pixels. However maybe we should be setting limits on how far into a panel gap a reflection can go before the fit is unreliable, and assigning a higher uncertainty for these cases.

Code:
```
[...]

PR #1297, opened 2020-06-09
Be more permissive in profile fitting

rjgildea:
- Fit profiles to reflections with invalid pixels: Profile fitting can handle masked pixels, so no obvious reason to exclude all reflections with masked from profile fitting.
- Remove checks on bbox limits: this dramatically increases the number of profile-integrated reflections for multi-panel detectors

See #1291, #1294 and #1295

PR #1640, opened 2021-03-30
In profile-fitted integration, don't integrate reflections with a high fraction of invalid pixels.

jbeilstenedmands:
For profile-fitted integration, only integrate reflections that have a high fraction of valid foreground pixels (currently set to ~~0.6~~ 0.75). Controlled by parameter valid_foreground_threshold (range 0 -> 1). For #1638.

This is intended so that reflections which have centroids in the panel gaps are not integrated by profile fitting. In my opinion the default parameter should be at least 0.5 so that the peak centre is approximately observed. Perhaps a higher threshold such as 0.75 is more appropriate, I plan to investigate how the data quality is affected depending on the threshold value.

### DIFF
--- a/algorithms/integration/integrator.py
+++ b/algorithms/integration/integrator.py
@@ -184,6 +184,12 @@ def generate_phil_scope():
           .type = bool
           .help = "Use profile fitting if available"
 
+        valid_foreground_threshold = 0.6
+          .type = float(value_min=0, value_max=1)
+          .help = "The minimum fraction of foreground pixels that must be valid"
+                  "in order for a reflection to be integrated by profile fitting."
+          .expert_level = 2
+
         sigma_b_multiplier = 2.0
           .type = float(value_min=1.0)
           .help = "Background box expansion factor"
@@ -356,6 +362,7 @@ class Parameters:
         def __init__(self):
             self.fitting = True
             self.sigma_b_multiplier = 2.0
+            self.valid_foreground_threshold = 0.6
             self.validation = Parameters.Profile.Validation()
 
     def __init__(self):
@@ -421,6 +428,9 @@ class Parameters:
 
         # Profile parameters
         result.profile.sigma_b_multiplier = params.profile.sigma_b_multiplier
+        result.profile.valid_foreground_threshold = (
+            params.profile.valid_foreground_threshold
+        )
 
         # Get the min zeta filter
         result.filter.min_zeta = params.filter.min_zeta
@@ -807,7 +817,9 @@ class IntegratorExecutor(Executor):
     The class to process the integration data
     """
 
-    def __init__(self, experiments, profile_fitter=None):
+    def __init__(
+        self, experiments, profile_fitter=None, valid_foreground_threshold=0.6
+    ):
         """
         Initialize the executor
 
@@ -816,6 +828,7 @@ class IntegratorExecutor(Executor):
         self.experiments = experiments
         self.overlaps = None
         self.profile_fitter = profile_fitter
+        self.valid_foreground_threshold = valid_foreground_threshold
         super().__init__()
 
     def initialize(self, frame0, frame1, reflections):
@@ -880,15 +893,24 @@ class IntegratorExecutor(Executor):
         # Check for invalid pixels in foreground/background
         reflections.contains_invalid_pixels()
 
+        # Exclude reflections where a high fraction of the foreground is masked
+        # e.g. due to a panel edge, as this will make the fitting unreliable.
+        sbox = reflections["shoebox"]
+        nvalfg = sbox.count_mask_values(MaskCode.Valid | MaskCode.Foreground)
+        nforeg = sbox.count_mask_values(MaskCode.Foreground)
+        fraction_valid = nvalfg.as_double() / nforeg.as_double()
+        selection = fraction_valid < self.valid_foreground_threshold
+        reflections.set_flags(selection, reflections.flags.dont_integrate)
+
         # Process the data
         reflections.compute_background(self.experiments)
         reflections.compute_centroid(self.experiments)
+
         reflections.compute_summed_intensity()
         if self.profile_fitter:
             reflections.compute_fitted_intensity(self.profile_fitter)
 
         # Compute the number of background/foreground pixels
-        sbox = reflections["shoebox"]
         reflections["num_pixels.valid"] = sbox.count_mask_values(MaskCode.Valid)
         reflections["num_pixels.background"] = sbox.count_mask_values(
             MaskCode.Valid | MaskCode.Background
@@ -896,9 +918,7 @@ class IntegratorExecutor(Executor):
         reflections["num_pixels.background_used"] = sbox.count_mask_values(
             MaskCode.Valid | MaskCode.Background | MaskCode.BackgroundUsed
         )
-        reflections["num_pixels.foreground"] = sbox.count_mask_values(
-            MaskCode.Valid | MaskCode.Foreground
-        )
+        reflections["num_pixels.foreground"] = nvalfg
 
         # Print some info
         fmt = " Integrated % 5d (sum) + % 5d (prf) / %5d reflections on image %d"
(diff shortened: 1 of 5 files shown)


algorithms/integration/integrator.py
```diff
@@ -356,6 +362,7 @@ def __init__(self):
         def __init__(self):
             self.fitting = True
             self.sigma_b_multiplier = 2.0
+            self.valid_foreground_threshold = 0.6
```
graeme-winter, line 365: Should read default value from `master_phil`?

algorithms/integration/integrator.py
```diff
@@ -807,7 +817,9 @@ class IntegratorExecutor(Executor):
     The class to process the integration data
     """
 
-    def __init__(self, experiments, profile_fitter=None):
+    def __init__(
+        self, experiments, profile_fitter=None, valid_foreground_threshold=0.6
```
graeme-winter, line 821: At risk of repeating myself, can we pull the default from PHIL?

algorithms/integration/integrator.py
```diff
@@ -880,25 +893,32 @@ def process(self, frame, reflections):
         # Check for invalid pixels in foreground/background
         reflections.contains_invalid_pixels()
 
+        # Exclude reflections where a high fraction of the foreground is masked
+        # e.g. due to a panel edge, as this will make the fitting unreliable.
+        sbox = reflections["shoebox"]
+        nvalfg = sbox.count_mask_values(MaskCode.Valid | MaskCode.Foreground)
+        nforeg = sbox.count_mask_values(MaskCode.Foreground)
+        fraction_valid = nvalfg.as_double() / nforeg.as_double()
+        selection = fraction_valid < self.valid_foreground_threshold
```
graeme-winter, line 902: Can we somehow report on #reflections no integrated here? Debug seems sensible

algorithms/integration/report.py
```diff
@@ -148,7 +148,7 @@ def binned_report(binner, index, data):
 
         try:
             report["rmsd_xy"] = list(
-                indexer_sum.mean(data["xyz.rmsd"].select(data["int"]))
+                indexer_sum.mean(data["xyz.rmsd"].select(data["sum"]))
```
graeme-winter, line 151: Seems like a change which is a peripheral bug fix?

graeme-winter: i.e. if this is the case, different PR?

test/command_line/test_integrate.py
```diff
@@ -254,6 +254,34 @@ def test_integration_with_sample_size(dials_data, tmpdir):
...
+            refls,
+            "profile.fitting=True",
+            "sampling.integrate_all_reflections=False",
+            "sampling.minimum_sample_size=500",
+            "prediction.padding=0",
+        ],
+        working_directory=tmpdir,
+    )
+    assert not result.returncode and not result.stderr
+    table = flex.reflection_table.from_file(tmpdir / "integrated.refl")
```
graeme-winter, line 276: Propose stronger test would be to check that there are no reflection centroids in tile join regions?

graeme-winter (commented):
Stage 0 review of looking at the code changes - seem sane - propose moving the default value to a unique place i.e. the master phil, not scattering 0.6 everywhere, and there seemed to be an unrelated change in the mix. 

I will run some actual integration tests now (no pun intended) to assess the impact of these changes. I will approve once that change is made. 

I think this is a large fanfare bug fix which warrants a 3.4.2 @ndevenish

jbeilstenedmands:
Tested on the x4-wide dataset. This PR gives significant improvement in merging R-factors, Isigma compared to current release (dials-3.4), with 10K fewer reflections. Compared to dials-3.3, R-factors are slightly worse however there are 6K more reflections.

This PR, with threshold value of 0.75:
```
High resolution limit                           1.09    2.95    1.09
Low resolution limit                           29.94   29.95    1.11
Completeness                                   96.6   100.0    71.1
Multiplicity                                    5.5     5.6     2.5
I/sigma                                        10.1    74.7     0.1
Rmerge(I)                                     0.091   0.034   2.577
Rmerge(I+/-)                                  0.075   0.020   2.637
Rmeas(I)                                      0.099   0.037   3.149
Rmeas(I+/-)                                   0.090   0.024   3.530
Rpim(I)                                       0.040   0.015   1.766
Rpim(I+/-)                                    0.048   0.013   2.329
CC half                                       0.999   0.999   0.297
Wilson B factor                              11.430
Anomalous completeness                         93.4    98.1    53.1
Anomalous multiplicity                          3.0     3.5     1.5
Anomalous correlation                         0.495   0.790   0.068
Anomalous slope                               0.381
dF/F                                          0.093
dI/s(dI)                                      0.627
Total observations                            82773    5011    1330
Total unique                                  15109     893     533
```

dials-3.4 (after #1297):
```
High resolution limit                           1.09    2.95    1.09
Low resolution limit                           29.94   29.95    1.11
Completeness                                   97.1   100.0    72.7
Multiplicity                                    6.1     5.7     2.8
I/sigma                                         9.3    66.4     0.1
Rmerge(I)                                     0.108   0.036   2.861
Rmerge(I+/-)                                  0.091   0.022   2.589
Rmeas(I)                                      0.118   0.040   3.445
Rmeas(I+/-)                                   0.107   0.027   3.409
Rpim(I)                                       0.046   0.016   1.864
Rpim(I+/-)                                    0.056   0.014   2.196
CC half                                       0.998   0.999   0.233
Wilson B factor                              12.480
Anomalous completeness                         95.3    98.4    63.1
Anomalous multiplicity                          3.3     3.5     1.6
Anomalous correlation                         0.479   0.776   0.017
Anomalous slope                               0.430
dF/F                                          0.128
dI/s(dI)                                      0.736
Total observations                            92365    5124    1538
Total unique                                  15187     896     545
```

dials-3.3.5 (before #1297), processed to same resolution:
```
High resolution limit                           1.09    2.96    1.09
Low resolution limit                           29.94   29.95    1.11
Completeness                                   96.0   100.0    67.8
Multiplicity                                    5.1     5.2     2.3
I/sigma                                        11.1    83.2     0.1
Rmerge(I)                                     0.086   0.029   3.887
Rmerge(I+/-)                                  0.072   0.019   4.490
Rmeas(I)                                      0.095   0.032   4.820
Rmeas(I+/-)                                   0.087   0.022   6.103
Rpim(I)                                       0.040   0.013   2.791
Rpim(I+/-)                                    0.048   0.012   4.109
CC half                                       0.999   0.999   0.169
Wilson B factor                              11.460
Anomalous completeness                         91.4    92.6    46.5
Anomalous multiplicity                          2.8     3.3     1.4
Anomalous correlation                         0.434   0.743   0.040
Anomalous slope                               0.435
dF/F                                          0.091
dI/s(dI)                                      0.640
Total observations                            76654    4660    1176
Total unique                                  14995     892     512
```

jbeilstenedmands:
Updated the default threshold to be 0.75, to be more conservative, and this is also the value for the equivalent idea in XDS.

algorithms/integration/integrator.py
jbeilstenedmands: Unfortunately this hoop jumping is needed due to the way the parameters are constructed in the integration code: `self.params = Parameters.from_phil(params.integration)`. I don't know why it needs to be this way but it won't work without it .....

graeme-winter: 🧐

OK, propose we live with this for now and then add an issue to resolve why this is the case? Because this _will_ end badly at some point in the future.

algorithms/integration/report.py
jbeilstenedmands: It's somewhat related, it affects the output values in the table after the changes in this PR (there is a size error causing the calculation to fail and this column is populated with zero values, before the changes it happens to work ok due to the data arrays being the same size), so best to include it in here I think.

graeme-winter (approved):
As above, change set looks good to go thank you

Agree with setting MIN_PK equivalent to 0.75

PR #1644, opened 2021-04-07
Ensure easy_mp child process log messages are handled with the correct threshold level in the parent process

benjaminhwilliams:


test/util/test_log.py
```diff
@@ -33,3 +36,72 @@ def test_LoggingContext():
...
+        """
+        dials.util.log.config_simple_cached()
+        logger = logging.getLogger("dials")
+        logger.info(test_log_message)
+        (handler,) = logger.handlers
+        return handler.records
+
+    # Check that the child processes cache INFO log messages when the parent process
+    # logger's threshold level is the default (DEBUG).
+    dials.util.log.config(verbosity=2)
```
Anthchirp, line 73: This (and other statements) will leak into the next test.

Anthchirp:
What happens to the records? Are they emitted later? Would it not be better to pass everything from the subprocess and have the parent worry about loglevel (and source) filtering?

benjaminhwilliams:
> What happens to the records? Are they emitted later? Would it not be better to pass everything from the subprocess and have the parent worry about loglevel (and source) filtering?

That's logging fix option 2 (probably the better option), which I haven't pushed yet.

benjaminhwilliams:
> What happens to the records? Are they emitted later? Would it not be better to pass everything from the subprocess and have the parent worry about loglevel (and source) filtering?

That's logging fix option 2 (probably the better option), which I haven't pushed yet.

test/util/test_log.py
benjaminhwilliams: Yes, I realised this as I was committing it but I wanted to get some CI checks for the benefit of my own understanding of things, so I pushed it like this and intend to tidy it up later.  Hence draft.

PR #1645, opened 2021-04-08
Correctly re-handle cached log records from easy_mp child processes

benjaminhwilliams:
`dials.find_spots` and `dials.integrate` use `libtbx.easy_mp` for multiprocessing.  Child processes forked/spawned in this way create log messages, including `INFO` messages.  These get emitted even if you begin with `logging.getLogger("dials").setLevel(logging.WARNING)`, despite the fact that the `INFO` messages don't meet this `WARNING` threshold.  The changes here fix that.

At present, log records created in `easy_mp` child processes are initially handled by a simple caching logger, created with `dials.util.log.config_simple_cached`, which saves the log records and doesn't emit them.  This is so that they can eventually be emitted in the parent process once the child process has completed.  Before emitting, we need to check whether the log record severity meets the threshold logging level for the relevant logger in the parent process.  This check has not previously been enforced, so that log records raised with low severity in child processes would get emitted in the parent process even if the logger in the parent process had its threshold set to a higher severity.

For the most part, this has never been a problem in DIALS itself, but when external tools like screen19 call DIALS functions and want to silence some of the DIALS logging, messages from spot finding and integration slip through because they don't obey any modified logging levels.

Fixes xia2/screen19#49
Closes #1644


Anthchirp (approved):

benjaminhwilliams:
I think these changes are preferable to those in #1644, but it was helpful for me to open #1644 by way of thinking out loud.

PR #1647, opened 2021-04-13
Cosym: remove clustering code

rjgildea:
Alternative/simpler method to determine the reindexing operations. Since moving to the pointless-style space group determination it shouldn't be necessary to rely on clustering to assign the reindexing ops.

Instead:
1. choose a high density point as a seed
2. for each dataset, find the nearest point to the current centroid corresponding to points in the same indexing mode
3. map each reindexing operator to the unique coset representative based on the coset left decomposition


phyy-nx:
Hi @rjgildea, I'm hoping you can hold this PR?  It's an API breaking change that breaks our dials.cosym wrappers in XFEL, as tested in xfel_regression.

I've been working on a PR to add xfel_regression to the dials azure system so it would get checked on PRs, which would help keep things synchronized better (more below).  In the meantime, here's how you can run xfel_regression tests using the dials bootstrap:

```
wget https://raw.githubusercontent.com/dials/dials/main/installer/bootstrap.py
python bootstrap.py
source dials
conda install -c conda-forge git-lfs distro pandas mpi4py
cd modules
git clone https://gitlab.com/cctbx/xfel_regression.git
cd xfel_regression
git lfs install --local
git lfs pull
cd ../../
mkdir test; cd test
libtbx.configure xfel_regression
libtbx.run_tests_parallel module=xfel_regression nproc=auto
```

For reference about dials/Azure/xfel_regression: I've forked dials, created a sandbox Azure system, and connected it to my fork.  I've then added an xfel_regression pipeline to a branch in that fork, and while it works, one of the macos pipelines is timing out and I don't know why.  You can my progress [here](https://github.com/phyy-nx/dials/runs/2301564824).

ndevenish:
It looks like the only API breakage is the removal of the PHIL parameters - I agree that it might be worth "keeping them in" but warning if they are used that it isn't doing anything

ndevenish:
Oh, I guess the `seed_clustering` class was removed also

rjgildea:
After re-adding a `cluster.n_clusters` no-op parameter, I get this far:
```
  Trying to abort all MPI processes because of exception in process 4:
    File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/cctbx_project/xfel/merging/command_line/merge.py", line 179, in <module>
      result = script.run()
    File "/Users/rjgildea/software/cctbx_py3/build/../../cctbx/modules/cctbx_project/xfel/merging/command_line/merge.py", line 142, in run
      experiments, reflections = worker.run(experiments, reflections)
    File "/Users/rjgildea/software/cctbx/modules/cctbx_project/xfel/merging/application/modify/cosym.py", line 157, in run
      COSYM.run()
    File "/Users/rjgildea/software/cctbx/modules/dials/util/observer.py", line 44, in notify
      r = method(self, *args, **kwargs)
    File "/Users/rjgildea/software/cctbx/modules/cctbx_project/xfel/merging/application/modify/aux_cosym.py", line 282, in run
      self.cosym_analysis.run()
    File "/Users/rjgildea/software/cctbx/modules/cctbx_project/xfel/merging/application/modify/aux_cosym.py", line 120, in run
      self._cluster_analysis()
    File "/Users/rjgildea/software/cctbx/modules/dials/util/observer.py", line 44, in notify
      r = method(self, *args, **kwargs)
    File "/Users/rjgildea/software/cctbx/modules/cctbx_project/xfel/merging/application/modify/aux_cosym.py", line 129, in _cluster_analysis
      self.cluster_labels = self._do_clustering(self.params.cluster.method)
  AttributeError: 'CosymAnalysis' object has no attribute '_do_clustering'
```

I note that you here override a private method, `_cluster_analysis()` which then calls further private methods:
https://github.com/cctbx/cctbx_project/blob/4238d5d91491d8d3bafbdbda9dbbed3126e065fc/xfel/merging/application/modify/aux_cosym.py#L97-L128

On a related note, you also override the `run()` method, which simply calls a sequence of private methods. The call to `_optimise()` in your version of the `run()` method will not behave as you expect, as it now expects the first parameter to be the `engine`, i.e. either a string `"scitbx"` or `"scipy"`

https://github.com/cctbx/cctbx_project/blob/4238d5d91491d8d3bafbdbda9dbbed3126e065fc/xfel/merging/application/modify/aux_cosym.py#L80-L95

https://github.com/dials/dials/blob/396ba98e3cec763c1543bd98514814610e6d049f/algorithms/symmetry/cosym/__init__.py#L291-L315

It isn't entirely obvious to me why it is necessary to override so many methods (including private methods, which should be discouraged) - many of them seem to be near duplicates of the existing dials code (or some former state of the code). This approach seems prone to issues such as this, and it would be better to find a cleaner approach to any specialisations you require.

nksauter:
I'd be happy to have the necessary methods cleanly implemented in xfel
without relying on the changing interface to dials.cosym.

graeme-winter:
@nksauter  @phyy-nx  I think if we had a better idea of what you needed / were using and were trying to achieve we could probably make the interface a bit more reliable and your usage less painful. Obviously copying and pasting the code will _work_ but does mean that you can't benefit if any bugs are found etc. 

Am minded to agree with @rjgildea that overriding private methods or depending on the implementation details is not ideal.

Regarding testing: if there is behaviour that you depend on in the code would suggest adding a specific lightweight test to the `dials` code base which exercises this rather than relying only on `xfel-regression` - the latter is substantial and non-trivial to run (I have experience with this) and would probably exceed e.g. our free Azure disk quota etc.

dwpaley:
It's not necessary to leave the vestigial clustering phil params. The test failures from removing those will be fixed by cctbx PR 607 and the following xfel_regression PR: https://gitlab.com/cctbx/xfel_regression/-/merge_requests/2

With those two corresponding PRs (plus cctbx commit f7ce08b), the xfel_regression test of our modify_cosym merging worker passes again. Unfortunately that test uses ~200 MB of input data so there is no immediately obvious path to adding that test to the dials tests.

phyy-nx:
Hey some clarification on Azure.  Each job job in a pipeline (aka each checkmark in the commit dropdown menu) is totally independent and gets its own storage and VM.  So my work in my [sandbox](https://github.com/phyy-nx/dials/runs/2301564824) would not impact anything because it only adds a new pipeline which only runs xfel_regression, not any of the dials tests, and it runs separate from the others.

Essentially a pipeline for dials_regression, xia2, dials, dxtbx, xfel_regression, and so forth, would mean each one can use the full storage available.  Docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#hardware

PR #1651, opened 2021-04-13
Fix error pickling log test function

ndevenish:
By moving it to the outer (and thus importable) scope. For testing.

Hopefully Fixes #1650.

PR #1652, opened 2021-04-14
re-enable Azure Windows builds

Anthchirp:
They will be horribly slow until we stop building cbflib

### DIFF
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -10,12 +10,12 @@ jobs:
 
   strategy:
     matrix:
-      python36:
-        PYTHON_VERSION: 3.6
-#     python37:
-#       PYTHON_VERSION: 3.7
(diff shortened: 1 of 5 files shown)


Anthchirp:
So the runtime is now on the same order of magnitude as linux and mac builds, so in principle this can be merged. Might want to look into that failing pickle test though. (Appreciate this is the aim of #1651)

Anthchirp:
And we should probably document somewhere that Python 3.6 is not supported on Windows

dagewa:
Sounds good. I'm happy to write a "Building DIALS on Windows" webpage. I suppose we are still some way off providing installable builds.

We should also clearly document that Windows support is still limited, due to lack of multiprocessing. That's something I want to look into, but I think it's going to be a case of making everything I look at pickleable, which is going to be quite some work.

Anthchirp:
To be honest we can document the lack of support in the code quite easily...

PR #1660, opened 2021-04-16
Spot-count histogram: use math.ceil instead of round

rjgildea:
Python uses a rounding half to even strategy, which means that `1.5` and `2.5` both round to `2`, which resulted inconsistency in the spot-count depending on whether there was an even or odd number of images.


graeme-winter:
🧐    It was that easy?

I spent a while looking at this in the past, was evidently looking in `${WRONG-PLACE}` 🤬

test/util/test_ascii_art.py
```diff
@@ -95,3 +95,32 @@ def test_spot_counts_per_image_plot(dials_regression):
 1 image 10"""
     output = "\n".join(line.rstrip() for line in output.splitlines())
     assert output == expected_output
+
```
graeme-winter, line 98: Is it worth having this test on e.g. 1...9 images of centroid test data to make sure that it works _right_ for all the cases?

graeme-winter (approved):
Thanks for fixing this - it was one of the more annoying "features" and I am genuinely shocked that the fix was this simple, but it presumably was so that's great!

test/util/test_ascii_art.py
rjgildea: The bug here was that the behaviour was inconsistent between even- and odd-length imagesets. We already had a test for output of even-length imagesets (1..10 above). I have added a test for an odd-length imageset below (1..3). We can change that to 1..9 if you think that would be better, but I don't think it would make any difference to the outcome.

rjgildea:
You can see the old behaviour by doing:
```
$ dials.import $(dials.data get -q x4wide)/X4_wide_M1S4_2_*.cbf

$ dials.find_spots imported.expt scan_range=1,2
...
Histogram of per-image spot count for imageset 0:
738 spots found on 2 images (max 396 / bin)
* 
**
**
**
**
**
**
**
**
**
12

$ dials.find_spots imported.expt scan_range=1,3
...
Histogram of per-image spot count for imageset 0:
958 spots found on 2 images (max 514 / bin)
* 
**
**
**
**
**
**
**
**
**
12

$ dials.find_spots imported.expt scan_range=1,4
...
Histogram of per-image spot count for imageset 0:
1190 spots found on 4 images (max 433 / bin)
*  *
*  *
*  *
*  *
*  *
*  *
** *
****
****
****
1  4

$ dials.find_spots imported.expt scan_range=1,5
...
Histogram of per-image spot count for imageset 0:
1449 spots found on 4 images (max 515 / bin)
*  *
*  *
*  *
*  *
*  *
*  *
****
****
****
****
1  4
```
With the changes in this PR, the behaviour is now as expected.

PR #1674, opened 2021-04-23
Allow finding dials_regression from environment

ndevenish:
As an alternative to copying into multiple places, or softlinking manually, it makes sense to follow the example of [dials-data](https://github.com/dials/data/blob/e17760480c363df08654e60dfc31de981c30419f/dials_data/datasets.py#L73) and let the be environment-configured so that we never need to think of it again, once it is setup.

Same change as https://github.com/cctbx/dxtbx/pull/350, but in dials, since this logic is duplicated for package-independence.


Anthchirp:
build failure unrelated and fixed on `main`

PR #1675, opened 2021-04-24
Partially fix cosym documentation build

ndevenish:
The merging of #1647 removed a module previously referenced by the documentation. That is fixed here.

Also a problem, however is this section in the [`dials.cosym`](https://dials.github.io/documentation/programs/dials_cosym.html) documentation:

> image

This output has now been [removed](https://github.com/dials/dials/commit/a7bb8c747c5b45f4b8880b0fdfb799ee90b33094#diff-4eb0e90752ec0c5de32be735a2b55c490308091767055a97f52eb6315a7fac07L195-L198), so requires reconsideration.


Anthchirp:
build failure unrelated and fixed on `main`

PR #1676, opened 2021-04-25
Update Azure builds

Anthchirp:
* Remove build caches, because they're not trustworthy anyway (cf. #1582)
* Simplify the build scripts
* Replace miniconda with micromamba (new bootstrap parameter `--mamba`)
* new bootstrap parameter `--clean` to remove the temporary miniconda/micromamba install straightaway

Preparatory step before we look into more complicated testing workflows on Azure (#1671)

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,3 +1,6 @@
+variables:
+  CACHE_VERSION: 20201102
+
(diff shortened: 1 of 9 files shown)


Anthchirp:
pytype errors are false positives

PR #1678, opened 2021-04-26
Fix procrunner deprecation warning

rjgildea:


Anthchirp (approved):
Bonus points for using `perf_counter` over `time`

PR #1688, opened 2021-05-06
Add absorption_level=high/medium/low options to dials.scale

jbeilstenedmands:
As part of allowing easier user-control of the absorption correction in `dials.scale`, introduce a three level option that sets `lmax` and `surface_weight` to more suitable values.
The low, medium and high are for relative absorption corrections of approximately 1%, 5% and 25%+.

If the `absorption_level` option is not chosen (the default is no value for this parameter), the two parameters default to the same values as for low.

This structure will allow setting by `xia2`, whether a low/medium/high is chosen or if there is some higher level logic taking into account sample and experiment details.

I will test this on a range of datasets to check that the `surface_weight` settings are suitable but quick tests indicate that the numbers are around the right level.


jbeilstenedmands:
Note, I have changed the parameter name to `absorption_level`, so that it can be the same as the proposed `xia2` option (where the parameter `absorption` is already taken and has a different function). I think this name is probably a little clearer also.

algorithms/scaling/model/model.py
```diff
@@ -122,15 +122,35 @@
...
+    .help = "Expected degree of relative absorption for different scattering"
+            "paths through the crystal(s). If an option is selected, the"
+            "scaling model parameters lmax and surface_weight will be set to"
+            "appropriate values."
+            "Relative absorption increases as crystal size increases,"
+            "increases as wavelength increases and is increased as the crystal"
+            "dimensions become less equal (i.e. is higher for needle shaped"
+            "crystals and zero for a spherical crystal)."
+            "Definitions of the levels and approximate correction magnitude:"
+            "low:    ~1%% relative absorption, expected for crystals on the order"
```
graeme-winter, line 136: protein crystals - 100 µm germanate (say) will absorb a lot more.

graeme-winter (approved):
Change set looks sane, made one small comment to the help str however I do wonder what happens if you set `absorption_level=high` and have a crystal with low symmetry (so the correction parameters are not well constrained.)

Overall though I think this is a valiable addition I will find myself using frequently. Thank you.

algorithms/scaling/model/model.py
jbeilstenedmands: perhaps 'typical protein crystal' would be a better description i.e. containing only CHNOS?

joserbngithub:
hi @graeme-winter , what are the specs of the test datasets? where are these sets?

cheers,
Jose'

> As part of allowing easier user-control of the absorption correction in `dials.scale`, introduce a three level option that sets `lmax` and `surface_weight` to more suitable values.
> The low, medium and high are for relative absorption corrections of approximately 1%, 5% and 25%+.
> 
> If the `absorption_level` option is not chosen (the default is no value for this parameter), the two parameters default to the same values as for low.
> 
> This structure will allow setting by `xia2`, whether a low/medium/high is chosen or if there is some higher level logic taking into account sample and experiment details.
> 
> I will test this on a range of datasets to check that the `surface_weight` settings are suitable but quick tests indicate that the numbers are around the right level.

Issue #1689, opened 2021-05-07
dials.spot_counts_per_image counts table/overall mismatch for indexed reflections


huwjenkins:
Hi,

I was interested in using `dials.spot_counts_per_image` to look at the fraction of indexed reflections but it appears this cannot work because only the indexed reflections are passed to the `per_image_analysis` code to make the per image table. 

As an example I took the first 10 images from the "Correcting poor initial geometry" tutorial dataset:

`dials.import x247398/t1.0*.img.bz2 image_range=1,10`

and ran spot finding:

`dials.find_spots imported.expt nproc=4`

and indexing:

`dials.index imported.expt strong.refl`

Now `indexed.refl` has the following (output from `dials.filter_reflections indexed.refl`):
```+--------------------+--------+
| flag               |   nref |
|--------------------+--------|
| centroid_outlier   |    819 |
| indexed            |   4083 |
| strong             |   5413 |
| used_in_refinement |   1000 |
+--------------------+--------+
```

`dials.spot_counts_per_image imported.expt indexed.refl resolution_analysis=false` gives:
```
+---------+----------+-----------------+-------------------+
|   image |   #spots |   #spots_no_ice |   total_intensity |
|---------+----------+-----------------+-------------------|
|       1 |      439 |             389 |           3412005 |
|       2 |      425 |             369 |           4363618 |
|       3 |      387 |             345 |           4323713 |
|       4 |      363 |             324 |           3633360 |
|       5 |      397 |             352 |           3702135 |
|       6 |      417 |             381 |           5352372 |
|       7 |      398 |             346 |           3804298 |
|       8 |      370 |             330 |           3602412 |
|       9 |      416 |             379 |           4989161 |
|      10 |      471 |             423 |           3546383 |
+---------+----------+-----------------+-------------------+
+------------------------+---------------+
| Overall statistics     |               |
|------------------------+---------------|
| #spots                 | 5413          |
| #spots_no_ice          | 4854          |
| d_min                  | -1.00         |
| d_min (distl method 1) | -1.00 (-1.00) |
| d_min (distl method 2) | -1.00 (-1.00) |
+------------------------+---------------+
```
but while the overall #spots is correct the sum of the #spots column in the table is 4083 - i.e. the number of _indexed_ reflections. Running `dials.spot_counts_per_image imported.expt strong.refl resolution_analysis=false` confirms this:
```
+---------+----------+-----------------+-------------------+
|   image |   #spots |   #spots_no_ice |   total_intensity |
[...]

PR #1690, opened 2021-05-07
dials.spot_counts_per_image: explicitly fail given indexed reflections

rjgildea:
Since behaviour isn't currently well-defined on indexed reflections, ensure that we are only given output of spotfinding.

See also #1689


graeme-winter:
The proper fix for this involves resolving #1029 which in turn will generate a lot of work around every step downstream of the indexing. In the interim, this pull request would resolve the issue of user surprise which would be an improvement. 

If we can resolve the underlying question then the changes in here may need to be reverted, but experience suggests that this is unlikely to happen in short order so justifies @rjgildea work here.

command_line/spot_counts_per_image.py
```diff
@@ -67,6 +67,12 @@ def run(args=None):
         sys.exit("Only one reflection list may be passed")
     reflections = reflections[0]
 
+    if "miller_index" in reflections:
```
graeme-winter, line 70:
```
if "miller_index" in reflctions or any(experiments.crystals())
```
->  catch both outcomes (because I can see someone passing indexed.refl imported.expt or something in response to one of the cases)

if both then maybe 

```
sys.exit("Only unindexed experiments / reflections are currently supported")
```

So you catch this in one go?

graeme-winter (approved):
I am not happy that this is the best solution, but I respect the point that this is the most pragmatic resolution to the situation we find ourselves in, thanks for making the changes.

I'll tag this commit once merged on #1029 as something which may need to be reverted if we get that far.

PR #1693, opened 2021-05-10
Faster numpy-based matching of miller indices

rjgildea:
Replace use of the cctbx `match_miller_indices()` method with a faster numpy-based approach.

1. Pre-calculate a mapping of miller indices to a 1D array (using `np.ravel_multi_index()`)
2. Pre-calculate epsilons
3. For each CC1/2 calculation, identify the matching pairs of indices using `np.intersect1d()`

Using a multi-crystal example with 133 datasets in `P 3 2 1` (i.e. an Rij matrix of size `(12 * 133) ** 2 == 2547216`), this PR improves the time taken to calculated the Rij matrix from 126 seconds to 79 seconds.

### DIFF
--- a/.conda-envs/linux.txt
+++ b/.conda-envs/linux.txt
@@ -24,6 +24,7 @@ conda-forge::msgpack-c
 conda-forge::msgpack-python
 conda-forge::numpy>=1.16
 conda-forge::orderedset
+conda-forge::pandas
 conda-forge::pillow>=5.4.1
 conda-forge::pip
 conda-forge::procrunner>=2.2.0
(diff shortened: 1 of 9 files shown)


benjaminhwilliams:
Ah, I see this got merged.  I think we intended to fix up the PR description, since a lot had changed since the PR was opened.

benjaminhwilliams:
Sorry, it dropped off my to-do list.

ndevenish:
Is
> ``dials.cosym``: Significantly faster calculation of Rij matrix of pairwise correlation coefficients
Still a reasonable description?

benjaminhwilliams:
Yes, the news fragment still makes sense.  It's just that the commit message doesn't match the diff.

PR #1694, opened 2021-05-10
Enumerate experiment index

graeme-winter:
Start making use of experiment.index from https://github.com/cctbx/dxtbx/pull/359

This is largely "housekeeping" however is preliminary work for some other work towards #1029. 

This is work in progress - some of the changes are trivial but some of the instances are more involved. N.B. none of these changes are _critical_ for the `dxtbx` PR (or indeed, even needed) however are important for demonstrating that the changes are correct therein, as the `dials` tests pass. 

Have also identified a lot of cases where the pattern:

```
for j, expt in enumerate(experiments):
  # ignore j
```

is used so those become cleaner. Also found cases where `j` in the above example is clobbered by a loop variable as well.

### DIFF
--- a/algorithms/refinement/refiner.py
+++ b/algorithms/refinement/refiner.py
@@ -196,7 +196,7 @@ def _trim_scans_to_observations(experiments, reflections):
         else:
             obs_start = int(min_exp_z)
             obs_stop = int(math.ceil(max_exp_z))
-            half_deg_in_images = int(math.ceil(0.5 / exp.scan.get_oscillation()[1]))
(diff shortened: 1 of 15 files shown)


graeme-winter:
Motivation for this:

If I want to make the changes necessary to make #1029 possible it will mean that some experiments will not have crystals - if this is the case then it further means you cannot simply enumerate over the experiment which are present and have crystals as these will not be a contiguous list. 

However - if we also want to split things up and process separately we will also not want to enumerate -> this. Having the "working `id`" be consistent across a run seems ... sensible.

graeme-winter:
Better solutions than this exist

PR #1696, opened 2021-05-10
Limit Sphinx to <4 to fix documentation build

ndevenish:
The brand new release doesn't work with parts of the documentation build. This fixes the version to remain on 3.x, so the documentation builds again.


Anthchirp (approved):
Lgtm. Any indication how much work upgrading to v4 involves?

PR #1698, opened 2021-05-11
Move tests into tests directory

graeme-winter:
Kept the structure the same as far as possible. Had to fix references in flight. I think all the tests should pass, will run regression tests in a mo. 

Motivation: consistency


Anthchirp:
If you move the tests into `/tests` with the aim of eventually removing them from the python package, then `__init__.py`s should not be moved along, which will also prevent the files from becoming importable. If the tests have import dependencies between them then they need to be rewritten/moved back later, and then it'd be easier to just leave them where they are.

graeme-winter:
Hmmm would seem there is a lot of weird crosstalk between tests (i.e. there are tests which are very not self-contained and imported bits of other tests from other folders e.g. 

```
――――――― ERROR collecting tests/command_line/test_search_beam_position.py ―――――――
ImportError while importing test module '/Users/graeme/git/dials/modules/dials/tests/command_line/test_search_beam_position.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../conda_base/python.app/Contents/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/command_line/test_search_beam_position.py:11: in <module>
    from dials.algorithms.indexing.test_index import run_indexing
E   ModuleNotFoundError: No module named 'dials.algorithms.indexing.test_index'
```

which weirdly _does not cause test failure_)

Anthchirp:
minimal command to reproduce the issue:
```
pytest doc/examples/test_boilerplate.py::test_boilerplate tests/algorithms/indexing/basis_vector_search/test_combinations.py::test_filter_known_symmetry_no_matches --regression
```

Anthchirp:
`xfel_regression` refers to `from dials.test.(...) import ...` in `tst_ensemble_refinement.py`, so that is the likely reason for the `x_r` failure.

But this PR is ready for merging.

phyy-nx:
I've patched xfel_regression to refer to the new location in tst_ensemble_refinement.py.

Anthchirp:
Thanks!

PR #1702, opened 2021-05-13
Update test to read scaled data

graeme-winter:
Note that this now fails.

Following up on #1700 suggestion that this does what we need.

```
Column background.dispersion not written to file
Column background.mse not written to file
Column background.sum.value not written to file
Column background.sum.variance not written to file
Column intensity.scale.value not written to file
Column intensity.scale.variance not written to file
Column inverse_scale_factor not written to file
Column inverse_scale_factor_variance not written to file
Column original_index not written to file
Column phi not written to file
Column s1 not written to file
Column zeta not written to file
```

also

```
        # Construct the crystal model
>       crystal = Crystal(real_space_a, real_space_b, real_space_c, space_group_symbol)
E       RuntimeError: cctbx Error: Space group symbol not recognized:  I 2 2 3

/Users/graeme/git/dials/modules/dials/util/nexus/nx_mx.py
```


phyy-nx:
I added 2 commits.  One fixes up reading Crystal objects in nx_mx.py.  The other adds to nx_reflections.py the ability to read and write arbitrarily named columns that are not in the NXreflections specification.  With these changes, test/util/test_nexus/test_nexus.py now passes using scaled data.

jbeilstenedmands:
Looks like this PR was ready to go after @phyy-nx 's commits, so I have updated and opened to trigger the full test suite and will merge in a few days unless there are objections.

PR #1705, opened 2021-05-14
Remove unneeded slicing and imageset creation in spotfinding, indexing, and integration

### LINKED ISSUE #1704: Extra data reads in dials.stills_process
Profiling dials.stills_process on a single image (from an hdf5 dataset, but it shouldn't matter) showed that image was being read 2 times instead of only once and being cached (verified by a print statement in dataset_as_flex in nexus.py).  In the case of the XFEL GUI, which adds a radial_average step to check on the solvent content in the data, the image was read 3 times, instead of only once and being cached.  I believe this is because the experiment.imageset is being replaced, which negates the cache.  This is a big hit on the file system when running at scale.

Note if the imageset is sliced (eg imageset[x:x+1]), and then the sliced imageset is saved on the experiment, that would also reset the cache since the sliced imageset is a new imageset.  I bet that's what's actually happening.

To debug this, I put a print statement in the imageset.h constructor and checked to see where it was hit.  This constructor should only be hit once for each image, but it's hit 3 times after the process_experiments method of dials.stills_process starts.  Below, I've put the logs for the 3 times:

Spotfinding:

```
Finding spots in image 1 to 1...
DEBUGCONSTRUCTOR2
Setting chunksize=1
```

Indexing:

```
Indexing Strong Spots
********************************************************************************
DEBUGCONSTRUCTOR2
Using max_cell: 137.3 Angstrom
```

Integrate:

```
Using multiprocessing with 1 parallel job(s)
DEBUGCONSTRUCTOR2
 Frames: 0 -> 1
```

I hope this can be fixed by tracking down the places where its slices the imageset.  For processing stills, it shouldn't be sliced because dials.stills_process pre-slices everything down to 1 image.

I'll keep digging.

phyy-nx:
Fixes #1704


algorithms/indexing/indexer.py
```diff
@@ -412,9 +412,10 @@ def from_parameters(
                 from dxtbx.imageset import ImageSet  # , MemImageSet
 
                 for experiment in experiments:
-                    experiment.imageset = ImageSet(
-                        experiment.imageset.data(), experiment.imageset.indices()
-                    )
+                    if isinstance(experiment.imageset, ImageSequence):
+                        experiment.imageset = ImageSet(
+                            experiment.imageset.data(), experiment.imageset.indices()
+                        )
```
rjgildea, line 418: It isn't clear to me why indexing is even creating imagesets in the first place... it appears to only be doing this in the stills indexer, so presumably you might have a better idea why this is the case?

algorithms/integration/processor.py
```diff
@@ -499,7 +499,8 @@ def __call__(self):
 
             # I am 99% sure this is implied by all the code above
             assert (frame1 - frame0) <= len(imageset)
-            imageset = imageset[frame0:frame1]
+            if len(imageset) > 1:
+                imageset = imageset[frame0:frame1]
```
rjgildea, line 503: It seems slightly concerning that this apparently results in an extra image read - is this the case for rotation data, or just still imagesets?

algorithms/spot_finding/finder.py
```diff
@@ -821,7 +821,10 @@ def _find_spots_in_imageset(self, imageset):
 
             logger.info("\nFinding spots in image %s to %s...", j0, j1)
             j0 -= 1
-            r, h = extract_spots(imageset[j0:j1])
+            if len(imageset) == 1:
+                r, h = extract_spots(imageset)
+            else:
+                r, h = extract_spots(imageset[j0:j1])
```
rjgildea, line 827: As above - it would probably be instructive to understand _why_ this results in an extra image read.

algorithms/indexing/indexer.py
phyy-nx: Right, so in `dials.stills_process`, [imageset_from_anyset](https://github.com/cctbx/dxtbx/blob/main/imageset.py#L564) is [called](https://github.com/dials/dials/blob/main/command_line/stills_process.py#L297) on all images right at the start.  So by the time the indexing is ran, there's no way to have an ImageSequence.

However, if `dials.index` is ran, with `indexing.stills.indexer=stills`, then it's possible an ImageSequence makes it this far, even though the user wishes it to be treated as a still.  Therefore, the ImageSequence is reset as an ImageSet, and things like the StillsReflectionPredictor are [properly picked](https://github.com/dials/dials/blob/main/algorithms/spot_prediction/reflection_predictor.py#L88).

In the `dials.stills_process` case, before this changeset, ImageSets were always recreated, which invalidated the cache and caused a data read later on.  With this changeset, they are never recreated, because of the imageset_from_anyset call.

A data read wouldn't be triggered in the `dials.index` case because the data is never read in that program, so the cache doesn't matter.

algorithms/integration/processor.py
phyy-nx: I didn't profile `dials.integrate`, but I don't think it has the same problem.  I believe profile fitting is done from the strong shoeboxes, and data reading then only happens during integration itself.

In `dials.stills_process`, the same imageset is handed from program to program and so it cannot be reset.  In the `dials.integrate` it is invoked fresh and used once.  I believe :)

algorithms/spot_finding/finder.py
phyy-nx: See above, but in this case it's because of the extra radial_average step we do in the XFEL GUI.  That causes a data read, so when spotfinding happens and the imageset is sliced, the cache is lost.  In this changeset, the same imageset is used at each step, preventing the cache loss.

phyy-nx:
Thanks for the great questions @rjgildea.  Happy to clarify further or even provide the cProfile plots as needed.

algorithms/indexing/indexer.py
rjgildea: I see. It seems a bit odd that refinement should care at all about the imageset - it _should_ only depend on the models and input parameters. Perhaps a better (at least longer term) solution would be for refinement not to depend on the imageset type when choosing which predictor to use, then constructing a new imageset in indexing shouldn't be necessary at all.

phyy-nx: Totally agree.  My googling indicates that picking the StillsReflectionPredictor is the only place where ImageSequence is tested for explicitly, and I think it should instead be choosing the predictor based on if the scan is a set of stills:

Instead of 

```
if isinstance(experiment.imageset, ImageSequence):
```

It should be

```
if experiment.scan and not experiment.scan.is_still():
```

Like it is elsewhere.

But that seemed like a bigger change than this fix warranted?  Maybe not?  If folks like that better I can add it to this PR.

graeme-winter: TBH I think this is not a question of "like or dislike" - I think very shortly this will become necessary for sane processing of grid scan data beyond simply spot finding. As such, I'd probably suggest we put together a second PR where we actively eliminate 

```
if isinstance(experiment.imageset, ImageSequence):
```

type checks from the code base - this would be work which was started back in the load before heat death days but never (yet) completed.

algorithms/integration/processor.py
graeme-winter: Profile fitting is _not_ just done from the strong shoeboxes - for rotation data they are only used to determine the profile parameters - all reflections are then gathered to generate the reference profiles. 

We should probably check whether the rest is called for `dials.integrate` on rotation data as a separate exercise and raise an issue if so.

graeme-winter:
Questions raised above suggest we should probably have a better collective insight into _what is going on_ 🙂

algorithms/indexing/indexer.py
graeme-winter: On a practical note @phyy-nx  if you could put a comment in there as to _why_ this check is in place would probably help for the medium term.

newsfragments/1705.bugfix
```diff
@@ -0,0 +1 @@
+Speed up processing still images by reducing the number of file reads per image
```
graeme-winter, line 1: `dials.stills_process`: improve processing performance by preventing re-reading of image data

?

graeme-winter (approved):
There are questions raised in here which are bigger in scope than the PR so seems unfair to ask them to be addressed in here. 

Overall, I think the changes are limited in their scope & _hopefully_ we will be revisiting the slightly questionable bits (checking if image sets are sequences) so I am not too worried.

phyy-nx:
Any other comments?  I'll do the other ImageSequence PR after this goes through.

phyy-nx:
Updates:

1. I was wrong about there only being one instance of checking for ImageSequence. It's done all over the place.  Not worth chasing down at the moment or near future, IMO.
2. I profiled dials.integrate using 540 cbfs (the thaumatin 540 dataset).  I ran dials.import, dials.find_spots, dials.index, dials.refine, all using default arguments.  I then ran this:

```
libtbx.python -m cProfile -o integrate_540_nproc1.prof `libtbx.find_in_repositories dials`/command_line/integrate.py refined.* nproc=1
```

If you run without nproc=1, parallel map isn't profiled in such a way that you can see what's going on with get_raw_data.  However, with nproc=1, it's clear that get_raw_data is called 1080 times, twice per image.

Folks interested in a second issue on this?

phyy-nx:
Find spots is fine, 540 calls to get_raw_data:

```
libtbx.python -m cProfile -o find_spots_540_nproc1.prof `libtbx.find_in_repositories dials`/command_line/find_spots.py imported.expt nproc=1
```

phyy-nx:
More on dials.integrate.  As I mentioned, get_raw_data is called twice per image, but that's because the reflection profile modeler reads each image in turn, and then the integrator reads each image in turn.  Since only one image at a time is cached, there's no way to avoid 2 data reads as currently implemented.

Unless the profile modeler is creating shoeboxes that are then discarded, and then recreated.  In which case that does seem like a waste.  The second read shouldn't be needed.

(for stills it doesn't matter since we don't use the profile modeler)

(moving on from this, FYI :)

graeme-winter:
> More on dials.integrate. As I mentioned, get_raw_data is called twice per image, but that's because the reflection profile modeler reads each image in turn, and then the integrator reads each image in turn. Since only one image at a time is cached, there's no way to avoid 2 data reads as currently implemented.
> 

yup:

https://github.com/dials/dials/pull/1705#discussion_r634056320

> Unless the profile modeler is creating shoeboxes that are then discarded, and then recreated. In which case that does seem like a waste. The second read shouldn't be needed.

too much memory - already this is very heavy on memory just storing the boxes for the "live" block

> (for stills it doesn't matter since we don't use the profile modeler)
> 
> (moving on from this, FYI :)

yup, fair

PR #1713, opened 2021-05-18
Move location of test data for kapton integration testing

### LINKED ISSUE #1712: move integration/kapton test data into dials/data
This dials test is not run during as part of the main tests, as `xfel_regression` is not available at this point:
```
SKIPPED [1] test/algorithms/integration/test_kapton.py:20: test requires xfel_regression
```
It is also not run as part of the PR builds where `xfel_regression` is available, as it is not part of the `xfel_regression` test suite.
So we never run this test here.

I would suggest that either the test is moved into `xfel_regression`, or the test data is moved into [dials-data](https://github.com/dials/data). At ~6.6 MB the latter shouldn't really be an issue, all it really needs is a description file.

_Originally posted by @Anthchirp in https://github.com/dials/dials/issues/1707#issuecomment-842067617_

phyy-nx:
Closes #1712


phyy-nx:
Draft PR because I think a new version of dials-data needs to be published in conda before the data can be downloaded from the new location?

Anthchirp:
Yes. That package built overnight and is now ready.

Anthchirp (approved):

PR #1716, opened 2021-05-19
`dials.image_viewer`: add option to display in the "best fit" frame

dagewa:
- This should be moved so that it is done once per detector, not every time an image is displayed
- It should also be made optional with a checkbox to control


dagewa:
Also, need to check if this breaks the beam centre and other overlays

graeme-winter:
> Also, need to check if this breaks the beam centre and other overlays

Checking now

graeme-winter:
@dagewa 

Screenshot 2021-05-19 at 16 11 26

Looking good!

graeme-winter:
2 theta > 90° though 

```
Grey-Area fix-image-viewer :( $ dials.image_viewer split_12.*

input {
  experiments = split_12.expt
  reflections = split_12.refl
}

  File "/Users/graeme/git/dials/build/../modules/dials/command_line/image_viewer.py", line 208, in <module>
    run()
  File "/Users/graeme/git/dials/conda_base/python.app/Contents/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/Users/graeme/git/dials/build/../modules/dials/command_line/image_viewer.py", line 204, in run
    show_image_viewer(params=params, reflections=reflections, experiments=experiments)
  File "/Users/graeme/git/dials/build/../modules/dials/command_line/image_viewer.py", line 169, in show_image_viewer
    wrapper.display(experiments=experiments, reflections=reflections)
  File "/Users/graeme/git/dials/modules/dials/util/image_viewer/spotfinder_wrap.py", line 81, in display
    self.frame.load_image(chooser_wrapper(imagesets[0], 0))
  File "/Users/graeme/git/dials/modules/dials/util/image_viewer/spotfinder_frame.py", line 591, in load_image
    super().load_image(
  File "/Users/graeme/git/dials/modules/dials/util/image_viewer/slip_viewer/frame.py", line 423, in load_image
    self.pyslip.tiles.set_image(
  File "/Users/graeme/git/dials/modules/dials/util/image_viewer/slip_viewer/tile_generation.py", line 306, in set_image
    self.flex_image = get_flex_image_multipanel(
  File "/Users/graeme/git/dials/modules/dials/util/image_viewer/slip_viewer/tile_generation.py", line 116, in get_flex_image_multipanel
    beam_center /= npanels / 1e-3
  File "/Users/graeme/git/dials/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in __truediv__
    return rec([e/other for e in self.elems], self.n)
  File "/Users/graeme/git/dials/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in <listcomp>
    return rec([e/other for e in self.elems], self.n)
ZeroDivisionError: Please report this error to dials-support@lists.sourceforge.net: float division by zero
```

graeme-winter:
Looking at that one now

graeme-winter:
OK, beam centre nonsense is baked in everywhere which causes massive failures if the two-theta angle is > 90 - because _there is no beam centre_

However this current state is a long way along the road to fixing. But the rest of the road is a rocky and pothole strewn mudbath, so may be hard going.

phyy-nx:
I re-ran my rotation tests from previous attempts to fix this.  Here are the commands again:

```
dials.import `libtbx.find_in_repositories dials_regression`/image_examples/ALS_831/q315r_lyso_001.img output.experiments=singlepanel.expt
dials.import `libtbx.find_in_repositories dials_regression`/image_examples/LCLS_cspad_nexus/idx-20130301060858401.cbf output.experiments=multipanel.expt
libtbx.python `libtbx.find_in_repositories dials`/util/image_viewer/slip_viewer/rotate_detector.py singlepanel.expt
mv rotated.expt singlepanel_rotated.expt
libtbx.python `libtbx.find_in_repositories dials`/util/image_viewer/slip_viewer/rotate_detector.py multipanel.expt
mv rotated.expt multipanel_rotated.expt
```

Then use dials.image_viewer on each of the .expt files in turn to see what happens. On both master and this project_2d branch, both single and multi panel images appear rotated correctly.

👍

dagewa:
I completed additional work as per the first comment.

It might be a bit messy still. For instance, I tried setting the default to `"image"` rather than `"lab"`, but it still initially displays the `"lab"` form. Maybe this is because the calculation is not properly performed until the option is toggled at least once? Anyway, I leave that to anyone with more idea about the image viewer code to look at.

The whole image may shift between the two modes. This is a minor annoyance I don't have time to look at. Finally, as @graeme-winter points out, there are still issues at high angle, which are not due to the best-frame projection but due to problems calculating beam centre.

dagewa:
Demo:
rotated-cspad

phyy-nx (approved):
LGTM, but I don't have the 2theta tests.  On my rotation tests things look good.

graeme-winter:
@dagewa taking a quick look at the default

graeme-winter:
Some confusion in the code...

```
        self.settings.project_onto = self.params.project_onto
```

will set it as a string value however 

```
            self.settings.project_onto = self.projection_ctrl.GetSelection()
```

later on assigns as the _index_ of the selection. I do not think that this feature is unique to this new option.

graeme-winter:
> Some confusion in the code...
> 
> ```
>         self.settings.project_onto = self.params.project_onto
> ```
> 
> will set it as a string value however
> 
> ```
>             self.settings.project_onto = self.projection_ctrl.GetSelection()
> ```
> 
> later on assigns as the _index_ of the selection. I do not think that this feature is unique to this new option.

Fortunately this is never actually _read_ anyway 🙄

dagewa:
Yes, I admit to getting confused by this while copy-and-pasting, but I could not resolve in the time frame. I would welcome anyone with a better idea of what is going on in the image viewer code to tidy up, but I fear starting down that route might end up like an episode of Hoarder Homes.

graeme-winter:
> Yes, I admit to getting confused by this while copy-and-pasting, but I could not resolve in the time frame. I would welcome anyone with a better idea of what is going on in the image viewer code to tidy up, but I fear starting down that route might end up like an episode of Hoarder Homes.

Am looking

rjgildea:
The issue is that the attribute `detector.projected_2d` is only set as a consequence of calling `frame.update_settings()`:
https://github.com/dials/dials/blob/a8050e232eb8c363d69eb7809715be09ca357d03/util/image_viewer/slip_viewer/frame.py#L497
however this is only called _after_ the image is generated here:
https://github.com/dials/dials/blob/a8050e232eb8c363d69eb7809715be09ca357d03/util/image_viewer/slip_viewer/frame.py#L423-L431
Unfortunately the method can't be called _before_ this line, as it appears to depend on attributes defined as part of generating the image...

rjgildea:
The reason it can't be called before the image is created, is because this PR explicitly adds an attribute to `self.pyslip.tiles.raw_image.get_detector()` which does not exist until `self.pyslip.tiles.set_image()` has been called:
https://github.com/dials/dials/pull/1716/files#diff-3b34f595b35ea14a7777f614778f9da3c90d11f059de766d5d7d3c4d3ec9ff41R1018-R1025

rjgildea:
Right, so what happens is we call [`SpotFrame.load_image()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/spotfinder_frame.py#L555) which calls [`XrayFrame.load_image()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/slip_viewer/frame.py#L384) which calls [`SpotFrame.update_settings()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/spotfinder_frame.py#L1008) which is the method that sets the `detector.projected_2d` attribute. However, `update_settings()` is only called _after_ calling [`self.pyslip.tiles.set_image `](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/slip_viewer/frame.py#L423-L431) which is what results in `self.pyslip.tiles.raw_image` being set, which is required by `update_settings()`, meaning that this can't be called before setting the image.

graeme-winter:
> Right, so what happens is we call [`SpotFrame.load_image()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/spotfinder_frame.py#L555) which calls [`XrayFrame.load_image()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/slip_viewer/frame.py#L384) which calls [`SpotFrame.update_settings()`](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/spotfinder_frame.py#L1008) which is the method that sets the `detector.projected_2d` attribute. However, `update_settings()` is only called _after_ calling [`self.pyslip.tiles.set_image `](https://github.com/dials/dials/blob/d96ecfcc766aa9363b0a22ac9678016c1bd88a54/util/image_viewer/slip_viewer/frame.py#L423-L431) which is what results in `self.pyslip.tiles.raw_image` being set, which is required by `update_settings()`, meaning that this can't be called before setting the image.

So... if we want this to work the only sane option is to evaluate it as part of the image loading. That seems a long way from ideal.

graeme-winter:
I hold the opinion that though this change set is not perfect, the image viewer is a _lot better_ with the changes in than out -> am minded to suggest we merge in it's current form and add a new issue about the annoying behaviour. 

I welcome other perspectives.

command_line/image_viewer.py
```diff
@@ -32,6 +32,8 @@
   .type = int
 color_scheme = *grayscale rainbow heatmap invert
   .type = choice
+project_onto = *lab image
```
graeme-winter, line 35: Probably I would prefer `projection` than `project_onto`

newsfragments/1716.feature
```diff
@@ -0,0 +1 @@
+``dials.image_viewer``: Add an option to display in a best-fit "image" coordinate frame.
```
graeme-winter, line 1: Probably worth @ndevenish poking at this a little to get some kind of "snapped to grid" idea in there. I have no inspiration otherwise would do myself. Worth making this readable by the general reader.

util/image_viewer/slip_viewer/frame.py
```diff
@@ -524,70 +524,7 @@ def get_key(self, file_name_or_data):
             return super().get_key(file_name_or_data)
 
     def update_settings(self, layout=True):
-        # XXX The zoom level from the settings panel are not taken into
```
graeme-winter, line 527: @rjgildea improving his -ve line count, approve! 

This aliasing did completely confuse me while I was trying to change the default, so appreciate it being removed. 

That's a couple of haunted trees felled anyway.

util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -114,6 +114,14 @@ def get_flex_image_multipanel(
             pass
     beam_center /= npanels / 1e-3
 
+    # Calculate 2D origin, fast and slow vectors for detector projected onto
+    # a best fit frame. FIXME this should be done once for the detector, not
+    # here for every image we want to display!
+    try:
```
graeme-winter, line 120: EAFP is OK, but wonder if this is a little harder to read than just a "hasattr" or even setting the attribute to `None` when the object is created (which would involve a little more exploration but helps to leave breadcrumbs for a longer term improvement)

dagewa: oh the `FIXME` comment is out of date too btw. It came from the first commit where `project_2d` was called every time this function run rather than checking the detector for pre-calculated results. I did do this in a rush yesterday...

dagewa: `panels` is a misnomer here, AFAICT. It is actually a `Detector` object. We _could_ assign this attribute when the experiments are first loaded in, to ensure it is there whenever it matters.

An alternative is what I suggested at the bottom of https://github.com/cctbx/dxtbx/pull/224#issue-494737806, which is that the `project_2d` function gets moved to the `Detector` object itself. Then any detector can tell a viewer how to lay out its panels in 2D. This would have the advantage of providing a method that could be overridden in special cases, like that of the P12M, avoiding the special case code here in the image viewer.

dagewa: hmm, but actually it would probably have to be in the `Format`, because I don't think a `Detector` knows what detector it is...

Anthchirp: until then: 
`projected_axes = getattr(panels, "projected_2d", None)`
replaces the 4 lines

util/image_viewer/spotfinder_frame.py
```diff
@@ -1014,6 +1015,15 @@ def update_settings(self, layout=True):
         ):
             self.pyslip.tiles.update_brightness(new_brightness, new_color_scheme)
 
+        detector = self.pyslip.tiles.raw_image.get_detector()
+        if self.params.project_onto == "image":
+            if not hasattr(detector, "projected_2d"):
+                detector.projected_2d = project_2d(detector)
+            elif detector.projected_2d is None:
+                detector.projected_2d = project_2d(detector)
```
Anthchirp, line 1023:
```suggestion
            if not getattr(detector, "projected_2d", None):
                detector.projected_2d = project_2d(detector)
```

graeme-winter: ☝️  this is kinda what I was getting at 🙂

I appreciate I am not always clear though

newsfragments/1716.feature
dagewa: Indeed, should also now say that the option is the new default

graeme-winter:
Following the latest change set @dagewa  something is looking good - imported / integrated multi-two-theta offset scan with overlay of reflections:

Screenshot 2021-05-21 at 09 57 59
Screenshot 2021-05-21 at 09 58 05
Screenshot 2021-05-21 at 09 58 15
Screenshot 2021-05-21 at 09 58 22
Screenshot 2021-05-21 at 09 58 27

bonus points - for reasons that are opaque no longer crashes for two theta > 90° 🤔  🤷‍♂️  👍

dagewa:
I am rather puzzled by the fact that the fix fixes more than I thought it fixed :thinking:

PR #1722, opened 2021-05-24
Deprecate `preserve_exception_message` in `dials.util.mp`

Anthchirp:
I believe there exists no sensible scenario in which this would be set to `False`.


util/mp.py
```diff
@@ -53,7 +54,7 @@ def parallel_map(
     asynchronous=True,
     callback=None,
     preserve_order=True,
-    preserve_exception_message=True,
+    preserve_exception_message=...,
```
graeme-winter, line 57: TIL

util/mp.py
```diff
@@ -63,6 +64,13 @@ def parallel_map(
     """
     from dials.util.cluster_map import cluster_map as drmaa_parallel_map
 
+    if preserve_exception_message is not Ellipsis:
```
graeme-winter, line 67: OK, this is a really nice use of a NIL value which is not truth-like.

util/mp.py
```diff
@@ -83,7 +91,7 @@ def parallel_map(
             qsub_command=qsub_command,
             asynchronous=asynchronous,
             preserve_order=preserve_order,
-            preserve_exception_message=preserve_exception_message,
+            preserve_exception_message=True,
```
graeme-winter, line 94: Shouldn't it be removed from `drmaa_parallel_map` while we're there?

graeme-winter (approved):
Sane and sensible - I cannot imagine a use case where not preserving the exception makes things better.

PR #1724, opened 2021-05-24
Windows parallelism

### LINKED ISSUE #1721: Enable parallel processing on windows
Currently we rely on `easy_mp` which works only on UNIX-like systems - propose -

- [ ] assess desired usage patterns
- [ ] simplify content of `dials.util.mp`
- [ ] unify all multiprocessing to use `dials.util.mp` interface
- [ ] remove use of easy_mp
- [ ] replace with `concurrent.futures`

#discuss

graeme-winter:
Allow parallel processing for `dials.find_spots` and `dials.integrate` on Windows. 

Currently draft while I evaluate the performance impact. 

Fixes #1721


graeme-winter:
Spot finding - 500 image Eiger 2 XE 16M data set, from i03:
- single process on `main` took 438s wall clock (added `print()` with `time.time()` around `run()`)
- using all threads (12) took 210s 

So: much longer than it should have taken but still faster, suspect that the penalty for spawn multiprocessing is really killing us here (i.e. would not surprise me if we are making a copy of the process for _every image_). Will evaluate integration in a mo.

graeme-winter:
Integration - same 500 image data set:
- `main` - 392s
- `branch` - 159s over 12 processes

So doing _something_

graeme-winter:
This is a long way from perfect, but it (i) works and (ii) is faster than before so I think a net improvement. Much scope for making parallelism _better_.

dagewa:
It is remarkable that this is so simple. I've just started looking at the old I04 "BAG training" data (Pilatus miniCBFs).

Spotfinding before:
```
 138.7: Saved 126786 reflections to strong.refl
```
after (log claims `nproc=20`):
```
  54.3: Saved 126786 reflections to strong.refl
```

Integration before:
```
299.6: Saving the experiments to integrated.expt
```
after (log claims `nproc=20`):
```
183.8: Saving the experiments to integrated.expt
```

so definitely _an improvement_.

dagewa:
I would like to look at a non-HDF5 image stack (`.mrc` format for example) to see if there are any file-locking issues on read. I need to copy the data to the right machine and set up the jobs, so will take a while.

dagewa:
It worked with an `.mrc` image stack, though performance improvement was nothing special.

In the particular case I tried (180 images) spotfinding went from `37.2` to `38.3` seconds so was not actually faster using multiprocessing. Integration went from `89.6` to `69.2` seconds, so there was at least a slight improvement there.

ndevenish:
As long as it works no worse, which it looks like it does, it seems okay to put this in.

dagewa (approved):
This works for me :+1:. Performance improvement in the cases I looked at was limited, but that is out of scope here. It's great to see that we don't have to force `nproc=1` on Windows!

Issue #1719, opened 2021-05-21
Strange result from `dials.background`


dagewa:
The attached [background.zip](https://github.com/dials/dials/files/6522783/background.zip) contains an ED image with high background due to low angle scatter. I was trying to look at this with `dials.background` but I got a strange result.
```
dials.generate_mask image_001.mrc mask.phil
dials.import image_001.mrc mask=pixels.mask slow_fast_beam_centre=1017,1080
dials.background imported.expt d_max=12 d_min=3 | grep "Mean"
dials.background imported.expt d_max=10 d_min=3 | grep "Mean"
dials.background imported.expt d_max=8 d_min=3 | grep "Mean"
```
I expected the first `dials.background` job to give the highest mean background value, because it contains more of the low angle scatter, and the last job to give the lowest value, but it is actually the other way round:
```
fcx32934@nedvedice:/tmp/background$ dials.background imported.expt d_max=12 d_min=3 | grep "Mean"
Mean background: 83.519
fcx32934@nedvedice:/tmp/background$ dials.background imported.expt d_max=10 d_min=3 | grep "Mean"
Mean background: 97.636
fcx32934@nedvedice:/tmp/background$ dials.background imported.expt d_max=8 d_min=3 | grep "Mean"
Mean background: 106.532
```
Am I missing something, or does `dials.background` have a bug?

dagewa:
The inner ring here is at 12 Å. 
![Screenshot from 2021-05-21 14-38-58](https://user-images.githubusercontent.com/11502083/119146188-4e591180-ba42-11eb-8cfc-6d390669be3a.png)

dagewa:
Try to figure out what is going on by plotting the background pixels.
```
    import matplotlib.pyplot as plt
    plt.imshow(background_pixels.as_numpy_array() * 1, cmap=plt.cm.gray)
```
`d_max=12 d_min=3`:
![background_pixels_12-3](https://user-images.githubusercontent.com/11502083/119334103-a0887580-bc82-11eb-8f0b-b21657523293.png)

`d_max=10 d_min=3`:
![background_pixels_10-3](https://user-images.githubusercontent.com/11502083/119334191-b72ecc80-bc82-11eb-85f6-8bb045571ebe.png)

d_max=8 d_min=3`:
![background_pixels_8-3](https://user-images.githubusercontent.com/11502083/119334200-bbf38080-bc82-11eb-8d31-54766e3b5f4a.png)

dagewa:
With
```
[...]

PR #1726, opened 2021-05-25
`dials.background` masking fixes

dagewa:
- Pre-calculated masks assigned to the `imageset` are no longer discarded
- Signal or background assignment for each pixel is done on the corrected image

With these changes the example in #1719 gives much more reasonable results:
```
fcx32934@nedvedice:~/Downloads/background$ dials.background imported.expt d_max=12 d_min=3 | grep "Mean"
Mean background: 306.010
fcx32934@nedvedice:~/Downloads/background$ dials.background imported.expt d_max=10 d_min=3 | grep "Mean"
Mean background: 295.407
fcx32934@nedvedice:~/Downloads/background$ dials.background imported.expt d_max=8 d_min=3 | grep "Mean"
Mean background: 279.802
```
where the background pixels for the final job are now located as shown in white:
background


dagewa:
I had to update a couple of tests for the new behaviour, but I think the PR is more the way I would expect the program to operate. I've requested review to make sure there is agreement on that.

PR #1728, opened 2021-05-26
Fix memory calculation for multiple sweep integration runs

### LINKED ISSUE #1727: dials.integrate fails if > 1 scan and later scans longer than earlier
This data:

https://zenodo.org/record/51405

first 4 scans - import all, find spots on all, index, refine, integrate ->

This error 

```
================================================================================

Integrating reflections

dials Internal Error: C:\Users\graeme\git\dials\modules\dials\algorithms\integration\boost_python\integration_integrator_ext.cc(243): DIALS_ASSERT(b[5] <= frame1) failure.
```

@ndevenish  blocks 3.5 release

graeme-winter:
Need to select the reflections for each experiment, not use all reflections for each image set. Fixes #1727.


algorithms/integration/integrator.py
```diff
@@ -1208,9 +1208,11 @@ def integrate(self):
         def _determine_max_memory_needed(experiments, reflections):
             max_needed = 0
 
-            for j, experiment in enumerate(experiments):
+            for experiment in experiments:
                 imageset = experiment.imageset
-                subset = reflections.select(reflections["id"] == j)
+                subset = reflections.select_on_experiment_identifiers(
```
graeme-winter, line 1213: 👍

jbeilstenedmands (approved):
Change makes sense for the bug, thanks for fixing.

I thought it would be useful to start using the experiment identifier selection method, even though your selection wasn't wrong. I haven't verified this extra change on the data in question but I'm hoping it's trivial for you to do that?

graeme-winter:
@jbeilstenedmands testing

jbeilstenedmands:
Having thought about this further, I think the proposed change set would not have correctly handled multi-lattice data, where there are several reflection tables associated with the imageset, but which get processed as one 'block' of data for integration. So I've updated to handle this too.

graeme-winter:
I note there are now a lot of test failures in here 🤔

graeme-winter:
> Oh, this is probably because a lot of the test data won't have any identifiers set. Let me think about this.

Ah that makes some sense - indicates that we should probably update the test data but also a quick `hasattr` workaround would probably do it, and fall back to the original enumeration?

algorithms/integration/integrator.py
```diff
@@ -1208,6 +1208,19 @@ def integrate(self):
         def _determine_max_memory_needed(experiments, reflections):
             max_needed = 0
             for imageset in experiments.imagesets():
+                # find all experiments belonging to that imageset, as each
+                # imageset is processed as a whole for integration.
+                if all(experiments.identifiers()):
+                    expt_ids = []
+                    for experiment in experiments:
+                        if experiment.imageset == imageset:
+                            expt_ids.append(experiment.identifier)
```
Anthchirp, line 1217:
```suggestion
                    expt_ids = [
                        experiment.identifier
                        for experiment in experiments:
                        if experiment.imageset == imageset
                    ]
```
just a suggestion: this feels like it should be a list comprehension (or a tuple or even just a generator)

PR #1733, opened 2021-06-02
Add hdf5plugin dependency

Anthchirp:
over the previously used (and abandoned) hdf5-external-filter-plugins

### DIFF
--- a/.conda-envs/README.MD
+++ b/.conda-envs/README.MD
@@ -1,6 +1,6 @@
 This directory contains 3 groups of files:
 
-DIALS bootstrap - Python 3.6-3.8
+DIALS bootstrap - Python 3.6-3.9
 ================================
 linux.txt
 macos.txt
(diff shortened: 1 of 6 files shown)


Anthchirp:
Successfully ran regression tests on a newly bootstrapped environment

Anthchirp:
If you have an existing DIALS development installation then you should follow these steps when convenient:
* Update the dxtbx and dials repositories
* Run 
  `conda remove hdf5-external-filter-plugins`
  and
  `conda install -c conda-forge hdf5plugin`
  to change the installed packages.
* This can then be followed by a 
  `make reconf`
  in the build directory to make sure the build is up-to-date.

Future bootstrapped environments will have these changes included automatically.

Anthchirp:
There is a chance that conda gets its packages in a twist during the remove step and will run for a very long time before offering to upgrade many packages (and install new ones). You can also use the command
`conda remove --force hdf5-external-filter-plugins hdf5-external-filter-plugins-bitshuffle hdf5-external-filter-plugins-bzip2 hdf5-external-filter-plugins-lz4` to remove the old plugins, which will not suffer from this issue.

PR #1735, opened 2021-06-08
Build a Python 3.9 environment by default

Anthchirp:
and update cctbx to 2021.5


graeme-winter:
Python 3.9 build worked fine for me on macOS 11.4 using system compilers (the ones in Conda no longer work quite right after an `xcode` update)

graeme-winter (approved):
Seems like a timely change

PR #1738, opened 2021-06-09
Thaumatin 2021 tutorial

### LINKED ISSUE #1737: Update DIALS "processing in detail" to 2021 Eiger data
Would seem sensible to bring this somewhat up to date

graeme-winter:
Work in progress but putting this here so people can see what I am up to and maybe even make helpful suggestions. This will be more lightweight / less comprehensive than the previous version. 

This will fix #1737

### DIFF
--- a/doc/sphinx/documentation/tutorials/processing_in_detail_thaumatin.rst
+++ b/doc/sphinx/documentation/tutorials/processing_in_detail_thaumatin.rst
@@ -339,53 +339,41 @@ damage.
 Integration
 ^^^^^^^^^^^
 
-After the refinement is done the next step is integration, which is performed
-by the program :doc:`dials.integrate <../programs/dials_integrate>`. Mostly,
-the default parameters are fine for Pilatus data, which will perform
-XDS-like 3D profile fitting while using a generalized linear model in order
-to fit a Poisson-distributed background model. We will also increase the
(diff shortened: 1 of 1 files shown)


rjgildea:
Just to point out we do have a slightly more up-to-date [tutorial using SARS-CoV-2 MPro data](https://dials.github.io/documentation/tutorials/mpro_x0692.html) from I04-1 (although obviously this is from a recently-retired PILATUS detector rather than an EIGER). I would recommend adding any new tutorial to the [`dials.generate_tutorial_text`](https://github.com/dials/dials/blob/main/command_line/generate_tutorial_text.py) command to ensure than the tutorial remains functional and the output current.

graeme-winter:
Totally valid suggestion which is beyond my reach and unlikely given my current deadline of resolving this by next week!

Also wanted to have a tutorial for well measured data...

rjgildea:
It shouldn't be _too_ difficult - see https://github.com/dials/dials/commit/f69fc319a4304ff54ed88805adb4af81b414a2bd for how it was done for the MPro tutorial. Should mostly be a copy/pasta job, right? :)

graeme-winter:
Closing for the moment - I think this may need to be done _properly_ so I will come up with something immediate for the needs for next week and revisit

PR #1750, opened 2021-06-17
Stack overlays

dagewa:
The image viewer code had an unused buffer to accumulate overlays across images -n to +n from the current image.

Here this is repurposed to accumulate overlays +n images, where n is the number of images in the stack. This helps tasks such as identifying dodgy reflections on images.

Before:
before

After:
after


dagewa:
Adding some usual image viewer critics to review :wink:

PR #1751, opened 2021-06-17
Numbers!

dagewa:
Excessive precision spotted while reading https://github.com/graeme-winter/dials_tutorials/blob/main/br-lyso-multi/multicrystal.md

BTW don't end a branch name in `!`. It does slightly break GitHub ;-)


dagewa:
@ndevenish you had suggestions regarding this?

PR #1754, opened 2021-06-22
Change output_unintegrated_reflections default

### LINKED ISSUE #1753: dials.integrate: change default for output of unintegrated reflections
Recent CCP4 workshop showed that memory use of downstream tasks in `dials.cosym` `dials.symmetry` and `dials.scale` caused issues for students. We should change default of `output_unintegrated_reflections` in `dials.integrate` from `True` to `False` as this option is mainly only useful for debugging and causes downstream programs to have problems. 

Example for a 360° cubic insulin data set (i.e. small unit cell etc.) - with current default:

```
Maximum resident set size (kbytes): 2391892
```

as measured by `/usr/bin/time -v` in `dials.symmetry` and 

```
Maximum resident set size (kbytes): 2283316
```

when the unintegrated reflections are _not_ included in the output. In this case it is onlt 5% of the data size, but in other cases it can be more significant - and most importantly this memory use is not necessary for performing the task in hand, only "nice to have" for debugging purposes (for which the input reflections are available anyway)

graeme-winter:
Setting output_unintegrated_reflections=False by default will reduce memory
usage for downstream steps while not changing the results of the processing.

Fixes #1753


graeme-winter:
OK, I have failing tests now to fix in here (moderately unsurprising) 

When fixing the tests I will try and make the tests actually test the correct behaviour.

dagewa (approved):

PR #1761, opened 2021-06-22
Remove cache of Kabsch debugging

### LINKED ISSUE #1734: `dials.image_viewer` does not update results of bottom buttons for series of '.expt'
When loading a series of _.expt_ in `dials.image_viewer` the buttons at the bottom (mean... threshold) don't seem to update results when moving through the images. 

If I use instead the raw files (`dials.image_viewer image_*.cbf`), the results of buttons at the bottom update as expected.

graeme-winter:
Was broken anyway for stills as image.index is always 0 so did not update when moving to next frame. 
 
Fixes #1734
    
Removed now redundant `request` dict as well


phyy-nx:
Wouldn't this re-open #1213? (@dagewa)

graeme-winter:
> Wouldn't this re-open #1213? (@dagewa)

Could be, but working and slow seems better to me than broken and quick

We know the image viewer as a whole needs speeding up 🙄

dagewa:
Yes, it would reopen #1213. What does change when moving through stills if not `image.index`? It seems this could be fixed less destructively if the right item is put in the `self._dispersion_debug_memo` dictionary.

dagewa:
Is there an example dataset available somewhere I can look at to demonstrate #1734?

graeme-winter:
> Yes, it would reopen #1213. What does change when moving through stills if not `image.index`? It seems this could be fixed less destructively if the right item is put in the `self._dispersion_debug_memo` dictionary.

Still process imported experiments each have length 1 imageset -> index always 0, you iterate through imagesets. I will come up with a less destructive fix

graeme-winter:
Reproduce: pick any sequence of still images, run through `dials.stills_process` then 

```
dials.image_viewer *imported.expt
```

_stills processing need not have worked_ as the data will still have been imported

graeme-winter:
@phyy-nx  @dagewa  I have restored the fix to #1213 but restructured the code rather than aiming for a minimal change. Motivation:

 - before I spent a good chunk of time trying to work out where `request` was sent - it was sent nowhere
 - computing a hash rather than a dictionary comparison seems more idiomatic for "what was intended"

FWIW simply including the `imageset` pointer in whatever the fix was _would_ have fixed the initial bug but would mean a future explorer in these woods would have had to retrace steps.

dagewa:
This looks like a better solution. I did wonder if we could use `functools.lru_cache` for this though, but I'm not sure if it would work with the type passed as `image` to `_calculate_dispersion_debug` (which might be something like a "`chooser_wrapper`")

graeme-winter:
@ndevenish  are the build errors related to PR or some external influence? I cannot tell 🤔

Anthchirp:
cctbx/cctbx_project#627 to blame

PR #1764, opened 2021-06-23
Add pre-commit checks for python comprehensions

Anthchirp:
This will now flag code such as
```
./tests/algorithms/spot_finding/test_per_image_analysis.py:64:16: C406 Unnecessary list literal - rewrite as a dict literal.
./tests/algorithms/spot_finding/test_per_image_analysis.py:89:16: C406 Unnecessary list literal - rewrite as a dict literal.
./tests/algorithms/image/filter/test_median.py:73:30: C414 Unnecessary list call within sorted().
./tests/algorithms/integration/profile/test_grid_sampler.py:47:13: C416 Unnecessary list comprehension - rewrite using list().
```
which have been fixed in this pull request.
Around 25 more remain.


.pre-commit-config.yaml
```diff
@@ -1,8 +1,8 @@
 repos:
 
 # Automatically sort imports
-- repo: https://github.com/PyCQA/isort.git
-  rev: 5.5.3
+- repo: https://github.com/PyCQA/isort
+  rev: 5.6.4
```
ndevenish, line 5: Was there a reason this was updated also, or just an opportunity to update? Why to 5.6.4?

ndevenish:
I've also gone through and updated the rest of the places where this warns. It looks like a good check. I also approve of consolidating configurations.

.pre-commit-config.yaml
Anthchirp: Just an upportunity update to the most recent version that supports pre-commit <2.9, and that is also already referenced in other repositories such as ISPyB, zocalo, workflows, dlstbx, ...

PR #1770, opened 2021-06-30
Improve nearest point labels in `dials.reciprocal_lattice_viewer`

dagewa:
Add the d-spacing and put ID and panel on one line.

Before:
before

After:
after


benjaminhwilliams:
Would it be helpful to include the resolution units? I mean, I know everyone uses Å, but it might just help to emphasise that 'res' stands for 'resolution'.

dagewa:
Unfortunately the font this thing uses won't display an Â properly
Screenshot from 2021-06-30 12-31-02

Anthchirp:
There's enough space to write out "Angstrom"

PR #1771, opened 2021-06-30
Show path in rlv title bar

dagewa:
When comparing `dials.reciprocal_lattice_viewer` windows side by side it could be difficult to keep track of which was which because the title bar just says "Reflection data viewer". This change makes it more similar to the `dials.image_viewer`, by printing a useful path up there (here the path to reflections)

Screenshot from 2021-06-30 16-11-43


dagewa:
Oops looks like I branched off #1770, so included those changes here too

Anthchirp:
`git rebase --onto master 002dc1835e71a5cbe649551c47361798b35d7fd1 && git push -f`
should now be the correct set of commits?

dagewa:
Looks right. Thanks for untangling.

ndevenish:
The test on previous commit are because it looks like the feature branch was inadvertantly rebased onto `master` instead of `main`. Merged in again.

Anthchirp:
> rebased onto master instead of main

:facepalm: sorry

ndevenish:
It's hard to train out that muscle memory sometimes :)

PR #1774, opened 2021-06-30
Image viewer brightness control

dagewa:
There may be different opinions about this, but I find the `dials.image_viewer` brightness to usually start too dark, to step too much on clicking in the scroll bar, and all the useful range being up the left end of the scale. This PR adjusts defaults to improve things a little. I welcome other views.


dagewa:
I think we can do better. I'll try a non-linear function when I get some time.

ndevenish:
image

ndevenish:
This seems to lose a lot of the controllability and maximum contrast. The selector (running via NX to diamond) seems only to "jump" when dragging between 1, 2, 6, 10, 14, 18, 22, 26, 30 making it darker, and a similar but slightly different range when dragging lighter. Fast-dragging frustratingly seems to allow spaces in-between.

In addition, it seems to lose a lot of the maximum contrast end:
image

I feel that it should definitely allow you to make it "Darker" than is useful, because then you can see the extreme ends and drag somewhere in the middle.

dagewa:
Thanks for the comments. I found the dark end of the scale fairly useless, so I cut it back earlier, but I can certainly extend it darker again. I'm not sure about the jumping behaviour. I don't think I saw that locally, but I'll investigate

ndevenish:
Actually, the behaviour that I'm observing is that it's not moving until I'm off the edge of the handle. This might be a specific linux/wx interaction problem


https://user-images.githubusercontent.com/529963/124601163-6bfe0100-de5f-11eb-84ef-0053a2ad4dd0.mp4

dagewa:
Ah right, actually I see the same. I had taken to clicking in the bar either side of the button, which steps by +/- 2

command_line/image_viewer.py
```diff
@@ -28,7 +28,7 @@
 
 phil_scope = iotbx.phil.parse(
     """\
-brightness = 100
+brightness = 20
```
graeme-winter, line 31: Approve of this change (though personally I would have set 10)

newsfragments/1774.feature
```diff
@@ -0,0 +1 @@
+``dials.image_viewer``: adjust brightness control for better defaults.
```
graeme-winter, line 1: Better is an opinion, would say they are certainly better for photon counting detectors

graeme-winter (changes requested):
I am not certain I fully agree with the changes in here - while the "slider" is integral it becomes very hard to find good values with the changes in this PR for a stack of 10 images

The newly nonlinear nature of the scaling is to me troublesome... but I am not sure what to suggest in its place

Please could you try it out yourself with a stack of 10 images then see how you feel about the use of dynamic range (hint: only the left hand extreme of the slider is useful)

graeme-winter:
> I am not certain I fully agree with the changes in here - while the "slider" is integral it becomes very hard to find good values with the changes in this PR for a stack of 10 images
> 
> The newly nonlinear nature of the scaling is to me troublesome... but I am not sure what to suggest in its place
> 
> Please could you try it out yourself with a stack of 10 images then see how you feel about the use of dynamic range (hint: only the left hand extreme of the slider is useful)

In fairness, this accusation should also be levelled at the existing implementation

dagewa:
It still looks a lot better to me :thinking:

**Main `stack=1`**
main-stack-1

**Branch `stack=1`**
branch-stack-1

Neither very much use, but I can at least start to see diffraction spots on the branch

**Main `stack=10`**
main-stack-10

**Branch `stack=10`**
branch-stack-10

The branch behaviour is _much_ better in this case. On main it's just a dark mess. On the branch I get a clean diffraction image.

dagewa:
@graeme-winter I've been discussing this with @ndevenish. I'm not able to reproduce anything concerning with the example I tried, so could you please provide an example to illuminate your request for changes? Thanks

ndevenish:
Since there's lack of consensus on this now we'll leave for this release.

PR #1784, opened 2021-07-02
Return `size_t` index arrays from `reflection_table.match`

dagewa:
I suspect the reason these were `flex.int` is because `AnnAdaptorSelfInclude.nn` is `flex.int`, so `mm` was made equivalent. I couldn't see a reason for `nn` to be a signed type though, so I changed both.

As a result, `reflection_tables` can be `select`ed directly with the returned arrays without requiring this conversion step outside of the `match` function.


array_family/flex_ext.py
```diff
@@ -645,8 +645,8 @@ def match(
         ann = AnnAdaptorSelfInclude(r, 3)
         ann.query(x)
 
-        mm = cctbx.array_family.flex.int(range(xyz.size()))
-        nn, distance = ann.nn, cctbx.array_family.flex.sqrt(ann.distances)
+        mm = cctbx.array_family.flex.size_t_range(xyz.size())
+        nn, distance = ann.nn.as_size_t(), cctbx.array_family.flex.sqrt(ann.distances)
```
graeme-winter, line 649: Thanks, makes much more sense - I am not sure why I did not do this before, but it is possible (as mentioned by @ndevenish ) that `flex.size_t` was not a thing when I added the original implementation

graeme-winter (approved):
Makes a lot of sense, clear improvement for very little cost -> win

PR #1785, opened 2021-07-02
Clear rings display in `dials.image_viewer`

### LINKED ISSUE #1777: unit cell tool in `dials.image_viewer` could use a clear rings button
Opening the unit cell tool after closing it and setting new unit cell parameters results in new rings over-imposed over old ones. There seems to be no way to clear the old rings?

dagewa:
Adds buttons to unit cell and generic ring display to allow clearing their display. Fixes #1777


elena-pascal (approved):
Looks good :smiley_cat:

PR #1787, opened 2021-07-05
Remove support for legacy versions of SciPy

benjaminhwilliams:
No longer accommodate older SciPy versions that lacked the `scipy.spatial.transform.Rotation.from_matrix` method.  That method was introduced in SciPy version 1.4, released on 16th December 2019, so it's long enough in the tooth that I don't think there's anything controversial in this change.


ndevenish (approved):

benjaminhwilliams:
@ndevenish, I'm not sure of the reason for [this test failure](https://dev.azure.com/azure-dials/dials/_build/results?buildId=4388&view=logs&j=3c9906cf-2496-5c56-e123-fc2f891a1e66&t=f3bcc04e-a9e0-58a3-5cc8-9a5529afdfb3&l=113), it seems unrelated to the changes here.

ndevenish:
I'm slightly confused by it. It doesn't look like an intermittent error, but definitely doesn't look related to the changes here.

Anthchirp:
No idea on the build error. I don't believe there is a need to set the minimum versions in the environment files though. Leaving it open would be just fine.

PR #1788, opened 2021-07-05
Improve documentation of goniometer rotation function

benjaminhwilliams:
Try to introduce a little clarity into the documentation of the `dials.command_line.anvil_correction.goniometer_rotation` function, with a view to making it easier to reuse elsewhere.


benjaminhwilliams:
Uh... @ndevenish, I don't think this got the review it needed from @jbeilstenedmands.

PR #1789, opened 2021-07-05
Add pybind11 to conda environments

ndevenish:
In order to merge https://github.com/cctbx/dxtbx/pull/377, we need to add pybind11 to the dials conda environment. This is because dxtbx doesn't have it's own, separate dependency list.

This does that.


Anthchirp (approved):
Looks good to me

PR #1803, opened 2021-07-12
Changes to spot finding to improve things for ED data

jmp1985:
This seems to improve the spotfinding for ED data somewhat.


ndevenish:
Doesn't that do the wrong thing with pedestal-subtracted data?

jmp1985:
Hi Nick. 

The spot finding algorithm assumes a Poisson distribution and looks at the variance / mean to see that it is equal to 1 (within some expected variance). If negative pixel values are included in the mean and variance calculation then typically the variance is unaffected bu the mean will be substantially closer to zero which has the effect of making the dispersion higher than expected. That can lead to background pixels and spot pixels being more difficult to distinguish. 

But having said that, yes, maybe David could take a look at some of his data to see if this pull request makes an improvement... or indeed makes things worse!

graeme-winter:
Surely as soon as you have a negative value for a Poissonian variant you've left the domain already right?

jmp1985:
Yes, I agree that statistically the assumptions are definitely violated by having negative values but these data are supposed to be from an electron counting detector so in theory should be well modelled by a Poisson. So my intention here was to make the algorithm work as well as possible with this data even if we think it's not strictly correct!

jmp1985:
> pot-finding for ED can be very much a trial-and-error process using t

Yes, I updated each class. I'm also taking a closer look at other aspects of spot finding to see if I can make it slightly better for ED and will update the pull request with my changes later. So maybe take a look tomorrow after I've updated it!

ndevenish:
>  typically the variance is unaffected bu the mean will be substantially closer to zero which has the effect of making the dispersion higher than expected

Isn't this just because your pedestal-subtraction level is uncertain? Surely if you truncate your offset poisson distribution it ceases to be a poisson distribution....

graeme-winter:
> > One thing that comes to mind though, does this update the KabschDebug... classes too? Spot-finding for ED can be very much a trial-and-error process using the dials.image_viewer. I think it is important that the spot-finding visualisations match what dials.find_spots will produce
> 
> Yes, I updated each class. I'm also taking a closer look at other aspects of spot finding to see if I can make it slightly better for ED and will update the pull request with my changes later. So maybe take a look tomorrow after I've updated it!

I would feel more comfortable if this was "better for ED _whilst making no difference for X-ray diffraction with photon counting detectors"

I do however note that this discussion probably applies equally to Jungfrau / integrating detector X-ray data collection

jmp1985:
I agree! Of course, if anything I change here makes things worse for X-ray then definitely don't merge the pull request! However, I would say that in the majority of cases for X-ray data with Pilatus/Eiger it will have no effect. Also, spot finding with CCD data has a similar issue so I suspect this will also make spot finding more robust for CCDs (although I acknowledge that this is an untested statement backed up by zero evidence).

dagewa:
Agree. The issue is with integrating detectors in general, which do not have a Poisson response anyway. "ED" is just an extreme manifestation of the issues, because there are some detector types in use that have calibration problems.

dagewa:
We might eventually end up with different algorithms for spot-finding with counting detectors vs spot-finding with integrating detectors. This would be appropriate and would mean we don't need to be looking for a compromise algorithm that spans both cases.

jmp1985:
To summarize the changes in the pull request:

1. Ignore negative pixel values in the mean and variance calculations
2. Make the erosion less severe before calculating mean of background
3. Use the original dispersion mask for calculating strong pixel mask

The spot finding algorithm works by:
1. computing the local mean and variance at each pixel
2. computing the index of dispersion at each pixel
3. calculating a dispersion mask where the variance  / mean is > 1.
4. the dispersion mask shows potential strong pixels in the neighbourhood
5. the dispersion mask is then eroded slightly
6. the mean is then calculated ignoring the potential strong pixels under eroded dispersion mask
7. the strong spot mask is then calculated by looking at pixels greater than the mean (which is now an esimate of the local background).

The rationale for these changes are as follows:

1. As per the discussion above, negative pixel values cause the assumptions of the spot finding algorithm to break down. Ignoring them seems to make the spot finding for ED data more robust.

2. During the background mean calculation step (6). We ideally only want
to consider pixels which are background. If we erode too far then we end
up using all the pixels which defeats the purpose of the second mean
calculation step.

3. During the final threshold step (7) if we use the eroded mask we have
fewer potential strong pixels to choose from. We are essentially
selecting only the peak if anything. It is probably better to use the
original uneroded dispersion mask and select strong pixels from within
that in the final step.

I've attached some figures illustrating the improvement for some ED data.

The diffraction image
Parameters used to do spot finding
The old spot finding results
The new spot finding results

jmp1985:
> We might eventually end up with different algorithms for spot-finding with counting detectors vs spot-finding with integrating detectors. This would be appropriate and would mean we don't need to be looking for a compromise algorithm that spans both cases.

Sure, this is probably a good idea in the long run.

graeme-winter:
Discussion in team meeting right now we would collectively like this to go in as a separate experimental algorithm which can be switched to, so that we do not change the default behaviour and can actively investigate with real data.

ndevenish:
Since this needs reworking, and I don't know if you want that to be here or in a separate PR, I'm degrading this to a draft PR

tests/algorithms/image/threshold/test_local.py
```diff
@@ -227,8 +229,45 @@ def test_dispersion_extended_threshold(self):
         result4 = debug.final_mask()
         assert result2 == result4
 
+    def test_dispersion_electron_threshold(self):
+        from dials.algorithms.image.threshold import (
+            DispersionElectronThreshold,
+            DispersionElectronThresholdDebug,
+        )
```
Anthchirp, line 236: This is already done at the top level?

tests/algorithms/image/threshold/test_local.py
```diff
@@ -227,8 +229,45 @@ def test_dispersion_extended_threshold(self):
...
+        from dials.array_family import flex
```
Anthchirp, line 237: this clobbers the scitbx import. Do you need both? Probably better to float this and add `as dials_flex` or something

tests/algorithms/image/threshold/test_local.py
```diff
@@ -227,8 +229,45 @@ def test_dispersion_extended_threshold(self):
...
+            self.mask,
+            self.gain,
+            self.size,
+            nsig_b,
+            nsig_s,
+            0,
+            self.min_count,
+        )
+        result4 = debug.final_mask()
+        assert result2 == result4
```
Anthchirp, line 267: Slightly more descriptive names than result[1234] might give the casual reader more insight what this test is actually testing. (This could also be encoded in the test name)

Anthchirp (commented):
Quick review without any knowledge about the background to the PR

dagewa:
> Hi, sorry for taking a while to update this pull request. I've not added the modified algorithm as a new algorithm and added a test etc.

Thanks James. It seems this isn't quite ready for review yet. The entry point needs declaring explicitly in `libtbx_refresh.py`, but even after that the code does not run because 
```
from dials.interfaces.dispersion_extended_spotfinder_threshold_ext import (
    estimate_global_threshold,
)
```
results in
```
ModuleNotFoundError: No module named 'dials.interfaces'
```
Are you intending to push some more commits to this branch?

ndevenish:
It looks like it still has the changes to `DispersionThreshold` that we discussed not doing?

I'm confused how you managed to test this locally without the changes David has needed?

algorithms/spot_finding/threshold.py
```diff
@@ -155,3 +155,72 @@ def __call__(self, image, mask):
...
+    A class implementing a 'gain' threshold.
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Set the threshold algorithm up
+        """
+
+        # Initialise the base class
+        ThresholdStrategy.__init__(self, **kwargs)
```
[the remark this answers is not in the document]
jmp1985: Now changed to use super().__init__(**kwargs)

algorithms/spot_finding/threshold.py
```diff
@@ -155,3 +155,72 @@ def __call__(self, image, mask):
...
+        :param mask: The mask to use
+        :return: The thresholded image
+        """
+        from dials.algorithms.image import threshold
+        from dials.array_family import flex
+
+        # Initialise the algorithm
+        try:
+            algorithm = self.algorithm[image.all()]
+        except Exception:
```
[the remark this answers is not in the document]
jmp1985: Changed to check for presence of image.all() in self.algorithm dict

tests/algorithms/image/threshold/test_local.py
jmp1985: Now using top level imports

tests/algorithms/image/threshold/test_local.py
jmp1985: Removed the flex import. The test seems to pass the same anyway

jmp1985:
> > Hi, sorry for taking a while to update this pull request. I've not added the modified algorithm as a new algorithm and added a test etc.
> 
> Thanks @jmp1985. It seems this isn't quite ready for review yet. The entry point needs declaring explicitly in `libtbx_refresh.py`, but even after that the code does not run because
> 
> ```
> from dials.interfaces.dispersion_extended_spotfinder_threshold_ext import (
>     estimate_global_threshold,
> )
> ```
> 
> results in
> 
> ```
> ModuleNotFoundError: No module named 'dials.interfaces'
> ```
> 
> Are you intending to push some more commits to this branch?

This is now fixed, tests/algorithms/image/threshold/test_local.py passes and I've verified that the algorithm is selectable on the command line.

dagewa:
Thanks @jmp1985. I will review soon. One comment, reading through the discussion above there may be justification for applying this method for integrating detectors in general, perhaps particularly for difficult cases with unknown bias. ED is out most obvious manifestation of that, but the same could surely be true for X-rays in some circumstances. I'm just wondering if `dispersion_electron` is the most appropriate name?

dagewa (changes requested):
Thanks for adding this as a separate algorithm and hooking it up to the image viewer, which makes it very easy to compare the thresholding results from the different algorithms. I'm happy for this to go in, but I am held up on the name `dispersion_electron` (see other comments). Is there a name that suggests more what is _does_ rather than what it should be used _for_? I think we would end up regretting `dispersion_electron`, so better fix it now.

dagewa:
👍 thanks

dagewa (approved):
I've tried this out on some more datasets and do see an improvement in some cases. It is very useful to have an extra tool in the kit and I appreciate the name change. I'm approving now, though I have not yet checked why the CI is erroring

dagewa:
None of the failures look relevant, so I've requested a re-run

ndevenish:
Otherwise, I don't currently understand where these test failures are coming from

src/dials/extensions/dispersion_zero_truncated_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,68 @@
...
+
+from dials.algorithms.spot_finding.threshold import (
+    DispersionZeroTruncatedThresholdStrategy,
+)
+from dials.extensions.dispersion_extended_spotfinder_threshold_ext import (
+    estimate_global_threshold,
+)
+
+logger = logging.getLogger(
+    "dials.extensions.dispersion_zero_truncated_spotfinder_threshold_ext"
```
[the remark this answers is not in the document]
jmp1985: I've added a commit to make this change now.

dagewa:
Hi @jmp1985, 5 tests currently fail on this branch, even if I merge in the latest main:
```
Results (449.94s):
    1328 passed
       5 failed
         - tests/test_scan_varying_integration_bug.py:10 test
         - tests/command_line/test_background.py:6 test
         - tests/command_line/test_export_best.py:6 test_export_best
         - tests/command_line/test_find_bad_pixels.py:6 test_find_bad_pixels
         - tests/command_line/test_stills_process.py:117 test_sacla_h5[False]
``` 
Do you have any idea why?

dagewa:
My understanding is that @jmp1985 does not want to merge this. The future in ED is with better detectors after all. Feel free to re-open if my understanding is incorrect.

PR #1807, opened 2021-07-14
Correct spelling erros

ndevenish:
Ran [codespell](https://github.com/codespell-project/codespell): 

    codespell -w --skip "*.js,*.html" -L hist -L als -L homogenous -L ser

And then manually "git add -p" to verify changes. There were a few false positives and places where it tried to change variable names, which I ignored.


util/options.py
```diff
@@ -438,7 +438,7 @@ def parse_args(
         # Parse the command line phil parameters
         user_phils = []
         unhandled = []
-        interpretor = self.system_phil.command_line_argument_interpreter()
+        interpreter = self.system_phil.command_line_argument_interpreter()
```
Anthchirp, line 441: static analysis found fault

ndevenish: Darn, I removed it but it snuck back in

PR #1809, opened 2021-07-14
Include DANO/SIGDANO in anomalous merged.mtz

rjgildea:
This is required by some downstream phasing software, eg. autoSHARP.


rjgildea:
Thanks for fixing the test @jbeilstenedmands

jbeilstenedmands (approved):

PR #1813, opened 2021-07-20
Numpy/scipy implementation of calc_crystal_frame_vectors

rjgildea:


algorithms/scaling/scaling_utilities.py
```diff
@@ -6,7 +6,9 @@
 import logging
 from math import acos
 
-import scitbx
+import numpy as np
+from scipy.spatial.transform import Rotation
```
elena-pascal, line 10: FWIW scipy also has [vector.ReferenceFrame](https://docs.sympy.org/latest/modules/physics/vector/api/classes.html), which is neat because yone just define reference frames with respect to one another and then call the vector in the wanted reference frame, and it does quaternion coordinate transform in the background.

rjgildea: That appears to be `SymPy` (which I hadn't come across before) rather than `SciPy`...

algorithms/scaling/scaling_utilities.py
```diff
@@ -72,31 +74,35 @@ def calc_crystal_frame_vectors(reflection_table, experiment):
...
+    non_zero = np.where(lengths > 0.0)
+    sel_s1 = s1[non_zero]
+    sel_z = reflection_table["xyzobs.px.value"].parts()[2].as_numpy_array()[non_zero]
+    s1n = sel_s1 / lengths[non_zero][:, np.newaxis]
+    phi = experiment.scan.get_angle_from_array_index(
+        flex.double(sel_z), deg=False
+    ).as_numpy_array()
+    rotation_matrix = Rotation.from_rotvec(phi[:, np.newaxis] * rotation_axis)
+    R_inv = (setting_rotation * rotation_matrix * fixed_rotation).inv().as_matrix()
+    s0c[non_zero] = R_inv @ s0
```
elena-pascal, line 101: yey for vectorization :sparkles:

algorithms/scaling/scaling_utilities.py
```diff
@@ -72,31 +74,35 @@ def calc_crystal_frame_vectors(reflection_table, experiment):
...
+    s1c[non_zero] = np.einsum("...ij,...j", R_inv, s1n)
```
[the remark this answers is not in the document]
elena-pascal: You should also be able to use `Rotation.apply`. It can apply multiple rotations to multiple vectors. [Each rotation is applied on the corresponding vector](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.apply.html)

algorithms/scaling/scaling_utilities.py
elena-pascal: Ah fair

algorithms/scaling/scaling_utilities.py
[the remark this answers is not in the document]
elena-pascal: oh, `Rotation.apply` does  `np.einsum('ijk,ik->ij', matrix, vectors)`:joy:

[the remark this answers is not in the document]
toastisme: Agree that using standard numpy multiplication is easier to read, but probably because I've not used `einsum` enough to find the syntax intuitive... Might be worth at least having comment explaining what is being done at 102

algorithms/scaling/scaling_utilities.py
toastisme: Might as well use `beam.get_sample_to_source_direction` at 85 if you're explicitly saying that's what you want.

toastisme (approved):
Code is clear and easy to understand. All tests involving `calc_crystal_frame_vectors` pass for me, with the same behaviour as `scaling_absoprtion_model`. Probably not relevant here as it's not likely to be a bottleneck, but for swapping out flex for numpy in general it might be useful to profile things, just to see what is being gained/lost in terms of speed.

jbeilstenedmands:
Here's a script to time the differences:
```import timeit
from dials_scaling_ext import calc_crystal_frame_vectors as calc_vectors_cpp
from dials.algorithms.scaling.scaling_utilities import calc_crystal_frame_vectors as calc_vectors_np
from dials.util.options import OptionParser, reflections_and_experiments_from_files
from typing import List
import functools

def run(args: List[str] = None) -> None:
    parser = OptionParser(
        usage="",
        read_experiments=True,
        read_reflections=True,
        check_format=False,
        phil=None,
    )
    params, options = parser.parse_args(args=args, show_diff_phil=False)

    reflections, experiments = reflections_and_experiments_from_files(
        params.input.reflections, params.input.experiments
    )
    reflections = reflections[0].split_by_experiment_id()

    for refl, expt in zip(reflections, experiments):
        print("C++ timings:")
        print(timeit.timeit(functools.partial(calc_vectors_cpp, refl, expt), number=10))
        print("Numpy timings:")
        print(timeit.timeit(functools.partial(calc_vectors_np, refl, expt), number=10))
if __name__ == "__main__":
    run()

Running this on betalactamase, (total time for 10 repeats): 
C++ timings:
0.8630936289999998
Numpy timings:
24.748112079000002

And on a 4-sweep I19 dataset (total time for 10 repeats):
C++ timings:
0.5192671039999999
Numpy timings:
15.083686385
C++ timings:
0.3737126879999977
Numpy timings:
9.062704091
C++ timings:
0.32305729200000144
Numpy timings:
9.151707523000002
C++ timings:
0.32131162099999955
Numpy timings:
9.040262024999997

So the numpy version is about 30x slower. I haven't profiled where this time is spent yet.

**Comment by rjgildea** [comment]
> Here's a script to time the differences:
> 
> ```
> from dials_scaling_ext import calc_crystal_frame_vectors as calc_vectors_cpp
> from dials.algorithms.scaling.scaling_utilities import calc_crystal_frame_vectors as calc_vectors_np
> from dials.util.options import OptionParser, reflections_and_experiments_from_files
> from typing import List
> import functools
> 
> def run(args: List[str] = None) -> None:
>     parser = OptionParser(
>         usage="",
>         read_experiments=True,
>         read_reflections=True,
>         check_format=False,
>         phil=None,
>     )
>     params, options = parser.parse_args(args=args, show_diff_phil=False)
> 
>     reflections, experiments = reflections_and_experiments_from_files(
>         params.input.reflections, params.input.experiments
>     )
>     reflections = reflections[0].split_by_experiment_id()
> 
>     for refl, expt in zip(reflections, experiments):
>         print("C++ timings:")
>         print(timeit.timeit(functools.partial(calc_vectors_cpp, refl, expt), number=10))
>         print("Numpy timings:")
>         print(timeit.timeit(functools.partial(calc_vectors_np, refl, expt), number=10))
> if __name__ == "__main__":
>     run()
> 
> Running this on betalactamase, (total time for 10 repeats): 
> C++ timings:
> 0.8630936289999998
> Numpy timings:
> 24.748112079000002
> 
> And on a 4-sweep I19 dataset (total time for 10 repeats):
> C++ timings:
> 0.5192671039999999
> Numpy timings:
> 15.083686385
> C++ timings:
> 0.3737126879999977
> Numpy timings:
> 9.062704091
> C++ timings:
> 0.32305729200000144
> Numpy timings:
> 9.151707523000002
> C++ timings:
> 0.32131162099999955
> Numpy timings:
> 9.040262024999997
> 
> So the numpy version is about 30x slower. I haven't profiled where this time is spent yet.
> ```

Thanks. That looks like there must be something wrong with the numpy version as I struggle to believe it should be that much slower than the C++ version. What where the input files to this script? I.e. are they in dials.data, or somewhere on disk at DLS?

**Comment by jbeilstenedmands** [comment]
That was just me testing on some local data. For a more reproducible test, try 
`python time_calc_crystal_vectors.py ~/dials/build/dials_data/l_cysteine_4_sweeps_scaled/scaled_20_25.*`
```
C++ timings:
0.010114565000000297
Numpy timings:
0.31883611599999995
C++ timings:
0.015045736999999892
Numpy timings:
0.45266281600000013
```
Based on what the function is doing, I expect the time will scale linearly with n_reflections, with a constant timing ratio ~30.

**Comment by rjgildea** [comment]
Profiling with [pprofile](https://github.com/vpelletier/pprofile) suggests that nearly 80% of the runtime is spent constructing a `flex.vec3_double` from the equivalent numpy array:
```
   105|        20|     0.237518|    0.0118759| 39.98%|    reflection_table["s0c"] = flex.vec3_double(s0c)
   106|        20|     0.232622|    0.0116311| 39.16%|    reflection_table["s1c"] = flex.vec3_double(s1c)
```

Using:
```
from dials.algorithms.scaling.scaling_utilities import calc_crystal_frame_vectors as calc_vectors_np
from dials.util.options import OptionParser, reflections_and_experiments_from_files
from typing import List

import pprofile

def run(args: List[str] = None) -> None:
    parser = OptionParser(
        usage="",
        read_experiments=True,
        read_reflections=True,
        check_format=False,
        phil=None,
    )
    params, options = parser.parse_args(args=args, show_diff_phil=False)

    reflections, experiments = reflections_and_experiments_from_files(
        params.input.reflections, params.input.experiments
    )
    reflections = reflections[0].split_by_experiment_id()

    prof = pprofile.Profile()
    with prof():
        for refl, expt in zip(reflections, experiments):
            for i in range(10):
                calc_vectors_np(refl, expt)
    prof.dump_stats("np.prof")
if __name__ == "__main__":
    run()
```

```
$ dials.python tst.py $(dials.data get -q l_cysteine_4_sweeps_scaled)/scaled_20_25.*
```

**Comment by rjgildea** [comment]
With 8f47dc0 the time difference is now down to less than a factor of 3. Nearly half of the runtime appears to be spent in the `np.linalg.inv(R)` function...
```
$ dials.python time_calc_crystal_vectors.py $(dials.data get -q l_cysteine_4_sweeps_scaled)/scaled_20_25.*
C++ timings:
0.019707225263118744
Numpy timings:
0.05824899673461914
C++ timings:
0.01879800856113434
Numpy timings:
0.05769295245409012

```

**Comment by rjgildea** [comment]
With the current implementation in this PR the time difference is a factor of ~1.5x for the numpy version vs the C++ version. Approximately half of the time appears to be spent in the line `R = setting_rotation @ rotation_matrix @ fixed_rotation`.
```
$ dials.python time_calc_crystal_vectors.py $(dials.data get -q l_cysteine_4_sweeps_scaled)/scaled_20_25.*
C++ timings:
0.020216181874275208
Numpy timings:
0.03371530398726463
C++ timings:
0.020444795489311218
Numpy timings:
0.032798103988170624
```
@jbeilstenedmands do you think this performance is acceptable? I presume either version is still a lot faster than the original python loop using `scitbx.matrix`...

jbeilstenedmands:
I'm a little undecided on this one. If I also adopt changing the inverse to transpose, then that speeds up the c++ too, I'm seeing a difference of 2.5x-3x between the two branches. To try and see if this is ever really going to be a problem, I found a dataset with 3.7M reflections, the runtime per-call of the function for me was ~1.0s vs ~2.9s, this is on a job which takes 270s overall, so only 1% of the overall runtime is in this function...

ndevenish:
After discussion with @jbeilstenedmands, we've decided that the difference is negligible enough, and value of using scipy enough - that we're happy for this to be merged in.

PR #1818, opened 2021-07-21
Fix image viewer for load_models=False

phyy-nx:
Load and hdf5 file with load_models=False and get this exception

```
  File "/dev/shm/aaron/modules/dials/util/image_viewer/spotfinder_frame.py", line 104, in __init__
    detector.projected_2d = None
AttributeError: Please report this error to dials-support@lists.sourceforge.net: 'NoneType' object has no attribute 'projected_2d'
```

Detector can be none if load_models=False, which is necessary for big hdf5 files and xtc streams.


dagewa (approved):
Looks fine, I wasn't aware of that use case before.

PR #1836, opened 2021-08-04
Fix a format-truncation compiler warning.

nksauter:
gcc compiler produces a warning: 'snprintf' output truncated before the last format character,
and a note: 'snprintf' output 16 bytes into a destination of size 15.
The fix is to adjust the format string to 14 whitespaces, allowing the
15th position to be used for null-termination.

### DIFF
--- a/newsfragments/1836.misc
+++ b/newsfragments/1836.misc
@@ -1,0 +1,1 @@
+Fix a compiler warning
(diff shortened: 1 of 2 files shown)


ndevenish:
It looks like this line possibly isn't necessary at all. Or at least `ostringstream` would be cleaner and less confusing.

However, I'm pretty sure this is the reflection viewer and basically disused, so not really worth understanding what the routine is trying to do in the first place.

Issue #438, opened 2017-09-12
dials.reciprocal_lattice_viewer: beam centre control stops working


dagewa:
It seems possible to get `dials.reciprocal_lattice_viewer` into a state in which changing the beam centre controls no longer has any effect. I think this may occur after a requested change causes an exception, but following that it is impossible to recover - typing the correct values in the beam centre controls does nothing.

To demonstrate, pick any dataset and enter some crazy value for the beam centre (1000 mm in X for the i04-BAG-training data for example). In other cases much less dramatic changes are sufficient to cause the problem

dagewa:
Here is an example in which smaller changes trigger the issue: [demo.zip](https://github.com/dials/dials/files/1295633/demo.zip). Changing the beam centre in X from 13.66 mm to 14.0 mm is enough.

dagewa:
Just checking this is still an issue (it is).

dagewa:
Fixed by 26a7341

PR #1842, opened 2021-08-12
Beam centre control

dagewa:
Fix long-standing bug #438 by providing a new beam centre control that will not set impossible beam centres, yet allows setting beam centre on any panel for multi-panel detectors.


dagewa:
I was mulling this over and I think the beam centre control does the wrong thing, in that it shifts the `s0` direction rather than shifting the detector. As a result the results are different to setting the beam centre on import. I may try to rework this to shift the detector instead. Leaving this open as it is an improvement over the current behaviour (fixes a bug), but I may supplant it with a new PR.

PR #1843, opened 2021-08-12
Check if template was given a `#` token.

### LINKED ISSUE #1840: `dials.import` with template could fail more gracefully when `#` is absent
`dials.import template=0p85_5s_0001.mrc` for instance fails with unfriendly traceback

DIALS 3.dev.531-g4e2b2bd3b

input {
  template = "0p85_5s_0001.mrc"
}

  File "/home/elena/Dials/build/../modules/dials/command_line/dials_import.py", line 951, in <module>
    run()
  File "/home/elena/Dials/conda_base/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/elena/Dials/build/../modules/dials/command_line/dials_import.py", line 947, in run
    importer.import_image(args)
  File "/home/elena/Dials/build/../modules/dials/command_line/dials_import.py", line 806, in import_image
    imagesets = _extract_or_read_imagesets(params)
  File "/home/elena/Dials/build/../modules/dials/command_line/dials_import.py", line 220, in _extract_or_read_imagesets
    experiments = ExperimentListFactory.from_templates(
  File "/home/elena/Dials/modules/dxtbx/src/dxtbx/model/experiment_list.py", line 836, in from_templates
    raise ValueError(f"Template '{template}' does not match any files")
ValueError: Please report this error to dials-support@lists.sourceforge.net: Template '0p85_5s_0001.mrc' does not match any files

elena-pascal:
Look for `#` in all instances in list (?) of templates, if not found fail politely. Trying to fix #1840


elena-pascal:
I though this was impossibly straightforward, but then I found out template is a list. So not really sure if this is indeed the expected behaviour.

ndevenish:
🤔  Rather than error out in this case, how about making it so a filename without "#" is still a valid template? I mean, if I passed in a single filename as a template, I'd expect it to at least match that one file

elena-pascal:
Should a template without `#` then just be treated as an expt file?

ndevenish:
It feels more literal behaviour to me, instead of ugly. Ask for something of shape: `/path/to/file` - get something literally in that shape - `/path/to/file`. Maybe worth printing a warning but I can see an argument that it would be useful for e.g. -> "I don't know if I got a filename or a template from the user, just give it to `template=` and both work"

ndevenish:
Since the template is in principle just expanded to a list of filenames, a string with no "#" _must_ be expanding in some erroneous way otherwise it'd still match. So I'd start tracing back where that calculation is done.

ndevenish:
Don't worry about the windows build errors btw, it's a problem upstream in https://github.com/cctbx/cctbx_project/issues/639

ndevenish (approved):
We discussed the possible improvements to this handling, and decided that it's better to just get this QOL improvement in now rather than trying to Fix All The Things.

PR #1846, opened 2021-08-13
dials.stills_process performance improvements

dwpaley:
* Speed up Kapton correction by returning once two Kapton faces intersecting the beam have been found.
* Filter out rare cases where sigma_b (a spot shape parameter) is nonphysically large. This avoids integration hanging in the Kapton correction step when millions of pixels are inside predicted reflections.

Co-authored-by: Aaron Brewster <asbrewster@lbl.gov>
Co-authored-by: Asmit Bhowmick <abhowmick@lbl.gov>

### DIFF
--- a/algorithms/integration/boost_python/kapton_ext.cc
+++ b/algorithms/integration/boost_python/kapton_ext.cc
@@ -51,6 +51,9 @@ scitbx::af::shared<double> get_kapton_path_cpp(
       } catch (dxtbx::error) {
         // Do nothing
       }
+      if (intersection_xy_list.size() == 2) {
+        break;
+      }
     }  // kapton faces
     /*
      * Now find out the maximum path length through the kapton. That is the path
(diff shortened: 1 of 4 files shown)


phyy-nx:
One of the performance tweaks here throws out images with sigma b > 0.1 degrees.  This is because the kapton correction code churns on these data for images with big unit cells and lots of pixels to integrate with sigma b is large.

Here is a histogram of sigma b values from nearly 600K images from a recent XFEL experiment across several different samples.  The two peaks come from two different binning sizes of a Rayonix detector.

image

We are close enough to the 0.1 degree cutoff in this PR that I'd rather move it to a phil parameter instead of a hard coded parameter.  Maybe profile.gaussian_rs.parameters.sigma_b_cutoff = None?  We'd override to 0.1 in dials.stills_process.  Problem with this is it would only be checked in dials.stills_process, but the parameter is in a dials.integrate phil scope...  Might be ok though?

phyy-nx:
Moved to stills_process-specific phil param.  Will merge tomorrow unless there are any comments.  Thanks.

PR #1851, opened 2021-08-17
Image viewer 2d projection

toastisme:
Changes to image_viewer scripts to match [#422](https://github.com/cctbx/dxtbx/pull/422), removing specific edge case behaviour for detectors with unique projections.


util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -58,7 +59,6 @@ def get_flex_image_multipanel(
 
     from iotbx.detectors import generic_flex_image
     from libtbx.test_utils import approx_equal
-    from xfel.cftbx.detector.metrology import get_projection_matrix
```
dagewa, line 61: Ah, I see this moved from DIALS to dxtbx in https://github.com/cctbx/dxtbx/pull/422/. As per comment there, it would be preferable for dxtbx to not depend on xfel, so we can work towards a separately distributable dxtbx

util/image_viewer/slip_viewer/tile_generation.py
```diff
@@ -89,9 +89,9 @@ def get_flex_image_multipanel(
             assert approx_equal(saturation, panel.get_trusted_range()[1])
     assert saturation is not None
 
-    # Create rawdata and my_flex_image before populating it.
+    # Create rawdata and flex_image_multipanel before populating it.
     rawdata = flex.double(flex.grid(len(detector) * data_padded[0], data_padded[1]))
-    my_flex_image = generic_flex_image(
+    flex_image_multipanel = generic_flex_image(
```
dagewa, line 94: :+1: the name here bothered me every time I came across it

dagewa (changes requested):
I got some test failures (though have not looked why yet)
```
       8 failed
         - tests/command_line/test_export_bitmaps.py:6 test_export_single_bitmap
         - tests/command_line/test_export_bitmaps.py:18 test_export_multiple_bitmaps
         - tests/command_line/test_export_bitmaps.py:38 test_export_bitmap_with_prefix_and_no_padding
         - tests/command_line/test_export_bitmaps.py:52 test_export_bitmap_with_prefix_and_extra_padding
         - tests/command_line/test_export_bitmaps.py:66 test_export_bitmap_with_specified_output_filename
         - tests/command_line/test_export_bitmaps.py:94 test_export_still_image
         - tests/command_line/test_export_bitmaps.py:107 test_export_multi_panel
         - tests/command_line/test_export_bitmaps.py:127 test_export_restricted_multiimage
```

util/image_viewer/spotfinder_frame.py
```diff
@@ -103,15 +103,13 @@ def __init__(self, *args, **kwds):
...
-                    195,
-                ):
+                if detector.has_projection_2d():
                     self.params.projection = None
+                    continue
                 else:
-                    detector.projected_2d = project_2d(detector)
+                    detector.projection_2d_axes = get_detector_projection_2d_axes(
+                        detector
+                    )
```
rjgildea, line 112: Rather than set an attribute on the detector that is then used later, could we just move the code from `tile_generation.py` that calls `get_panel_projection_2d_from_axes` here and explicitly set the projections in the panel objects. And then the code in `tile_generation.py` can simply look up the projections and use them directly. That way the logic is in a single place, rather than spread over two files.

rjgildea: Although I realise the exact projection matrices depend on whether or not the projection is in the lab or image frame. Does this particular point in the code get run again if you've changed the projection mode in the GUI?

rjgildea (approved):
As mentioned in https://github.com/cctbx/dxtbx/pull/422 this will be a useful change to remove the hardcoded handling of special cases such as the P12M 👍

util/image_viewer/spotfinder_frame.py
toastisme: Had a look at this, and this code does only gets called on startup. I agree it would be cleaner to isolate where projections are set, but it's a bit fiddly in its current state as `get_panel_projection_2d_from_axes` requires image data. I'm also not entirely clear on the benefits of using image projection vs lab. Bit hesitant to extend the surface area of this much further so it might be better tidying up the logic as a separate PR.

PR #1855, opened 2021-08-19
Basis vector colours

dagewa:
While poking around inside the image viewer for another branch, I noticed that when there are multiple experiments, the crystal basis vectors were always shown with one colour, whereas the predictions cycle through a set of colours. This PR changes the basis vector display to match the prediction colours.

**Before**
main

**After**
branch


newsfragments/1855.feature
```diff
@@ -0,0 +1 @@
+``dials.image_viewer`` the basis vector colours of multiple crystals now match the prediction colours.
```
rjgildea, line 1:
```suggestion
``dials.image_viewer``: The basis vector colours of multiple crystals now match the prediction colours.
```

util/image_viewer/spotfinder_frame.py
```diff
@@ -1693,7 +1693,11 @@ def map_coords(x, y, p):
                             continue
                         x, y = detector[panel].get_ray_intersection_px(s1)
                         x, y = map_coords(x, y, panel)
-                        vector_data.append((((beam_x, beam_y), (x, y)), vector_dict))
+                        basis_vector_dict = dict(vector_dict)
+                        basis_vector_dict["color"] = self.prediction_colours[i_expt]
+                        vector_data.append(
+                            (((beam_x, beam_y), (x, y)), basis_vector_dict)
+                        )
```
rjgildea, line 1700: I think this should work, without the need to explicitly create a modified copy of `vector_dict`:
```suggestion
                        vector_data.append(
                            (((beam_x, beam_y), (x, y)), {**vector_dict, "color": self.prediction_colours[i_expt]}),
                        )
```

rjgildea (approved):
Seems like a sensible change 👍

util/image_viewer/spotfinder_frame.py
dagewa: :+1: I'll commit and try it

PR #1863, opened 2021-08-25
User messages for missing files

toastisme:
This adds a warning for dials.import if no input files are found when using a wildcard, and an error message if a specific requested file is missing.

Currently the behaviour when using wildcards is to show help if no files are found, or when using multiple wildcards to show nothing if some wildcards return nothing. This adds a warning to the user if a wildcard returns nothing.

The current behaviour for when a specific file is not found is to show the traceback from dxtbx. This pr hides the traceback and shows just the file not found error, followed by calling help.


util/options.py
```diff
@@ -258,26 +261,35 @@ def try_read_experiments_from_images(
         for arg in args:
             # Don't expand wildcards if URI-style filename
             if "*" in arg and not get_url_scheme(arg):
-                args_new.extend(glob(arg))
+                filenames = glob(arg)
+                if not filenames:
+                    logger.warning("No files found with pattern %s" % arg)
```
Anthchirp, line 266: Suggest f-string you always want to expand, logger comma notation if not - there shouldn't be a need to ever use `%` in the context of logging.

This aside - shouldn't `arg` end up in `unhandled`?
Should this only be a warning or should it be a `sys.exit(...)`?

newsfragments/1863.bugfix
```diff
@@ -0,0 +1 @@
+Error when trying to import a missing file now handled more elegantly. A warning is now given when trying to import files using a wildcard and none are found.
```
rjgildea, line 1:
```suggestion
dials.import: More helpful error message when trying to import a missing file. A warning is now given when trying to import files using a wildcard and none are found.
```

util/options.py
```diff
@@ -258,26 +261,35 @@ def try_read_experiments_from_images(
...
+                else:
+                    args_new.extend(glob(arg))
```
rjgildea, line 268: No need to perform the `glob()` twice. Might be preferable to switch the order of the if/else statements to avoid the need for `if not`:
```suggestion
                if filenames:
                    args_new.extend(filenames)
                else:
                    logger.warning(f"No files found with pattern {arg}")
```

util/options.py
```diff
@@ -258,26 +261,35 @@ def try_read_experiments_from_images(
...
+                scan_tolerance=scan_tolerance,
+                format_kwargs=format_kwargs,
+                load_models=load_models,
             )
+            if experiments:
+                self.experiments.append(
+                    FilenameDataWrapper(filename="<image files>", data=experiments)
+                )
+        except FileNotFoundError as e:
+            logger.error("File %s not found", e.filename)
```
rjgildea, line 291: Does the `FileNotFoundError` come from the `ExperimentListFactory.from_filenames` call or the `FilenameDataWrapper` call? If the former it might be slightly cleaner to do:
```suggestion

        try:
            experiments = ExperimentListFactory.from_filenames(
                args,
                unhandled=unhandled,
                compare_beam=compare_beam,
                compare_detector=compare_detector,
                compare_goniometer=compare_goniometer,
                scan_tolerance=scan_tolerance,
                format_kwargs=format_kwargs,
                load_models=load_models,
            )
        except FileNotFoundError as e:
            logger.error("File %s not found", e.filename)
        else:
            if experiments:
                self.experiments.append(
                    FilenameDataWrapper(filename="<image files>", data=experiments)
                )
```
Under what circumstances is a `FileNotFoundError` raise here?

rjgildea (changes requested):
Might be useful to add a simple test case to cover this (although possibly overkill)?

util/options.py
toastisme: I thought logger was an exception to using f-strings where it's [optimised](https://docs.python.org/3/howto/logging.html#optimization) for `%` style formatting.

I didn't use `sys.exit(...)` to allow the help to be called downstream, which is consistent with what happens when you try and open files from a single directory with a wild card.

rjgildea: To do this "properly" with logging it should be:
`logger.warning("No files found with pattern %s", arg)`
not
`logger.warning("No files found with pattern %s" % arg)`
i.e. you pass `arg` as an argument to the `logger.warning` function rather than using `%` string formatting.
See also https://docs.python.org/3/library/logging.html#logging.Logger.debug

This is only really advantage if `arg` is an object whose string representation is evaluated lazily (and relatively expensive). Otherwise you are probably best just using f-Strings now, i.e.
`logger.warning(f"No files found with pattern {arg}")`
(although opinions here may vary)

util/options.py
toastisme: Removed the `glob` call. 
I thought logging shouldn't should f strings - is this no longer the case?

> [internally the logging package uses %-formatting to merge the format string and the variable arguments](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles)

util/options.py
```diff
@@ -262,10 +262,10 @@ def try_read_experiments_from_images(
             # Don't expand wildcards if URI-style filename
             if "*" in arg and not get_url_scheme(arg):
                 filenames = glob(arg)
-                if not filenames:
-                    logger.warning("No files found with pattern %s" % arg)
+                if filenames:
+                    args_new.extend(filenames)
                 else:
-                    args_new.extend(glob(arg))
+                    logger.warning("No files found with pattern %s" % arg)
```
rjgildea, line 268: This is still the wrong way to use logging, as it should really either be:
`logger.warning("No files found with pattern %s", arg)` (which result in delayed formatting, which is only really needed if formatting is expensive)
or (my preference as f-String syntax is more modern/readable etc etc):
`logger.warning(f"No files found with pattern {arg}")`
Using `logger.warning("No files found with pattern %s" % arg)` gives the worst of both world as it uses the legacy `%` string formatting syntax, but without the potential advantages of delayed formatting within the logging library.

toastisme: Ah sorry I didn't realise there was an underlying difference between `%s", arg` and `%s" % arg`.

util/options.py
toastisme: Re: `arg` going into `unhandled`, I did I try this originally. But unhandled arguments here currently get raised at the options level, which just gives a `Sorry: Unable to handle the following arguments` with no explanation, which doesn't seem particularly helpful to the user. The IO should be being handled at the import (top) level really, but I assumed changing the behaviour of `parse_args` in `options` was a bit risky with the current quick_parse/ignore_unhandled/return_unhandled mix of behaviours...

toastisme:
> Might be useful to add a simple test case to cover this (although possibly overkill)?

Not quite sure what this would be testing if it's just logger statements based on `Glob` and a `FileNotFoundError`. I guess you could have a test to confirm that trying to import a non-existent file always gives a logger.error followed by calling help, and trying to import multiple wildcards where one returns nothing gives only a logger.warning. Does feel a bit overkill to me..

toastisme:
Sorry this had slipped my mind. Yes I agree the solution isn't ideal. It would be more consistent for missing files to be caught in unhandled. My original reason for not doing this was that it's not particularly helpful for the user to see an argument is unhandled but not know it's due to missing files. 

A better solution might changing unhandled arguments in general to be a dictionary which explain the reason for each unhandled argument (99% of which I'm guessing would just be unrecognised arguments).

util/options.py
ndevenish: As a minor note on this (admittedly old) conversation, logger is an exception where the logger-style formatting should _in theory_ be used, but I've long given up trying to force this since using f-strings in loggers is so prevalent.

ndevenish:
Even though I think we could do better here, this is absolutely an improvement to what we had before.

PR #1864, opened 2021-08-25
stills_indexer: Add a deterministic option for experiment identifiers

dwpaley:


dwpaley:
After https://github.com/cctbx/dxtbx/pull/438 we think this is no longer needed because we don't have to delete imagesets during `cctbx.xfel.merge` jobs. Thus we can always look at the imageset if we want to connect processed data to the original images, and no longer need to duplicate that information in the identifier.

PR #1866, opened 2021-08-27
Drop Python 3.7 support

Anthchirp:
As requested in the last core meeting here's some advance warning of dropping Python 3.7 support [as per DC3](https://dials.github.io/kb/proposals/dc3).

This change will take effect **after the release following the 26th of December 2021**.
This pull request will be updated once we can narrow this down further to a specific date range.

After this PR is merged [new language features such as `:=` assignment, positional-only parameters in functions, f-string `=`, etc...](https://docs.python.org/3/whatsnew/3.8.html) become available to developers.


dagewa:
CCP4 8.0 will have Python 3.7. The following release is planned for later in 2022 and will have Python 3.9.

__init__.py
```diff
@@ -1,8 +1,4 @@
 import logging
-import sys
-
-if sys.version_info.major == 2:
-    sys.exit("Python 2 is no longer supported")
```
graeme-winter, line 5: \o/

installer/bootstrap.py
```diff
@@ -1264,7 +1264,7 @@ def run():
         "--python",
         help="Install this minor version of Python (default: %(default)s)",
         default="3.9",
-        choices=("3.7", "3.8", "3.9"),
+        choices=("3.8", "3.9"),
```
graeme-winter, line 1267: There a reason we do not allow 3.10 yet?

graeme-winter (approved):
Good to see - helps us stay up to date and avoid the pain we had dropping Python 2.7 etc.

installer/bootstrap.py
Anthchirp: Lots of packages don't build on 3.10 yet, but yes we can probably enable this in bootstrap

Anthchirp: I've added 3.10 in this PR. The step currently holding this back is our numpy incompatibility due to https://github.com/cctbx/cctbx_project/issues/627. Other packages that are not yet on 3.10 include eg. bitshuffle (https://github.com/kiyo-masui/bitshuffle/issues/94, https://github.com/conda-forge/bitshuffle-feedstock/pull/22), so you may still want to hold off for a bit.

ndevenish:
These new warnings make me wonder if we've been missing things by only running the oldest version in CI

Anthchirp:
Yes. The reason we only run one version is because building cctbx takes ages. Fix that and (eg. via prebuilt cctbx) then we could build all versions simultaneously while still being faster than before.

Issue #1829, opened 2021-08-01
dials.reciprocal_lattice_viewer with multiple lattices in show in crystal frame doesn't 


graeme-winter:
![Screenshot 2021-08-01 at 10 20 47](https://user-images.githubusercontent.com/1241716/127765980-c161de6b-f5a6-487d-a782-1b611a5b28c6.png)
![Screenshot 2021-08-01 at 10 20 52](https://user-images.githubusercontent.com/1241716/127765982-48b032a9-abcd-4d06-bf2d-5c5126ee2763.png)

Looks like only 1st U matrix is used for mapping not right U matrix for each lattice - should have three different "shells"

graeme-winter:
Input

[1829.zip](https://github.com/dials/dials/files/6912421/1829.zip)

graeme-winter:
Taking a look at this now and it is not as obvious as I would have expected e.g. 

```
            vecs = [
                [
                    matrix.col(l)
                    for l in (100 * matrix.sqr(c.get_B()))
                    .transpose()
                    .as_list_of_lists()
                ]
                for c in crystals
            ]
```

is where these come from - there is no U matrix in the mix

graeme-winter:
Looks like this is because multiple lattices on one set share the image set, and it is that which is being used for the selections where the mapping to reciprocal space is calculated.

graeme-winter:
OK, unknown here is what to do if we have (i) unindexed reflections and (ii) more than one crystal lattice - _which_ U matrix to use for the mapping...

Issue #1104, opened 2020-01-22
Eliminate legacy 'imageset_id' in reflection tables


benjaminhwilliams:
Move to determination of image set from experiment ID → experiment → imageset.

ndevenish:
Do we have empty experiments pre-indexing now?

benjaminhwilliams:
I don't know what the state of play is on that.  Who's plate is that on?

ndevenish:
I mean, without being certain on that we can’t call this legacy? Because otherwise is required for unindexed reflections?

rjgildea:
Currently the experiment "id" is overloaded to mean two different things:
1) After spotfinding to associate a strong spot with an imageset/uninded experiment
2) After indexing to associate an indexed reflection with an indexed experiment (with unindexed reflections set to -1)
After indexing there is no way just using experiment ids/identifiers to keep track of which imageset unindexed reflections belong to. This is the role that imageset_ids currently perform. As such I don't think we can state that these are "legacy" until we have a valid alternative.

graeme-winter:
Humbly suggest that if we followed up on https://github.com/dials/dials/issues/1029 the problem would go away, as the old experiment ID would point at the imported experiment, the new one to the recently indexed reflections, so `imageset_id` would be truly redundant (and everything would be more easily understood)

rjgildea:
> Humbly suggest that if we followed up on #1029 the problem would go away, as the old experiment ID would point at the imported experiment, the new one to the recently indexed reflections, so `imageset_id` would be truly redundant (and everything would be more easily understood)

I still don't see how this would allow you to get "all reflections on the first imageset", which the existing `imageset_id` approach _does_.

PR #1868, opened 2021-09-01
If crystal frame, use U for each crystal

graeme-winter:
Weird side effect of this is you never map the unindexed reflections. Need to
iterate through and assign unindexed reflections a U matrix for mapping or
similar.

WIP for #1829

### DIFF
--- a/array_family/flex_ext.py
+++ b/array_family/flex_ext.py
@@ -1239,10 +1239,10 @@ Found %s"""
         panel_numbers = cctbx.array_family.flex.size_t(self["panel"])
 
         for i, expt in enumerate(experiments):
-            if "imageset_id" in self:
-                sel_expt = self["imageset_id"] == i
-            else:
+            if "id" in self:
                 sel_expt = self["id"] == i
+            else:
+                sel_expt = self["imageset_id"] == i
 
             for i_panel in range(len(expt.detector)):
                 sel = sel_expt & (panel_numbers == i_panel)
--- a/array_family/flex_ext.py
+++ b/array_family/flex_ext.py
@@ -1238,11 +1238,15 @@ Found %s"""
         self["rlp"] = cctbx.array_family.flex.vec3_double(len(self))
         panel_numbers = cctbx.array_family.flex.size_t(self["panel"])
 
+        # crystal_frame is not used in indexing, but is a feature of the
+        # reciprocal lattice viewer -> but if we are looking at the crystal
+        # coordinate frame we need to look at the experiments independently
+
         for i, expt in enumerate(experiments):
-            if "id" in self:
-                sel_expt = self["id"] == i
-            else:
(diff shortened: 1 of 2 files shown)


array_family/flex_ext.py
```diff
@@ -1239,10 +1239,10 @@ def map_centroids_to_reciprocal_space(
         panel_numbers = cctbx.array_family.flex.size_t(self["panel"])
 
         for i, expt in enumerate(experiments):
-            if "imageset_id" in self:
-                sel_expt = self["imageset_id"] == i
-            else:
+            if "id" in self:
                 sel_expt = self["id"] == i
+            else:
+                sel_expt = self["imageset_id"] == i
```
rjgildea, line 1245: Have you ran the indexing tests after making this change? I'm pretty sure this will result in some indexing failures as it explicitly relies on the existing behaviour - i.e. use the `imageset_id`s which are distinct, rather than the `id`s which will have been reset to `-1` at the beginning of indexing to indicate "unindexed". (Yes, I know #1029 #1104.)

graeme-winter: Of course: this is precisely why I flagged this very early as a draft PR as I thought this will be über destructive - though that to me indicates that we have a _different_ issue... but point well made.

graeme-winter: Unsurprisingly this breaks many things

```
Results (264.31s):
    1022 passed
       4 failed
         - tests/algorithms/indexing/test_index.py:753 test_index_known_orientation
         - tests/algorithms/indexing/test_assign_indices.py:247 test_local_multiple_rotations
         - tests/command_line/test_export_best.py:4 test_export_best
         - tests/command_line/test_search_beam_position.py:170 test_search_small_molecule
     157 skipped
```

rjgildea, line 1245: Perhaps the following would solve your problem in a less destructive way?
```suggestion
            if not crystal_frame and "imageset_id" in self:
                sel_expt = self["imageset_id"] == i
            else:
                sel_expt = self["id"] == i
```

graeme-winter: Am doing a little bit of hunting around - I have a suspicion that there are better ways to fix this problem, as there are other derived questions from this (like: is this part of the reason why re-running indexing on multi-lattice data does not re-index all lattiecs)

graeme-winter: BTW I recognise that this would fix the _behaviour_ but I worry that it is not fixing the underlying problem.

graeme-winter: This change also broke indexing with > 1 lattice...

graeme-winter:
Looking at the code @rjgildea  if there are multiple experiments with the same `imageset_id` then the points will be mapped multiple times... this is probably not what we want? I mention for your opinion, perhaps I should raise a new issue?

ndevenish (approved):

PR #1869, opened 2021-09-02
Changed default image viewer helper message to reflect current controls.

toastisme:
This just updates the default image viewer helper message, as it seems to be out of date with current controls. 

I can't find anywhere where the intensity plot is used (I'm guessing it's [OnShowPlot](https://github.com/dials/dials/blob/main/util/image_viewer/rstbx_frame.py#L333)), and the right click to zoom in [wx_viewer](https://github.com/dials/dials/blob/main/util/wx_viewer.py#L279) doesn't seem to be in use.

### DIFF
--- a/newsfragments/1869.misc
+++ b/newsfragments/1869.misc
@@ -1,0 +1,1 @@
+Updated default image viewer status bar message to reflect current controls
--- a/newsfragments/1869.misc
+++ b/newsfragments/1869.misc
@@ -1,1 +1,1 @@
-Updated default image viewer status bar message to reflect current controls
+dials.image_viewer: Updated default status bar message to reflect available controls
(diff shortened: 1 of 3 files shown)


newsfragments/1869.misc
```diff
@@ -0,0 +1 @@
+Updated default image viewer status bar message to reflect current controls
```
rjgildea, line 1:
```suggestion
dials.image_viewer: Updated default status bar message to reflect available controls
```

rjgildea (approved):
Are both of these actually required or (more likely) does one of them represent dead code. in which case it should probably be removed (not suggesting you should do that here in this PR)?

toastisme:
> Are both of these actually required or (more likely) does one of them represent dead code. in which case it should probably be removed (not suggesting you should do that here in this PR)?

It looks like only only `frame.py` is used, but felt it was best to at least keep it `rstbx_frame.py` updated in case this is used in some other context.

PR #1876, opened 2021-09-08
Bad pixel format

graeme-winter:
Details at https://jira.diamond.ac.uk/browse/I04-575

Changing output format from this to be more useful (TODO should write this out as e.g. JSON to be more portable, but that is for a different day)

### DIFF
--- a/command_line/find_bad_pixels.py
+++ b/command_line/find_bad_pixels.py
@@ -183,7 +183,7 @@ def run(args=None):
     hot_pixels = hot_mask.iselection()
 
     for h in hot_pixels:
-        print("    mask[%d, %d] = 8" % (h % nfast, h // nfast))
+        print("    mask[%d, %d] = 16" % (h // nfast, h % nfast))
 
 
 if __name__ == "__main__":
--- a/newsfragments/XXX.bugfix
+++ b/newsfragments/XXX.bugfix
@@ -1,0 +1,2 @@
+dials.find_bad_pixels: change output pixel format to write out input to Diamond
+bad pixel mask program correctly (C ordering, 16 which indicates noisy pixel)


tests/command_line/test_find_bad_pixels.py
```diff
@@ -18,7 +18,7 @@ def test_find_bad_pixels(dials_data, tmp_path):
     count = 0
     for record in result.stdout.decode().split("\n"):
         if "mask" in record:
-            assert record.split()[-1] == "8"
+            assert record.split()[-1] == "16"
```
rjgildea, line 21: We should probably test/verify the C-ordering vs Fortran-ordering... a little concerning that changing the ordering doesn't cause the test to fail. Either parse the output to build up the set of bad pixels to compare with known reference, or just test one or more expected row to assert that it is present in the output.

rjgildea: Obviously json output would make this easier, although probably still useful to check the stdout.

newsfragments/1876.bugfix
```diff
@@ -0,0 +1,2 @@
+dials.find_bad_pixels: change output pixel format to write out input to Diamond
+bad pixel mask program correctly (C ordering, 16 which indicates noisy pixel)
```
rjgildea, line 2:
```suggestion
dials.find_bad_pixels: update output for compatibility with Diamond Light Source bad pixel mask program (pixel coordinates in row-major order; set mask value to 16 which corresponds to a noisy pixel)
```

rjgildea (changes requested):
Could you update the help message to explain the output format?

tests/command_line/test_find_bad_pixels.py
graeme-winter: 👍

command_line/find_bad_pixels.py
```diff
@@ -12,6 +12,16 @@
 
 help_message = """
 
+Identify unreliable pixels in a sequence of images: defined to be those pixels
+which read as signal in spot finding in 50% or over of images. Outputs data in
+the order slow, fast (e.g. C ordering) as ::
```
rjgildea, line 17:
```suggestion
the order slow, fast (i.e. row-major order) as ::
```

rjgildea (approved):
Thanks for the updated test and help 👍

PR #1880, opened 2021-09-10
dials.index: support arbitrary P1 unit cells

### LINKED ISSUE #1878: P1 indexing fails if cell parameters are not in conventional setting
See xfail test here: https://github.com/dials/dials/commit/9b436918a175995b4c857040da94ed26ac94007d

rjgildea:
Map input unit cell to minimum cell for aP settings, to support passing in arbitrary P1 cells. The final unit cell will be reported in the requested setting.

Fixes #1878


dwpaley (approved):
Looks great, thank you @rjgildea. I'm surprised by the (non-)behavior of `change_of_basis_op_to_reference_setting` for triclinic, but this solves my problem.

rjgildea:
> I'm surprised by the (non-)behavior of `change_of_basis_op_to_reference_setting` for triclinic

`change_of_basis_op_to_reference_setting` is purely based on the space group, and independent of the unit cell, so for triclinic it is always the identity op:
https://github.com/cctbx/cctbx_project/blob/35f53cd89fbf93de66dc2ff55632cb5c88144937/cctbx/crystal/__init__.py#L216-L217

dwpaley:
Makes sense. Thanks for fixing this so quickly!

PR #1884, opened 2021-09-17
`dials.find_rotation_axis`

dagewa:
New program to search for the orientation of the rotation axis prior to indexing. The code is based on `edtools.find_rotation_axis` from @stefsmeets's [edtools](https://github.com/instamatic-dev/edtools).

The test requires https://github.com/dials/data-files/pull/33 and https://github.com/dials/data/pull/303 to be merged first.


dagewa:
As this uses a significant amount of @stefsmeets code I think the eventual merge commit should include co-authorship, if you are happy with that @stefsmeets?

stefsmeets:
> 
> 
> As this uses a significant amount of @stefsmeets code I think the eventual merge commit should include co-authorship, if you are happy with that @stefsmeets?

Cool, nice to see this made it in! That would work for me, thanks 👍

command_line/find_rotation_axis.py
```diff
@@ -0,0 +1,496 @@
...
+        azimuth_fine = azimuth_tmp = optimise(
+            arr, azimuth_tmp, wavelength, plusminus=1, step=0.1, hist_bins=hist_bins
+        )
+
+        azimuth_final = azimuth_tmp
+
+        logger.info("---")
+        logger.info(f"Best azimuth (global search): {azimuth_global:.3f}")
+        logger.info(f"Best azimuth (local search): {azimuth_local:.3f}")
+        logger.info(f"Best azimuth (fine search): {azimuth_fine:.3f}")
```
Anthchirp, line 439: Just a note (don't hold up the PR for this): logging isn't print - Things like this should be a single log call.

dagewa: Understood. This comes from a find-and-replace starting from the `print`s in `edtools.find_rotation_axis`. Happy to make this a single call before merge.

PR #1903, opened 2021-10-28
Add options to dials.damage_analysis to plot data and generate files for damage series

jbeilstenedmands:
I was thinking about implementing a tool to split a multi-crystal dataset into subsets based on dose, and thought this would fit nicely as part of `dials.damage_analysis`, where additional plots could also be generated.

This PR adds a `damage_series.dose_group_size` option to `dials.damage_analysis`, to generate some merging statistics plots for subsets of the data. 
The corresponding datafiles can also be output by turning on the options `output.damage_series` and `output.accumulation_series`.

**Example output**
For a multi-crystal dataset (5 sweeps of up to 500 images each with significant radiation damage):
`dials.damage_analysis scaled.{expt,refl} group_size=100 output.damage_series=True output.accumulation_series=True`

This generates additional plots after the pychef analysis which are output in the html (cc_half, r_pim, i_over_sigma and completness):
newplot (86)
newplot (87)
Output files are optionally generated for the damage group series:
`damage_series_0_100.{expt,refl} damage_series_100_200.{expt,refl}` etc.
and also optionally for the damage accumulation series:
`damage_series_0_100.{expt,refl} damage_series_0_200.{expt,refl}` etc.

### DIFF
--- a/newsfragments/1903.feature
+++ b/newsfragments/1903.feature
@@ -1,1 +1,1 @@
-``dials.damage_analysis``: Add dose_group_size option to allow generation of plots for subsets of the data based on dose
+``dials.damage_analysis``: Add ``dose_group_size=`` option to allow generation of plots for subsets of the data based on dose.


jbeilstenedmands:
Can't remember why I didn't mark this as ready at the time, I can't see anything particular missing so i'm now putting this out for review.

PR #1904, opened 2021-10-29
python-dateutils -> python-dateutil

phyy-nx:
Mamba is permissive of this but conda fails.  Affects the cctbx bootstrap.


ndevenish:
Huh, I that's definitely strange, neither my linux mamba nor micromamba actually accepts this name, but it's apparently working on the dxtbx CI?

ndevenish:
On reconsideration I remember seeing a bug at some point that micromamba ignored missing dependencies, which is possibly what happened here. (There was a ticket filed but I can't find it again).

PR #1911, opened 2021-11-04
Some corner cases: reflection bin critical values None not 0.0 & missing partiality

graeme-winter:
If the bin critical value is None rather than 0 because of missing all
reflections, replace None with 0.0 so that the flex.double call is sane.

If no partiality column in reflection table, make sum_partial_reflections a
no-op i.e. simply return the input table.


jbeilstenedmands (approved):
Looks sensible, thanks. Are there are other locations in `resolution_analysis.py` where we need to check for None values too, looks like some very similar code appears several times in the lines above the change?

ndevenish:
Am treating case above as "there might be more" rather than "there should be more in this PR"

PR #1912, opened 2021-11-04
Indexed as integrated

graeme-winter:
dials.indexed_as_integrated

Manipulate an indexed reflection file to look as if it were summation integrated. This simply takes the spot intensities that have been indexed, assigns a resolution and sets the summation integrated flag. Allows symmetry analysis using intensities and scaling on indexed data for very rapid feedback data processing.


jbeilstenedmands:
I wonder if we can make this idea redundant with a few simple changes in the other programs.
When the `intensity.sum.value` column is set in spot finding, shouldn't the `integrated_sum` flag be set at that point?
It makes sense to me that the `d` column can be set during indexing, i'm sure this would be helpful in other cases too.
Then you could probably run symmetry/scaling etc from a pure `indexed.{refl,expt}` set, as non-indexed/integrated spots will get filtered by the symmetry/scaling programs?

graeme-winter:
@jbeilstenedmands I thought the same myself, and that was actually my first implementation of this - however some thoughts - 

 - this is a very substantial change in the workflow and an abuse of the flags
 - when we filter reflections later in integration, the input ones would somehow pass through
 - this workflow _should not be possible by default_ 

But I agree that indexing really could set the `d` column. 

An alternative I had thought is to allow `dials.{symmetry|scale}` to have a PHIL option of

```
accept_indexed_as_integrated
```

or similar which could encode the changes in this file. Perhaps that would be a better option?

jbeilstenedmands:
In terms of filtering in integration, I don't think that is particularly a problem, as we could just add an 'unset_flags' call for integrated_sum at the start of integration as a safety measure.
In terms of flag abuse, before seeing this PR I was surprised that we had a case where "valid" values for `intensity.sum.value` exist but all `integrated_sum` flags are False, in my mind (and some of the scaling code!) these concepts are almost equivalent.
Should it not be possible to scale the strong spots that were summation integrated in spot finding? I guess if we allowed such a thing then more people might start doing it 😉

jbeilstenedmands (approved):
I appreciate this script fulfils a specific need, and although the implementation might not be how I would have done it, I see that it makes sense to keep this kind of thing separate and to "fork" the workflow at this point. So i've added a test and then it should be good to go.

command_line/indexed_as_integrated.py
```diff
@@ -0,0 +1,73 @@
+import iotbx.phil
+
+import dials.util
+from dials.array_family import flex
+from dials.util.options import OptionParser, reflections_and_experiments_from_files
+
+help_message = """
+Small modifications to an indexed spot list to allow it to be treated as if
+it were integrated, for rapid feedback analysis.
+"""
```
rjgildea, line 10: Can we add a warning (to both the help and the program stdout) that this is very much experimental and not (yet) intended for routine use, etc etc?

command_line/indexed_as_integrated.py
```diff
@@ -0,0 +1,73 @@
...
+    reflections, experiments = reflections_and_experiments_from_files(
+        params.input.reflections, params.input.experiments
+    )
+
+    if len(reflections) != 1 or len(experiments) != 1:
+        parser.print_help()
+        return
+
+    stronger = indexed_as_integrated(reflections[0], params, experiments)
+    stronger.as_file(params.output.reflections)
```
rjgildea, line 69: Possibly slightly clearer variable name to aid comprehension:
```suggestion
    pseudo_integrated = indexed_as_integrated(reflections[0], params, experiments)
    pseudo_integrated.as_file(params.output.reflections)
```

command_line/indexed_as_integrated.py
```diff
@@ -0,0 +1,73 @@
+import iotbx.phil
+
+import dials.util
+from dials.array_family import flex
+from dials.util.options import OptionParser, reflections_and_experiments_from_files
```
rjgildea, line 5:
```suggestion
from dials.util.options import ArgumentParser, reflections_and_experiments_from_files
```

command_line/indexed_as_integrated.py
```diff
@@ -0,0 +1,73 @@
...
+    usage = "dials.indexed_as_integrated [options] indexed.refl indexed.expt"
+
+    parser = OptionParser(
+        usage=usage,
+        phil=phil_scope,
+        read_reflections=True,
+        read_experiments=True,
+        check_format=False,
+        epilog=help_message,
+    )
```
rjgildea, line 56: OptionParser is deprecated, and `args` is not passed through:
```suggestion
    parser = ArgumentParser(
        args=args,
        usage=usage,
        phil=phil_scope,
        read_reflections=True,
        read_experiments=True,
        check_format=False,
        epilog=help_message,
    )
```

tests/command_line/test_indexed_as_integrated.py
```diff
@@ -0,0 +1,48 @@
...
+    refl = data_dir / "indexed.refl"
+    expt = data_dir / "indexed.expt"
+
+    command = [
+        "dials.indexed_as_integrated",
+        refl,
+        expt,
+        "output.reflections=output.refl",
+    ]
+    result = procrunner.run(command, working_directory=tmp_path)
```
rjgildea, line 17: I would probably prefer to call the command line `run` method directly, rather than with an additional indirection layer via `procrunner` that complicates any tracebacks or warnings etc, but that isn't critical to merging this PR.

rjgildea (approved):
As per JBE's comment, happy for this to be merged as long as it is clearly marked as experimental/not for production use.

command_line/indexed_as_integrated.py
graeme-winter: Wilco

graeme-winter: You happy with the health warnings just added?

Since I had to figure out the logging I added a few extra things as well:

```

input {
  experiments = indexed.expt
  reflections = indexed.refl
}

DIALS 3.dev.633-g55edc3010

----------------------------------------------------------
Do not use the output for structure solution or refinement
----------------------------------------------------------

Saved 10883 of 11185 reflections


Saved pseudo-integrated data to: pseudo_integrated.refl


```

PR #1913, opened 2021-11-04
Add missing item in scale-and-filter results to/from json

rjgildea:
The `expids_and_image_ranges` key was missing from the deltacchalf filtering results json file. This item is needed for generating some of the plots in `image_group` mode.


jbeilstenedmands (approved):
Thanks for fixing!

PR #1917, opened 2021-11-09
Document the DXTBX rotation operator convention

benjaminhwilliams:
DXTBX uses an in-house convention for the representation of the goniostat rotation operator _R_.  Here, I've added documentation of that convention to the `conventions.rst` page of the docs.

Questions for reviewers:
1. Is this the right place for this information?  It is more developer documentation than user documentation, so it may want to live on some other page.
2. Is the content correct?
2. Does the content make sense?  Is it intelligible?
3. Could it be more concise?

Partial fix for https://github.com/cctbx/dxtbx/issues/128.


benjaminhwilliams:
This has informal :+1: offline from @elena-pascal & @dagewa.

Issue #1429, opened 2020-09-24
More helpful help


dagewa:
At the moment, `dials.program -h` gives essentially the same output as `dials.program`. Meanwhile, getting useful help about the program parameters requires something like `dials.program -c -e2 -a1`, which is annoying.

How about if `dials.program -h` is changed to do the equivalent of `cat <(dials.program) <(dials.program -c -e2 -a1)`? What about piping the output to `less` as well, on platforms that support it?

rjgildea:
I'm very much in favour of improving the command line help and how users access help for the parameters, and I very much dislike having to tell users to type magic spells like `dials.program -c -e2 -a1`. However, I note that your proposal would generate e.g. over 600 lines of output for `dials.refine`, which would bury the program description among hundreds of lines of output. Unfortunately I don't have a better suggestion...

```
$ dials.refine -c -e2 -a1 | wc -l
583
$ dials.refine | wc -l
28
```

ndevenish:
For a while I've thought that having a "top parameters" for each program would be useful, such that asking for --help would print their names and descriptions.

dagewa:
I think we could partially fix this with better use of `expert_levels`, but also the pipe to `less` where appropriate

rjgildea:
> like `man` pages

This is what I was thinking. Perhaps better use of bold/colour formatting may help as well with aiding visual comprehension.

Compare the output of `man git` with `dials.refine -c -e2 -a1`:

![Screenshot 2020-09-24 at 10 50 25](https://user-images.githubusercontent.com/2810624/94129831-c8a1a980-fe53-11ea-9a80-cdc8b9bb8b6e.png)
![Screenshot 2020-09-24 at 10 49 57](https://user-images.githubusercontent.com/2810624/94129840-cb040380-fe53-11ea-99fd-dc81e49173c9.png)

dagewa:
if we're actually changing the format of the phil output then I would go further - removing the quotes around the help strings for a start.

dagewa:
[Using less as a pager](https://chase-seibert.github.io/blog/2012/10/31/python-fork-exec-vim-raw-input.html#using-less-as-a-pager), like `git log`

PR #1920, opened 2021-11-10
More verbose cli help by passing -h multiple times

rjgildea:
`-h` is now equivalent to `-ce1`
`-hh` is equivalent to `-ce1a1`
`-hhh` is equivalent to `-ce2a1`
`-hhhh` is equivalent to `-ce3a1`
etc...

Fixes https://github.com/dials/dials/issues/1429


phyy-nx:
Awesome

PR #1924, opened 2021-11-12
dials.refine: fix track_parameter_correlation bug

### LINKED ISSUE #1923: Crash writing history.json from dials.refine
```
$ dials.refine ../refined.{expt,refl} correlation_plot.filename=corrgram.pdf track_condition_number=True history=history.json scan_varying=False
```
ended with
```
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/refine.py", line 477, in <module>
    run()
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/refine.py", line 473, in run
    history.to_json_file(params.output.history)
  File "/home/fcx32934/sw/cctbx/modules/dials/algorithms/refinement/engine.py", line 134, in to_json_file
    json.dump(d, f)
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Please report this error to dials-support@lists.sourceforge.net: Object of type double is not JSON serializable
```
The issue is that the `parameter_correlation` is kept in the journal as `flex.double`, which was fine when history was output as pickle, but breaks now it is JSON. The fix is easy, but should add a test as well.

dagewa:
Fixes #1923 by storing tracked parameter correlations using Python lists rather than `flex.double`, making it easier to export to JSON format.


ndevenish:
Since you were making the return type explicit, I thought that I would add a type annotation. In order to actually work out what this type is, I had to work backwards a little bit. Can you double-check that the types in https://github.com/dials/dials/pull/1924/commits/dfef605ea2a40dc3dd77b480513e315b46e48d6f make sense?

Note: `_packed_corr_mat` accepts `refiner._jacobian`, which is assumed in multiple places to _sometimes_ be a `sparse.matrix`. However, `scitbx.sparse.matrix` doesn't have an `.all()` method, so https://github.com/dials/dials/blob/dfef605ea2a40dc3dd77b480513e315b46e48d6f/algorithms/refinement/engine.py#L300-L304 actually _fails_ if it's parsed a sparse array, which since this works is apparently never. This is a type error, so I'm not sure where the assumption is failing here (e.g. that this can be never called in cases where the type is known to be sparse).

dagewa:
Yes in practice the matrix always passes through `split_jacobian_into_blocks` first, which already does the `j = self._jacobian.as_dense_matrix()`. In that case, I think `_packed_corr_mat` could be restricted to `flex.double` only so that the function types reflect the actual usage

dagewa:
Oh wait, that's fun. If the jacobian _is_ sparse, it is split into dense vectors in `split_jacobian_into_blocks`, and then converted back to sparse in `_packed_corr_mat`, which certainly seems unnecessary. I'll take a closer look at this...

dagewa:
Oh in resolving a merge I lost my original commit message. For reference, that was

```
    Avoid sparse<->double tennis
    
    split_jacobian_into_blocks always returns flex.double, so change
    _packed_corr_mat to always operate on that type, rather than
    adding a needless conversion to sparse.```

PR #1926, opened 2021-11-12
Specifying intensity_choice=auto fails for some dials.export output formats

huwjenkins:
Some of the output formats for `dials.export` fail when `intensity_choice=auto` is specified on the command line with the cryptic error:

```
argument of type 'AutoType' is not iterable
```

This change fixes that.


jbeilstenedmands (approved):
Thanks for finding and fixing, I learnt something new here!

ndevenish:
The CI tests seem to be an unrelated issue. Merging now to put into release. Thanks for this.

PR #1930, opened 2021-11-16
Unmark xpass dials.tests.util.test_nexus.test_run

benjaminhwilliams:
This partially reverts commit 885a2f6ea3900dd0c9fcc15c03561fb45452c3bb.


benjaminhwilliams:
Ah, @rjgildea has reminded me that we're stuck with broken CCTBX for a while, so we don't in fact want to merge this straight away.

PR #1938, opened 2021-11-23
DUI tutorial update

dagewa:
Update for DUI 2021.11.1 with DIALS 3.7


dagewa:
The test failures are unrelated. The path to tutorial data at Diamond may still be incorrect, and the tutorial makes no mention of the DUI report view workaround required for the workshop. Nevertheless, going to merge now because we need the website updated by the workshop and I don't want to cut it too fine.

PR #1943, opened 2021-11-30
Read MemoryProvisioned from _CONDOR_MACHINE_AD env var

rjgildea:
When determining available memory for integration, take into account
`MemoryProvisioned` from the HTCondor machine ad if the
`_CONDOR_MACHINE_AD` environment variable is set.

See also:
https://htcondor.readthedocs.io/en/latest/users-manual/services-for-jobs.html#extra-environment-variables-htcondor-sets-for-jobs


ndevenish:
Do we also want to read this for CPU provisioning when guessing the number of parallel processes?

isikhar (approved):

isikhar:
> Do we also want to read this for CPU provisioning when guessing the number of parallel processes?

Could be useful in case nproc isn't set.

PR #1950, opened 2021-12-08
Add ability to output datafiles for dials.cluster_unit_cell

jbeilstenedmands:
Adds a new option, `output.clusters=False/True`.
If True, then a `cluster_*.expt`,`cluster_*.refl` pair is created for each cluster, where the clusters are those separated at the threshold point. So changing the threshold controls how the data is split for output.


tests/command_line/test_cluster_unit_cell.py
```diff
@@ -27,6 +28,77 @@ def test_dials_cluster_unit_cell_command_line(dials_regression, run_in_tmpdir):
...
+        command=[
+            "dials.split_experiments",
+            tmp_path / "combined.refl",
+            tmp_path / "combined.expt",
+        ],
+        working_directory=tmp_path,
+    )
+    assert not result.returncode
+    experiments = glob.glob(os.path.join(tmp_path, "split_*.expt"))
+    reflections = glob.glob(os.path.join(tmp_path, "split_*.refl"))
```
rjgildea, line 86: I think the following should work (and be much simpler):
```suggestion
    experiments = tmp_path.glob("split_*.expt")
    reflections = tmp_path.glob("split_*.refl")
```

tests/command_line/test_cluster_unit_cell.py
```diff
@@ -27,6 +28,77 @@ def test_dials_cluster_unit_cell_command_line(dials_regression, run_in_tmpdir):
...
 
+def test_dials_cluster_unit_cell_command_line_output_files(dials_regression, tmp_path):
+    pytest.importorskip("scipy")
+    pytest.importorskip("xfel")
+
+    data_dir = os.path.join(
+        dials_regression, "refinement_test_data", "multi_narrow_wedges"
+    )
+    experiments = glob.glob(os.path.join(data_dir, "data/sweep_*/experiments.json"))
+    reflections = glob.glob(os.path.join(data_dir, "data/sweep_*/indexed.pickle"))
```
rjgildea, line 39: I think the following should work:
```suggestion
    data_dir = pathlib.Path(dials_regression) / "refinement_test_data" / "multi_narrow_wedges"
    experiments = data_dir.glob("data/sweep_*/experiments.json")
    reflections = data_dir.glob("data/sweep_*/indexed.pickle")
```

command_line/cluster_unit_cell.py
```diff
@@ -70,12 +84,81 @@ def run(args=None):
...
+                reflections = _assign_and_return_joint(experiments, reflections)
+
+        elif len(reflections) > 1:
+            if not len(reflections) == len(experiments):
+                reflections = parse_multiple_datasets(reflections)
+                if len(reflections) != len(experiments):
+                    raise ValueError(
+                        f"Mismatched number of reflection tables (f{len(reflections)}) and experiments (f{len(experiments)})"
+                    )
+            reflections = _assign_and_return_joint(experiments, reflections)
```
rjgildea, line 124: This seems like something that is probably used in other command line (or very similar code) - could it be extracted somewhere so it can be more easily re-used? (not necessarily as part of this PR)

newsfragments/1950.feature
```diff
@@ -0,0 +1 @@
+``dials.cluster_unit_cell``: Add option output.clusters=True/False to generate output files for each cluster
```
rjgildea, line 1:
```suggestion
``dials.cluster_unit_cell``: Add option ``output.clusters=True/False`` to generate output files for each cluster generating by splitting the dendrogram at the given ``threshold``.
```

command_line/cluster_unit_cell.py
```diff
@@ -29,6 +36,10 @@
     .type = bool
     .help = 'Display the dendrogram with a log scale'
 }
+output.clusters = False
+    .type = bool
+    .help = "If True, the data will be split at the threshold value and a pair"
```
rjgildea, line 41: "the data"?

rjgildea (approved):
Thanks for adding this, I think it will be a useful addition to the multi-crystal workflow 👍

command_line/cluster_unit_cell.py
jbeilstenedmands: Sure, will keep this in mind when reviewing the other relevant tools for multicrystal work

PR #1951, opened 2021-12-13
Match by hkle

graeme-winter:
Add a method to `flex.reflection_table` to match `other` by `miller_index` and `entering` flags


graeme-winter:
Added @benjaminhwilliams and @rjgildea as reviewers as I have tried to add type annotations and a test to this so would welcome some feedback... And ofc @ndevenish needs to rewrite the news fragment

graeme-winter:
Oh, I note that `import pandas as pd` is in here...

array_family/flex_ext.py
```diff
@@ -469,6 +471,48 @@ def subsort(self, key0, key1, reverse=False):
...
+        self, other: dials_array_family_flex_ext.reflection_table
+    ) -> Tuple[cctbx.array_family.flex.size_t, cctbx.array_family.flex.size_t,]:
+        """
+        Match reflections with another set of reflections by the h, k, l
+        and entering values. Uses pandas dataframe merge method to match
+        the columns: assumes the key h, k, l, e is unique which is false
+        if > 360 degree rotation.
+
+        :param other: reflection table to match against
+        :return: indices in self, indices in other for matches
```
benjaminhwilliams, line 484: I think elsewhere we've tended to use something like Google style for docstrings:
```suggestion
        Args:
            other: reflection table to match against

        Returns:
            Indices in self, indices in other for matches
```

benjaminhwilliams (changes requested):
Breaking this down to the basic objectives, we are trying to turn (part of) a Flex `reflection_table` into a Pandas `DataFrame`.  This is made more difficult than it could be by the limited interoperability of Flex and NumPy.  In an ideal world, the Flex types would all have `__array__` methods, exposing the logic that is currently written into their `as_numpy_array` methods (better still, replacing these methods altogether), so that anything that expects a `ndarray` sees a `ndarray`, with that `ndarray` being a view of the data in the Flex array rather than a copy.  As it is, none of the Flex types have an `__array__` method, and only some have an `as_numpy_array` method.  Worse still, although `flex.miller_index` has an `as_vec3_double` method, it has no integer-type equivalent, in spite of the fact that the data in a `flex.miller_index` are naturally integers.

I have had a few goes at different ways of doing this, without delving into adding an `__array__` method to `flex.miller_index` and the other Flex types.  I've come full circle back to something very closely resembling your suggestion, so I apologise that this is a very wordy review with few actual proposed changes.

array_family/flex_ext.py
```diff
@@ -469,6 +471,48 @@ def subsort(self, key0, key1, reverse=False):
...
+        n1 = np.array(range(len(e1)))
+
+        x1 = {"h": h1, "k": k1, "l": l1, "e": e1, "n1": n1}
+
+        p1 = pd.DataFrame(data=x1, columns=["h", "k", "l", "e", "n1"])
+
+        merged = pd.merge(p0, p1, on=["h", "k", "l", "e"], how="inner")
+
+        n0 = cctbx.array_family.flex.size_t(merged["n0"].to_numpy())
+        n1 = cctbx.array_family.flex.size_t(merged["n1"].to_numpy())
```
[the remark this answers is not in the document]
benjaminhwilliams: Though this is another example of poor interoperability of Flex with standard Python data handling tools.  I haven't delved into why `flex.size_t(merged.n0.values)` is so much quicker than `flex.size_t(merged.n0)`.

array_family/flex_ext.py
```diff
@@ -469,6 +471,48 @@ def subsort(self, key0, key1, reverse=False):
             ref_tmp.sort(key1, reverse)
             self[min(val) : (max(val) + 1)] = ref_tmp
 
+    def match_by_hkle(
+        self, other: dials_array_family_flex_ext.reflection_table
+    ) -> Tuple[cctbx.array_family.flex.size_t, cctbx.array_family.flex.size_t,]:
```
benjaminhwilliams, line 476: Somehow, this dropped out of my collection of suggested changes.
```suggestion

```    ) -> Tuple[2 * (cctbx.array_family.flex.size_t,)]:

**Review by benjaminhwilliams — CHANGES_REQUESTED** [review]

File: array_family/flex_ext.py (hunk 1 above)
Comment (rjgildea) [reply array_family/flex_ext.py:484]: +1

File: array_family/flex_ext.py [hunk 4]
```diff
@@ -469,6 +471,48 @@ def subsort(self, key0, key1, reverse=False):
...
+        hkl = other["miller_index"].as_vec3_double().parts()
+        h1 = hkl[0].iround().as_numpy_array()
+        k1 = hkl[1].iround().as_numpy_array()
+        l1 = hkl[2].iround().as_numpy_array()
+        e1 = other["entering"].as_numpy_array()
+        n1 = np.array(range(len(e1)))
+
+        x1 = {"h": h1, "k": k1, "l": l1, "e": e1, "n1": n1}
+
+        p1 = pd.DataFrame(data=x1, columns=["h", "k", "l", "e", "n1"])
```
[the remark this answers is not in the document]
rjgildea: An alternative approach, using `flumpy` could be:
```
from dxtbx import flumpy
...
h, k, l = np.hsplit(flumpy.to_numpy(hkl.as_vec3_double()).astype(int), 3)
```

It looks like `flumpy` doesn't natively handle `flex.miller_index` types, but since this is something can control we should be able to easily add this, then we can skip the flex.miller_index -> flex.vec3_double() -> int dance.

array_family/flex_ext.py
rjgildea: An alternative, possibly more readable, option would be to declare a [type alias](https://docs.python.org/3/library/typing.html#type-aliases), e.g.:

```
FlexSizeT = cctbx.array_family.flex.size_t
...
def match_by_hkle(...) -> Tuple[FlexSizeT, FlexSizeT]
```

benjaminhwilliams: Indeed.

array_family/flex_ext.py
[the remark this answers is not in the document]
graeme-winter: On this specific question: my working assumption was that (h, k, l, e) makes a unique key -> it should therefore uniquely match from set 1 to set 2

However that is an assertion which I probably need to test on entry to this routine.

array_family/flex_ext.py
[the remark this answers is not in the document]
graeme-winter: I like the idea of avoiding the dance. I like it so much: should we add that to `flumpy` and work into this PR _before_ consider merging?

array_family/flex_ext.py
graeme-winter: I cargo culted in from other local areas; happy to update.

benjaminhwilliams (approved):
LGTM, though I think @rjgildea is right that [his suggestion](https://github.com/dials/dials/pull/1951#discussion_r769671763) for type hints is more readable.

ndevenish:
I somewhat agree that the type hint here could be a little clearer, but `FlexSizeT` doesn't seem that much better than aliasing to, say, `flex.size_t`, Which I don't think we do here specifically because we want to distinguish between different flavours of the `flex` module. It doesn't feel like a significant enough blocker to merging.

ndevenish:
Okay, I didn't see `Tuple[2 * (cctbx.array_family.flex.size_t,)]` :P

ndevenish:
`flumpy` was suggested, and is probably a bit more efficient, but happy to PR to change that separately.

PR #1952, opened 2021-12-16
Import mtz: update_experiments_and_reflection_table

dagewa:
This PR adds a function `update_experiments_and_reflection_table` to `dials.import_mtz` that calculates detector geometry from reflections and adds mm positions of the reflections to the table. This is intended to allow `dials.two_theta_refine` to run on data imported from MTZ without an image template (working towards allowing `xia2.multiplex` to run on such data). In order to do so, an additional change is made to `dials.two_theta_refine`, to use all reflections when we don't know have the `strong` flag.

I have not assessed to what extent this affects the accuracy of results from `dials.two_theta_refine` yet. But essentially this is the means to the end of getting usable results out of multiplex in the situation where only MTZ files, and not images, are available. There must be some level of compromise implicit in that, but it is not clear yet how much worse the results could be.

 I set this up as a PR rather than pushing to the `import_mtz` branch, as that branch was set up and therefore "owned" by @jbeilstenedmands

### DIFF
--- a/command_line/two_theta_refine.py
+++ b/command_line/two_theta_refine.py
@@ -176,7 +176,7 @@ class Script:
     def filter_integrated_centroids(reflections):
         """Filter reflections to include only those with the integrated and the
         strong flag set, but only if there are apparently some integrated
-        reflections"""
+        and strong reflections"""
 
         orig_len = len(reflections)
         mask = reflections.get_flags(reflections.flags.integrated)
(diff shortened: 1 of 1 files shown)


jbeilstenedmands (approved):

jbeilstenedmands:
Thanks @dagewa for picking this up. Confirmed that this change, along with https://github.com/xia2/xia2/pull/641, allows the multiplex processing to finish and looks ok at first glance, so merged in to allow further testing.

PR #1953, opened 2021-12-17
Use FuncFormatter to convert ds2 to d-spacings

rjgildea:
See also xia2/xia2#636


command_line/background.py
```diff
@@ -127,12 +128,7 @@ def run(args=None):
                     label = f"image {indx}" if len(images) > 1 else ""
                 ds2 = 1 / flex.pow2(d)
                 ax.plot(ds2, I, label=label)
-            xticks = ax.get_xticks().tolist()
-            ax.xaxis.set_major_locator(mticker.FixedLocator(xticks))
-            x_tick_labs = [
-                "" if e <= 0.0 else f"{math.sqrt(1.0 / e):.2f}" for e in xticks
-            ]
-            ax.set_xticklabels(x_tick_labs)
+                ax.xaxis.set_major_formatter(resolution_formatter)
```
benjaminhwilliams, line 131: I think the original indentation was correct.  We are putting all these plots on the same axes, so we only need to set the tick format once.
```suggestion
            ax.xaxis.set_major_formatter(resolution_formatter)
```

benjaminhwilliams (changes requested):
Just one suggestion, see inline comment below.

~Also, I notice that `fig` and `ax` are [defined early](https://github.com/dials/dials/blob/e3406ec5b47a2a937b84556ddd8f74a9f0cb08fa/command_line/background.py#L82-L83) in `dials.background` but only used if `params.output.plot` is truthy.  Should `fig = ...` and `ax = ...` be moved down to \~L117?~

Please ignore, I wasn't reading it thoroughly enough.

benjaminhwilliams (approved):

rjgildea:
> Also, I notice that `fig` and `ax` are [defined early](https://github.com/dials/dials/blob/e3406ec5b47a2a937b84556ddd8f74a9f0cb08fa/command_line/background.py#L82-L83) in `dials.background` but only used if `params.output.plot` is truthy. Should `fig = ...` and `ax = ...` be moved down to ~L117?

I think this is because the plots are added to `ax` within a loop over imagesets - so if multiple imagesets are passed, then they all appear on the same plot.

PR #1954, opened 2021-12-17
Use RawDescriptionHelpFormatter in ArgumentParser

rjgildea:
The default formatter line-wraps the description and epilog text.

Bug introduced by #1920


dagewa (approved):
LGTM :+1:

PR #1958, opened 2021-12-27
Fix Kapton Absorption GUI to properly draw absorption max/min lines

voklejas2:
Fixed inability to draw proper max absorption edge by replacing the 'map_coords' function inside DrawEdgeMax with a new class method, _map_coords. The new class method was based on a different sub-function 'map_coords' that was located inside the UpdateAbsorption function. For consistency, the UpdateAbsorption function was modified to use the '_map_coords' class method.

### DIFF
--- a/util/image_viewer/slip_viewer/kapton_2019_correction_frame_frame_plugin.py
+++ b/util/image_viewer/slip_viewer/kapton_2019_correction_frame_frame_plugin.py
@@ -1,0 +1,657 @@
+from __future__ import division
+# -*- Mode: Python; c-basic-offset: 2; indent-tabs-mode: nil; tab-width: 8 -*-
+#
+# $Id
+from six.moves import range
+
+"""
+Plugin to calibrate kapton correction parameters for an XFEL image
+"""
+
+import wx
+from dials.algorithms.integration.kapton_2019_correction import KaptonTape_2019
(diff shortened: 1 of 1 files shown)


util/image_viewer/slip_viewer/kapton_2019_correction_frame_frame_plugin.py
```diff
@@ -0,0 +1,657 @@
...
+      pass
+    dc.SetBrush(wx.Brush("black"))
+    dc.SetPen(wx.Pen("black"))
+    for segment in data:
+      lon1, lat1, lon2, lat2 = segment
+      x1, y1 = self._pyslip.ConvertGeo2View((lon1, lat1))
+      x2, y2 = self._pyslip.ConvertGeo2View((lon2, lat2))
+      dc.DrawLine(x1, y1, x2, y2)
+
+  def _map_coords(self, x, y, p):
```
voklejas2, line 396: This is the new class method that replaces two redundant functions both called "map_coords" inside the DrawMaxEdge and UpdateAbsorptionData.  This new class method is nearly identical to that in the UpdateAbsorptionData function but somewhat different from that in DrawMaxEdge.

voklejas2:
Moving this to CCTBX

PR #1966, opened 2022-01-11
Update ssx test name

jbeilstenedmands:
Update name following https://github.com/dials/data/pull/320


Anthchirp:
I have stopped the conda-forge deployment of dials-data updates, so would suggest to hold this until the DIALS release has gone out

jbeilstenedmands:
No problem, will leave this here in anticipation

Anthchirp:
Conda builds have resumed. I expect the new version will become available in an hour or so

PR #1969, opened 2022-01-13
Multiplex GUI

dagewa:
Update PHIL `short_caption`s and `help` strings for auto-generated ccp4i2 GUI.

See also https://github.com/xia2/xia2/pull/644


util/resolution_analysis.py
```diff
@@ -374,25 +377,31 @@ def miller_array_from_mtz(unmerged_mtz, anomalous=False, labels=None):
...
     .expert_level = 1
+    .short_caption = "Space group"
   reference = None
     .type = path
+    .short_caption = "Reference"
   emax = 4
     .type = float(value_min = 0)
-    .help = "Reject reflecitons with normalised intensities E^2 > emax^2"
+    .help = "Reject reflections with normalised intensities E^2 > emax^2"
+    .short_caption = "Maximum normalised intensity E"
```
rjgildea, line 404: "E" probably doesn't add anything here...
```suggestion
    .short_caption = "Maximum normalised intensity"
```

rjgildea (approved):
Thanks, these look sensible, although I do wonder if you could automatically determine a suitable `short_caption` in the case that none is explicitly set, e.g. capitalise the first character and replace `_` with ` `, i.e. `space_group` -> `Space group`, since a lot of the changes here are of the type:
```
space_group = None
  .type = space_group
  .short_caption = "Space group"
```

dagewa:
> I do wonder if you could automatically determine a suitable short_caption in the case that none is explicitly set

Yes, it turns out there's almost something for doing that already in the i2 code. But I would need to fix that up first and then return to DIALS and xia2 to strip out the trivial captions

PR #1972, opened 2022-01-13
Allow installation with Numpy 1.21.5+

Anthchirp:
as the underlying incompatibility issue was identified and worked
around. For more information see cctbx/cctbx_project#708.
Minimum version requirement set to the oldest currently supported
version according to NEP-29.

This is also a required step for Python 3.10 support.


ndevenish:
I'm happy that everything non-HTTP-related passed here to merge without rerunning

PR #1973, opened 2022-01-13
Allow building against HDF5 1.12

Anthchirp:
HDF5 1.12 has been out for nearly two years (February 2020), so I think we can start looking into running on it and see what breaks.

I welcome comments of caution 😄


graeme-winter:
This absolutely needs trying out, to find out what the impact is. Will investigate.

Anthchirp:
merging in main, so that investigations would cover current numpy+hdf5 versions

ndevenish:
This will get better testing in the main repo, and in any case the conda-forge package appears to be using HDF5 1.12 already.

ndevenish:
Correction: I had thought this was the dxtbx CI, in which case the merge title would have been correct. I apologies for the somewhat confusing mistake.

PR #1978, opened 2022-01-14
Only run docker build action on dials/dials repo

rjgildea:
So that it won't attempt to run on forked repos (where it likely won't have docker credentials available).


Anthchirp (approved):
looks plausible

.github/workflows/github-registry.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build-and-push-image:
+    if: ${{ github.repository == 'dials/dials' }}
```
graeme-winter, line 13: I assume that the syntax is correct -> this is helpful thank you

graeme-winter (approved):
Appreciate the fix :)

PR #1982, opened 2022-01-20
flex.reflection table: remove support for saving string data as messagepack

### LINKED ISSUE #1858: reflection_table std::string messagepack writing  is completely broken
It supports writing of `std::string` arrays https://github.com/dials/dials/blob/e129e6c3870e2d606ac8c39e65f2e5aa1027eed8/array_family/reflection_table_msgpack_adapter.h#L665-L666
but does so in a broken way; It uses the same "ref/versa" writing as all the standard objects:
https://github.com/dials/dials/blob/e129e6c3870e2d606ac8c39e65f2e5aa1027eed8/array_family/reflection_table_msgpack_adapter.h#L171-L173

This means that it writes to the messagepack file, an area of memory `n_items * sizeof(std::string)` starting from the `&const_ref<std::string>[0]`; which is not how std::strings work; e.g. this will definitely contain the first string (but not read it), and if you are lucky this will contain _some_ of the  subsequent strings, but basically you are reading random memory.

I don't know if this functionality was ever used for writing strings in the old `.pickle` format, but it's broken here and we've obviously not used it. Are there any examples of us wanting/needing to write string columns in the reflection tables?

If not, my suggestion is that we just drop the `versa<std::string>` support.

graeme-winter:
Remove the appearance of support for saving string columns to messagepack. This appears to work but it does not. 

Resolves #1858


graeme-winter:
Discussion in meeting concluded we should actually fix this -> closing

PR #1968, opened 2022-01-12
2nd round of modifications to improve/extend the Kapton correction tool.

voklejas2:
This round includes modifications to enable Kapton absorption correction
for higher angles of rotation of the Kapton tape.

Note: that in order to see the effects of these modifications, you must also change to cctbx_project branch 'katon_plugin_pt2'.

PR #1986, opened 2022-01-24
Update pre-commit environments

ndevenish:
Post-release, and also the recent PR https://github.com/dials/dials/pull/1968 flagged up a bug with the version of isort we were using (isort output depended on where you were running it from).


Anthchirp:
don't merge this yet

PR #1988, opened 2022-01-24
DIALS 3.8: Allow bootstrapping all supported python versions

ndevenish:
**This PR is into `dials-next`, so not into `main`**

This adds the capability into release `bootstrap.py` to use per-python-version lock lists, which means that it is now possible to bootstrap python versions 3.7, 3.8 and 3.9. Previously, it has only been possible to bootstrap the release versions e.g. DIALS 3.8 bootstrap could _only_ be Python 3.9. However, since DIALS 3.8 is an LTS release specifically for Python 3.7 support, we need to be able to continue testing with python 3.7.

Installation using Python 3.9 is identical to release, however Python 3.8 and Python 3.7 lock files have been generated post-release, so have some minor version differences - these should be in minor dependencies, because I explicitly pinned the most important packages to the version originally locked in the release of DIALS 3.8.


Anthchirp (approved):
lgtm

PR #1993, opened 2022-02-02
Fix dials.cosym target fail bug

jbeilstenedmands:
In `dials.cosym`, I found a case where a low resolution cutoff in the resolution filter caused at least one full dataset to be removed, leading to a stacktrace during the minimisation at:
```
>       assert (x.size // self.dim) == (len(self._lattices) * len(self.sym_ops))
E       AssertionError: Please report this error to dials-support@lists.sourceforge.net:

/dials/modules/dials/algorithms/symmetry/cosym/target.py:271: AssertionError
```
Of course a suitable d_min could just be set, but I still thought it would be good to avoid crashing out and handle this more gracefully.

I have added a test to reproduce this, and the fix is fairly straightforward, just making sure to use the correct number of datasets in the procedure (which could be reduced compared to upon initialisation). In the example test, only 7/8 datasets remain after the resolution filter, so only 7 reindexing operators can be determined. The only outstanding question is whether the filtered-out dataset should remain in the output `refl`, `expt` files. As no reindexing operator can be determined, I don't believe they should, so this proposed changeset removes them as they may not be consistently indexed.

p.s. I updated all the tests to use `pathlib` objects


command_line/cosym.py
```diff
@@ -232,6 +234,19 @@ def _apply_reindexing_operators(self, reindexing_ops, subgroup=None):
...
             )
             refl["miller_index"] = cb_op.apply(refl["miller_index"])
+        # Allow for the case where some datasets are filtered out.
+        if len(reindexing_ops) < len(self._experiments):
+            to_delete = []
+            for i in range(len(self._experiments)):
+                if i not in unique_ids:
+                    to_delete.append(i)
+            to_delete.sort(reverse=True)
+            for idx in to_delete:
```
rjgildea, line 244: It _might_ be slightly cleaner to sort inline (but this is just personal preference). It's not entirely obvious to me why the `reverse=True` is necessary here?
```suggestion
            for idx in sorted(to_delete):
```

tests/command_line/test_cosym.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import os
+from pathlib import Path
```
rjgildea, line 3: I generally prefer to import the module and then use `pathlib.Path` in the code then it is obvious where it comes from - I think this is generally how we are using `pathlib` throughout our source code.
```suggestion
import pathlib
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
 def test_cosym_partial_dataset(dials_data, run_in_tmpdir):
     """Test how cosym handles partial/bad datasets."""
-    mcp = dials_data("multi_crystal_proteinase_k")
+    mcp = dials_data("multi_crystal_proteinase_k", pathlib=True)
     args = []
     for i in [1, 2]:
-        args.append(mcp.join("experiments_%d.json" % i).strpath)
-        args.append(mcp.join("reflections_%d.pickle" % i).strpath)
+        args.append(str(mcp / f"experiments_{i}.json"))
+        args.append(str(mcp / f"reflections_{i}.pickle"))
```
rjgildea, line 56: I think `os.fspath` is generally recommended vs `str` in cases where this is necessary:
https://docs.python.org/3/library/os.html#os.fspath
Is it actually necessary here? If so, we should probably make the underlying dials/dxtbx file-handling code support `PathLike` objects (but outside the scope of this PR).
```suggestion
        args.append(os.fspath(mcp / f"experiments_{i}.json"))
        args.append(os.fspath(mcp / f"reflections_{i}.pickle"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
     # Make one dataset that will be removed in prefiltering
-    r = flex.reflection_table.from_file(mcp.join("reflections_8.pickle").strpath)
+    r = flex.reflection_table.from_file(str(mcp / "reflections_8.pickle"))
```
rjgildea, line 58:
```suggestion
    r = flex.reflection_table.from_file(os.fspath(mcp / "reflections_8.pickle"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
     r["partiality"] = flex.double(r.size(), 0.1)
     r.as_file("renamed.refl")
     args.append("renamed.refl")
-    args.append(mcp.join("experiments_8.json").strpath)
+    args.append(str(mcp / "experiments_8.json"))
```
rjgildea, line 62:
```suggestion
    args.append(os.fspath(mcp / "experiments_8.json"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
     # Add another good dataset at the end of the input list
-    args.append(mcp.join("experiments_10.json").strpath)
-    args.append(mcp.join("reflections_10.pickle").strpath)
+    args.append(str(mcp / "experiments_10.json"))
+    args.append(str(mcp / "reflections_10.pickle"))
```
rjgildea, line 65:
```suggestion
    args.append(os.fspath(mcp / "experiments_10.json"))
    args.append(os.fspath(mcp / "reflections_10.pickle"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
 
     dials_cosym.run(args=args)
-    assert os.path.exists("symmetrized.refl")
-    assert os.path.exists("symmetrized.expt")
+    assert Path("symmetrized.refl").is_file()
+    assert Path("symmetrized.expt").is_file()
```
rjgildea, line 69:
```suggestion
    assert pathlib.Path("symmetrized.refl").is_file()
    assert pathlib.Path("symmetrized.expt").is_file()
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
     experiments = load.experiment_list("symmetrized.expt", check_format=False)
     assert len(experiments) == 3
 
 
+def test_cosym_resolution_filter_excluding_datasets(dials_data, run_in_tmpdir):
+    mcp = dials_data("multi_crystal_proteinase_k", pathlib=True)
+    args = ["space_group=P4", "seed=0", "d_min=20.0"]
+    for i in [1, 2, 3, 4, 5, 7, 8, 10]:
+        args.append(str(mcp / f"experiments_{i}.json"))
+        args.append(str(mcp / f"reflections_{i}.pickle"))
```
rjgildea, line 79:
```suggestion
        args.append(os.fspath(mcp / f"experiments_{i}.json"))
        args.append(os.fspath(mcp / f"reflections_{i}.pickle"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
+    dials_cosym.run(args=args)
+    assert Path("symmetrized.refl").is_file()
+    assert Path("symmetrized.expt").is_file()
```
rjgildea, line 82:
```suggestion
    assert pathlib.Path("symmetrized.refl").is_file()
    assert pathlib.Path("symmetrized.expt").is_file()
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
+
+
 def test_cosym_partial_dataset_raises_sorry(dials_data, run_in_tmpdir, capsys):
     """Test how cosym handles partial/bad datasets."""
-    mcp = dials_data("multi_crystal_proteinase_k")
-    args = ["renamed.refl", mcp.join("experiments_8.json").strpath]
-    r2 = flex.reflection_table.from_file(mcp.join("reflections_10.pickle").strpath)
+    mcp = dials_data("multi_crystal_proteinase_k", pathlib=True)
+    args = ["renamed.refl", str(mcp / "experiments_8.json")]
+    r2 = flex.reflection_table.from_file(str(mcp / "reflections_10.pickle"))
```
rjgildea, line 98:
```suggestion
    args = ["renamed.refl", os.fspath(mcp / "experiments_8.json")]
    r2 = flex.reflection_table.from_file(os.fspath(mcp / "reflections_10.pickle"))
```

tests/command_line/test_cosym.py
```diff
@@ -49,37 +49,57 @@ def test_cosym(dials_data, run_in_tmpdir, space_group, engine):
...
     r2["partiality"] = flex.double(r2.size(), 0.1)
     r2.as_file("renamed2.refl")
     args.append("renamed2.refl")
-    args.append(mcp.join("experiments_10.json").strpath)
+    args.append(str(mcp / "experiments_10.json"))
```
rjgildea, line 102:
```suggestion
    args.append(os.fspath(mcp / "experiments_10.json"))
```

tests/command_line/test_cosym.py
```diff
@@ -148,10 +168,10 @@ def test_synthetic(
...
 
     dials_cosym.run(args=args)
-    assert os.path.isfile("symmetrized.refl")
-    assert os.path.isfile("symmetrized.expt")
-    assert os.path.isfile("cosym.html")
-    assert os.path.isfile("cosym.json")
+    assert Path("symmetrized.refl").is_file()
+    assert Path("symmetrized.expt").is_file()
+    assert Path("cosym.html").is_file()
+    assert Path("cosym.json").is_file()
```
rjgildea, line 174:
```suggestion
    assert pathlib.Path("symmetrized.refl").is_file()
    assert pathlib.Path("symmetrized.expt").is_file()
    assert pathlib.Path("cosym.html").is_file()
    assert pathlib.Path("cosym.json").is_file()
```

command_line/cosym.py
```diff
@@ -232,6 +234,19 @@ def _apply_reindexing_operators(self, reindexing_ops, subgroup=None):
...
+                logger.info(
+                    f"Removing dataset {idx} as unable to determine reindexing operator"
+                )
+                del self._experiments[idx]
+                del self._reflections[idx]
```
rjgildea, line 249: A slightly shorter, possibly cleaner, approach requiring only a single explicit loop:
```suggestion
        # Allow for the case where some datasets are filtered out.
        input_ids = set(range(len(self._experiments)))
        to_delete = sorted(input_ids - unique_ids)
        for idx in to_delete:
            logger.info(
                f"Removing dataset {idx} as unable to determine reindexing operator"
            )
            del self._experiments[idx]
            del self._reflections[idx]
```

rjgildea (approved):
Thanks for the bug fix and moving tests to use `pathlib.Path`. A few very minor comments about the pathlib usage, but otherwise looks good to me 👍

command_line/cosym.py
jbeilstenedmands: Thanks, the reason for `reverse=True` is to avoid issues when deleting in-place multiple experiments, reflections from the lists.

tests/command_line/test_cosym.py
ndevenish: In all these cases, it's obvious from reading that it's a Path object  or some kind of smart-path. While `os.fspath` is technically correct, it does make it look strange. My brain: "str - that's nice and obvious". "fspath - why is this necessary instead of str"

PR #1997, opened 2022-02-03
Bootstrap: Preserve standard packaging behaviour

ndevenish:
On request, https://github.com/cctbx/dxtbx/pull/474 changes the default installation behaviour so that standard python package installation will only be used when explicitly requested - the previous read-only `base/` fallback will now be the default. This means that we have to explicitly set this in DIALS bootstrap to preserve the existing behaviour.

This does that on configure, and checks for the marker file on `bootstrap.py update` and writes it if not present.

Users who do not use bootstrap to update their installations will need to manually `touch build/TBX_INSTALL_PACKAGE_BASE` before running `libtbx.refresh` in order to preserve the existing behaviour (and possibly to avoid a duplicate installation of dxtbx metadata?).


Anthchirp (approved):
Code-wise this looks reasonable.

Anthchirp:
On reflection: 
https://github.com/cctbx/dxtbx/pull/474#issuecomment-1029749842

ndevenish:
I think it's cleaner and simpler long term to just not do this. https://github.com/cctbx/dxtbx/pull/474#issuecomment-1032663296.

PR #1999, opened 2022-02-05
Avoid broken/untested pytest release

ndevenish:
It looks like pytest doesn't test on Python 3.9 because they release 7.0.0 and it was broken: https://github.com/pytest-dev/pytest/issues/9608


The-Compiler:
> It looks like pytest doesn't test on Python 3.9

That's incorrect - we [test Python 3.9 on CI](https://github.com/pytest-dev/pytest/actions/runs/1799100180) and have done so even before Python 3.9 was released. 

>  because they release 7.0.0 and it was broken

Only in very specific circumstances, which apparently nobody ran into when we asked people to try the prerelease.

PR #2000, opened 2022-02-06
Remove UUID

graeme-winter:
Use the `dxtbx.util.ersatz_uuid4` method in place of "proper" UUIDs - no change in behaviour expected beyond not importing `uuid` and hence not causing problems with large MPI ranks


graeme-winter:
Depends on cctbx/dxtbx/#477

graeme-winter:
Tests here will obviously fail until cctbx/dxtbx#477 is merged

PR #2002, opened 2022-02-08
Compute threshold with models

### LINKED ISSUE #2001: Spot-finding: `compute_threshold` can't use information about the experiment
According to https://dials.github.io/documentation/tutorials/developer_tutorial.html, new pixel thresholding methods can be created by implementing a `SpotFinderThresholdIface`. This has a `compute_threshold` method that just takes an `image` and a `mask`.

This means that at the point of `compute_threshold`, information about the experiment is not available. For example, it is not possible to identify the panel that the `image` came from and therefore we can't call methods like `panel.get_two_theta_array`. I would like to do this for a new spot-finding algorithm using a radial profile threshold.

dagewa:
Pass relevant experimental models into `compute_threshold` for future use.
    
The existing implementations of `compute_threshold` take `**kwargs`, which they ignore. A future implementation can make use of the `imageset`, panel number and ROI definition, all of which are needed to identify exactly which pixels we are trying to threshold. Fixes #2001.

### DIFF
--- a/algorithms/spot_finding/finder.py
+++ b/algorithms/spot_finding/finder.py
@@ -92,7 +92,7 @@ class ExtractPixelsFromImage:
         # Add the images to the pixel lists
         num_strong = 0
         average_background = 0
-        for im, mk in zip(image, mask):
+        for i_panel, (im, mk) in enumerate(zip(image, mask)):
             if self.region_of_interest is not None:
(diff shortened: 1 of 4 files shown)


ndevenish:
Thanks; this is useful - I have run into issues not having this information in the past.

PR #2009, opened 2022-02-11
radial_profile threshold algorithm

dagewa:
This PR adds an extension to calculate a radial profile threshold for use in `dials.find_spots`, and interactively in the `dials.image_viewer`.

This method calculates background value and sigma in 2θ shells, then sets a threshold at a level `n_sigma` above the radially-averaged background. As such, it is important to have the beam centre correct and to mask out any significant shadows. The 
method may be particularly useful for electron diffraction images, where there can be considerable inelastic scatter around low resolution spots, and it is difficult to tune parameters of the dispersion algorithms (see e.g. #1803). The algorithm is relatively insensitive to noise properties of the detector, which helps for the case of integrating detectors with poorly known gain and response statistics.

For example, default spot-finding on an electron diffraction image provided by a collaborator gives poor results, with many noise peaks both at the low resolution region around the backstop and at high resolution:
`dials.find_spots ../masked.expt`
disp-ext-default

Default spot-finding with the new method is much better:
`dials.find_spots ../masked.expt method=radial_profile`
radial-prof

A similar algorithm is available in other programs. The description of 'peakfinder 8' in https://doi.org/10.1107/S1600576714007626 was helpful in the development of this method, however the code is based on the approach taken in `dials.background`.

While `n_sigma` is the main parameter to optimise, a `blur=narrow|wide` option is also included. This helps to reduce noise peaks and to recombine split spots by convolving the image first with a Gaussian kernel.

Scope for future work includes:
- Optimisation for speed. The current implementation is rather slow, partly because it does a round of `dispersion` filtering first as part of determining the background pixels.
- Robust statistics. If noise peaks or shadows are included in the background determination, this will affect the threshold value. A possible alternative to an `n_sigma` cutoff would be to use the median and IQR of background values as robust measures of the central tendency and variation of background in shells. In this case, there might not be any need to pre-filter by the `dispersion` threshold, thus this approach may be faster (despite the expense of calculating the median in shells).
- Determination of best parameter defaults for various data sets.


extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if self.kernel:
+            image = convolve(image, self.kernel)
+
+        try:
+            imageset = kwargs["imageset"]
+            i_panel = kwargs["i_panel"]
+        except KeyError:
+            raise ValueError("Missing required arguments")
+        region_of_interest = kwargs.get("region_of_interest")
```
rjgildea, line 121: I think you can use the [keyword-only asterisk syntax](https://www.python.org/dev/peps/pep-3102/) to simplify this section:
```suggestion
    def compute_threshold(self, image, mask, *, imageset, i_panel, region_of_interest=None, **kwargs):
        """
        Compute the threshold.

        :param image: The image to process
        :param mask: The pixel mask on the image
        :**kwargs: Arbitrary keyword arguments
        :returns: A boolean mask showing foreground/background pixels
        """

        if self.kernel:
            image = convolve(image, self.kernel)
```

extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+        threshold = I + n_sigma * sig
+
+        # Shift the full 2θ array to the lower bound and truncate
+        infos = list(h0.slot_infos())
+        lower_bound = infos[0].low_cutoff
+        lookup = full_two_theta_array - lower_bound
+        lookup.set_selected(lookup < 0, 0)
+
+        # Truncate just under the shifted upper bound and rescale
+        upper_bound = infos[-1].high_cutoff - lower_bound
```
rjgildea, line 173: `slots` would probably be a more meaningful variable name than `infos` here:
```suggestion
        # Shift the full 2θ array to the lower bound and truncate
        slots = list(h0.slot_infos())
        lower_bound = slots[0].low_cutoff
        lookup = full_two_theta_array - lower_bound
        lookup.set_selected(lookup < 0, 0)

        # Truncate just under the shifted upper bound and rescale
        upper_bound = slots[-1].high_cutoff - lower_bound
```

extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Determine the threshold value for each bin
+        n_sigma = self.params.spotfinder.threshold.radial_profile.n_sigma
+        threshold = I + n_sigma * sig
+
+        # Shift the full 2θ array to the lower bound and truncate
+        infos = list(h0.slot_infos())
+        lower_bound = infos[0].low_cutoff
+        lookup = full_two_theta_array - lower_bound
+        lookup.set_selected(lookup < 0, 0)
```
rjgildea, line 170: What is `lookup` here? This probably justifies a more descriptive variable name, or at least a comment as to what it is a lookup for.

rjgildea: I see there is a comment a few lines below, but I think it would be helpful to have a meaningful name or comment the first time it is declared.

tests/command_line/test_find_spots.py
```diff
@@ -377,3 +377,36 @@ def test_find_spots_with_per_image_statistics(dials_data, tmpdir):
...
+    assert len(reflections) == 554
+
+
+def test_find_spots_radial_profile_with_blur(dials_data, tmpdir):
+    result = procrunner.run(
+        [
+            "dials.find_spots",
+            "nproc=1",
+            "threshold.algorithm=radial_profile",
+            "blur=narrow",
```
rjgildea, line 402: Would it be worth parametrising this test in order to test both `blur=narrow` and `blur=wide`?
```suggestion
            "blur=narrow",
```

command_line/image_viewer.py
```diff
@@ -83,8 +83,14 @@
...
   .type = float(value_min=0)
-  .help = "Set gain for the thresholding algorithm. This does not override the"
+  .help = "Set gain for the dispersion algorithm. This does not override the"
           "detector's panel gain, but acts as a multiplier for it."
+n_sigma = 8
+  .type = float(value_min=0)
+blur = narrow wide
+  .type = choice
+n_bins = 100
+  .type = int
```
rjgildea, line 93: Could these be shared with the definitions in the algorithm itself? So that we can benefit from the help strings here, as well as ensuring that both sets of parameters remain in sync.

extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+    A similar algorithm is available in other programs. The description of
+    'peakfinder 8' in https://doi.org/10.1107/S1600576714007626 was helpful
+    in the development of this method.
+    """
+
+    name = "radial_profile"
+
+    @staticmethod
+    def phil():
+        from libtbx.phil import parse
```
rjgildea, line 30: Is there a specific reason for the inline import here?

tests/command_line/test_find_spots.py
```diff
@@ -377,3 +377,36 @@ def test_find_spots_with_per_image_statistics(dials_data, tmpdir):
     assert (
         b"|   image |   #spots |   #spots_no_ice |   total_intensity |" in result.stdout
     )
+
+
+def test_find_spots_radial_profile(dials_data, tmpdir):
+    result = procrunner.run(
```
rjgildea, line 383: Could these tests call the `dials.command_line.find_spots.run()` method directly then we can skip the additional `procrunner` dependency/indirection, as well as removing the need to explicitly check for errors?
https://github.com/dials/dials/blob/54cd6499dcd8af113bb138756bffc89297a2463d/command_line/find_spots.py#L239

extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+        Initialise the algorithm.
+
+        :param params: The input parameters
+        """
+        self.params = params
+
+        # Set approximate Gaussian kernel for blurring
+        if self.params.spotfinder.threshold.radial_profile.blur == "narrow":
+            self.kernel = flex.double(
+                (0.0625, 0.125, 0.0625, 0.125, 0.25, 0.125, 0.0625, 0.125, 0.0625)
```
rjgildea, line 63: Possibly this 3x3 array would also benefit from the formatting suggested below.

rjgildea (approved):
It's interesting to see a new spotfinding algorithm in DIALS, it will be exciting to see how it performs on a variety of data!

I had a quick play with it in the image viewer - looking at the standard x4wide dataset (relatively high background by modern standards) it seemed to do quite well. However, when picking a random ProtK dataset from I03 with low background (mostly 0s and 1s) it appeared to find a lot of signal in the background region according to the threshold map. There is also a very clearly defined change cutoff at around 1.3 Å? I guess these low-background counting detector datasets aren't the kind of data that the algorithm is designed for, but I wonder if there are any potential improvements to handle this kind of data better?
Screenshot 2022-02-11 at 17 47 29

I also noticed that changing blur in the image viewer didn't appear to have any effect - is this to be expected?

I think it would be worth adding a more comprehensive description of the algorithm to the newsfragment for inclusion in the release notes.

dagewa:
> However, when picking a random ProtK dataset from I03 with low background (mostly 0s and 1s) it appeared to find a lot of signal in the background region according to the threshold map. There is also a very clearly defined change cutoff at around 1.3 Å? I guess these low-background counting detector datasets aren't the kind of data that the algorithm is designed for, but I wonder if there are any potential improvements to handle this kind of data better?

That's interesting... could you send me the path to that dataset? I'd like to take a look

rjgildea:
> That's interesting... could you send me the path to that dataset? I'd like to take a look

Can you see this?
`/dls/i03/data/2022/cm31105-1/TestProteinaseK/protk_1/protk_1_1.nxs`

dagewa:
> I also noticed that changing blur in the image viewer didn't appear to have any effect - is this to be expected?

Yes, the blur happens internally, when the image is passed in the `compute_threshold` function, so you don't get to see the effect in the image viewer. I was thinking that if blurring has any general use in spot-finding, it could potentially go at a higher level - i.e. before any thresholding algorithm is applied. Alternatively, I also thought about removing all the buttons for different image types when `radial_profile` is selected, except `image` and `threshold`, and then perhaps adding in `blur` and `background` buttons. I wasn't sure about doing too much image viewer work within this PR though.

extensions/radial_profile_spotfinder_threshold_ext.py
dagewa: Copied from `DispersionSpotFinderThresholdExt`. I'll float them both

extensions/radial_profile_spotfinder_threshold_ext.py
```diff
@@ -0,0 +1,186 @@
...
+                        26,
+                        16,
+                        4,
+                        1,
+                        4,
+                        7,
+                        4,
+                        1,
+                    )
+                )
```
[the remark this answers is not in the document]
dagewa: TIL

extensions/radial_profile_spotfinder_threshold_ext.py
dagewa: There _might_ be a valid reason for doing this for extensions. @ndevenish might anyway look at running an auto-float script across `dials/`. I'll leave it as-is here then.

tests/command_line/test_find_spots.py
dagewa: ok, got it

dagewa:
Thanks @rjgildea for the review. I've addressed your comments, minus the import floating issue. I've also changed the algorithm to use robust statistics (median and IQR) rather than mean and standard deviation. This means I could get rid of the initial dispersion method step that was used to determine which pixels were in the background region. The result is not actually faster though, because of the expense of calculating the median. There are more efficient methods for this (like `median_functor` from cctbx), but this does not immediately give access to the quartiles as well (for the IQR calculation). I've stuck with the naive calculation involving a full sort for now.

By setting a minimum value for the threshold, the algorithm is now better behaved for very low background images, like the ProtK data you shared. However, you are right, this algorithm is not designed for such cases. In fact, in this case it acts as an overly-expensive global threshold filter only:
image

though, `blur` does seem to have some benefit for reduction of noise peaks:
image

extensions/radial_profile_spotfinder_threshold_ext.py
ndevenish: Yes, I think the floating scripts were never run on DIALS. It probably should at some point.

biochem-fan:
@dagewa
When processing a SACLA SFX dataset, I came across many false positives with your radial_profile method. This happened when the user inserted a very thick attenuator by mistake, leading to very low background images like the one @rjgildea mentioned above. After the user changed the attenuator, the algorithm worked well, much better than the non-radial default algorithm, because images had rings from the carrier medium. Thus, for right attenuators, the algorithm works fine but I would like to see if there are any options that make it more robust against strange images.

> By setting a minimum value for the threshold, the algorithm is now better behaved for very low background images

How do I set this minimum value? Is it `dispersion.global_threshold`? I am confused because you also wrote the implementation no longer used the dispersion filter. And what is the unit, raw pixel values, (gain) corrected pixell values or n_sigma?

dagewa:
Yes, it would be nice if this algorithm worked better for low background cases, or perhaps had a special mode for this.

Sorry, my comment above was vague. I did not mean to suggest that there is a user option to adjust the threshold, but that I hardcoded a minimum:

https://github.com/dials/dials/blob/ca46d4b7d8a57a809832a3bdfd4bebd1603a9784/extensions/radial_profile_spotfinder_threshold_ext.py#L129-L136

The value `adu` is the gain-corrected "count" for a single quantum. If the threshold above median is only 1 `adu` or less, then it is set to 2 `adu`. Maybe there is a better way. Perhaps for very weak backgrounds we should not use robust measures like median and interquartile range, but should use mean and sigma. The median estimates the weak background to be zero, which in this case is unhelpful.

PR #2016, opened 2022-02-21
Powder calibrate with widget

elena-pascal:
Use pyFAI calibrate for electron powder diffraction. 
Since pyFAI was developed for X-rays the expected ring density is lower than that of electron powder patterns. To circumvent this I proposed using a matplotlib widget to guesstimate an initial beam geometry.

Would be worth in the future to look at pyFAI's spot picker instead to feed the algorithm the correct ring instead.


command_line/powder_calibrate.py
```diff
@@ -0,0 +1,467 @@
...
+        __import__(module_name)
+    except ModuleNotFoundError as err:
+        if module_name == "circle_fit":
+            module_name = "circle-fit"
+        print(
+            err,
+            f"""
+            This script requires the {module_name} library. Try:
+
+                conda install -c conda-forge {module_name}
```
dagewa, line 38: I gave up with this after some minutes and used `mamba` instead. Perhaps we should push the user to `conda install` `mamba` first and then `mamba install -c conda-forge pyFAI`?

command_line/powder_calibrate.py
```diff
@@ -0,0 +1,467 @@
...
+        else:
+            pass
+
+
+if __name__ == "__main__":
+
+    Al_data = DialsData(
+        expt_file="/home/elena/Work/Diamond/2021/Al_data/Yun_cameralength_cal/cameralength_cal/Al-standard_0p67_5s/imported.expt",
+        refl_file="/home/elena/Work/Diamond/2021/Al_data/Yun_cameralength_cal/cameralength_cal/Al-standard_0p67_5s/strong.refl",
+    )
```
dagewa, line 461: When I ran `dials.powder_calibrate` with no arguments I got
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/elena/Work/Diamond/2021/Al_data/Yun_cameralength_cal/cameralength_cal/Al-standard_0p67_5s/imported.expt'
```
Would prefer for new programs to use [the boilerplate](https://github.com/dials/dials/wiki/Writing-a-DIALS-program) to get consistent presentation of help messages, option switches etc.

dagewa (changes requested):
I'm not sure how to run `dials.powder_calibrate` and `dials.powder_calibrate_widget`. As noted on the in-line comment, could these programs use the boilerplate, and exit gracefully with a suitable help message when the input is missing or wrong?

I'm also not very keen on the name `dials.powder_calibrate_widget`. Could this just be a single command-line program, `dials.powder_calibrate`, with the interactive option being chosen by a parameter? I note you have `eyeball=True` in the PHIL parameters for `dials.powder_calibrate_widget`. When this is false, does this just operate as `dials.powder_calibrate`? In that case, it could be false by default and the user can switch on interactivity when required.

I might be missing something, because I did not successfully run the programs. I'll continue to review upon further instruction.

command_line/powder_calibrate.py
elena-pascal: My bad, `dials.powder_calibrate` was not meant to be in this branch at all.

PR #2023, opened 2022-03-03
Remove scan floating point comparisons

toastisme:
This just changes some implicit stills Scan checks to explicitly use `Scan.is_still()`.


toastisme:
Corresponding changes made in dxtbx in [495](https://github.com/cctbx/dxtbx/pull/495)

PR #2028, opened 2022-03-07
Run test on windows

dagewa:
Make `test_dials_cluster_unit_cell_command_line_output_files` test work on Windows.

- Avoid "command line too long errors" by writing `input.phil`
- Also use `reflections.pickle` rather than `indexed.pickle` to match other tests using this data (to reduce new dependencies on files in `dials_regression`)


jbeilstenedmands (approved):
Looks good, thanks.

Anthchirp:
Could we write a phil file in all cases? That would keep the test simpler

PR #2031, opened 2022-03-09
Parallelise outlier rejection in refinement

### LINKED ISSUE #1427: Parallelise outlier rejection in refinement
Outlier rejection in refinement is done independently for reflections split by experiment (if `seperate_experiment=True`), by panel (if `separate_panels=True`) and by phi (depending on `block_width`). This can potentially lead to a large number of independent outlier rejection jobs, especially in the case of multi-panel detectors, which can significantly increase (and dominate) the runtime (see #1424). This seems an embarrassingly parallel task for which it should be possible to get significant performance increases.

https://github.com/dials/dials/blob/414c84f0b7cde2effcc50b3af2ed813da9c6c05d/algorithms/refinement/outlier_detection/outlier_base.py#L74-L270

dagewa:
Fixes #1427 

The new PHIL parameter `outlier.nproc` is set to `1` by default when using the `Refiner` as a library, but overridden to be `Auto` in `dials.refine`. I'll tag @rjgildea and @phyy-nx for review, as they may also want to set `outlier.nproc` to something other than `1` in some uses of refinement in `dials.index`, `stills_process` etc.

### DIFF
--- a/algorithms/refinement/outlier_detection/mcd.py
+++ b/algorithms/refinement/outlier_detection/mcd.py
@@ -17,6 +17,7 @@ class MCD(CentroidOutlier):
         separate_experiments=True,
         separate_panels=True,
         block_width=None,
+        nproc=1,
         alpha=0.5,
         max_n_groups=5,
         min_group_size=300,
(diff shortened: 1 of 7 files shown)


phyy-nx:
Thanks. For stills_process, the multiprocessing is indeed handled by the client/server dispatcher so we'll keep it at nproc=1 for outlier rejection.  I look forward though to trying multiprocessing when we do big dials.refine jobs of thousands of stills simultaneously.

phyy-nx (approved):

PR #2034, opened 2022-03-17
Avoid Scan dereference in parallel_integrator and parallel_reference_profiler

toastisme:
This changes explicit `Scan` dereferences in parallel_integrator.h and parallel_reference_profiler.h to instead call methods from a `ScanBase` pointer. This only make sense if [503](503) is merged.


ndevenish:
If cctbx/dxtbx#503 is being approached differently, does this PR also need to be reworked?

toastisme:
Closing this as it is no longer needed ([different approach](https://github.com/cctbx/dxtbx/pull/620) with Scan objects now taken compared to [503](https://github.com/cctbx/dxtbx/pull/503))

PR #2036, opened 2022-03-18
`separate_images` outlier rejection

dagewa:
Special case of outlier rejection, discussed on dials-support Slack. FAO @PrinceWalnut & @kmdalton.

This is spun off from the https://github.com/dials/dials/pull/2031 branch to reduce future merge conflicts.

### DIFF
--- a/algorithms/refinement/outlier_detection/outlier_base.py
+++ b/algorithms/refinement/outlier_detection/outlier_base.py
@@ -202,13 +202,19 @@ class CentroidOutlier:
         header.extend(["Nref", "Nout", "%out"])
         rows = []
 
-        # now loop over the lowest level of splits and run outlier detection
-        with concurrent.futures.ProcessPoolExecutor(max_workers=self.nproc) as pool:
-            futures = [
-                pool.submit(self._run_job, job, i) for i, job in enumerate(jobs3)
+        # Now loop over the lowest level of splits and run outlier detection
+        if len(jobs3) > 1:
+            with concurrent.futures.ProcessPoolExecutor(max_workers=self.nproc) as pool:
+                outlier_detection_runs = [
+                    pool.submit(self._run_job, job, i) for i, job in enumerate(jobs3)
+                ]
+            outlier_detection_runs = [e.result() for e in outlier_detection_runs]
+        else:
+            # If there's a single job keep it in the main process
+            outlier_detection_runs = [
+                self._run_job(jobs3[0], 0),
             ]
-        for f in futures:
-            result = f.result()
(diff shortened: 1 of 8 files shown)


algorithms/refinement/outlier_detection/outlier_base.py
```diff
@@ -438,14 +508,31 @@ def from_parameters_and_colnames(params, colnames):
...
+                params.outlier.separate_blocks = False
+            if params.outlier.separate_experiments:
+                logger.info("Resetting separate_experiments=False")
+                params.outlier.separate_experiments = False
+
         if not params.outlier.separate_blocks:
             params.outlier.block_width = None
+
+        if params.outlier.nproc is libtbx.Auto:
+            params.outlier.nproc = available_cores()
```
phyy-nx, line 525: Hi, I'm excited by this pull request.  Does this line mean that dials.stills_process will use available_cores() when doing outlier rejection?  This doesn't work in MPI multiprocessing, where each process would independently call available_cores and then use too many.  But I haven't tested it :)

dagewa: Hi @phyy-nx, the PR for _that_ change specifically is https://github.com/dials/dials/pull/2031, where you are tagged for review :) (I appreciate you are in beam time mode so won't have had chance to look).

In #2031, `dials.refine` is changed to automatically select `nproc` for outlier rejection, but that default is restricted to the command line program. Use of the `Refiner` as a library has unchanged defaults. In fact, when `nproc=1` the outlier rejection is done in the main process. I found that was necessary to get identical results for those outlier algorithms that use random sampling (like `mcd`), plus it avoids overhead of setting up a subprocess.

phyy-nx: Thanks @dagewa for clarifying!  Not surprising you considered this.  Indeed beamtimes :)

PrinceWalnut:
@dagewa, I've run this code on our data + analyses and, upon inspection of the experiment, reflection, and log files, it appears that everything has worked appropriately. The survivorship bias we were seeing in our data has also disappeared now that systematic errors in regions of the dataset are being addressed appropriately. This solution works for us. I do have one question: in the log files, there are "Outlier rejections per job," with the same number of jobs as images. Are these jobs representative of the same number image, or is this a glob-like random ordering of images?

dagewa:
Good, I'm glad this is working for you. The jobs might not be run in order, but the results are reconstructed in order so that table is sequential for the images in your dataset. As such, it should match your image number, if your dataset starts with image 1.

phyy-nx (approved):

PR #2043, opened 2022-03-23
Try to play nice with phils

elena-pascal:
Unify phil usage in main class

### DIFF
--- a/command_line/powder_calibrate.py
+++ b/command_line/powder_calibrate.py
@@ -150,18 +150,6 @@ def parse_to_tuples(
 
     params, options = parser.parse_args(args=args, show_diff_phil=False)
 
-    # Configure the logging.
-    dials.util.log.config(options.verbose, logfile=params.output.log)
-
-    # Log the dials version
-    logger.info(dials_version())
-
-    # Log the difference between the PHIL scope definition and the active PHIL scope,
-    # which will include the parsed user inputs.
-    diff_phil = parser.diff_phil.as_str()
-    if diff_phil:
-        logger.info("The following parameters have been modified:\n%s", diff_phil)
-
     # populate experimental parameters
     experiments = flatten_experiments(params.input.experiments)
 
@@ -600,15 +588,12 @@ class PowderCalibrator:
                 "pyfai_improvement=" + pyfai_improvement,
                 "straight_lines=" + straight_lines,
             ]
-            self.expt_params, self.user_args = parse_to_tuples(args=args)
-        elif not starting_geom and not standard:
-            # command line usage
-            self.expt_params, self.user_args = parse_to_tuples()
-        elif not standard:
+        if not standard:
             exit("Standard missing")
-        elif not starting_geom:
+        if not starting_geom:
             exit("Starting geometry missing")
 
+        self.expt_params, self.user_args = parse_to_tuples(args=args)
         self.geometry = Geometry(self.expt_params)
         self.detector = self.geometry.detector
 
@@ -670,7 +655,7 @@ class PowderCalibrator:
             unit="q_nm^-1",
         )
         fig, ax = plt.subplots()
-        fig.set_size_inches(8, 8)
+        fig.set_size_inches(10, 10)
         pfjupyter.plot2d(int2, calibrant=self.calibrant, ax=ax)
         plt.title("Are those lines straight?")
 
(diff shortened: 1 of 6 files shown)


elena-pascal:
I think there was an easy way and a hard way to do that merge and it seems I choose the hard way

command_line/powder_calibrate.py
```diff
@@ -723,15 +697,52 @@ def calibrate_with_calibrant(
...
+
+    # Log the dials version
+    logger.info(dials_version())
+
+    # Log the difference between the PHIL scope definition and the active PHIL scope,
+    # which will include the parsed user inputs.
+    diff_phil = parser.diff_phil.as_str()
+    if diff_phil:
+        logger.info("The following parameters have been modified:\n%s", diff_phil)
+
```
dagewa, line 735: If run without command-line arguments it's nice for the program to provide basic help:
```suggestion
    if len(experiments) == 0:
        parser.print_help()
        return
```

dagewa: Similarly, `standard="something"` is a mandatory argument. With the default `standard=None` the program exits with a stacktrace. Could this be a nice apologetic message instead?

dagewa:
The slider widget does not seem to be working for me. I get 
```
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.9/site-packages/matplotlib/cbook/__init__.py", line 287, in process
    func(*args, **kwargs)
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.9/site-packages/matplotlib/widgets.py", line 585, in <lambda>
    return self._observers.connect('changed', lambda val: func(val))
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/powder_calibrate.py", line 365, in update
    new_geometry = self.geometry.__deepcopy__()
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/powder_calibrate.py", line 300, in __deepcopy__
    new = self.__class__(self.expt_params)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/powder_calibrate.py", line 199, in __init__
    self.expt_params = get_expt_params(expt)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/command_line/powder_calibrate.py", line 145, in get_expt_params
    detector = expt.detector[0]
AttributeError: 'str' object has no attribute 'detector'
```
whenever I change the slider.

elena-pascal:
Hmn those are unhappy widget noises. I'll have a look soon 👌

command_line/powder_calibrate.py
elena-pascal: I had a dream that I did cover those bases. I think I have officially looked too long at this code.

elena-pascal:
Ok, I am done with this version. There will be a future version that does detector distance eyeballing but that will be a separate PR.

dagewa:
I saw a warning printed after doing the widget calibration. Might be worth looking at...
```
Starting calibration using Al:
-----
Al Calibrant with 22 reflections at wavelength 2.5082356047905177e-12

Using the eyeballing widget.
Drag sliders to roughly overlap rings and then save for further fitting. 

No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.
Geometry fitted by pyFAI:
-----

```

elena-pascal:
Yes, ignore the warning I've send [a patch](https://github.com/silx-kit/pyFAI/issues/1665) to pyFAI for this, should go away soon. It is also inconsequential.

dagewa:
There's a typo in the default filename for one of the output files
```
pyfai_improvement = pyfai_imporovement.png
```
it's outside the PR changes so I can't make a commit suggestion, I think

dagewa:
What happens if the geometry has the wrong distance as well as beam centre? You can align the centre of the rings with the widget, but they would have the wrong size. Is that a problem, or does the automated PyFAI calibration then fix that?

dagewa (approved):
There's a couple of suggestions here, but the program ran as expected so I'm approving

command_line/powder_calibrate.py
```diff
@@ -723,15 +756,59 @@ def calibrate_with_calibrant(
...
+    parser = ArgumentParser(
+        usage=usage, phil=phil, read_experiments=True, check_format=True, epilog=__doc__
+    )
+
+    parameters, options = parser.parse_args(args=args, show_diff_phil=False)
+
+    experiments = flatten_experiments(parameters.input.experiments)
+
+    # Check user is holding the tool right
+    if experiments == 0 or not parameters.standard:
```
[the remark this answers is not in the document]
Anthchirp: Out if interest - why not `if not experiments`?

elena-pascal:
There are 7 parameters that can be fit: beam position x, beam position y, distance, wavelength (these are coupled for ED), 3 axis detector tilt. For the fit to give meaningful results the starting values ought to be not very far. For the data I have from Yun the distance is not too bad, so it fits well after getting a better estimate for beam position. But the data from Adam has completely meaningless distance information. Which is why I want to put a PR for eyeballing the beam distance as well.

elena-pascal:
Ah, I was hoping the pfps will show up next to the commit rather than just have the co-author names in the commit text.

PR #2044, opened 2022-03-24
Test less geometry

dagewa:
When importing MTZ we have to make some assumptions about the geometry. As a result, we can't test the exact goniometer and detector orientation. But we should still be able to test detector distance and beam centre. The required detector distance tolerance is a bit surprising. Need to check what happens to the detector distance when exporting and then importing from MTZ. Feels like something that should be kept precisely...

Switch off `dials.two_theta_refine` when running `xia2.multiplex`. This is what I will also add to the i2 interface


jbeilstenedmands (approved):
Thank you, looks good!

PR #2048, opened 2022-03-30
Fix dials.show for unknown columns of type vec3_double or miller_index

phyy-nx:
Found by running dials.show on reflection tables [modified by cctbx.xfel.merge](https://github.com/cctbx/cctbx_project/blob/master/xfel/merging/application/model/crystal_model.py#L312-L317) to include a column named `miller_index_asymmetric` of type `flex.miller_index`.


dagewa (approved):
Looks like the right fix for now, but I note `dials.show` has slightly surprising behaviour for 3-vector columns. For example,
```
>>> from cctbx.array_family import flex
>>> foo=flex.vec3_double([(1,2,3),(3,2,1),(1,3,2)])
>>> foo.max()
(3.0, 3.0, 3.0)
```
The max is done on columns separately, so you get a value that is not actually present in the array. This is a bit odd if for example you are looking at Miller indices. I'm not sure how best to approach this, but I appreciate this PR is just fixing a bug, not changing that behaviour.

phyy-nx:
Agree the max of a miller_index array doesn't seem well defined.  But ya, this fixes a specific bug so merging as is.  Thanks.

rjgildea:
> Looks like the right fix for now, but I note `dials.show` has slightly surprising behaviour for 3-vector columns. For example,
> 
> ```
> >>> from cctbx.array_family import flex
> >>> foo=flex.vec3_double([(1,2,3),(3,2,1),(1,3,2)])
> >>> foo.max()
> (3.0, 3.0, 3.0)
> ```
> 
> The max is done on columns separately, so you get a value that is not actually present in the array. This is a bit odd if for example you are looking at Miller indices. I'm not sure how best to approach this, but I appreciate this PR is just fixing a bug, not changing that behaviour.

I'm not sure that this behaviour is entirely surprising - it is consistent with numpy `array.max(axis=0)`, which makes sense when you're dealing with primarily columnar data:
```
import numpy as np
>>> a = np.array([(1,2,3),(3,2,1),(1,3,2),(1,1,1)])
>>> a.max()
3
>>> a.max(axis=0)
array([3, 3, 3])
>>> a.max(axis=1)
array([3, 3, 3, 1])
```

phyy-nx:
👍

PR #2056, opened 2022-04-04
dials.show without imageset

dagewa:
`dials.show` was crashing when passed an experiment list that had an empty imageset (for example, from `dials.import_mtz`). In that case skip the display of format class and template.

### DIFF
--- a/command_line/show.py
+++ b/command_line/show.py
@@ -251,16 +251,14 @@ def show_experiments(experiments, show_scan_varying=False):
     text = []
     for i_expt, expt in enumerate(experiments):
         text.append("Experiment %i:" % i_expt)
+        template = None
         if expt.imageset:
             format_class = expt.imageset.get_format_class()
+            template = expt.imageset.get_template()
             if not format_class.is_abstract():
                 text.append(f"Format class: {format_class.__name__}")
-            if expt.identifier != "":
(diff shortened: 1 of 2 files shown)


command_line/show.py
```diff
@@ -251,15 +251,14 @@ def show_experiments(experiments, show_scan_varying=False):
     text = []
     for i_expt, expt in enumerate(experiments):
         text.append("Experiment %i:" % i_expt)
-        format_class = expt.imageset.get_format_class()
-        if not format_class.is_abstract():
-            text.append(f"Format class: {format_class.__name__}")
+        template = None
+        if expt.imageset:
+            format_class = expt.imageset.get_format_class()
+            template = expt.imageset.get_template()
```
rjgildea, line 257: Is `get_template()` defined for _all_ `ImageSet`s, or is it just an `ImageSequence` thing? If there wasn't explicit support already for `expt.imageset` being `None` then I suspect the `try`/`except AttributeError` was there for a reason...

rjgildea: History repeats itself 😄 
https://github.com/dials/dials/pull/1133#discussion_r374706496

dagewa: well spotted :-) I'll put that back then

PR #2070, opened 2022-04-10
Explicitly configure xfel before dials in bootstrap

ndevenish:
Currently, the module order between xfel/dials (as recorded in libtbx_env and the `build/SConstruct`) is dependent on the order of configuration - configuring `"dials [xfel]"`[^1] means that xfel is listed first in the module order, and configuring `"xfel [dials]"` means that dials ends up first in the module order.

This is because libtbx doesn't topologically sort dependencies[^2], it only checks to see if the dependency is already present in the module list. Since xfel and dials have a circular dependency, if the dials module is configured before xfel, then when [processing](https://github.com/cctbx/cctbx_project/blob/dc9c18d354d9d395f2381c7ed7c86bb9bc034060/libtbx/env_config.py#L2399-L2411) the dependencies of dials the xfel module is explicitly [inserted](https://github.com/cctbx/cctbx_project/blob/dc9c18d354d9d395f2381c7ed7c86bb9bc034060/libtbx/env_config.py#L755-L766) before dials in the module list. If xfel is configured before dials, then the inverse happens - xfel is added to the module list, then the dials module is registered, and inserted before the xfel module.

This doesn't currently cause any issues because xfel assumes that it is first and sets up `env_etc.dials*` in it's [`xfel/SConscript`](https://github.com/cctbx/cctbx_project/blob/dc9c18d354d9d395f2381c7ed7c86bb9bc034060/xfel/SConscript#L7-L8). These are then replaced in the [`dials/SConscript`](https://github.com/dials/dials/blob/cf62107c7d843c139868270309ca52089ad51dad/SConscript#L10-L11). This obviously means that there is no way that DIALS can influence or change these variables, which are currently defined in two places.

This PR simply makes the DIALS bootstrap explicitly configure xfel, instead of waiting for it to be implicitly pulled in. This means that the xfel dependency list is evaluated first, so dials gets places first in the module and `SConstruct` evaluation order, meaning that the DIALS `SConscript` will get to configure the variables first. The xfel module overwriting the variables is resolved by https://github.com/cctbx/cctbx_project/pull/753.

[^1]: The `[module]` implies that specifying the `module` isn't necessary - it's implicitly pulled in through the `module/libtbx_config` dependency specification.

[^2]: It's actually impossible to topologically sort the total web of dependencies, including option modules, because there are several circular dependencies in the web of cctbx_project/dials requirements. If you want to build a DAG, then you need to e.g. [carefully break](https://github.com/ndevenish/tbxtools/blob/main/src/tbxtools/tbx2cmake/read_scons.py#L264-L273) the dependencies until the dependency graph is acyclic. Thus, this is possible breaking only a couple of `optional_modules` edges.


phyy-nx:
👍

PR #2071, opened 2022-04-11
Allow bootstrap update on git worktree checkouts

ndevenish:
These have a `.git` in the root repository folder that is a file pointing to the real checkout. Rather than try and work out what to do, just assume that the user knows what they are doing in this scenario.


Anthchirp (approved):
Sensible, although the message is a bit misleading then

ndevenish:
Okay, so I fixed the message, with the assumption that if you have a `.git` file then that's what is going on - if you have a non-worktree git checkout where the `.git` file exists then I guess you'll have to live with the confusion 🤷 

I've also fixed precommitbx which also fell over on the same situation.

PR #2073, opened 2022-04-12
Remove magic number in dials integrate

graeme-winter:
Include derivation to aid future reader


benjaminhwilliams (approved):
LGTM, thanks.

graeme-winter:
Actually, hold the line. I have a small suspicion that the magic number is the wrong magic number (not by much) which I would like to check...

graeme-winter:
Graeme's hit-list += `EPS = 1e-7`

graeme-winter:
> tests/algorithms/profile_model/ellipsoid/test_model.py

I think I got all these

PR #2075, opened 2022-04-12
Calibrate distance in widget as well

elena-pascal:
For very poor knowledge of detector distance one needs to start by guesstimating the distance.


command_line/powder_calibrate.py
```diff
@@ -349,20 +363,23 @@ def __init__(
...
-        self.beam_slow_slider = self.make_slider("slow")
+
+        self.beam_fast_slider = self._make_slider("fast")
+        self.beam_slow_slider = self._make_slider("slow")
+        self.distance_slider = self._make_slider("distance")
 
     def __repr__(self):
         calibrant = os.path.splitext(os.path.basename(self.calibrant.filename))[0]
-        return f"Eyeballing Widget using {calibrant} Calibrant starting from Geometry: \n {self.geometry}"
+        return f"Eyeballing Widget instance using {calibrant} Calibrant starting from Geometry: \n {self.geometry}"
```
ndevenish, line 376:
```suggestion
        return f"<Eyeballing Widget instance using {calibrant} Calibrant starting from Geometry: \n {self.geometry}>"
```
Is this used for direct output anywhere? I tend to try for this sort of `__repr__` to vaguely follow python's plain object convention, it can be hard to separate objects in the terminal otherwise.
```
>>> object()
<object object at 0x1028288a0>
```
If it's actually used for user-output anywhere then it's fine as-is.

newsfragments/2075.misc
```diff
@@ -0,0 +1 @@
+`dials.powder_calibrate`: Add detector distance rough calibration to widget.
```
ndevenish, line 1:
```suggestion
``dials.powder_calibrate``: Add detector distance calibration.
```

ndevenish (approved):

command_line/powder_calibrate.py
elena-pascal: Ooh, yes good call!

command_line/powder_calibrate.py
```diff
@@ -349,20 +363,23 @@ def __init__(
...
         self.fig, self.ax = self.set_up_figure()
         self.calibrant_image = self.calibrant_rings_image()
-        self.beam_fast_slider = self.make_slider("fast")
-        self.beam_slow_slider = self.make_slider("slow")
+
+        self.beam_fast_slider = self._make_slider("fast")
+        self.beam_slow_slider = self._make_slider("slow")
+        self.distance_slider = self._make_slider("distance")
 
     def __repr__(self):
```
ndevenish, line 374:
```suggestion
    def __str__(self):
```

ndevenish: Can you test that this works okay?

ndevenish (changes requested):

command_line/powder_calibrate.py
elena-pascal: I got sidetracked. 
I did move `__repr__`s  to `__str__`s but then I got hit by loss aversion bias and decided I also want to have `__repr__`s.

PR #2077, opened 2022-04-13
Switch DIALS to `src/` layout

ndevenish:
This is a WIP branch for switching DIALS to `src/` layout. This will enable it to be installed as a standard python package when being used outside the libtbx build system. Behaviour under bootstrap should remain as identical as possible, but `pip install dials/` will also be possible and install DIALS as a standard `site-packages/` python package. Similar to dxtbx, extension compilation isn't altered here and it's assumed is dealt with separately.

The commits in this PR:
- Move all non-tests to `src/dials` https://github.com/dials/dials/commit/1906a318c1fa0ca0a3fd0b5217b068533dcc1018
- Update the ecosystem files - `libtbx_refresh.py`, `libtbx_config`, `SConscript`, `setup.py` - https://github.com/dials/dials/commit/079f6ac28f90b4aa889329f01fc3fcaa36c3b141. The libtbx_refresh is written in a similar style to dxtbx, and installs into the `build/` folder without touching `conda_base/`.
- Fixup the tests to no longer assume that they are located in `dials.tests`. https://github.com/dials/dials/commit/c4b2ad5359628da4ac0d0b6576961d03a22c94b4

Things to do before this is ready:

- [x] All tests pass in xia2, dials, dxtbx
- [x] All tests in `xfel_regression` pass. `xfel` compiles, but the regression tests have not been run yet.
- [x] Testing DIALS autocomplete
- [x] Discussion of migration strategies

Once this is merged, I will go through and assist in rebasing/remerging any outstanding PRs. This is probably relatively simple, as git is pretty good at detecting changes to moved folders, though github is less helpful.

### Transitioning to `src/`

A `libtbx.refresh` will definitely be necessary. Possibly two. The sources will be moved around, but there will be some non-repository python cache files left behind, that you can probably either ignore or remove with e.g.

```bash
find modules/dials -name __pycache__ | xargs rm -rf
find modules/dials -depth -type d -empty -delete
# or
cd dials && git clean -nxd # Replace the -n with -f if happy
```


ndevenish:
- `xfel_regression` tests failed because it [imported](https://gitlab.com/cctbx/xfel_regression/-/blob/19b392fe8295361173cea9f6699924b13376aef2/test/command_line/tst_ensemble_refinement.py#L4) a single string from [dials.stills_process](https://github.com/dials/dials/blob/0cda7353e6c8a5facb5705399f3cfbabd83eee3a/tests/command_line/test_stills_process.py#L35-L61), and in the `src/` refactor `dials.tests` is no longer an actual installed module. For now, I've included _just_ the portion of the tests that uses in the `src/dials/tests` subpackage in https://github.com/dials/dials/pull/2077/commits/f26b953c1d342166c5d1dc8f04a100a80c8f5507. This should probably be migrated. 
- There is also a failed test on my test machine for `mpiexec` not being available - I'm not sure what the CI tests do that makes this pass. I'll see if it runs in here, but don't imagine this is a blocker

ndevenish:
To aid migration, it now continues to install a blank libtbx.dials package with `libtbx.pkg_utils.define_entry_points({})`. This prevents double-registering of entry_points.

The other issue is lingering `__pycache__/` folders in DIALS - which should not cause any problems. They can be removed with usage of `git clean -nffdx` or e.g.
```
find dials -type d -name __pycache__ | xargs rm -rf
find dials -depth -type d -empty -delete
```

ndevenish:
So, because of the moved `src/` layout, the xfel module will fail to compile if it's configured first in the module list, which it previously did because of the issue in https://github.com/dials/dials/commit/cafaf4df9454b1a63cac2940322784d0ca020fc7. To manage this migration, I have:
- Removed `xfel` from the DIALS "optional_dependencies" list. This means that it will always end up configured afterwards (because it's no longer a circular dependency), but at the cost that it now needs to be specified in the initial `libtbx.configure` - I think this is always the case?
- Checked for this misconfiguration case and print an explicit error. If you run a `libtbx.refresh` and it's configured this way, it'll exit with this error:
  ```
  Error: xfel module is configured before dials, but now requires dials first.

  To fix this, please run:

          libtbx.configure --exclude=xfel $(libtbx.list_modules) xfel
  ```

I'm... not sure that there is a better way to manage this inside dials. It's not easy to re-order these at time of refresh, because it's in a [literal loop over the list](https://github.com/cctbx/cctbx_project/blob/47f0fd9eaaba1f05ebbc2e2c2492d68d9106260c/libtbx/env_config.py#L2252-L2253), and changing that while running seems to be the worse of the options. But leaving it as-is means that if xfel is configured first, it will fail, because it doesn't know about the `src/` layout and thus won't be able to find the headers.

ndevenish:
On the plus side, the `xfel_regression` CI tests appear to be passing now.

ndevenish:
And autocomplete works.

build.py
```diff
@@ -0,0 +1,77 @@
...
+    entry_points = setup_kwargs.setdefault("entry_points", {})
+    console_scripts = entry_points.setdefault("console_scripts", [])
+    # Work out what dispatchers to add
+    all_dispatchers = sorted(
+        itertools.chain(
+            *[
+                get_entry_point(f, "dials", "dials.command_line")
+                for f in (package_path / "command_line").glob("*.py")
+            ]
+        )
```
Anthchirp, line 65:
```suggestion
        itertools.chain.from_iterable(
            get_entry_point(f, "dials", "dials.command_line")
            for f in (package_path / "command_line").glob("*.py")
        )
```

libtbx_refresh.py
```diff
@@ -1,53 +1,66 @@
...
+\033[31;1mError: \033[34mxfel\033[31m module is configured before \033[34mdials\033[31m, but now requires dials configured first.
+
+To fix this, please run:
+
+        \033[0;1mlibtbx.configure --exclude=xfel {' '.join(module_names)}\033[0m
+"""
+        )
 
 if sys.version_info.major == 2:
     sys.exit("Python 2 is no longer supported")
```
Anthchirp, line 62: can probably drop this now?
```suggestion
```

setup.py
```diff
@@ -1,45 +1,92 @@
...
+    "version": __version_tag__,
+    "long_description": Path(__file__).parent.joinpath("README.md").read_text(),
+    "description": "Diffraction Integration for Advanced Light Sources",
+    "author": "Diamond Light Source",
+    "license": "BSD-3-Clause",
+    "author_email": "dials-support@lists.sourceforge.net",
+    "project_urls": {
+        "homepage": "https://dials.github.io",
+        "repository": "https://github.com/dials/dials",
+    },
```
Anthchirp, line 22: I would suggest moving the bulk of the static information into a `setup.cfg` - happy to provide this as commit

Anthchirp (approved):
Overall looking good.

I'm sure you've already thought of this, but: when merging this I'd be keen to squash it down to two commits, one just for the renames and one with changes that also adds the rename commit to that ignore file.

libtbx_refresh.py
Anthchirp: -> #2086

ndevenish:
So, this is now the message if you `libtbx.refresh` but `xfel` is resolved before DIALS:

image

This should not be a problem for any new environments, because `xfel` is now an implicit rather than explicit dependency in the `libtbx_refresh`.

ndevenish:
I'm also happy to consider just "Run libtbx.configure automatically then exit" but this feels a little prone to just complicating things, and I don't know if it will break anything for custom environments.

setup.py
ndevenish: I thought I recalled us doing this for dxtbx, but it appears not to be the case? Did something not work?

ndevenish:
On Markus' suggestion I've reworded to be more explicit about what the DIALS change was:

image

Anthchirp:
If you get
```
  File "modules/cctbx_project/libtbx/env_config.py", line 3224, in <module>
    warm_start(sys.argv)
  File "modules/cctbx_project/libtbx/env_config.py", line 3129, in warm_start
    _warm_start(env, args)
  File "modules/cctbx_project/libtbx/env_config.py", line 3099, in _warm_start
    env.refresh()
  File "modules/cctbx_project/libtbx/env_config.py", line 2253, in refresh
    module.process_libtbx_refresh_py()
  File "modules/cctbx_project/libtbx/env_config.py", line 2570, in process_libtbx_refresh_py
    exec(to_str(fh.read()), global_vars)
  File "<string>", line 12, in <module>
ImportError: bad magic number in 'dials': b'\x03\xf3\r\n'
make: *** [reconf] Error 1
```
then you have to delete `modules/dials/__init__.pyc`.

PR #2078, opened 2022-04-13
Update dials.export format=mmcif 

jbeilstenedmands:
Updates to include unmerged data conforming to the latest version of the `mmcif_pdbx.dic` used for depositions.
Makes `v5` the default for `format=mmcif` (rather than `v5_next`, which is less useful for deposition...).
Validated against the pdbx dictionary using the beta lactamase and https://zenodo.org/record/6380022#.Ylakl27MK-o example datasets and the instructions at https://sw-tools.rcsb.org/apps/MMCIF-DICT-SUITE/README-source


rjgildea:
It's probably worth seeing if we can upload the resulting mmcifs to the [wwPDB Validation Service](https://validate.wwpdb.org/validservice). I'm just attempting to generate some suitable files to upload. Do we have the beta lactamase and Zenodo insulin data available somewhere centrally at DLS?

PR #2080, opened 2022-04-14
Refactor find_spots and import to not depend on __main__, deprecate ImageImporter

ndevenish:
Both `dials.find_spots` and `dials.import` check if they are imported as `__main__` to decide whether to configure logging (and thus show any output). This assumption fails if running from a standard `console_scripts` entry point.

This refactors both scripts away from the legacy `a = Script(); a.run()` pattern more into the style of https://github.com/dials/dials/blob/main/doc/examples/boilerplate.py, which advocates splitting the `run()` method from a `do_X` method that assumes the logging is already set up.

Import was... slightly complex here - because the heavy lifting for import is implicitly done when parsing the arguments, and it does a first pass to determine the logging level, the function is quite intertangled with the ArgumentParser flow. Thus, I've added a `configure_logging=` function parameter which toggles this. I've kept but deprecated the `ImageImporter` class because it seems more likely that this is used elsewhere. I've updated the tests that used this directly.

There are also tests that assume that the `find_spots.run` function returns the experiments and reflection table, and passes in argument-style parameters. For this reason, `find_spots.run` here returns if passed `return_results=True`. The tests should probably be refactored into running as-a-subprocess for this.


command_line/dials_import.py
```diff
@@ -739,215 +742,235 @@ def convert_to_grid_scan(self, imageset_list, params):
...
-        """
-        logger.info(
-            compare.sequence_diff(
-                sequence1, sequence2, tolerance=params.input.tolerance
-            )
+
+class ImageImporter:
+    def __init__(self, phil=phil_scope) -> None:
+        warnings.warn(
+            "ImageImporter class is deprecated. Please use dials.command_line.dials_import.run instead",
```
elena-pascal, line 949: "Please use dials.command_line.dials_import.run"
Why not dials.command_line.dials_import.do_import? 

It looks like you fixed the only usage of `ImageImporter`, the one in `test_generate_mask`, so this is mostly a useless comment.

ndevenish: Reasoning:
1. `.do_import` doesn't do the same thing as `ImageImporter` did, but `run` does - it assumes you have done no pre-parsing, so it's the smallest change that would give the same results.
2. It's the only usage Inside DIALS - I cannot discount the possibility that someone else is using `ImageImporter`, as a public API, to wrap image importing. This might be overly cautious, but it costs nothing to have a transition period. This does, however, point out that I didn't make a "Remove after X.Y" comment alongside this deprecation, so at least that needs to change.

ndevenish: Scratch #1, I was remembering the reasoning from an earlier stage of the branch. `do_import` would indeed have the same behaviour as importing and using, because it wouldn't configure logging (whereas `.run` would). So that's a good catch.

Issue #2076, opened 2022-04-12
DIALS has a cyclic dependency on the xfel module


ndevenish:
At some point we developed a circular dependency on the "optional" xfel module. This takes the form of usage of`xfel.clustering.cluster.Cluster` and ` xfel.clustering.cluster_groups.unit_cell_info`. Since the conda-forge/cctbx[-base] conda packages don't include xfel, this is a blocker to `conda install dials` - at the very least, we should make it an optional dependency again.

The places that appear to use unqualified imports are:
- [ ] [`dials.cosym`](https://github.com/dials/dials/blob/cafaf4df9454b1a63cac2940322784d0ca020fc7/command_line/cosym.py#L11) uses `unit_cell_info` for output.
- [ ] [`dials.combine_experiments`](https://github.com/dials/dials/blob/cafaf4df9454b1a63cac2940322784d0ca020fc7/command_line/combine_experiments.py#L20) uses `unit_cell_info` for output, and imports `.cluster` for `Cluster`.
- [ ] [`dials.cluster_unit_cell`](https://github.com/dials/dials/blob/cafaf4df9454b1a63cac2940322784d0ca020fc7/command_line/cluster_unit_cell.py#L13-L14) uses both,
- [ ] and so does [`algorithms/indexing/analysis.py`](https://github.com/dials/dials/blob/cafaf4df9454b1a63cac2940322784d0ca020fc7/algorithms/indexing/ssx/analysis.py#L12-L13).
- [ ] Finally, [`algorithms/clustering/unit_cell.py`](https://github.com/dials/dials/blob/cafaf4df9454b1a63cac2940322784d0ca020fc7/algorithms/clustering/unit_cell.py#L6) uses `Cluster` only.

The `unit_cell_info`-only cases like cosym look reasonably easy to deal with - it's just parts of the output, so could pretty easily be moved to import-dependent output, in case the `xfel` modules is missing. The usage of `Cluster` is a little more complex - I don't have a good understanding of how easy/hard it'd be to replace these, either moving some of the code into dials or using something in scipy/scikit-learn for replacing it.

ndevenish:
Other, embedded uses (e.g. doesn't get pulled in by simply importing, so less problematic):

[...]

PR #2081, opened 2022-04-14
Remove `xfel.clustering` dependencies

rjgildea:
Importing `xfel` in `dials` causes a cyclic dependency as highlighted in #2076.
We only use a limited subset of the behaviour in `xfel.clustering`, so move it in house.

Use new clustering code in:
* `dials.cosym`
* `dials.cluster_unit_cell`

For #2076


ndevenish:
Pytype errors are just because new imports have been added (that aren't present when pytype runs), so not an issue.

PR #2091, opened 2022-04-22
Add a pytest mark to xfail tests when xfel is not present

ndevenish:
Some of these cases outright fail on imports, some just fail because the module is missing but the error is caught in deeper layers. This allows tests to pass even if xfel is missing.


phyy-nx:
Hi @ndevenish, this makes a lot of sense.  Question, will the current PR azure test pipeline remain a dev build as the dials conda package is developed?  If so, that's great because it can easily pull in xfel so these tests remain tested.

Anthchirp:
Fwiw you can achieve the same with an `xfel` fixture wothout declaring a marker, much like the `dials_regression` one, or `xds` and `ccp4` in xia2.

ndevenish:
> Hi @ndevenish, this makes a lot of sense. Question, will the current PR azure test pipeline remain a dev build as the dials conda package is developed? If so, that's great because it can easily pull in xfel so these tests remain tested.

I'm not sure how to solve this yet. It would be nice to not have ~2-3 hours of CI time for every commit, both from a 🐻‍❄️ perspective and a useful feedback perspective (a a glance, the build takes longer than the tests). But the xfel regression tests are useful.

I imagine that we'll end up moving to a prebuilt cctbx, but then either putting a libtbx install/develop layer for xfel or pulling in whatever system xfel ends up using for publishing on conda-forge (it might work if it's the _only_ layered libtbx package, but I don't think that supports arbitrary composition order).



> Fwiw you can achieve the same with an `xfel` fixture without declaring a marker, much like the `dials_regression` one, or `xds` and `ccp4` in xia2.

I dislike using fixtures if they aren't actually providing any arguments to the function

phyy-nx:
What about using cached builds?  I think the cctbx CIs use cached builds that are reset once a week.  Saves on build time a lot.  I mean, we still run make, but the build folder itself is reused for a week.

PR #2096, opened 2022-04-25
Add a CMake build to DIALS

ndevenish:
Exists in parallel to existing SCons-based libtbx build, and intended to be used on a prebuilt cctbx installation with or without dxtbx. Will install libraries directly into the python site-packages, similarly to dxtbx, with the intention that we can revisit this later.

The `FindCCTBX.cmake` is based on the one the in in [dxtbx](https://github.com/cctbx/dxtbx/blob/main/cmake/Modules/FindCCTBX.cmake), but has been expanded to find cctbx slightly more robustly, and cover the new cases of potentially renamed libann. DXTBX can be found either directly, if importable, or as part of cctbx otherwise.


ndevenish:
(Cancelled build because since this doesn't touch existing systems, is literally 🔥 🐻‍❄️ 🔥 )

PR #2099, opened 2022-04-28
Revert the value of small_scale_cutoff in dials.scale

jbeilstenedmands:
As part of https://github.com/dials/dials/pull/2074, the threshold for excluding reflections due to unfeasibly low scale factors was reduced. This PR reverts the value to it's previous default.


rjgildea (approved):
Confirmed this fixes the failing xia2 test, thanks for the quick fix!

Issue #2029, opened 2022-03-07
Tutorial error for `dials.refine_bravais_settings`


dagewa:
A user reports one of the tutorials is not working with up-to-date DIALS. We get to this point
https://dials.github.io/documentation/tutorials/centring_vs_pseudocentring.html#converting-to-a-centred-lattice

where the next command to run is
```
dials.refine_bravais_settings indexed.expt indexed.refl
```
For convenience I attach `indexed.{expt,refl}` [here](https://github.com/dials/dials/files/8198531/indexed.zip)

With current DIALS the output table is different to what the tutorial reports, with the change-of-basis op `b+c,-b+c,a` being at the top of the table.

We can reindex with that operator:
```
dials.reindex indexed.refl change_of_basis_op=b+c,-b+c,a
```
But the resulting reflections don't match `bravais_setting_5.expt`, as shown by

```
dials.refine bravais_setting_5.expt reindexed.refl
...
+-------------------+--------+--------+---------+-------+-------+
|                   |    Min |     Q1 |     Med |    Q3 |   Max |
|-------------------+--------+--------+---------+-------+-------|
| Xc - Xo (mm)      | -325.8 | -27.63 |  -6.483 | 18.42 | 208.4 |
| Yc - Yo (mm)      |  -1027 | -17.72 | -0.6885 | 17.41 | 821.6 |
...
```

Why does the change-of-basis op reported by `dials.refine_bravais_settings` for `bravais_setting_5.expt` not match that crystal?

dagewa:
Interested in getting @rjgildea's input into why the above happens and whether it is a bug

dagewa:
Maybe this is the same as #1280?

dagewa:
Ok, no, this is not the same, but another case in which the `cb_op` appears to be wrong

dagewa:
This will be made redundant by #2064, as `indexed.expt` above is in A222 (as per the tutorial).

dagewa:
With current DIALS, using the [dataset](https://github.com/dials/dials/files/8198531/indexed.zip) posted above, RBS now gives this output.
```
+------------+--------------+--------+--------------+----------+-----------+-------------------------------------------+----------+-------------+
|   Solution |   Metric fit |   rmsd | min/max cc   |   #spots | lattice   | unit_cell                                 |   volume | cb_op       |
|------------+--------------+--------+--------------+----------+-----------+-------------------------------------------+----------+-------------|
[...]

PR #2105, opened 2022-05-05
RBS: reported cb_op now correctly maps input to Bravais setting

### LINKED ISSUE #1280: dials.refine_bravais_settings misleading if input not P1
Data processed in I23, want to map to H3, RBS says:

```
+------------+--------------+--------+--------------+----------+-----------+----
---------------------------------------+----------+-------------------+
|   Solution |   Metric fit |   rmsd | min/max cc   |   #spots | lattice   | uni
t_cell                                 |   volume | cb_op             |
|------------+--------------+--------+--------------+----------+-----------+----
---------------------------------------+----------+-------------------|
|   *     22 |            0 |  0.106 | 0.879/0.939  |    18000 | cI        | 78.07  78.07  78.07  90.00  90.00  90.00  |   475808 | b+c,a+c,a+b       |
|   *     21 |            0 |  0.08  | 0.893/0.898  |    18000 | hR        | 110.39 110.39  67.53  90.00  90.00 120.00 |   712724 | a+2*b+c,-b+c,a    |
|   *     20 |            0 |  0.103 | 0.885/0.886  |    18000 | hR        | 110.40 110.40  67.58  90.00  90.00 120.00 |   713349 | a+b+2*c,a-c,b     |
|   *     19 |            0 |  0.103 | 0.882/0.888  |    18000 | hR        | 110.41 110.41  67.64  90.00  90.00 120.00 |   714044 | 2*a+b+c,-a+b,c    |

```

but this is misleading as the `cb_op` for cI needs to be inverted first i.e.

```
>>> cb = sgtbx.change_of_basis_op
>>> i = cb("b+c,a+c,a+b")
>>> ii = i.inverse()
>>> h = cb("a+2*b+c,-b+c,a")
>>> c = h * ii
>>> c.as_abc()
'a+c,b-c,-1/2*a+1/2*b+1/2*c'
```

is the correct reindexing op. 

Assert: the reindexing operation in the output should _work_

rjgildea:
Ensure that the reported reindexing operators correctly map the input symmetry to the given Bravais settings, regardless of whether the input symmetry was a primitive or non-primitive setting

Fixes #1280 (and probably #2029)


src/dials/algorithms/indexing/bravais_settings.py
```diff
@@ -89,6 +89,12 @@ def __init__(self, other):
 
 
 class RefinedSettingsList(list):
+    def __init__(self, *args, cb_op_to_primitive=None, **kwargs):
+        if cb_op_to_primitive is None:
+            cb_op_to_primitive = sgtbx.change_of_basis_op()
+        self.cb_op_to_primitive = cb_op_to_primitive
```
graeme-winter, line 95: This warning is perhaps legitimate? Is there a way we can trigger this code in the regression tests to ensure that this important fix remains fixed?

src/dials/algorithms/indexing/bravais_settings.py
```diff
@@ -190,7 +196,9 @@ def lowest_symmetry_space_group_for_bravais_lattice(
     ).group()
 
 
-def refined_settings_from_refined_triclinic(experiments, reflections, params):
+def refined_settings_from_refined_triclinic(
+    experiments, reflections, params, cb_op_to_primitive=None
```
graeme-winter, line 200: Is `cb_op_to_primitive` not always set?

Even if it is `a,b,c` or `h,k,l` or whatever? I.e. a no-op?

tests/command_line/test_refine_bravais_settings.py
```diff
@@ -11,27 +12,29 @@
...
-                pickle_path,
-                experiments_path,
-                "reflections_per_degree=5",
-                "minimum_sample_size=500",
-                "beam.fix=all",
-                "detector.fix=all",
-                "prefix=tst_",
-            ]
-        )
+    result = procrunner.run(
```
graeme-winter, line 19: This `codecov` warning is ... odd?

graeme-winter (approved):
Change set looks good - only concern I have is there is an ambiguity in the definition of a no-op from either None or identity -> would probably make sense to settle on either / or?

Also, I suspect there is something you have learned / already knew about, about cb ops, which would be very useful to informally share _somewhere_ if you could think of a way of doing that? A long form comment in the top of `bravais_setting` code or something would seem legitimate to aid the reader once they get further down. Distinctions between `hkl` and `abc` indexing etc. 

Overall though changes look good thank you

tests/command_line/test_refine_bravais_settings.py
rjgildea: This test uses dials_regression hence isn't included in the codecov statistics.

src/dials/algorithms/indexing/bravais_settings.py
rjgildea: This was just backwards compatibility for consistency with the previous function signature - in practice, at least as called within dials code, `cb_op_to_primitive` is never `None`. The actual fix is covered by the test in `test_refine_bravais_settings_non_primitive_input`.

src/dials/algorithms/indexing/bravais_settings.py
rjgildea: As above, this was mainly for backwards compatibility.

src/dials/command_line/refine_bravais_settings.py
```diff
@@ -130,10 +130,17 @@ def map_to_primitive(experiments, reflections):
         .info()
         .change_of_basis_op_to_primitive_setting()
     )
-    experiments[0].crystal.update(
-        experiments[0].crystal.change_basis(cb_op_to_primitive)
-    )
-    reflections["miller_index"] = cb_op_to_primitive.apply(reflections["miller_index"])
+    if not cb_op_to_primitive.is_identity_op():
```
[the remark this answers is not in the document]
rjgildea: `cb_op_to_primitive` is never `None` here, as it is set just above by lines 127-132.

PR #2106, opened 2022-05-05
stills_process: crash on unhandled command line argument

dwpaley:
Currently if I type

```
$ dials.stills_process data.h5 params.phil abcd=efgh
```
or worse, one of:
```
$ dials.stills_process data.h5 params.phll
$ dials.stills_process data.h5 params.phil spotfinding.filter.d_max=10 # oops, should be spotfinder
```
the unhandled argument is ignored with possibly confusing consequences. Currently all unhandled arguments are assumed to be input paths. Can we validate that they are actually paths?


src/dials/command_line/stills_process.py
```diff
@@ -402,7 +402,9 @@ def run(self, args=None):
                 all_paths.extend(params.input.glob)
             globbed = []
             for p in all_paths:
-                globbed.extend(glob.glob(p))
+                g = glob.glob(p)
+                assert g, f"Unhandled path or option: {p}"
```
ndevenish, line 406: Would this be better as a sys.exit, which is more readable than a stack trace?
```suggestion
                if not g:
                    sys.exit(f"Error: Unhandled path or option: {p}")
```

PR #2108, opened 2022-05-06
WIP: Spotfinder: new method ``omit`` for resolution filtering

dwpaley:
Still testing and cleaning up but would appreciate any comments.

Instead of applying a mask before spotfinding, the new method finds
all spots and then filters on resolution. This proved important for
serial data with high background where any low-resolution mask would
give a ring of spots wherever the mask ended.

- Add new phil parameter resolution_filter_method in spotfinder.filter

- Refactor the generation of resolution ranges out of util.generate_mask
and into a new function util.generate_skip_ranges

- Add new SpotFinder init arg omit_ranges

- Implement the filter step in SpotFinder.find_spots

### DIFF
--- a/src/dials/algorithms/spot_finding/finder.py
+++ b/src/dials/algorithms/spot_finding/finder.py
@@ -642,6 +642,7 @@ class SpotFinder:
         no_shoeboxes_2d=False,
         min_chunksize=50,
         is_stills=False,
+        omit_ranges=None,
     ):
         """
         Initialise the class.
@@ -673,6 +674,7 @@ class SpotFinder:
         self.no_shoeboxes_2d = no_shoeboxes_2d
         self.min_chunksize = min_chunksize
         self.is_stills = is_stills
+        self.omit_ranges = omit_ranges
 
     def find_spots(self, experiments: ExperimentList) -> flex.reflection_table:
         """
@@ -720,6 +722,30 @@ class SpotFinder:
             assert missed.count(True) == 0, "Failed to remap {} experiment IDs".format(
                 missed.count(True)
             )
+
+            if self.omit_ranges is not None:
+                table_filtered = flex.reflection_table()
+                for expt_id, experiment in enumerate(experiments):
+                    refls = table.select(table["id"] == expt_id)
+                    s0 = experiment.beam.get_s0()
+                    for panel_id, panel in enumerate(experiment.detector):
+                        panel_refls = refls.select(refls["panel"] == panel_id)
+                        sel = flex.bool(len(panel_refls), True)
+                        for d1, d2 in self.omit_ranges:
+                            d1, d2 = sorted(d1, d2)
+                            subsel = flex.bool(
+                                [
+                                    d1
+                                    < panel.get_resolution_at_pixel(
+                                        s0, panel_refls["xyzobs.px.value"][i][0:2]
+                                    )
+                                    < d2
+                                    for i in range(len(panel_refls))
+                                ]
+                            )
+                            sel &= subsel
+                        table_filtered.extend(panel_refls[sel])
+                table = table_filtered
 
             reflections.extend(table)
             # Write a hot pixel mask
--- a/src/dials/util/masking.py
+++ b/src/dials/util/masking.py
@@ -154,38 +154,68 @@ def lru_equality_cache(maxsize=10):
     return _decorator
 
 
-def generate_ice_ring_resolution_ranges(beam, panel, params):
+def generate_ice_ring_skip_ranges(
+    params, detector=None, beam=None, disable_parallax=False
+):
     """
     Generate a set of resolution ranges from the ice ring parameters
     """
-    if params.filter is True:
+    if not params.filter:
+        return
 
-        # Get the crystal symmetry
-        crystal_symmetry = crystal.symmetry(
-            unit_cell=params.unit_cell, space_group=params.space_group.group()
-        )
+    if params.d_min is not None:
+        d_min = params.d_min
+    else:
+        assert detector is not None
+        assert beam is not None
+        panel_d_mins = []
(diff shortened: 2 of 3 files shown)


rjgildea:
We do something along these lines in per-image analysis, mainly for performance reasons (to avoid the costly calculation of per-pixel resolution masks):
https://github.com/dials/dials/blob/c645ecc624938d3fce81a31b5eebe3fca513bc38/src/dials/command_line/find_spots_server.py#L79-L95
The version there maps the reflections to reciprocal space and calculates the d-spacing that way. I don't know how it compares performance wise to your version, but it does avoid a Python loop over every reflection.

dwpaley:
@rjgildea Thanks. Not sure if this is faster (probably!) but certainly more clear.

I note that this code in `finder.py` may now leave a couple extra columns in the reflection table depending on the value of `self.omit_ranges`. Does this seem like a problem? Could move the four lines ending `table["d"] = ...` outside the if test if there is a risk of confusing behavior; looks like that is what `find_spots_server` does.

Can you comment your preferred co-authored-by line?

src/dials/util/masking.py
```diff
@@ -154,38 +154,68 @@ def _generate_cache_info():
...
+    if params.d_min is not None:
+        yield (0, params.d_min)
+    if params.d_max is not None:
+        d_max = params.d_max
+        d_inf = max(d_max + 1, 1e9)
+        yield (d_max, d_inf)
+    for range in params.resolution_range:
+        if range is None:
+            continue
+        for d1, d2 in range:
```
dwpaley, line 211: This loop got mangled in editing, will fix

src/dials/algorithms/spot_finding/finder.py
```diff
@@ -10,6 +10,7 @@
 from typing import Iterable, Tuple
 
 import libtbx
+import uctbx
```
rjgildea, line 13:
```suggestion
from cctbx import uctbx
```

phyy-nx:
Bump to remove stale, and to note this came up again as useful for me during some recent reprocessing.

dagewa:
This does sound useful, though I wonder if the use case would anyway be satisfied by running something like
```
dials.filter_reflections strong.refl d_max=10
```
after standard spot-finding?

ndevenish:
> We do something along these lines in per-image analysis, mainly for performance reasons (to avoid the costly calculation of per-pixel resolution masks):

This difference actually came up when we were trying to reimplement spotfinding for our fast-feedback service, that prevented us from matching exact results.

Thinking about it, is this the _correct_ way to be filtering resolution? With a pixel-based mask, won't any reflection that is cut in half will have it's center biased towards the inside of the detector?

Does anyone recall discussions about this in the past? @jbeilstenedmands @dimitrivlachos @graeme-winter ?

graeme-winter:
> > We do something along these lines in per-image analysis, mainly for performance reasons (to avoid the costly calculation of per-pixel resolution masks):
> 
> This difference actually came up when we were trying to reimplement spotfinding for our fast-feedback service, that prevented us from matching exact results.
> 
> Thinking about it, is this the _correct_ way to be filtering resolution? With a pixel-based mask, won't any reflection that is cut in half will have it's center biased towards the inside of the detector?
> 
> Does anyone recall discussions about this in the past? @jbeilstenedmands @dimitrivlachos @graeme-winter ?

@ndevenish this is nuanced: if you filter post finding then you _do_ find more sensible spots (rather than half-spots as you mention) but you also have the problem that e.g. ice rings which are essentually huge spots, particularly on detectors without tile join, kill the spot finding by making for mahoosive spot shoeboxes before the filtering is applied. If we mask the image then find, this problem is avoided

jbeilstenedmands:
Is a potential solution the addition of a buffer region (of say ~5 pixels) that we add to the current resolution mask, so that we can more accurately find spot centroids around the resolution cutoff (and then still apply the correct resolution filtering to the spot centroids). Then we don't encounter the problems that come with doing the full-image spot-finding but reduce the bias in the centroid values for near-resolution-limit spots?

dimitrivlachos:
I think what @jbeilstenedmands is suggesting is a reasonable hybrid approach. Do both per-pixel and per-spot-centroid filtering but give the pixel region enough of a buffer as to properly determine edge spot centroids and not bias them toward the centre.

Though if we do add a buffer region, I think it is important that we make that abundantly obvious somewhere so _when_ this gets revisited later down the line it doesn't have to be re-discovered.

phyy-nx:
A buffer region sounds useful and would probably fix the issue @dwpaley originally posted about.  Would that be applicable to both the per-pixel based method and the resolution based method?  Asking because a secondary benefit of resolution based filtering is speed, as @ndevenish commented, especially for stills with a varying wavelength per shot where resolution masking per-pixel is slow.

dimitrivlachos:
Using a buffer region like this would necessitate using both filtering methods: a first per-pixel pass with the added buffer to determine the shoeboxes, followed by a second centroid resolution pass to cull outliers in the buffered region.

So, this hybrid method would, unfortunately, not come with the speed increase that using only centroid resolution filtering would provide, @phyy-nx. However, it would be the most correct method, as it properly categorises all spots within the resolution limits.

Another point that came up during our in-person discussion: relying solely on centroid resolution filtering when strong ice rings are present can cause the analysis to fail, as it attempts to construct an enormous dataset encompassing the entire ice ring before filtering it out. This would suggest that using only centroid resolution filtering may not be a viable option at all. Though please do correct me if I am wrong @graeme-winter @jbeilstenedmands @ndevenish

phyy-nx:
Note the serial use case this was designed for was a 'crown' of spots at like 15 Å around the beam stop (Rayonix room temperature XFEL data) and it worked quite well for us.  Granted that's quite a narrow use case.

Issue #521, opened 2018-03-13
dials.index fails on its own example


ndevenish:
```
$ dials.index
...[help output]...
Examples:
  dials.index datablock.json strong.pickle unit_cell=37,79,79,90,90,90 space_group=P43212
```
Which is also the example on https://dials.github.io/documentation/programs/dials_index.html. With Lysozyme data set (though don't think it matters):
```
$ dials.index datablock.json strong.pickle unit_cell=37,79,79,90,90,90 space_group=P43212
...
Traceback (most recent call last):
  File "/dls/science/users/mep23677/dist/build/../modules/dials/command_line/index.py", line 182, in <module>
    run(sys.argv[1:])
  File "/dls/science/users/mep23677/dist/build/../modules/dials/command_line/index.py", line 153, in run
    params=params)
  File "/dls/science/users/mep23677/dist/modules/dials/algorithms/indexing/indexer.py", line 577, in from_parameters
    idxr = indexer_fft3d(reflections, imagesets, params=params)
  File "/dls/science/users/mep23677/dist/modules/dials/algorithms/indexing/fft3d.py", line 35, in __init__
    super(indexer_fft3d, self).__init__(reflections, imagesets, params)
  File "/dls/science/users/mep23677/dist/modules/dials/algorithms/indexing/indexer.py", line 507, in __init__
    self._setup_symmetry()
  File "/dls/science/users/mep23677/dist/modules/dials/algorithms/indexing/indexer.py", line 655, in _setup_symmetry
    space_group=target_space_group.info().as_reference_setting().group())
  File "/dls/science/users/mep23677/dist/modules/cctbx_project/cctbx/crystal/__init__.py", line 101, in __init__
    "Space group is incompatible with unit cell parameters."
AssertionError: Space group is incompatible with unit cell parameters.
```
Whereas swapping the unit cell parameters passes this test:
```
$ dials.index datablock.json strong.pickle unit_cell=79,79,37,90,90,90 space_group=P43212
...[stuff]...
```

PR #2113, opened 2022-05-13
Move flex_table to dxtbx

toastisme:
This PR along with [#521](https://github.com/cctbx/dxtbx/pull/521) moves `flex_table` to dxtbx, with the few `reflection_table` specific functions (i.e those involving `experiment_identifiers`) moved to `reflection_table_suite.h`. The motivation for this is to use `flex_table` for other purposes in dxtbx (specifically a properties table in `Scan`), as it's essentially a generic dictionary with some nice functionality that `reflection_table` extends. 

I'm not really happy with the `flex_table_suite.h` and `reflection_table_suite.h` setup. At the c++ level you could accidentally use methods from the former and might not notice the lack of experiment identifiers updating. As nearly all `flex_table` functions are not actually part of the `flex_table` class, clean inheritance isn't really possible..So I have just copied the convention.

Tests will not pass without [#521](https://github.com/cctbx/dxtbx/pull/521) being merged first. No additional tests added as all functionality should be the same, and should be covered by `tests/algorithms/array_family/test_identifiers_handling.py` and `tests/algorithms/array_family/test_reflection_table.py`


dagewa:
During the DIALS catch up at UCLA we noticed there is some duplicated code in `flex_table.h` and `flex_table_suite.h` files in dxtbx and DIALS. Looking back, we realised that this is part of a transfer of the flex table from DIALS to dxtbx, but we have only taken the first step. This PR would complete the process of transfer started by https://github.com/cctbx/dxtbx/pull/521, but it has merge conflicts. Can we get this resolved soon?

ndevenish:
Discussed in DIALS core. Probably when this merge is fixed, we'd want to have the header files remain, but just import the dxtbx versions and print a deprecation warning when the old headers are included. This way, downstream dependencies will warning but not break.

toastisme:
Deprecation warnings have been added. I've kept `flex_reflection_table.cc` pointing at `dials/array_family/boost_python/flex_table_suite.h` for now just to confirm I've not missed redirecting any functions that might get picked up with tests outside of the CI.

There's currently a build failure for Windows, but this looks to be due to the Windows Azure build using an older version of dxtbx.

src/dials/algorithms/integration/boost_python/integration_integrator_ext.cc
```diff
@@ -22,6 +22,9 @@ namespace dials { namespace algorithms { namespace boost_python {
    * Split the reflection table where the blocks are given.
    */
   inline void job_list_split(const JobList &self, af::reflection_table data) {
+    using dxtbx::af::flex_table_suite::reorder;
+    using dxtbx::af::flex_table_suite::setitem_column;
```
ndevenish, line 26: Since there are now referring to namespaces defined outside of dials, would it be worth adding an explicit `#include <dxtbx/array_family/flex_table_suite.h>` instead of relying on the explicit one from `integrator.h`?

src/dials/algorithms/integration/boost_python/integration_integrator_ext.cc
```diff
@@ -125,13 +128,11 @@ namespace dials { namespace algorithms { namespace boost_python {
     data.resize(bbox_new.size());
 
     // Reorder the reflections
-    af::boost_python::flex_table_suite::reorder(data, indices.const_ref());
+    reorder(data, indices.const_ref());
```
ndevenish, line 131: Although... since it is used exactly once (or twice, for below), it's probably clearer not to use `using` but to qualify it at the point of call.

src/dials/algorithms/integration/integrator.h
```diff
@@ -574,7 +574,7 @@ namespace dials { namespace algorithms {
      * Accumulate the results.
      */
     void accumulate(std::size_t index, af::reflection_table result) {
-      using namespace af::boost_python::flex_table_suite;
+      using dxtbx::af::flex_table_suite::set_selected_rows_index;
```
ndevenish, line 577: Likewise, this is used once, 20 lines ahead in this function. Just qualify it.

src/dials/array_family/boost_python/flex_table_suite.h
```diff
@@ -12,1328 +12,400 @@
...
-    void operator()(T &column) {
-      DIALS_ASSERT(index < column.size());
-      column[index] = extract<typename T::value_type>(item);
-    }
-  };
-
-  /**
-   * A visitor to append column data from 1 table to another
-   */
+  using column_to_object_visitor [[deprecated(
```
ndevenish, line 26: No comment, just - hurrah for C++14!

ndevenish:
Okay, yes, the specific issue is that dxtbx now imports the [scitbx/.../ref_pickle_double_buffered.h](https://github.com/cctbx/dxtbx/blob/59405f6b38e86fc958fd2d9d3c19cd8645f8c8e8/src/dxtbx/array_family/flex_table_suite.h#L15) directly, because the header-inclusion-guard version wasn't moved across. Thus, even though dials only uses the inclusion-guard version, everything inside still gets declared twice. This needs a DXTBX change to fix.

ndevenish:
https://github.com/cctbx/dxtbx/pull/636 will fix this, but I will also update the DIALS references

PR #2116, opened 2022-05-20
dials.image_viewer n_iqr fix

dagewa:
I noticed that the `n_iqr` setting for `radial_profile` thresholding appeared to do nothing when viewing images with "Threshold pixels" selected. It turns out this control was not actually connected to the algorithm. This fixes that.


dagewa:
@ndevenish :eyes: for the news fragment. Otherwise this can be merged straight away.

PR #2118, opened 2022-05-21
Respect user supplied resolution cutoff in Summary from dials.scale

### LINKED ISSUE #2117: Layout of dials.scale merging stats table does not respect user resolution cutoff
Despite @graeme-winter's comment in #1312:

https://github.com/dials/dials/pull/1312#issuecomment-649994211

```
dials.scale symmetrized.{expt,refl} [options] dmin=0.83
```

sometimes reports outer shell with resolution bin determined by CC1/2 fit even though a resolution limit was supplied by the user. Even worse the CC 1/2 cutoff that decides this is *hard coded*:

https://github.com/dials/dials/blob/25055c1f6164dc33e672e7c5c6a9c5a35e870660/src/dials/algorithms/scaling/observers.py#L156

Please could this behaviour be fixed? Like @dagewa in #1524 I find this behaviour very annoying. If I specify a resolution cutoff I expect to see a summary where the outer shell is at that resolution. At the very least I would like a way to disable the CC 1/2 based suggested resolution without having to edit the code.

huwjenkins:
fixes #2117

### DIFF
--- a/src/dials/algorithms/scaling/observers.py
+++ b/src/dials/algorithms/scaling/observers.py
@@ -153,7 +153,10 @@ were considered for use when refining the scaling model.
             anom_stats = script.anom_merging_statistics_result
         logger.info(make_merging_statistics_summary(stats))
         try:
-            d_min = resolution_cc_half(stats, limit=0.3).d_min
+            if script.params.cut_data.d_min is None:
+                d_min = resolution_cc_half(stats, limit=0.3).d_min
+            else:
+                d_min = script.params.cut_data.d_min
         except RuntimeError as e:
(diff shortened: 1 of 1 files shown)


jbeilstenedmands (approved):
I am on board with these changes. If no user resolution limit is specified, then you get the current behaviour, where the suggestion of the limit from the fit is informative. But if a resolution limit is given, just use that and don't perform a fit. I would welcome approval from @graeme-winter before merging this in.

graeme-winter (approved):

jbeilstenedmands:
Thanks @huwjenkins for this contribution!

PR #2126, opened 2022-06-01
Fix dials.export, format=mtz for ssx data

jbeilstenedmands:
The dials data reduction programs are untested for ssx data, so the changes in https://github.com/dials/dials/pull/2054 cause issues when calling goniometer methods on null objects. Apologies as I should have thought to test those changes on this kind of data.

To fix dials.export, I've added a few extra checks to skip certain calculations if no goniometer is present (also, a log call has been moved to a debug log call, as this is logged for every experiment which becomes very verbose).
I've also added a test using integrated ssx data to run all standard data reduction programs (`dials.cosym`, `dials.scale` etc), to avoid breaking changes slipping in. This also revealed a few places in `dials.compute_delta_cchalf` where errors can occur, so these have ben fixed.


src/dials/algorithms/scaling/scale_and_filter.py
```diff
@@ -498,6 +498,8 @@ def _add_new_histogram(d, hist, index):
         )
 
     for c, deltas in enumerate(delta_cc_half_lists):
+        if not (max(deltas) > min(deltas)):
```
dagewa, line 501: Is this not just a slightly roundabout way of checking that all elements are equal? If you're going to make a `flex.double` of the `deltas` anyway, you could do `deltas.all_eq(deltas[0])`?

dagewa (approved):
lgtm

src/dials/algorithms/scaling/scale_and_filter.py
jbeilstenedmands: Good point, i'll update the code

PR #2128, opened 2022-06-01
Raster fix for dials.stills_process

### LINKED ISSUE #2127: Raster scans broken for dials.stills_process
User bug report, originally from cctbx.xfel.small_cell_process, which wraps dials.stills_process:

```
Error integrating eiger_master_XXXXX dials Internal Error:
<pathto>/modules/dials/algorithms/integration/boost_python/integration_integrator_ext.cc(242):
DIALS_ASSERT(b[4] >= frame0) failure.
```

This eiger dataset is a raster scan of stills:

```
dxtbx.print_header eiger_master.h5
Using header reader: FormatHDF5EigerNearlyNexus
...
Scan:
    number of images:   XXXXX
    image range:   {1,XXXXX}
    oscillation:   {0,0}
    exposure time: 0.025
```

The error is from here:
https://github.com/dials/dials/blob/b285fe35c6b598ac198f7082b4d5d14721d880df/src/dials/algorithms/integration/integrator.py#L1237-L1248

Let's say we are processing image 100.  Because dials.stills_process slices imagesets up during processing, `frame0` and `frame1` from `imageset.get_scan().get_array_range()` will be 100 to 101.  However, again because of imageset slicing, the reflection bboxes will be `f1 f2 s1 s2 0 1`.  This causes the assert in the error message.

There are a couple of ways to fix this, but I don't know if they affect raster scan processing at DLS.  I'll make a PR.

phyy-nx:
Two specific fixes:
- In _determine_max_memory_needed, check if the scan is a still, in which case its bbox z coordinates should be 0, 1
- In dials.import, in the manual geometry updater, fix convert_sequences_to_stills. This was likely fallout from cctbx/dxtbx#438, where the imageset indices were redone to use less memory.

Fixes #2127


phyy-nx:
@graeme-winter @jbeilstenedmands I'm not actually sure about the first fix, in `_determine_max_memory_needed`.  When processing ImageSequences of stills, what are the bbox z coordinates for reflections?  For dials.stills_process they are 0, 1.

jbeilstenedmands:
@phyy-nx for the bbox z coordinates on ImageSequences, I looked at two examples. A .cbf file from I24 beamline had bbox z coordinates of -1, 0, whereas a .h5 file from VMXi beamline had bbox z coordinates of 0,1. 🤔

src/dials/algorithms/integration/integrator.py
```diff
@@ -1235,7 +1235,7 @@ def _determine_max_memory_needed(experiments, reflections):
                         if experiment.imageset == imageset:
                             subset.extend(reflections.select(reflections["id"] == j))
                 try:
-                    if imageset.get_scan():
+                    if imageset.get_scan() and not imageset.get_scan().is_still():
                         frame0, frame1 = imageset.get_scan().get_array_range()
```
jbeilstenedmands, line 1239:
```suggestion
                    if imageset.get_scan():
                        if not imageset.get_scan().is_still():
                            frame0, frame1 = imageset.get_scan().get_array_range()
                        else:
                            frame0, frame1 = (0, len(imageset))
                    else:
                        frame0, frame1 = (0, len(imageset))
```
When testing, I found that this line raised a `'NoneType' object has no attribute 'is_still'` error (my process was import, find_spots, ssx_index, integrate), so I think this should be guarded against in this function. It looks like the current logic assumes that the first call to `imageset.get_scan()` will raise a RuntimeError, but I don't think that is always the case anymore.
In terms of the effect on ImageSequences, we don't usually have an ImageSequence at this point, as not all images are indexable so the sequence has already been split into single-image imagesets (and I don't usually trigger this code path anyway). So I'm happy that this is a safe change for current workflows.

jbeilstenedmands (changes requested):
I think these changes are fine with a small tweak to catch the case of `imageset.get_scan()` being `None` in `integrator.py`

src/dials/algorithms/integration/integrator.py
phyy-nx: Hi, thanks for the catch.  I've taken your proposal and modified it slightly and checked that in. It now removes the try/except block and does all the checks explicitly.  Can you review/test this version?  Thanks!

jbeilstenedmands: Thanks, confirmed that your modification works, and is much cleaner code 🙂

jbeilstenedmands (approved):

PR #2129, opened 2022-06-03
reflection table performance tuning

dwpaley:
Performance improvements in `centroid_px_to_mm` and `map_centroids_to_reciprocal_space`:

- Cache the `panel_numbers==i_panel` bool arrays.
- Skip panels with no reflections.

On an example with 1100 experiments and a 256-panel detector, this gave a 14x speedup, around 15s per call to 1.2s per call.


dwpaley:
@graeme-winter I would appreciate your comment on the "dirty hack" added in f6f620cb5. It is definitely funky but I'm not sure if it's too funky.

src/dials/array_family/flex_ext.py
```diff
@@ -1297,14 +1299,44 @@ def map_centroids_to_reciprocal_space(
         # reciprocal lattice viewer -> but if we are looking at the crystal
         # coordinate frame we need to look at the experiments independently
 
+        if hasattr(experiments, "detectors"):
+            detectors = experiments.detectors()
```
graeme-winter, line 1303: I think this is excessively defensive:

```
Python 3.9.15 | packaged by conda-forge | (main, Nov 22 2022, 08:55:37) 
[Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dxtbx.model.experiment_list import ExperimentList
>>> el = ExperimentList()
>>> el.detectors()
[]
>>> hasattr(el, "detectors")
True
```

even an empty experiment list would pass that...

src/dials/array_family/flex_ext.py
```diff
@@ -1297,14 +1299,44 @@ def map_centroids_to_reciprocal_space(
...
+        else:
+            detectors = list(set(expt.detector for expt in experiments))
+        try:
+            test_subcache = self.detector_panel_sels[detectors[0]]
```
graeme-winter, line 1307: Does `test_` here mean test in any way, or is this a (cached) selection?

If latter suggest sel_(name) is better?

src/dials/array_family/flex_ext.py
```diff
@@ -1297,14 +1299,44 @@ def map_centroids_to_reciprocal_space(
...
+            cache_valid = False
+        else:
+            test_key = list(test_subcache.keys())[0]
+            cache_valid = len(test_subcache[test_key]) == len(self)
+        if not cache_valid:
+            self.detector_panel_sels = {det: {} for det in detectors}
+            # Dirty hack: without the following line, the reflection_table
+            # (a boost.python class) will become unpickleable because the
+            # object's python __dict__ is not empty and we have not declared
+            # that the boost-defined __getstate__ method knows about the stuff
```
graeme-winter, line 1318: Could we explicitly wipe the cache on pickle (do we even use pickle?)

Or you talking about general serialisation? Even so, we could explicitly dump the cache before we try and serialise the object, or name it in such a way as it can be ignored at that point?

graeme-winter (commented):
I feel like we should do something to avoid `self.__getstate_manages_dict__ = True`

This feels like something we will come to regret later on.

We have explicit code in the pickle suite etc. code -> could delete the cache explicitly? Feels like that would be a better solution than something "magic"

PR #2133, opened 2022-06-08
put phil strings inside functions

dwpaley:
Wouldn't it be nice if your editor would fold phil strings? Then you wouldn't have to scroll past hundreds of lines at the top of a file. Yes, that would be nice.

I've just done this for stills_process, but if this idea is appreciated, then I could make a more comprehensive effort.


ndevenish:
I think if we were going to do this more generally, it might be better to consider a "library module" of places that we store the PHIL configuration in. But no objections to making the stills_process code easier to navigate 😃

PR #2135, opened 2022-06-13
LevMar: terminate with zero expected objective decrease

dagewa:
Avoid divide-by-zero in the special case of starting refinement already at the minimum.


jbeilstenedmands (approved):
Thanks for the fix

PR #2143, opened 2022-06-20
Remove failing rs_mapper test

rjgildea:
This has been failing since cctbx/dxtbx#455 was merged, however I don't think the test makes sense as is, either before or after these changes. dials/dxtbx does not handle screening data collections from eigers gracefully - before this change it interprets as 3 x 45-degree images (so the starting angle is correct, but the oscillation range of each image is wrong), and after as 3 x 0.15-degree oscillations (so the starting angles are incorrect, but the oscillation ranges are correct). See also cctbx/dxtbx#13.

Release:
```
$ dials.show imported.expt | grep Scan -A 3
Scan:
    number of images:   3
    image range:   {1,3}
    oscillation:   {0,45}
```
Nightly build:
```
$ dials.show imported_nightly.expt | grep Scan -A 3
Scan:
    number of images:   3
    image range:   {1,3}
    oscillation:   {0,0.15}
```


dagewa:
I guess we could replace with a different dataset that includes a mask. `dectris_eiger_master.h5` from the `image_examples` dataset in `dials_data` might be okay. This is a still image, but I don't think that matters for the test, which is just checking that `dials.rs_mapper` does the right thing when an image contains a mask (#955).

I can add that test later today. It does appear to exercise the issue:
```
dials.rs_mapper imported.expt map_file=junk.ccp4
mapdump mapin junk.ccp4 <<<GO |grep "Maximum den"
```
gives
```
           Maximum density .................................   289.11111
```

whereas
```
dials.rs_mapper imported.expt map_file=junk.ccp4 ignore_mask=true
mapdump mapin junk.ccp4 <<<GO |grep "Maximum den"
```
gives 
```
           Maximum density ................................. 65535.00000
```

dagewa:
Apparently back in 2018 it was different https://github.com/dials/dials/pull/955#issuecomment-539580595, but the test wasn't checking the ignore case.

dagewa (approved):
I added a revised test, which passed locally

Issue #2139, opened 2022-06-16
Confused by dials.find_rotation_axis output


huwjenkins:
I'm a bit confused by the output of `dials.find_rotation_axis` with a vertical goniometer. 

For example I have a dataset which after dials.import has:

```
Goniometer:
    Rotation axis:   {0,1,0}
    Fixed rotation:  {1,0,0,0,1,0,0,0,1}
    Setting rotation:{1,0,0,0,1,0,0,0,1}
```

`dials.find_rotation_axis` reports:

```
Azimuth (current): 90.00000 degrees
                 1.57080 radians
```
and finishes with:
```
Rotation axis found: -93.20 deg. / -1.627 rad.
Goniometer:
    Rotation axis:   {-0.0558215,0.998441,0}
    Fixed rotation:  {1,0,0,0,1,0,0,0,1}
    Setting rotation:{1,0,0,0,1,0,0,0,1}
```

but if I run 
```
dials.find_rotation_axis optimised.expt strong.refl
```

where `optimised.expt` is the experiments file from the previous run `dials.find_rotation_axis` reports:

```
Azimuth (current): 93.20000 degrees
                 1.62665 radians
```
and finishes with:
```
Rotation axis found: -93.20 deg. / -1.627 rad.
Goniometer:
    Rotation axis:   {-0.0558215,0.998441,0}
    Fixed rotation:  {1,0,0,0,1,0,0,0,1}
    Setting rotation:{1,0,0,0,1,0,0,0,1}
```

I guess it's to do with this:

https://github.com/dials/dials/blob/ded5135c772afa3754fc4bdcdcf6a03abc51a252/src/dials/command_line/find_rotation_axis.py#L95

but I probably don't understand (the dxtbx) geometry enough to see why y is flipped here but not here:

https://github.com/dials/dials/blob/ded5135c772afa3754fc4bdcdcf6a03abc51a252/src/dials/command_line/find_rotation_axis.py#L378-L379 @dagewa can you help me here? Thanks!

dagewa:
This was adapted from `edtools.find_rotation_axis`, which seems to be where the flip comes from:
https://github.com/instamatic-dev/edtools/blob/807a0cc70a0b1c0912e40857535bc42a43b962c9/edtools/find_rotation_axis.py#L9-L25
and the angle (renamed azimuth from omega) was set here:
[...]

PR #2145, opened 2022-06-20
`dials.find_rotation_axis`: `roty` flip on input

dagewa:
Thanks to @huwjenkins in #2139 for diagnosing the issue.


dagewa:
The test that fails is the unrelated one fixed by #2143

PR #2151, opened 2022-06-22
dials.cosym: don't apply cb_op to crystal twice

rjgildea:
This bug appears to have been introduced in #1647. Since it was only the crystal model, and not the miller indices, that were affected, I don't think it should cause any problems in most cases, although it may affect the resulting unit cells in cases of pseudo-symmetry.

Co-authored-by: James Beilsten-Edmands <30625594+jbeilstenedmands@users.noreply.github.com>


jbeilstenedmands (approved):
I agree, I don't think this should be applied twice

PR #2154, opened 2022-06-27
Allow use of a reference dataset in dials.cosym

jbeilstenedmands:
Add the option `reference=`, to allow specifying a reference intensity file/model for consistent reindexing. This dataset is included in the cosym analysis, and the `x,y,z` reindexing mode of the reference used as the seed for determining the reindexing operators to ensure consistent indexing. 

This option only makes sense if we are restricting the space group, so the space group from the reference is used for the cosym analysis.


src/dials/algorithms/symmetry/cosym/__init__.py
```diff
@@ -382,15 +391,48 @@ def _reindexing_ops(
...
+                        reindexing_op = (
+                            self.cb_op_inp_min.inverse() * cb_op * self.cb_op_inp_min
+                        ).as_xyz()
+                        if reindexing_op == "x,y,z":
+                            xis = np.array([X[i]])
+                        break
+                if xis is not None:
+                    break
+            if xis is None:
+                raise ValueError("Unable to determine x,y,z mode of reference")
```
rjgildea, line 415: 1. checking that `self.cb_op_inp_min.inverse() * cb_op * self.cb_op_inp_min` is equal to identify op, is equivalent to simply checking that `cb_op` is the identity op
2. I think this whole block isn't actually necessary, as the first copy of the dataset should be the one that hasn't had a change of basis applied, which I think it what you are looking for here, the following should be equiavelent:
```suggestion
            xis = np.array([coords[sel][0]])
```
The order of the symmetry copies of the dataset is determined by the order of `Target.sym_ops`, and therefore the first copy should be the one with `cb_op` = `x,y,z`:
https://github.com/dials/dials/blob/2abb8d42ee323e4b82623e2d564c134693fd3b60/src/dials/algorithms/symmetry/cosym/target.py#L92

src/dials/algorithms/symmetry/cosym/__init__.py
```diff
@@ -831,3 +873,35 @@ def as_dict(self):
...
+    )
+
+    with open(os.devnull, "w") as devnull:
+        reference_intensities, _ = reference_intensities.apply_change_of_basis(
+            str(ref_cb_op), out=devnull
+        )
+    reference_intensities = reference_intensities.expand_to_p1()
+    reference_intensities = (
+        reference_intensities.as_non_anomalous_array().merge_equivalents().array()
+    )
```
rjgildea, line 904: If you use `miller.array.change_basis()` rather than `miller.array.apply_change_of_basis()` then you don't need the `os.devnull` dance:
```suggestion
    reference_intensities = (
        reference_intensities.change_basis(
            ref_cb_op,
        )
        .expand_to_p1()
        .as_non_anomalous_array()
        .merge_equivalents()
        .array()
    )
```

rjgildea (changes requested):
Thanks, this looks like it should be a useful change. I think the code to choose the reference coordinates can be significantly simplified (see suggestions), and the `os.devnull` dance shouldn't be necessary if you use instead `miller.array.change_basis()`.

src/dials/algorithms/symmetry/cosym/__init__.py
```diff
@@ -382,15 +390,32 @@ def _reindexing_ops(
...
+            X = coords
+            nbrs = NearestNeighbors(
+                n_neighbors=min(11, len(X)), algorithm="brute", metric="cosine"
+            ).fit(X)
+            distances, indices = nbrs.kneighbors(X)
+            average_distance = np.array([dist[1:].mean() for dist in distances])
+            i = average_distance.argmin()
+            xis = np.array([X[i]])
+            coordstr = ",".join(str(round(i, 4)) for i in xis[0])
+            logger.info(f"High density cluster seed coordinates: {coordstr}")
```
rjgildea, line 418:
```suggestion
            # We no longer need the reference dataset, so remove it from the dataset ids
            self.dataset_ids = self.dataset_ids.select(
                self.dataset_ids != n_datasets - 1
            )
            n_datasets -= 1
            # FIXME trim coords and coords_reduced?
        else:
            # choose a high density point as seed
            X = coords
            nbrs = NearestNeighbors(
                n_neighbors=min(11, len(X)), algorithm="brute", metric="cosine"
            ).fit(X)
            distances, indices = nbrs.kneighbors(X)
            average_distance = np.array([dist[1:].mean() for dist in distances])
            i = average_distance.argmin()
            xis = np.array([X[i]])
```

jbeilstenedmands: This can be useful for debugging in trickier cases, as it's not usually obvious where the highest cos-angle density is, but I think the debug log is more appropriate, so I've moved it to `logger.debug`

jbeilstenedmands:
The `CosymAnalysis` class now takes a parameter `seed_dataset: Optional[int] = None`, to indicate the dataset in the intensities list that should be used to seed the clustering. If you're happy with this change then I think this is ready for merge.

rjgildea (approved):

PR #2155, opened 2022-06-28
Constraints mapping bug

dagewa:
Fixes an issue in which models which are shared by many experiments (such as detector models in serial crystallography) were being identified by `id` being set to the index in the list of those models, not the index of the experiment list.

For a concrete example using data shared by @nksauter this had the effect that that this `.phil` snippet:
```
      constraints {
        id = 1, 5000
        parameter=Dist
      }
```
attempted to apply constraints to the 2nd and 5001st detector, not the first and second detectors, picked out by experiments with `id == 1` and `id == 5000`.

### DIFF
--- a/newsfragments/xxx.bugfix
+++ b/newsfragments/xxx.bugfix
@@ -1,0 +1,1 @@
+``dials.refine``: correct an error in mapping constraint definitions to models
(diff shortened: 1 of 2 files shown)


dagewa:
I'm confident this bug fix is correct, so merging now.

PR #2181, opened 2022-07-29
Fix pytest-nunit version for windows to avoid test script crash on azure

jbeilstenedmands:


jbeilstenedmands:
Merging to get the windows azure test jobs back up and running, will continue to investigate the underlying cause.

PR #2182, opened 2022-08-01
Handle shared experiment models when converting to cambridge frame in mtz export

jbeilstenedmands:
If an experiment list has shared experiment models e.g. shared Beam model, the transformations end up being attempted multiple times as the function loops over experiments. I found that this caused a crash in the rotate beam function, but in theory it could lead to an issue in any shared model. So simply make a `deepcopy` of the model if the model is shared.


dagewa (approved):
Good spot. As MTZ does not handle shared models anyway, making distinct copies first to simplify later bookkeeping makes sense.

jbeilstenedmands:
I think this might also need backporting to the LTS branch, as it fixes a bug introduced in https://github.com/dials/dials/pull/2054 which was also backported to the LTS branch

Issue #2042, opened 2022-03-23
[Feature Request] dials.symmetry Should Use a Fourier Analysis to Detect Screw Axes


kmdalton:
As discussed in  #1742, Fourier analysis is an effective strategy to detect screw axes which is robust to systematic errors in background estimates. [Pointless](http://scripts.iucr.org/cgi-bin/paper?S090744491003982X), appears to use such a strategy to detect screw axes. Currently, `dials.symmetry` finds screw axes by testing whether systematic absences are significantly different from zero. It would be better if `dials.symmetry` _also_ implemented a Fourier approach. There is some example numpy code in #1742 which might be helpful.

dagewa:
Here is a link to a dataset from a user, for which `dials.symmetry` returns P2, but `pointless` (correctly) chooses P21. https://drive.google.com/file/d/1F-Epkowjmo9CMJ_ltqticgERs-sV6GNZ/view?usp=sharing

kmdalton:
I had a look into these data, @dagewa. Indexing actually fails for a different reason. Namely, these are not indexed in the appropriate convention when `dials.symmetry` passes them to `dials.algorithms.symmetry.absences.run_absences_checks.run_systematic_absences_checks`. This leads to the program searching the wrong axis for systematic absences. I'll try to look into this more this week, but I am afraid I am juggling a lot right now. It feels like it might be a bug.

kmdalton:
I _think_ [this](https://github.com/dials/dials/compare/main...kmdalton:dials:main) is a fix for the P21 data set. Although I'm a little bleary eyed.

dagewa:
Ah I see, sorry @kmdalton I didn't realise this was a different underlying issue. Thanks for the suggested fix. Perhaps it is easiest to set this up as a PR. I think @rjgildea and/or @jbeilstenedmands are probably best placed to look at it

kmdalton:
Yes, agreed. I'll try to submit a PR this afternoon.

kmdalton:
And thanks for that data set, it was helpful. Perhaps we can add it to the tests.

kmdalton:
resolved by #2184

Issue #1742, opened 2021-06-14
dials.symmetry says lysozyme is P4


graeme-winter:
[symm_wrong.zip](https://github.com/dials/dials/files/6649041/symm_wrong.zip)

That is not a problem per se

```
Estimation of resolution for Laue group analysis

Removing 17 Wilson outliers with E^2 >= 16.0
Resolution estimate from <I>/<σ(I)> > 4.0 : 2.43
Resolution estimate from CC½ > 0.60: 5.65
High resolution limit set to: 2.43
Selecting 3145 reflections with d > 2.43

Input crystal symmetry:
Unit cell: (37.4233, 79.3749, 79.6591, 89.8611, 89.6826, 89.9018)
Space group: P 1 (No. 1)
Change of basis op to minimum cell: a,b,c
Crystal symmetry in minimum cell:
Unit cell: (37.4233, 79.3749, 79.6591, 89.8611, 89.6826, 89.9018)
Space group: P 1 (No. 1)
Lattice point group: P 4/m m m (c,a,b)

Overall CC for 1559 unrelated pairs: 0.097
Estimated expectation value of true correlation coefficient E(CC) = 0.991
Estimated sd(CC) = 1.023 / sqrt(N)
Estimated E(CC) of true correlation coefficient from identity = 0.920

--------------------------------------------------------------------------------

Scoring individual symmetry elements

+--------------+--------+------+-----+----+---------------+
|   likelihood |   Z-CC |   CC |   N |    | Operator      |
|--------------+--------+------+-----+----+---------------|
|        0.423 |   4.23 | 0.62 |  48 |    | 1 |(0, 0, 0)  |
|        0.84  |   8.64 | 0.86 | 356 | ** | 4 |(1, 0, 0)  |
|        0     |   0    | 0    |   0 |    | 2 |(1, 0, 0)  |
|        0.456 |   6.94 | 0.69 | 398 |    | 2 |(0, 1, 0)  |
|        0.552 |   1.17 | 0.6  |   4 | *  | 2 |(0, 0, 1)  |
|        0.348 |   6.36 | 0.64 | 118 |    | 2 |(0, 1, 1)  |
|        0.459 |   6.95 | 0.69 | 296 |    | 2 |(0, -1, 1) |
+--------------+--------+------+-----+----+---------------+

--------------------------------------------------------------------------------

Scoring all possible sub-groups

+-------------------+----+--------------+----------+--------+--------+------+-------+---------+--------------------+
[...]

PR #2183, opened 2022-08-01
Reindex refls_for_sym to reference setting in dials.symmetry

kmdalton:
In #2042, @dagewa [shared a data set](https://github.com/dials/dials/issues/2042#issuecomment-1112213738) which should index as `P 1 21 1`. However, `dials.symmetry` reports `P 1 2 1`. Looking into this, I realized that the underlying cause is that the reflection table passed into the systematic absence search has not been reindexed to the canonical setting of this Laue group. This leads to `dials.symmetry` searching the wrong crystallographic basis for absent reflections. I believe this PR fixes the bug, but I need some more familiar eyes on the problem. 

I don't think I have the ability to assign reviewers. Can @jbeilstenedmands and perhaps @rjgildea have a look?

PR #2184, opened 2022-08-01
Use Fourier analysis in ScrewAxis.score

kmdalton:
As discussed in  #1742 and  #2042, Fourier analysis can be used to test for systematic absences due to screw axes. Here is a version of `dials.symmetry` which uses Fourier analysis. I have not extensively tested it, but I can confirm it works for [my data](https://github.com/dials/dials/issues/1742#issuecomment-1053811733) which was not successfully processed by the current algorithm. In this branch, I have overridden the original algorithm. Perhaps, @jbeilstenedmands can work out how to report the results from both methods.


jbeilstenedmands (changes requested):
Thanks @kmdalton for implementing this, it will be a good addition to the program.

I've made a few changes to help with the output reporting and to fix the crashes from the tests. For now I've added a phil parameter `method=*direct fourier` to help test and compare the options, once this is working robustly I can try to implement an auto option to run both and report any discrepancies.

One issue I found, I was testing on a dataset (sg=`P 41 21 2`) where there is only data between l=13 and l=20 along the c-axis. With the fourier method, the 41 axis is not found, I think because the fourier transform is given direct space data containing zeros from l=0 to l=12, when in fact this data is just missing so probably shouldn't be part of the input for the fourier transform. I wasn't sure how this should be fixed, but the data is attached here, I wondered if you had any thoughts? p.s. to run this dataset successfully we also need your fix from #2183 🙂 
[scaled.zip](https://github.com/dials/dials/files/9241649/scaled.zip)

kmdalton:
I think your intuition to drop `l < 13` is correct. Let me have a peek at your data and get back to you.

jbeilstenedmands:
By the way, my data is the same dataset ("x4_wide") that Richard mentioned in #2183, although here I have only processed the first 30 images, which might make the issue of incomplete data along the fourfold axis more apparent.

kmdalton:
A few observations
 - The probability of the `4_1` screw axis is already pretty high (~0.89), it may be the case that the appropriate thresholds are different for the Fourier model.
 - Removing the zero padding at low `l` (52356592bfb40aa09936aa1b418461aa849c16ee), increases the signal above the default threshold. 
 - I believe the `2_1` screw cannot be detected in these data, because there are no axial reflections to use along `k`. 
 
 I don't see any harm in removing the low resolution indices from the Fourier transform. It shouldn't, AFAIK, cause any issues detecting periodic signals.

kmdalton:
@jbeilstenedmands , let me know if there is more I should do on this PR. As far as I can tell, all that is left is to make choices about how to integrate the new method into the CLI.  I don't have strong opinions on that, but let me know if I can be helpful somehow.

jbeilstenedmands:
I have a few questions about the Fourier transform method, as I'm not sure it's suitable for incomplete data.
Consider this data for a 41 screw axis:
l = [11 12 13 14 16 17 18 19 20]
I/sig = [0.31, 96.77, 0.04, 0.29, 9.06, 0.46, 0.07, 0.12, 10.46].
newplot - 2022-08-04T111857 317

As I understand it, the `np.fft.fft` function requires evenly spaced values (centered about zero?). If we shift the hkl, we still need to make sure the fourfold periodicity is aligned. So should we insert some zeros near the origin, or center the transform on l=16 in this case, and how would this affect the transform? But what if the l=16 data point was missing, and do we need to insert a zero l=15? I wonder if a better way is to manually calculate a Fourier transform based only on the input data, at the frequencies we are interested in? Sorry this is a tricky example but I'm certain the programs see much worse data than this sometimes!

kmdalton:
All good questions. The location of the data along the `l-axis` actually doesn't matter. I constructed the statistical test using only the amplitudes of the Fourier transform which are translation invariant. Here's an example based on the data from your last post:
image
You can see that the amplitudes of the FT (top right) are the same no matter how I translate the data (top left). The phases (bottom right) are a different story. In short, I don't think we need to worry about how we zero pad or whether the data are zero-centered. 

The question of missing data is more interesting. There are a lot of things one could imagine doing which range from the very naive to the very fancy. Along this spectrum, I would put
 - line scan the missing data points with values in the observed i/sigma range (perhaps appropriately scaled by resolution) and take the average probability of the screw axis
 - use optimization to impute the missing data point in a way that maximizes the probability of the screw axis
 - use mcmc to compute a posterior probability of the screw axis 
FFTs are super cheap to compute. So, really anything you can imagine is on the table. I think this is a good project to work on in the future, but I don't think it should stop us from adding this code to DIALS in meantime. 

Looking back, I did notice that my null model is not quite right. I think need to exclude the points at `-n//3, -n//4, -n//6` from the set used to compute the moments of the null model. I'll get to this today. 

I would also like to make sure this works with some data with a `6_{1/5}` screw. Do you have anything like that? Maybe it's already in the tests? 

Source code for the plot: 
```python
import numpy as np
from matplotlib import pyplot as plt


l = [11, 12, 13, 14, 16, 17, 18, 19, 20]
isigi = [0.31, 96.77, 0.04, 0.29, 9.06, 0.46, 0.07, 0.12, 10.46]

l = np.array(l)
isigi = np.array(isigi)

l = l - l.min()
n = l.max() + 1 + 4
n = 6 * ((n // 6) + 1)

f,((ax1,ax2),(ax3,ax4)) = plt.subplots(2, 2)
cmap = plt.get_cmap('cividis')
for i in range(n - l.max()):
    direct_space = np.zeros(n)
    direct_space[l + i] = isigi
    color = cmap(i / (n - l.max()))
    freq_space = np.fft.fft(direct_space)
    ax1.plot(direct_space, color=color)
    ax2.plot(np.abs(freq_space), color=color)
    ax4.plot(np.angle(freq_space), color=color)

for i in [2, 3, 4, 6]:
    kwargs = {'color' : 'k', 'linestyle' : '--'}
    ax2.axvline(n // i, **kwargs)
    ax2.axvline(n - n // i, **kwargs)

plt.sca(ax1)
plt.title("Miller Space")
plt.xlabel("l")
plt.ylabel("$I / \sigma$")

plt.sca(ax2)
plt.title("Frequency Space")
plt.xticks([])
plt.xlabel("1 / l")
plt.ylabel("$|FT(I / \sigma)|$")

plt.sca(ax4)
plt.title("Frequency Space")
plt.xticks([])
plt.xlabel("1 / l")
plt.ylabel("$\Phi(FT(I / \sigma))$")

ax3.axis('off')

plt.show()
```

kmdalton:
There may be some issues detecting threefold axes with this code... investigating.

jbeilstenedmands:
Thanks for the explanation and plots regarding shifts in the data, that makes sense.
For tests, it might be better to slightly refactor the Fourier transform aspect of `score_axis_fourier` into a standalone function that takes in the miller indices, isigi and axis_repeat and returns p_screw, then we can test this on it's own with some example intensity patterns rather than a full dataset.
I agree the missing data is probably not going to affect the result much in most cases and we can quickly implement something sensible for now.
One final thing, given that we know what kind of axis we're testing for based on the axis_repeat, can we be a bit cleverer in working out what repeats can form the null model based on the axis symmetry? What I mean is that for a potential 4_1 axis, you will never have a threefold (or sixfold) in that direction, so n // 3 should be in the null model in this case. Similarly for a 6-fold, the 4-fold repeats will be in the null model?

kmdalton:
> One final thing, given that we know what kind of axis we're testing for based on the axis_repeat, can we be a bit cleverer in working out what repeats can form the null model based on the axis symmetry? What I mean is that for a potential 4_1 axis, you will never have a threefold (or sixfold) in that direction, so n // 3 should be in the null model in this case. Similarly for a 6-fold, the 4-fold repeats will be in the null model?

Yeah I think this is a good idea. I had the same thought.

kmdalton:
> For tests, it might be better to slightly refactor the Fourier transform aspect of score_axis_fourier into a standalone function that takes in the miller indices, isigi and axis_repeat and returns p_screw, then we can test this on it's own with some example intensity patterns rather than a full dataset.

As of e3b6b3f0551e21ec58f91050bde12b0c6f145691, there is a separate `@staticmethod` which takes as input the miller indices along an axis, the I/SigI (or whatever quality measure you like), and the periodicity. The method returns the probability of the screw axis under the model. This could be used for testing directly.

jbeilstenedmands:
Thanks, I added testing of the Fourier method to the existing tests in `tests/algorithms/symmetry/absences/test_screw_axes.py`, and it works for those examples, I notice there is no threefold example so if you have one that would be useful to add there. I'm generally happy with the Fourier implementation, I will add the amplitude vs frequency plot into the html report as I think that would be very nice and then find a way to run both and give suitable output 👍

kmdalton:
Awesome -- I think this is a nice addition to DIALS. We have a bunch of data sets in `P 31 2 1` at least two of them are deposited 
 - https://data.sbgrid.org/dataset/834/
 - https://data.sbgrid.org/dataset/845/
 
Either of these should be suitable for making a test.  These are pretty long sweeps though. I don't think you'd want to process the whole thing just for a test. The reflection table I have on hand (from my labmate) is ~1GB.

kmdalton:
@jbeilstenedmands I hadn't quite understood how your tests were written. I was able to add the `3_1` screw axis test you requested. Let me know if there is anything else that needs doing on this PR.

jbeilstenedmands:
Thanks for adding the extra test. I've now added the fourier amplitudes plot to the html report, which looks something like this for a fuller dataset, let me know if there is any obvious issue here:
newplot - 2022-08-09T131409 826

kmdalton:
One minor comment on the plot. Perhaps you can just show lines for the expected repeats for a particular screw axis. I think that'd be less confusing for users. What do you think?

kmdalton:
One more question -- does the x-axis label get updated to be "1 / h" and "1 / k" depending on the `axis_idx`? I'm having a little trouble figuring out where that gets set in the plot code.

jbeilstenedmands:
> One more question -- does the x-axis label get updated to be "1 / h" and "1 / k" depending on the `axis_idx`? I'm having a little trouble figuring out where that gets set in the plot code.

I actually just changed that to 'frequency' to be more general as I thought "1 / l" or equivalent could be a little confusing. Regarding the vertical lines, I find having them all helpful to gauge where the different frequencies lie on the x-axis, but maybe it's a personal preference. From a user perspective it's probably helpful to keep them there to help diagnose any unexpected behaviour and understand the plot? I had to think for a while to understand exactly what the plot was showing at first 🙂

kmdalton:
Can you make the expected frequencies a different style or color? I just want users to know where to expect signal to be. Maybe make the expected frequencies black and the others a lighter grey? 

The `frequency` works for me. I assume this is going right beside a plot of I/SigI vs {h,k,l}? So it should be clear from context what axis is being analyzed.

jbeilstenedmands (approved):
Thanks a lot for this work and addressing my concerns, I'm happy with the algorithm and know it will be a great addition to the program. 
I think a new point release of DIALS is due in the next couple of week (3.11), so I propose that the direct method should be left as the default for that release, and then once that is out we can switch to the fourier method being the default on the main branch, which enables sufficient time for wider user testing before it becomes the default in 3.12 onwards.
p.s. I have added you to the Authors list of DIALS, as is customary for contributions 🙂

kmdalton:
I think these are just harmonics. You also get a peaks at 1/3 & 1/2 for sixfold periodicity. 
image

This is why I wrote the code to exclude these peaks from the null model in the case of four and sixfold screws. 

Here's the source for that plot in case it is useful:
```python
import numpy as np
from matplotlib import pyplot as plt
import seaborn as sns
import pandas as pd

n=48

periodicity = 2

periodicities = [7, 8]
cmap = plt.get_cmap('cividis')

data = {'1 / h' : np.arange(n)}
for periodicity in periodicities:
    x = np.zeros(n)
    x[np.arange(n)%periodicity==0] = 1.
    data[f"{periodicity}"] = np.abs(np.fft.fft(x))

data = pd.DataFrame(data)
print(data)
sns.lineplot(data=data.melt('1 / h', var_name='Periodicity'), x='1 / h', y='value', hue='Periodicity', palette='Set1')
plt.xticks([])

peaks = {
    n/6    : "1/6",
    n/4    : "1/4",
    n/3    : "1/3",
    n/2    : "1/2",
    2*n/3  : "1/3",
    3*n/4  : "1/4",
    5*n/6  : "1/6",
}

for freq in peaks.keys():
    plt.axvline(freq, linestyle='--', color='k')

plt.xticks(list(peaks.keys()), list(peaks.values()))
plt.show()
```

Issue #2186, opened 2022-08-02
Wilson B factor missing from dials.merge summary statistics


rjgildea:
xia2 output includes the Wilson B factor in the summary statistics table:
```
$ grep " Overall " xia2.txt -A 21
For AUTOMATIC/DEFAULT/SAD                    Overall    Low     High
High resolution limit                           1.09    2.95    1.09
Low resolution limit                           29.94   29.95    1.11
Completeness                                   96.7   100.0    71.7
Multiplicity                                    5.5     5.7     2.5
I/sigma                                         9.9    71.9     0.1
Rmerge(I)                                     0.093   0.036   2.613
Rmerge(I+/-)                                  0.075   0.021   2.783
Rmeas(I)                                      0.102   0.040   3.193
Rmeas(I+/-)                                   0.090   0.025   3.712
Rpim(I)                                       0.041   0.016   1.790
Rpim(I+/-)                                    0.048   0.013   2.436
CC half                                       0.999   0.999   0.311
Wilson B factor                              10.900
Anomalous completeness                         93.5    99.8    53.7
Anomalous multiplicity                          3.0     3.5     1.5
Anomalous correlation                         0.577   0.795   0.028
Anomalous slope                               0.381
dF/F                                          0.098
dI/s(dI)                                      0.672
Total observations                            82977    5078    1345
Total unique                                  15115     893     539
```
however, this is absent from the otherwise equivalent table in dials.merge output.

I had a quick look at this, however it doesn't seem trivial to fix as although the Wilson B factor is calculated as part of the xtriage analysis, this is only performed if json and/or html output is being generated. Perhaps it would be better to simply perform the xtriage analysis independently of whether we are generating json or html output?

rjgildea:
Closed by https://github.com/dials/dials/pull/2188

PR #2188, opened 2022-08-02
Add Wilson B factor to merging stats output in dials.merge

jbeilstenedmands:
For #2186 
I think this is all we need, as we were getting the B-factor from the Wilson scaling analysis separate from the xtriage analysis.

### DIFF
--- a/newsfragments/2188.misc
+++ b/newsfragments/2188.misc
@@ -1,0 +1,1 @@
+``dials.merge``: Add Wilson B factor to merging stats summary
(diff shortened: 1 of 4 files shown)


rjgildea (approved):
Thanks for the quick fix!

I re-ran dials.merge on the dataset in #2186 and found only a minor difference in the 3rd decimal place (I think that is because xia2 is parsing the figure from the xtriage output, reported to 2 d.p., and then reporting in to 3 d.p. in the summary statistics table).

```
$ dials.merge DataFiles/AUTOMATIC_DEFAULT_scaled.{expt,refl}
$ diff <(grep " Overall " xia2.txt -A 21) <(grep " Overall " dials.merge.log -A 21)
1c1
< For AUTOMATIC/DEFAULT/SAD                    Overall    Low     High
---
>                                              Overall    Low     High
14c14
< Wilson B factor                              10.900
---
> Wilson B factor                              10.903
```

Issue #2190, opened 2022-08-03
`image_viewer`: beam center position is shifted by (0.5, 0.5)


elena-pascal:
The beam center in image_viewer is shown with a different definition of pixel positions, ie shifted by (-0.5, -0.5) wrt to the rest of the image

This is a direct beam image
![beam_center](https://user-images.githubusercontent.com/30300213/182590003-df745586-0d61-4a83-a042-8c8344bbb9ad.png)

The beam position saved is actually correct
![correct_beam_center](https://user-images.githubusercontent.com/30300213/182591130-9461f557-bd7c-427c-84d3-769e47d04760.png)

Here is what happens if I set image origin to (0, 0) as opposed to (-0.5, -0.5)
![wrong_beam_center](https://user-images.githubusercontent.com/30300213/182591415-edd3f846-75fe-463f-9725-57c3039d240f.png)

dagewa:
See also https://github.com/dials/dials/issues/1545

elena-pascal:
What is the intended behaviour? Center of pixel is integer or edge of pixel is integer?

dagewa:
I think edge is integer is correct, so top left corner should be 0,0

elena-pascal:
bottom left?

elena-pascal:
If pixel edges are desired to be integers, that is not what is happening at the moment.

dagewa:
> bottom left?

I meant top left in the `dials.image_viewer` (for most single panel detectors), which should read out `fast=0 / slow=0` but actually has both at `-0.5`

phyy-nx:
I think the readout of -0.5 for the corner of the 0,0 pixel is correct, sinc in dials the center of the pixel is defined as the 0,0 position in pixel and millimeter space, right?

nksauter:
A separate but related issue is that the mouse-over pixel intensity readout
is half pixel out of sync.  Would appreciate this diff being added to the
code:

*diff --git a/src/dials/util/image_viewer/slip_viewer/frame.py
b/src/dials/util/image_viewer/slip_viewer/frame.py*

*index 736c4e007..c10c7c92d 100644*

*--- a/src/dials/util/image_viewer/slip_viewer/frame.py*

*+++ b/src/dials/util/image_viewer/slip_viewer/frame.py*

@@ -222,8 +222,8 @@ def handle_position_event(self, event):

                 possible_intensity = None

                 fi = self.pyslip.tiles.raw_image

                 detector = fi.get_detector()

[...]

PR #2194, opened 2022-08-05
Fix beam center and intensity readout of pixel coordinates

### LINKED ISSUE #1545: dials.image_viewer coordinate offset
I'm pretty sure we've been here before, but I'm also fairly sure this was fixed at some point. Currently, the mouse pointer has an offset of about half a pixel (maybe exactly half) when displaying the intensity read out in the bottom status bar. Here's an example using `X4_wide_M1S4_2_0001.cbf`. The cursor is inside the masked region and should show intensity = -1, but instead the intensity of the pixel to the left is displayed.

image_viewer_offset

This is really annoying when you notice, and plainly misleading when you don't.

phyy-nx:
Fixes #1545 and #2190

Tested on dials_regression/ALS_831/q315r_lyso_001.img (single panel), dials_regression/LCLS_cspad_nexus/idx-20130301060858401.cbf (multi-panel) and dials_data/x4wide/X4_wide_M1S4_2_0001.cbf (cbf mini Pilatus).  The intensity readouts are now right, but I left in the -0.5 pixel offset, since I think it's right.  In other words, the 0,0 position should be the center of the pixel, as that's what dials uses to do things like predict x-rays.

I also added the beam center fix suggested by @elena-pascal, but I haven't tested it.  I do think it is right.

Co-authored-by: Elena Pascal <elenadpascal@gmail.com>


dagewa:
Thanks Aaron! The intensity readout seems fixed by this, which is great. However I remain confused about something:
```
                # the dials convention is that the center of the pixel is 0,0, so the extent
                # of the 0,0 pixel is from -0.5,-0.5 to 0.5,0.5 in pixel space.
```
I don't think this is true. I think the borders between the pixels are at integer values. That is, the corner of a panel at its origin should be at 0,0 in pixel space. The centre of the first pixel in the array should be at 0.5,0.5.

We don't have this convention written down it seems, which is a problem.

dagewa:
I think it is clear from the code though
https://github.com/cctbx/dxtbx/blob/e4bc7c4965cfbb7e43df46e9ddea56a2bbb27119/src/dxtbx/model/pixel_to_millimeter.h#L97-L100
for the `SimplePxMmStrategy` the conversion between pixels and millimetres (and vice versa) is just a multiplication (or division), without an offset of 0.5 pixels. I hope we agree on the convention that the active area of the panel has origin 0,0 in mm.

elena-pascal:
I agree with David, in the sense that while the fix fixes the current raised issue it is pushing the underlying problem further down the road. The changes between pixel units and mm are done very much not explicitly and it is indeed confusing. Since we are looking at this anyway could we not make it functional? Especially since there are available mapping functions?

rjgildea:
> I don't think this is true. I think the borders between the pixels are at integer values. That is, the corner of a panel at its origin should be at 0,0 in pixel space. The centre of the first pixel in the array should be at 0.5,0.5.
> 
> We don't have this convention written down it seems, which is a problem.

I have verified that the centroids calculated by spotfinding for one-pixel spots are `(x+0.5, y+0.5, z+0.5)` where `x, y` are the xth and yth integer coordinates of the pixel comprising the spot, and `z` is the image index:

Using the "centroid_test_data" example dataset, and overriding the spotfinding defaults to find one-pixel spots:
```
$ dials.import $(dials.data get -q centroid_test_data)/*.cbf
$ dials.find_spots imported.expt min_spot_size=1
```
and then using the following script to find some one-pixel spots:
```
$ cat one_pixel_spots.py 
import sys

refl = flex.reflection_table.from_file(sys.argv[1])

sel = refl["n_signal"] == 1
print(f"Number of one-pixel spots: {sel.count(True)}")

xyz_sel = refl["xyzobs.px.value"].select(sel)
for xyz in xyz_sel[:10]:
    print(f"{xyz=}")
```
gives the following output:
```
$ dials.python one_pixel_spots.py strong.refl 
Number of one-pixel spots: 206
xyz=(1260.5, 536.5, 0.5)
xyz=(911.5, 639.5, 0.5)
xyz=(1373.5, 715.5, 0.5)
xyz=(1375.5, 715.5, 0.5)
xyz=(580.5, 910.5, 0.5)
xyz=(1400.5, 984.5, 0.5)
xyz=(1174.5, 1012.5, 0.5)
xyz=(768.5, 1039.5, 0.5)
xyz=(767.5, 1040.5, 0.5)
xyz=(870.5, 1089.5, 0.5)
```

IMO the only behaviour that makes sense is for the origin of the panel to be (0,0) and for a panel of size (m,n) then any photon impinging on the panel would have coordinates in the range (0..m, 0..n).

graeme-winter:
> IMO the only behaviour that makes sense is for the origin of the panel to be (0,0) and for a panel of size (m,n) then any photon impinging on the panel would have coordinates in the range (0..m, 0..n).

+1

phyy-nx:
Hi folks, I think I agree with all the comments.  My memory of the 0,0 position being in the center of the pixel is coming from somewhere, but @rjgildea's evidence seems great.  Maybe I'm remembering the papers from @nksauter ([section 2.3.1](https://journals.iucr.org/d/issues/2020/02/00/tz5101/index.html#DIVSEC2.3.1)) or James Holton ([section 2.3](https://journals.iucr.org/d/issues/2019/02/00/ba5297/index.html#DIVSEC2.3)).

I'll revise this PR tomorrow and include updating the graphical display of the coordinates to match this convention.

src/dials/util/image_viewer/slip_viewer/frame.py
```diff
@@ -203,16 +203,16 @@ def handle_position_event(self, event):
             )
 
             posn_str = "Picture:  fast={:.3f} / slow={:.3f} pixels.".format(
-                fast_picture,
-                slow_picture,
+                fast_picture + 0.5,
```
voklejas2, line 206: Kind of a lazy choice to add the offset here- rather than fix the function that produces the coordinates. I briefly looked into changing some of the position functions, namely via the `view_offset` parameters, in pyslip.py, but this wasn't as straightforward as I had hoped and I reverted to the lazy code change.

voklejas2: Lazy correction to add 0.5 offset in pix location lines 206-207 and 215. This restores enforces the convection that the origin be located at the corners (0,0).

phyy-nx: Updated to only adjust coords[0] and [1], and to remove the extra half pixel add a bit later.  Comment moved and rewritten to reflect the above converations

phyy-nx:
Ready for review!

phyy-nx:
@elena-pascal can you post that image somewhere?  Thanks!

phyy-nx:
I see the png in the github post above, but I was hoping to download the original db_01_00001.cbf.gz file so we could test it here.  Unless that's linked somewhere I'm not seeing?

phyy-nx:
For reference, here was the testing we did on a single panel image.  Print header shows this:
```
Panel:
  name: Panel
  type: SENSOR_UNKNOWN
  identifier:
  pixel_size:{0.132812,0.132812}
  image_size: {2560,2560}
  trusted_range: {-400,65535}
  thickness: 0
  material:
  mu: 0
  gain: 1
  pedestal: 0
  fast_axis: {0,1,0}
  slow_axis: {1,0,0}
  origin: {-170,-170,-196.5}
  distance: 196.5
  pixel to millimeter strategy: SimplePxMmStrategy
```

Just to verify things make sense, there is a bit more precision in the pixel size in the image header, so 0.1328125*2560=340mm, and the relevant axis for the origin is being interpreted correctly:
```
 AXIS_L0M0_L1M0 rotation detector AXIS_L0M0_R 1.0 0.0 0.0 -170.0 170.0 -196.5 detector_level_1^M
```

So with an s0 of 0,0,-1, the beam center will be in the middle of the 340mm^2 detector, which should be at coordinate 1280,1280, which is exactly what I observe.

elena-pascal:
This might work [db_01_00001.cbf.gz](https://github.com/dials/dials/files/9822420/db_01_00001.cbf.gz)

phyy-nx:
Got it.  Thanks, we'll look at it.

phyy-nx:
Ok, I believe on the `pix_offset_fix` branch, the image viewer is displaying what dxtbx is telling it to display for image [db_01_00001.cbf.gz](https://github.com/dials/dials/files/9822420/db_01_00001.cbf.gz).

```
dxtbx.print_header db_01_00001.cbf.gz | grep origin
  origin: {-148.77,-127.33,-160}
```
Given pixel size of 0.172, thats 864.94186, 740.29070, which is what the image viewer displays.

So the problem actually is that dxtbx is reading the header wrong for MiniCBF!  ImageCIF specifies counting from the _center_ of the pixel, not the cornes, as shown in these two references:
http://www.bernstein-plus-sons.com/software/CBF/doc/cif_img_1.8.4.html#_diffrn_detector_element.reference_center_fast
http://www.bernstein-plus-sons.com/software/CBF/doc/cif_img_1.8.4.html#array_structure_list_axis

Compare to the relevant lines from FormatCBFMini, where we aren't subtracting 0.5 from the beam_xy parameter.
https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/format/FormatCBFMini.py#L128
https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/format/FormatCBFMini.py#L170

Likewise, FormatCBFFullPilatus uses [imgCIF_H](https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/model/detector.py#L771), which calls cbflib's get_pixel_coordinates(0, 0), which according to the [docs](http://www.bernstein-plus-sons.com/software/CBF/doc/CBFlib.html#2.4.44), returns the center of pixel 0,0.  So to go from ImageCIF to dxtbx we need to apply a half pixel correction.  So with this diff, the image displays properly:
```
diff --git a/src/dxtbx/model/detector.py b/src/dxtbx/model/detector.py
index 399fde70..ceeee2a0 100644
--- a/src/dxtbx/model/detector.py
+++ b/src/dxtbx/model/detector.py
@@ -768,7 +768,7 @@ class DetectorFactory:
             cbf_detector.get_inferred_pixel_size(2),
         )

-        origin = tuple(cbf_detector.get_pixel_coordinates(0, 0))
+        origin = tuple(cbf_detector.get_pixel_coordinates(-0.5, -0.5))
         fast = cbf_detector.get_detector_axis_fast()
         slow = cbf_detector.get_detector_axis_slow()
         size = tuple(reversed(cbf_handle.get_image_size(0)))
```

Also note, @elena-pascal, your original matplotlib graph that shows a correct beam center (#2190) likely used [matplotlib's coordinate system](https://matplotlib.org/stable/tutorials/intermediate/imshow_extent.html#default-extent), where 0,0 is the center of a pixel.  So if you draw a cross in matplotlib at cbf_detector.get_pixel_coordinates(0, 0), it will look right.  But in our image viewer where the corner of the pixel is 0,0, it's wrong, without the above correction to the d0 vector in dxtbx.

Note, this is like the cause of the 0.5 pixel offset problem @nksauter worked through in [2020](https://journals.iucr.org/d/issues/2020/02/00/tz5101/index.html) (see section 2.3.1).

Also note if we fix all the CBF format classes by subtracting this half pixel offset, we'll need to make a corresponding change to [cbf_writer](https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/format/cbf_writer.py#L308-L354), which writes CBFs following the dxtbx convention instead of the ImageCIF convention.  It will need to add a half pixel to the vectors it writes!

Whew!!!!

voklejas2:
Nice, Aaron!

elena-pascal:
@phyy-nx Thanks for documenting what the source of inconsistency was!

phyy-nx:
Thanks!  I've copied the above discussion about dxtbx and half pixel offsets for cbfs to a new issue, cctbx/dxtbx#564.  I think that's independent of this PR.

src/dials/util/image_viewer/slip_viewer/frame.py
```diff
@@ -203,22 +203,26 @@ def handle_position_event(self, event):
...
             coords = self.pyslip.tiles.get_flex_pixel_coordinates(lon, lat)
             if len(coords) >= 2:
                 if len(coords) == 3:
                     readout = int(round(coords[2]))
                 else:
                     readout = -1
-
+                # the dials convention is that the center of the pixel is 0.5,0.5, so the extent
+                # of the 0,0 pixel is from 0.0,0.0 to 1.0,1.0 in pixel space.  Here we are
+                # get_flex_pixel_coordinates defines the center of the pixel as 0.0,0.0, so we
```
graeme-winter, line 217: And here is where we have the inconsistency betwixt what we think in DIALS and what the image viewer says.

We should probably add a comment to the implementation of `get_flex_pixel_coordinates` that it defines the 0,0 postion as the centre

newsfragments/2194.bugfix
```diff
@@ -0,0 +1 @@
+Fix intensity readout mouseover and beamcenter half-pixel error in the dials image viewer
```
graeme-winter, line 1: `dials.image_viewer`:

please

src/dials/util/image_viewer/slip_viewer/frame.py
```diff
@@ -450,6 +454,7 @@ def map_coords(x, y, p):
                 y, x = self.pyslip.tiles.flex_image.tile_readout_to_picture(
                     p, y - 0.5, x - 0.5
                 )
+                # y, x = self.pyslip.tiles.flex_image.tile_readout_to_picture(p, y, x)
```
graeme-winter, line 457: @phyy-nx  says will delete

graeme-winter (approved):
Dang, lots of discussion for a very modest change. 

Modulo pls could you add a comment to the `get_flex_pixel_coordinates` doc string looks good to me

phyy-nx:
Windows failures look unrelated, so merging.  Thanks all.

phyy-nx:
Note, I left #2190 open, as this PR didn't fix it.  Fixing cctbx/dxtbx#564 will fix #2190.

PR #2195, opened 2022-08-08
Rigorous definition of `trusted_range`

dagewa:
Work in progress to define the `trusted_range` as the inclusive range [min-valid-value, max-valid-value] and use this consistently throughout the codebase.

Requires https://github.com/cctbx/dxtbx/pull/536

Currently 11 test failures. Some of these are due to loading old experiment lists with trusted range starting from -1.

```
         - tests/algorithms/background/test_gmodel.py:26 test_simple
         - tests/algorithms/background/test_gmodel.py:87 test_robust
         - tests/command_line/test_background.py:6 test
         - tests/command_line/test_integrate.py:365 test_multi_sweep
         - tests/command_line/test_integrate.py:18 test_basic_integrate
         - tests/command_line/test_integrate.py:145 test_basic_threaded_integrate
         - tests/command_line/test_integrate.py:174 test_basic_integrate_output_integrated_only
         - tests/command_line/test_integrate.py:210 test_integration_with_sampling
         - tests/command_line/test_integrate.py:242 test_integration_with_sample_size
         - tests/command_line/test_integrate.py:337 test_basic_integration_with_profile_fitting
         - tests/command_line/test_model_background.py:11 test_model_background
```

### DIFF
--- a/src/dials/algorithms/shoebox/overload_checker.h
+++ b/src/dials/algorithms/shoebox/overload_checker.h
@@ -32,10 +32,10 @@ namespace dials { namespace algorithms { namespace shoebox {
       typedef Shoebox<>::float_type float_type;
 
       /**
-       * @param overload - The overload values for each pixel
+       * @param saturation - The saturation values for each pixel
        */
-      Checker(const af::const_ref<double> &overload)
-          : overload_(overload.begin(), overload.end()) {}
+      Checker(const af::const_ref<double> &saturation)
+          : saturation_(saturation.begin(), saturation.end()) {}
 
       /**
        * Mark all pixels that are overloaded as invalid
@@ -46,13 +46,13 @@ namespace dials { namespace algorithms { namespace shoebox {
       bool operator()(std::size_t panel,
                       const af::const_ref<float_type, af::c_grid<3> > &data,
                       af::ref<int, af::c_grid<3> > mask) const {
-        DIALS_ASSERT(panel < overload_.size());
+        DIALS_ASSERT(panel < saturation_.size());
         DIALS_ASSERT(data.accessor().all_eq(mask.accessor()));
         DIALS_ASSERT(data.size() == mask.size());
         bool result = false;
-        double overload_value = overload_[panel];
+        double saturation_value = saturation_[panel];
         for (std::size_t i = 0; i < data.size(); ++i) {
-          if (data[i] >= overload_value) {
+          if (data[i] > saturation_value) {
             mask[i] &= ~Valid;
             result = true;
           }
@@ -61,16 +61,16 @@ namespace dials { namespace algorithms { namespace shoebox {
       }
 
     private:
-      af::shared<double> overload_;
+      af::shared<double> saturation_;
     };
 
   public:
     /**
-     * Add the overload values for this detector
-     * @param overload The overloads for each panel
+     * Add the saturation values for this detector
+     * @param saturation The saturation value for each panel
      */
-    void add(const af::const_ref<double> &overload) {
-      checker_.push_back(Checker(overload));
+    void add(const af::const_ref<double> &saturation) {
+      checker_.push_back(Checker(saturation));
     }
 
     /**
--- a/src/dials/algorithms/spot_finding/factory.py
+++ b/src/dials/algorithms/spot_finding/factory.py
@@ -313,7 +313,7 @@ class BackgroundGradientFilter:
                         and x < (ex2 - buffer_size)
                     ):
                         mask[0, i_y, i_x] = False  # foreground
-                    elif value > trusted_range[0] and value < trusted_range[1]:
+                    elif value >= trusted_range[0] and value <= trusted_range[1]:
                         mask[0, i_y, i_x] = True  # background
 
             model = modeller.create(data.as_double(), mask)
--- a/src/dials/algorithms/integration/kapton_2019_correction.py
+++ b/src/dials/algorithms/integration/kapton_2019_correction.py
@@ -50,7 +50,7 @@ class KaptonTape_2019:
             p.set_local_frame(fast.elems, slow.elems, ori.elems)
             p.set_pixel_size((pixel_size, pixel_size))
             p.set_image_size(image_size)
-            p.set_trusted_range((-1, 2e6))
+            p.set_trusted_range((0, 2e6))
             p.set_name(f"KAPTON_{name}")
             return d
 
--- a/src/dials/algorithms/integration/parallel_integrator.h
+++ b/src/dials/algorithms/integration/parallel_integrator.h
@@ -1096,12 +1096,12 @@ namespace dials { namespace algorithms {
 
       // Get the starting frame and the underload/overload values
       int zstart = scan.get_array_range()[0];
-      double underload = detector[0].get_trusted_range()[0];
-      double overload = detector[0].get_trusted_range()[1];
-      DIALS_ASSERT(underload < overload);
+      double min_trusted = detector[0].get_trusted_range()[0];
+      double max_trusted = detector[0].get_trusted_range()[1];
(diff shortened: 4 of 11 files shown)


dagewa:
DIALS tests now pass by patching experiment list files on the fly to update the `trusted_range`. It will be better to replace these tests eventually, but this is quicker than adding `dials_data` to the dependencies for this PR

src/dials/algorithms/shoebox/overload_checker.h
```diff
@@ -32,10 +32,10 @@ namespace dials { namespace algorithms { namespace shoebox {
       typedef Shoebox<>::float_type float_type;
 
       /**
-       * @param overload - The overload values for each pixel
+       * @param saturation - The saturation values for each pixel
```
rjgildea, line 35: I appreciate that `saturation` is probably clearer than `overload` but it still seems ambiguous (without reading the code) as to whether this corresponds to the `max_trusted_value` elsewhere. At the very least a more detailed docstring/comment would be useful.

src/dials/algorithms/spot_finding/factory.py
```diff
@@ -313,7 +313,7 @@ def run(self, flags, sequence=None, shoeboxes=None, **kwargs):  # noqa: U100
                         and x < (ex2 - buffer_size)
                     ):
                         mask[0, i_y, i_x] = False  # foreground
-                    elif value > trusted_range[0] and value < trusted_range[1]:
+                    elif value >= trusted_range[0] and value <= trusted_range[1]:
```
rjgildea, line 316:
```suggestion
                    elif trusted_range[0] <= value <= trusted_range[1]:
```

tests/algorithms/background/test_gmodel.py
```diff
@@ -31,11 +33,17 @@ def test_simple(dials_data, model, tmpdir):
     reflns_simple.dirpath().ensure(dir=1)
     reflns_g_simple.dirpath().ensure(dir=1)
 
+    exp = load.experiment_list(experiments)
+    panel = exp[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    exp.as_json(tmpdir.join("trusted_range_patch.expt"))
```
rjgildea, line 40: Could you add a brief comment as to why this is necessary? I.e. what was the min_trusted_value in the unpatched experiments?

tests/algorithms/background/test_gmodel.py
```diff
@@ -92,11 +100,17 @@ def test_robust(dials_data, model, tmpdir):
     reflns_robust.dirpath().ensure(dir=1)
     reflns_g_robust.dirpath().ensure(dir=1)
 
+    exp = load.experiment_list(experiments)
+    panel = exp[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    exp.as_json(tmpdir.join("trusted_range_patch.expt"))
```
rjgildea, line 107: As above, a brief comment here would be helpful

tests/command_line/test_integrate.py
```diff
@@ -22,6 +22,12 @@ def test_basic_integrate(dials_data, tmpdir):
         dials_data("centroid_test_data", pathlib=True) / "experiments.json"
     )
     exp[0].identifier = "foo"
+
+    panel = exp[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    # exp.as_json(tmpdir.join("trusted_range_patch.expt"))
```
rjgildea, line 29:
```suggestion
```

tests/command_line/test_integrate.py
```diff
@@ -75,6 +81,13 @@ def test_basic_integrate(dials_data, tmpdir):
     with tmpdir.join("models.expt").open("w") as fh:
         json.dump(j, fh)
 
+    exp = load.experiment_list(tmpdir.join("models.expt"))
+    panel = exp[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    # exp.as_json(tmpdir.join("trusted_range_patch.expt"))
```
rjgildea, line 88:
```suggestion
```

tests/command_line/test_integrate.py
```diff
@@ -177,6 +196,12 @@ def test_basic_integrate_output_integrated_only(dials_data, tmpdir):
         dials_data("centroid_test_data", pathlib=True) / "experiments.json"
     )
     exp[0].identifier = "bar"
+
+    panel = exp[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    # exp.as_json(tmp_path / "trusted_range_patch.expt")
```
rjgildea, line 203:
```suggestion
```

tests/command_line/test_integrate.py
```diff
@@ -370,6 +413,12 @@ def test_multi_sweep(dials_regression, tmpdir):
     experiments = load.experiment_list(expts)
     for i, expt in enumerate(experiments):
         expt.identifier = str(100 + i)
+
+    panel = experiments[0].detector[0]
+    max_trusted = panel.get_trusted_range()[1]
+    panel.set_trusted_range((0, max_trusted))
+    # experiments.as_json(tmpdir.join("trusted_range_patch.expt"))
```
rjgildea, line 420:
```suggestion
```

rjgildea (approved):
Thanks again for doing this. Given the various tests where the original input experiment file is now patched in-situ, would it be worth generating an issue to record that these test data files need updating, and then referring to the issue at each point where the files are patched to aid finding them later on?

phyy-nx (approved):

dagewa:
Thanks! I have applied all those suggestions now and tests still pass, as long as https://github.com/cctbx/dxtbx/pull/536 is used.

dagewa:
I think there's a bit more to look at here. I tried a simple processing script on some datasets in `dials_data` but I see some differences in processing between the branch and main.

## insulin

- Exercises `FormatSMVADSCSN`
- Differences in spot-finding and beyond.

## x4wide

- Exercises `FormatCBFMiniPilatus`
- Identical processing.

## vmxi_thaumatin

- Exercises `FormatNXmxDLS`
- Differences in spot-finding and beyond. For some reason the `trusted_range` on import is still `{-1,65535}` on the branch as well as main.

dagewa:
I also checked:

## four_circle_eiger/01_CuHF2pyz2PF6b_Phi
- Exercises `FormatNXmxDLSI19_2`
- Identical processing

Let me know if there are any other datasets on which it would be prudent for me to do a before/after comparison.

phyy-nx:
@dagewa did any of your processing tests use reference_geometry from earlier processing?  That would have the trusted range wrong, and reference_geometry in dials.import replaces the detector model.

dagewa:
Hi @phyy-nx, no this is one of those changes that invalidates earlier processing. That's also why this PR patches various `.expt` files used in tests, because what's in those files would no longer be correct.

dagewa:
How concerned are we about this issue (change of definitions invalidates previous processing)? One possible way around this would be to introduce a version string into `.expt` files and then have a version check in DIALS that will refuse to proceed with old or unversioned files. Does that seem at all useful?

graeme-winter:
Personally I would say that reducing complexity at the cost of some lack of backwards compatibility is a trade worth making. 

Independently not hostile to idea of versioning the format but do t think we should change the behaviour here on that basis.

dagewa:
We discussed this earlier today and are keen to go forward without yet versioning `.expt` files. So, I intend to merge https://github.com/cctbx/dxtbx/pull/536 and this early next week unless any objections are raised.

PR #2197, opened 2022-08-09
Add unit cell angle distribution plots to cluster plotting code

jbeilstenedmands:
This PR add plots of unit cell angle distributions to the unit cell cluster plotting code, e.g.
newplot - 2022-08-09T094814 213
Distributions are only plotted for non-symmetry constrained angles, e.g. won't appear for cubic crystals, only beta distribution plotted for P2.

I have also implemented a heatmap version of the unit cell scatter plot to the cluster plotting code. For now I have implemented this as an option and kept the default as before for dials.cosym, xia2.multiplex etc, but I have added the new version to the dials.merge report :
New option:
newplot - 2022-08-09T094923 494
Default:
newplot - 2022-08-09T095209 856
As you can see, the default is probably more appropriate for up to several hundred crystals (the plots here are 34 crystals), but when you get into the 1000s, I think the heatmap is a more useful plot due to the point crowding.


newsfragments/2197.feature
```diff
@@ -0,0 +1 @@
+Add unit cell angle distribution plots to unit cell angle plots
```
rjgildea, line 1:
```suggestion
Unit cell clustering: Add unit cell angle distribution plots
```

src/dials/algorithms/clustering/plots.py
```diff
@@ -1,16 +1,106 @@
...
+            "grid": {"rows": 1, "columns": 3, "subplots": [["xy", "x2y2", "x3y3"]]},
+            "xaxis": {"title": "a (Å)"},
+            "yaxis": {"title": "b (Å)"},
+            "xaxis2": {"title": "b (Å)"},
+            "yaxis2": {"title": "c (Å)"},
+            "xaxis3": {"title": "c (Å)"},
+            "yaxis3": {"title": "a (Å)"},
+        },
+        "help": """\
+The distribution of the unit cell parameters: a vs. b, b vs. c and c vs.a respectively.
```
rjgildea, line 95:
```suggestion
The distribution of the unit cell parameters: a vs b, b vs c and c vs a respectively.
```

rjgildea (approved):
Thanks for adding this, it would be good to see some examples for SSX data, i.e. with 1000s of unit cells. For the scatter plot version improvements it might be worth exploring adding alpha levels to help reduce overcrowding of the plot (although probably outside of the scope of this PR).

PR #2201, opened 2022-08-10
Remove reflections

luisodls:
add one more way to run dials.filter_reflections, now by telling `remove_by_index=n1,n2,n3...`

this is useful if the user wants to remove reflections one by one manually from the GUI


dagewa (approved):
@luisodls wants this for DUI2 and the feature seems to fit well enough in the existing `dials.filter_reflections` program. I'll approve now and will merge early next week unless there are further comments.

ndevenish:
My only real problem is that it's not a generic operation. It's _extremely_ specific, based on the exact structure of the reflection table e.g. only helpful exactly after reading the table in a separate program (which presumably could have then just re-written the table itself). It doesn't fit into any kind of pipeline.

That said, It doesn't affect anything else. I might be happy just putting an extremely high expert level e.g. 5?

dagewa:
@ndevenish I've gone for your pragmatic suggestion. Is this okay now?

PR #2209, opened 2022-08-23
Rename dials.split_experiments

graeme-winter:
It can split experiments and reflections so really dials.split is more appropriate -> make the change, update the tests and so on. 

I don't know if this needs to go through a deprecation process as it is not really a mainstream `dials` user tool, but if it is could make the changes to tests etc.


graeme-winter:
Second thoughts: probably needs a deprecation process as it is used so named in xia2.

@ndevenish what is the protocol for that?

ndevenish:
We should definitely deprecate, we've mentioned this tool before. FWIW we can rename the script and have multiple `# LIBTBX_SET_DISPATCHER_NAME` lines so that both dispatchers are created.

I'd probably want this printing a message for a looooong time but am not entirely sure how safe it would be to actually remove.

graeme-winter:
> I don't have particularly strong views on this either way, but if you're changing this (to whatever) then you should probably also consider `dials.combine_experiments` (and possibly `dials.slice_sequence`) at the same time...

Noted, and not wrong, will do so. 

@ndevenish ofc yes I can leave the alias in place -> will do that

graeme-winter:
@rjgildea you may be shocked to hear that slice_sequence does not quite behave as expected 🤦‍♂️

graeme-winter:
Not going to tackle `slice_sequence` at this time - that may need a separate look at and addition of tests.

phyy-nx (changes requested):
Hello, I'm receptive to the changes, but this PR will break [cctbx.xfel.strip_experiments](https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/striping.py#L78), the wrapper that does [time-dependent ensemble refinement](https://journals.iucr.org/d/issues/2018/09/00/lp5037/lp5037fig7.html) in the XFEL GUI, due to a direct import from combine_experiments.py.  If the command dials.combine_experiments is removed later, that will break [cspad.cbf_meterology](https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/cspad_cbf_metrology.py#L279), a perhaps now legacy pipeline that does a bunch of dials steps in sequence.

Both programs are covered by tests in the Azure XFEL CI piepline as part of the xfel_regression suite so they will break too ([one](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/command_line/tst_cspad_cbf_metrology.py)) ([two](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/command_line/tst_ensemble_refinement.py)).

Before merging, I'd ask that corresponding PRs be made to these repositories as well.  Thanks.

graeme-winter:
> Hello, I'm receptive to the changes, but this PR will break [cctbx.xfel.strip_experiments](https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/striping.py#L78), the wrapper that does [time-dependent ensemble refinement](https://journals.iucr.org/d/issues/2018/09/00/lp5037/lp5037fig7.html) in the XFEL GUI, due to a direct import from combine_experiments.py. If the command dials.combine_experiments is removed later, that will break [cspad.cbf_meterology](https://github.com/cctbx/cctbx_project/blob/master/xfel/command_line/cspad_cbf_metrology.py#L279), a perhaps now legacy pipeline that does a bunch of dials steps in sequence.
> 
> Both programs are covered by tests in the Azure XFEL CI piepline as part of the xfel_regression suite so they will break too ([one](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/command_line/tst_cspad_cbf_metrology.py)) ([two](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/command_line/tst_ensemble_refinement.py)).
> 
> Before merging, I'd ask that corresponding PRs be made to these repositories as well. Thanks.

Just as a points of philosophy - 
 - I am changing command-line programs not library routines so would not generally expect thiings to be imported therefrom
 - if I change DIALS code am I responsible for making the changes to any and all places where that code is used?
 - I am also not sure I am comfortable changing `cctbx.xfel` without an easy view of the regression tests

As a compromise and a probable net improvement I could look at removing the work-code from these to some where more kin to a library then you could make the corresponding changes in the comfort of knowing that the code is unlikely to change again. 

Partly my motivation was triggered when I was explaining "how to use split_experiments to split up reflection files" which I realised was dumb.

ndevenish:
In theory the `dials.command_line` shouldn't be an API, but we know it is used as one.

How about falling back to just a deprecation - keeping `dials.command_line.split_experiments` etc as an API shim that prints a deprecation warning?

ndevenish:
Okay, so I added back in [combine_experiments.py](https://github.com/dials/dials/blob/9731c1d363bc8277806fa4dab46d1b17e0d7b6e9/src/dials/command_line/combine_experiments.py) and [split_experiments.py](https://github.com/dials/dials/blob/9731c1d363bc8277806fa4dab46d1b17e0d7b6e9/src/dials/command_line/split_experiments.py). They are now shells that do e.g. `from dials.command_line.split import *`, but also print (if `__main__`) or UserWarning (if imported) to note that it's deprecated.

This should hopefully allow a smoother deprecation cycle?

phyy-nx:
Thanks, deprecations are better!  Your points on philosophy are great @graeme-winter.  Here's a bit of noodling:

> * I am changing command-line programs not library routines so would not generally expect things to be imported therefrom

Agree regarding imports.  It would be better if the phil scopes were defined near algorithms so that code doesn't import anything from command_line.  That's the case for some parts of DIALS which is great.

I do think the answer changes when talking about pipelines though.  Pipelines like the two XFEL programs listed above rely on program names, so when those names go away the pipelines break.  It's fine! It's just an API change that needs to be accounted for.

> * if I change DIALS code am I responsible for making the changes to any and all places where that code is used?

Well, I dunno.  cctbx_project at least seems reasonable to to check at least?  Many thanks for cctbx/cctbx_project#781.

> * I am also not sure I am comfortable changing `cctbx.xfel` without an easy view of the regression tests

Sure, I get it. FWIW the XFEL tests are [easy to run](https://github.com/cctbx/cctbx_project/tree/master/xfel/conda_envs#user-content-cctbxxfel-tests), but you shouldn't have to.  Aren't they part of the PR test pipeline for dials?

phyy-nx:
Ah, nope, the xfel_regression tests are [not running](https://dev.azure.com/azure-dials/8f9a0b14-1c65-4a82-9c06-1ec7513061ae/_apis/build/builds/9339/logs/68) on PRs.  I'll dig a bit into this.

ndevenish:
> Aren't they part of the PR test pipeline for dials?

Yes, this was my recollection; it looks like it only run once all the other tests have passed - it looks like this has moved too fast for that to happen (yet?)

ndevenish:
> Ah, nope, the xfel_regression tests are [not running](https://dev.azure.com/azure-dials/8f9a0b14-1c65-4a82-9c06-1ec7513061ae/_apis/build/builds/9339/logs/68) on PRs. I'll dig a bit into this.

Ah, you are right - it is running and showing green but not doing anything. Will hold on digging if you are looking at.

phyy-nx:
Well, I think I've reached the limit of my understanding of it.  Looks like it's [missing an artifact](https://github.com/dials/dials/blob/main/.azure-pipelines/azure-pipelines.yml#L152-L153)?

phyy-nx (approved):
Approving this version with deprecations (thanks again), but I think maybe a separate issue regarding xfel_regression not running would be good?

ndevenish:
> Approving this version with deprecations (thanks again), but I think maybe a separate issue regarding xfel_regression not running would be good?

Yes, I'll have a look tomorrow otherwise. That step is just grabbing (and publishing) the results from the job run, which does appear to be happening (e.g. [earlier](https://dev.azure.com/azure-dials/dials/_build/results?buildId=9339&view=logs&jobId=728b758c-216d-5daa-c333-7c882b66b365&j=728b758c-216d-5daa-c333-7c882b66b365&t=783be0db-8d0e-53f0-f2ff-be366fafaa3d).

phyy-nx:
Thanks!!

graeme-winter:
`xia2.ssx` will need some work as well (learns he, running xia2 tests)

graeme-winter:
"People" seem to prefer an egregiously misnamed `dials.combine_experiment` which works on reflection data to `dials.combine` -> give up

PR #2213, opened 2022-08-25
Mark dials.find_spots_{server,client} as deprecated

rjgildea:
I am not aware that these are actively used (at least not at Diamond) and retaining them only gives rise to technical debt (see e.g. #2211). In any case, this would be much better implemented using e.g. Flask or FastAPI.


phyy-nx:
Sounds good.  I'll rifle through the ol' rolodex and get back to you :)

phyy-nx:
Thanks @mhilgart!

graeme-winter:
@rjgildea has a replacement in the works for providing REST-type interfaces to dials for processing. Once that is available, maybe then we can close this one as there would be a good alternative to use.

PR #2225, opened 2022-09-21
dials.find_rotation_axis plot fix

dagewa:
#2160 inadvertently broke the output of a plot from `dials.find_rotation_axis`. This recovers that output.


dagewa:
I don't understand the test failures here. Doesn't happen locally and seems to be nothing to do with this PR :thinking:

PR #2229, opened 2022-09-27
Move dev.dials.napari_rlv here from dials_scratch

dagewa:


dagewa:
This is a bit more broken than I thought due to https://github.com/napari/napari/issues/5111. I'll convert to draft

dagewa:
The partial workaround for text positioning means this works well enough now for wider testing.

PR #2248, opened 2022-10-17
dials.reciprocal_lattice_viewer: set maximum number of experiment toggles

### LINKED ISSUE #1605: for large list of ids dials.reciprocal_lattice_viewer does not show view space
For more than maybe few 10 ids the list becomes too large for the view screen on the right to fit in the window. This is what it looks like for 100. Looking at 1k ids (not sure how common a use case this is) makes it impossible to see the view space.
rec_space_view

graeme-winter:
Limit the number to 10 by default; can override with `max_experiments=` on the command line, e.g. set to 10000 to restore current behaviour. 10 chosen as "reasonable default"

Fixes #1605


graeme-winter:
Chose 15 in retrospect as some standard runs on i19 have 13 and 15 should still be "sane"

jbeilstenedmands (approved):
I tried main & PR on a 70-sweep multi-crystal dataset, and without this fix the reciprocal lattices couldn't even be displayed as the list of ids was wider than my screen..... And now I can see the first $n$ again, so I think this is an improvement.

graeme-winter:
Request in team chat: please can we instead show a text box with "showing X lattices"

Issue #2262, opened 2022-10-26
dials.refine: please be more imaginative with interval_width_degrees


graeme-winter:
```
Grey-Area screen19-2 :) $ time dials.refine indexed.* interval_width_degrees=60
Sorry: Ambiguous parameter definition: interval_width_degrees = 60
Best matches:
  refinement.parameterisation.beam.smoother.interval_width_degrees
  refinement.parameterisation.crystal.unit_cell.smoother.interval_width_degrees
  refinement.parameterisation.crystal.orientation.smoother.interval_width_degrees
  refinement.parameterisation.detector.smoother.interval_width_degrees
  refinement.parameterisation.goniometer.smoother.interval_width_degrees
```

To me it seems fairly obvious that in default scan varying refinement I want to set 

```
  refinement.parameterisation.crystal.unit_cell.smoother.interval_width_degrees
  refinement.parameterisation.crystal.orientation.smoother.interval_width_degrees
```

since the others are fixed - however this then means I need to add ~ 100 characters to the command line output (which is admittedly explicit, so good) 

Any chance we could make the more obvious interpretation?

dagewa:
I think the way to do this would be to have another `interval_width_degrees` at the higher level, i.e. `parameterisation` scope. Then make that `expert_level=0` (because the others are all at `1`). That would then get picked automatically with your command line. The behaviour of that one would be to globally set all the others.

I'll do this - just making notes here so I don't forget what to do

PR #2268, opened 2022-11-04
Overall interval_width_degrees

dagewa:
Set an overall `interval_width_degrees` to adjust the defaults for all models.

```
$ dials.refine indexed.expt indexed.refl interval_width_degrees=60
Warning: Ambiguous parameter definition: interval_width_degrees = 60
Best matches:
  refinement.parameterisation.interval_width_degrees
  refinement.parameterisation.beam.smoother.interval_width_degrees
  refinement.parameterisation.crystal.unit_cell.smoother.interval_width_degrees
  refinement.parameterisation.crystal.orientation.smoother.interval_width_degrees
  refinement.parameterisation.detector.smoother.interval_width_degrees
  refinement.parameterisation.goniometer.smoother.interval_width_degrees
Assuming refinement.parameterisation.interval_width_degrees was intended.
```

Fix for #2262.

By the way, you can also try `interval_width_degrees=Auto`. No guarantee this gives sensible results in all cases, but it is most likely to work well for wide scans (full or multi-turn).


graeme-winter:
@dagewa thank you; noted; will review

graeme-winter:
```
Warning: Ambiguous parameter definition: interval_width_degrees = 18
Best matches:
  refinement.parameterisation.interval_width_degrees
  refinement.parameterisation.beam.smoother.interval_width_degrees
  refinement.parameterisation.crystal.unit_cell.smoother.interval_width_degrees
  refinement.parameterisation.crystal.orientation.smoother.interval_width_degrees
  refinement.parameterisation.detector.smoother.interval_width_degrees
  refinement.parameterisation.goniometer.smoother.interval_width_degrees
Assuming refinement.parameterisation.interval_width_degrees was intended.
```

verbose but accurate 🙂

I wonder though whether that would worry people?

newsfragments/2268.feature
```diff
@@ -0,0 +1 @@
+``dials.refine``: use an overall ``interval_width_degrees`` parameter to set defaults for all models simultaneously.
```
graeme-winter, line 1: Does this set it for _all_ models?

For example I would _not_ expect 

```
refinement.parameterisation.detector.smoother.interval_width_degrees
```

to be affected?

graeme-winter: As in: I had expected that this would affect all _crystal_ models

src/dials/algorithms/refinement/parameterisation/configure.py
```diff
@@ -80,6 +80,12 @@
       .type = bool
       .short_caption = "Scan-varying refinement"
 
+    interval_width_degrees = None
+        .help = "Overall default value of the width of scan between checkpoints"
+                "in degrees for scan-varying refinement. If set to None, each"
+                "model will use its own specified value."
+        .type = float(value_min=0.)
```
graeme-winter, line 87: Would we want to set a maximum value? 90° or something? Or is that an exercise for the user?

src/dials/algorithms/refinement/parameterisation/configure.py
```diff
@@ -790,6 +796,22 @@ def build_prediction_parameterisation(
         A prediction equation parameterisation object
     """
 
+    # Set overall default interval_width_degrees values, if requested
+    if options.interval_width_degrees:
+        options.beam.smoother.interval_width_degrees = options.interval_width_degrees
```
graeme-winter, line 801: As comment above: I had anticiated this _only_ applying to the crystal parameters. I would further mention right here that we cannot (IIUC) serialise scan varying beam, goniometer and detector parameters -> it is probably not correct to sweep them up in this?

dagewa:
That comes from cctbx. Not sure I can easily hide it as it is all in the option parser stuff

dagewa:
Alternative would be to have a different named parameter and then not use the PHIL disambiguation mechanism, but overall I thought that seemed like a worse interface.

graeme-winter (changes requested):
At first glance this does exactly what I had expected, so thank you. On more thorough investigation I think this does probably _more_ than we would want? By default I would only expect the crystal parameters to vary, as the others (I had understood) were assumed to be static however... maybe I had that wrong?

```
Grey-Area screen19-2 :) $ dials.refine -c -e2 | grep interval_width
    interval_width_degrees = None
        interval_width_degrees = 36.0
          interval_width_degrees = 36.0
          interval_width_degrees = 36.0
        interval_width_degrees = 36.0
        interval_width_degrees = 36.0
```

This looks a lot like they are all scan varying? Since we cannot save scan varying detector positions etc. this confuses me some. 

Anyway: requested change is to ensure that this default only applies to the crystal parameters... very happy to discuss this though.

dagewa:
What the new parameter does is to override the default values for all the smoothers, which are otherwise set to 36.0. The choice of whether to allow scan-varying parameterisation still remains with the models where each has a `force_static` parameter. For crystals these are default `False`, for other models they are `True`.

This all interacts with `scan_varying=(False/True/Auto)` which actually controls whether scan-varying refinement is done.

src/dials/algorithms/refinement/parameterisation/configure.py
dagewa: We can serialise scan-varying models other than crystals, but `dials.integrate` does not use them properly.

dagewa: actually, I _might_ have fixed that at least for scan-varying beams. IIRC it is still problematic because things like the mosaicity model ignore the scan-varying parameters. So, while prediction in integration is scan-varying, the overall model remains global

dagewa:
For small molecule data with P1 cell I sometimes do this:

```
dials.refine indexed.expt indexed.refl detector.fix=distance crystal.unit_cell.force_static=true
```

That way you get the usual static then scan-varying macrocycles, but the cell remains fixed to its static value and only the orientation refines scan-varying. This is because I found that the P1 scan-varying parameters often looked unreasonable (too much freedom) and I would not expect the cell to change anyway. Admittedly this is with ED data with narrow wedges and high centroid errors and would be less of an issue with good quality X-ray data

graeme-winter:
> What the new parameter does is to override the default values for all the smoothers, which are otherwise set to 36.0. The choice of whether to allow scan-varying parameterisation still remains with the models where each has a `force_static` parameter. For crystals these are default `False`, for other models they are `True`.
> 
> This all interacts with `scan_varying=(False/True/Auto)` which actually controls whether scan-varying refinement is done.

Ah, in which case the effective behaviour is precisely what I would have expected thank you. Would then simply ask if this could be explicit in the news fragment.

newsfragments/2268.feature
graeme-winter: With an additional sentence explaining the default behaviour here as discussed in the conversation for the PR looks great, thank you.

graeme-winter (approved):
With additional comment in news -> 👍 looks good thank you

PR #2271, opened 2022-11-09
Integration inequality

### LINKED ISSUE #2084: Hardcoded inequality inconsistent with error message
In https://github.com/dials/dials/blob/0cda7353e6c8a5facb5705399f3cfbabd83eee3a/algorithms/integration/integrator.py#L559-L563 there is a `<=0` check on a reflection column, but the error message suggests it should be simply a `<0` check. This can break certain reflection tables where this column is set with hypothetically 0 variance (e.g. overridden values).

dagewa:
Change `<=` to `<` as suggested by the existing error message, to fix #2084 

Waiting to see if tests fail on the CI.


dagewa:
Ok, tests didn't fail, but the macos pipeline was cancelled because of running too long

PR #2275, opened 2022-11-09
stills_process: reflection subsampling

phyy-nx:
Add new feature to dials.stills_process: reflection_subsampling. If reflection_subsampling.enable=True, and an image fails to index, the program will randomly subsample the reflections and try again. It keeps trying, randomly subsampling in smaller amounts until a threshold is reached or the image successfully indexes.

The default is descending from 100% to 50% of the strong spots by 2% steps, randomly subsampling from the original pool of strong reflections at each step.

We have observed nearly doubling the number of indexed images for some samples using this approach, and we verified that the additional lattices are high quality (ie we are not adding merely weak lattices that failed to index).

Co-authored-by: Daniel Paley <dwpaley@lbl.gov>
Co-authored-by: Asmit Bhowmick <abhowmick@lbl.gov>


src/dials/command_line/stills_process.py
```diff
@@ -1190,33 +1213,59 @@ def index(self, experiments, reflections):
...
+                        idxr.index()
+                    else:
+                        ml_indexing_error = None
+                        for method in params.indexing.stills.method_list:
+                            params.indexing.method = method
+                            try:
+                                idxr = Indexer.from_parameters(
+                                    reflections,
+                                    experiments,
+                                    known_crystal_models=known_crystal_models,
```
dwpaley, line 1249: Should this line be here? Indexing with known_crystal_models was already attempted.

src/dials/command_line/stills_process.py
```diff
@@ -1190,33 +1213,59 @@ def index(self, experiments, reflections):
...
+                                    params=params,
+                                )
+                                idxr.index()
+                            except Exception as e_ml:
+                                logger.info("Couldn't index using method %s", method)
+                                ml_indexing_error = e_ml
+                            else:
+                                ml_indexing_error = None
+                                break
+                        if ml_indexing_error is not None:
```
dwpaley, line 1259: For consistency with line 1267 below

```suggestion
                        if ml_indexing_error:
```

dwpaley (changes requested):
Looks good. My question on line 1249 is the only really important one. I can do the suggested test rewrite if you're amenable to the suggestion.

src/dials/command_line/stills_process.py
phyy-nx: Nope, that can go

dwpaley:
@phyy-nx see the edits as we discussed last week, everything look ok?

dwpaley (approved):

PR #2287, opened 2022-11-30
wx.Timer.Start() now expects an integer

rjgildea:
As of wxPython 4.2.0 passing a float results in an error:

  Timer.Start(): argument 1 has unexpected type 'float'


rjgildea:
The test failures don't look to be related to the changes

PR #2290, opened 2022-12-02
Fix check annotations CI warnings by including types

jbeilstenedmands:


jbeilstenedmands:
This was sniped by https://github.com/dials/dials/pull/2293 but I think it's still an improvement, so resurrecting

Issue #2291, opened 2022-12-03
Boost 1.78 requires c++14


phyy-nx:
Hi all, conda, as of yesterday or today, pulls boost 1.78 instead of 1.74.  1.78 seems to require c++14, which dxtbx fails to compile under.

I can reproduce it using the standard installation for developers:
```
wget https://raw.githubusercontent.com/dials/dials/main/installer/bootstrap.py
python bootstrap.py
```

Pinning boost to 1.74 solves the problem for now (see #2292), but it's not a long term solution.

graeme-winter:
Testing now, though this would explain build failures which started appearing yesterday

rjgildea:
It looks like Boost.Geometry requires C++14 from 1.75 onwards:
https://www.boost.org/doc/libs/1_80_0/libs/geometry/doc/html/geometry/release_notes.html

ndevenish:
> seems to require c++14, which dxtbx fails to compile under.

As a point to update this issue - the problem is not dxtbx being incompatible - dxtbx compiles perfectly fine under C++14 - (which it does happily under CMake) it's that the cctbx build system only supports `-std=c++11`, to the point that it will rewrite attempts to force the standard via environment variables.

So this should probably be closed here and reopened in https://github.com/cctbx/cctbx_project.

If there are further dials issues after that, then suggest we reopen once it's actually possible to test.

PR #2292, opened 2022-12-03
Pin boost to 1.74

phyy-nx:
conda currently picks 1.78, which requires c++14.  This is failing for dxtbx/dials.

See #2291


graeme-winter:
Testing myself, now

graeme-winter:
Worked for me: am tempted to merge-with-prejudice

newsfragments/2292.bugfix
```diff
@@ -0,0 +1 @@
+Pin boost to 1.74
```
graeme-winter, line 1: Add comment linking to https://github.com/boostorg/geometry/issues/590 for background?

graeme-winter (approved):
LGTM

Wonder if we should add link to upstream change but other than that worked for me

phyy-nx:
Thanks for the legwork @graeme-winter

PR #2298, opened 2022-12-06
Allow `reflection_table.remove_on_experiment_identifiers` when `len==0`

Baharis:
I encountered what _I believe_ is a bug in `flex` behavior, and since it stops my jobs, I decided to suggest a fix. This PR is tiny, but it is my first one here, so please let me know if I'm missing anything.

When filtering problematic dataset in `cctbx.xfel`, I encountered multiple instances of the following error:

    File "[...]/cctbx_project/xfel/merging/application/filter/experiment_filter.py", line 50, in remove_experiments
      reflections.remove_on_experiment_identifiers(experiment_ids_to_remove)
    File "[...]/dials/src/dials/array_family/flex_ext.py", line 1190, in remove_on_experiment_identifiers
      assert "id" in self

The dataset I'm processing has very few reflections, so the error most likely stems from the fact, that an empty `flex.reflection_table()` does not have key `id`. However, all methods utilized within `remove_on_experiment_identifiers()` can act on empty table. In my understanding, the `assert "id" in self` statement is supposed to catch tables which have reflections, but the reflections have no `id`.

This PR suggests to add a conditional statement before the assertion, so that the presence of `id` keyword is required only if the table has any reflections to begin with. As a result, the method should act as a null-operation for a zero-length table. Alternatively, the method could just `return self` if `len(self)==0`, but this would create two return points, which is generally undesirable.


jbeilstenedmands (approved):
Thanks!

PR #2300, opened 2022-12-12
bugfix: counting wrong thing

irisdyoung:
This was counting number of clusters instead of size of each cluster.


jbeilstenedmands (approved):
Thanks for spotting and fixing, looks like this was accidentally changed in [this PR](https://github.com/dials/dials/pull/2081) earlier this year

phyy-nx:
Ok to merge?

Issue #2216, opened 2022-08-30
3.11 release: reciprocal lattice viewer broken


graeme-winter:
`wxPython` error:

```
sleeper-service work :( $ sleeper-service work :) $ dials.reciprocal_lattice_viewer indexed.*
DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
The following parameters have been modified:

input {
  experiments = indexed.expt
  reflections = indexed.refl
}

Gtk-Message: 16:01:02.265: Failed to load module "atk-bridge"
Gtk-Message: 16:01:02.267: Failed to load module "canberra-gtk-module"
TypeError: Please report this error to dials-support@lists.sourceforge.net: Timer.Start(): argument 1 has unexpected type 'float'
Traceback (most recent call last):
  File "/home/graeme/x/dials-v3-11-0/build/../modules/dials/src/dials/command_line/reciprocal_lattice_viewer.py", line 79, in <module>
    run()
  File "/home/graeme/x/dials-v3-11-0/conda_base/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/graeme/x/dials-v3-11-0/build/../modules/dials/src/dials/command_line/reciprocal_lattice_viewer.py", line 65, in run
    f = ReciprocalLatticeViewer(
  File "/home/graeme/x/dials-v3-11-0/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 78, in __init__
    self.create_viewer_panel()
  File "/home/graeme/x/dials-v3-11-0/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 150, in create_viewer_panel
    self.viewer = RLVWindow(
  File "/home/graeme/x/dials-v3-11-0/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 487, in __init__
    super().__init__(*args, **kwds)
  File "/home/graeme/x/dials-v3-11-0/modules/dials/src/dials/util/wx_viewer.py", line 810, in __init__
    wxGLWindow.__init__(self, *args, **keyword_args)
  File "/home/graeme/x/dials-v3-11-0/modules/dials/src/dials/util/wx_viewer.py", line 148, in __init__
    self.spintimer.Start(1000 / self.spintimer_maxfps)
TypeError: Please report this error to dials-support@lists.sourceforge.net: Timer.Start(): argument 1 has unexpected type 'float'

```

[...]

PR #2306, opened 2022-12-16
Fix TypeError: these functions expect ints

rjgildea:
https://docs.wxpython.org/wx.DC.html#wx.DC.DrawCircle
https://docs.wxpython.org/wx.DC.html#wx.DC.DrawText


rjgildea:
Look related to https://github.com/dials/dials/issues/2216

jbeilstenedmands:
Note I also came across this error in a different place when trying to run the image viewer, so I've added a commit to fix, although I'm not sure which feature of the viewer this corresponds to.

PR #2310, opened 2023-01-03
Use standard C++11 features instead of boost equivalents

rjgildea:
We expect a C++11 compiler, so we may as well use standard C++11 features where available instead of legacy boost equivalents.

* Use ``std::atomic`` instead of ``boost::atomic``
* Use ``std::shared_ptr`` instead of ``boost::shared_ptr``
* Use ``std::make_shared`` instead of ``boost::make_shared``
* Use standard ``<random>`` instead of boost equivalent
* Use ``std::dynamic_pointer_cast``
* Remove some seemingly unused boost includes


rjgildea:
Tests failing because of https://github.com/cctbx/cctbx_project/pull/837

ndevenish:
While we have commits that we don't want to merge this... isn't ready to merge.

rjgildea:
It's mainly waiting now for debugging help from someone who knows a bit more about debugging Windows builds.

PR #2311, opened 2023-01-04
Prevent default `clustering.max_clusters = None` from raising error

Baharis:
By default `dials.combine_experiments` will raise `TypeError` if `clustering.max_clusters` is not set, as default `clustering.max_clusters = None` is compared with number of clusters, which is an integer. Since there are no checks for `max_clusters` being set and the parameter is not marked as mandatory, I assume this is not an expected behavior. As such, this PR suggests comparing with `max_clusters` only if it has been set.


ndevenish:
My rule of thumb is a User STacktrack is Always Bad (admittedly, this is more of an aspiration sometimes), so this is a good catch.

Also: Do you want to add yourself to the root AUTHORS file?

Baharis:
@ndevenish I have been thinking about it and came to conclusion that while this parameter might have been intended as "mandatory", it cannot be marked as one, because then it would be required even if the clustering were to be turned off! In this case it might be better to raise a dedicated error, but right now not setting `max_clusters` is equivalent to setting it to a very high value. I guess it is OK, although I haven't worked with this part of the code too much.

As for AUTHORS, I haven't really noticed this is a thing, and of course I'd be honored! Thank you!

ndevenish:
> `max_clusters` is equivalent to setting it to a very high value. I guess it is OK, although I haven't worked with this part of the code too much.

Yes, not setting is semantically equivalent to "Don't limit", I think that this is the least surprising result

ndevenish:
I also initially thought `n_clusters` could be removed, but it turns out that is used later (after the size has changed... which I think is a bug).

src/dials/command_line/combine_experiments.py
```diff
@@ -752,9 +752,9 @@ def save_in_batches(
                 threshold=params.clustering.threshold,
             )
             n_clusters = len(clustered)
-            clusters = sorted(clustered.clusters, key=len, reverse=True)[
-                : min(params.clustering.max_clusters, n_clusters)
-            ]
+            max_c = params.clustering.max_clusters
+            to_c = min(max_c, n_clusters) if max_c is not None else n_clusters
+            clusters = sorted(clustered.clusters, key=len, reverse=True)[:to_c]
```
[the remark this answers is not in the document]
Baharis: Absolutely, this looks so much better! I focused on solving the issue at hand, but in hindsight my solution is not very readable. However, I'd suggest strictly checking for `None`. Although it might make little sense for a human to `clustering.use=True` with `clustering.max_clusters=0`, I'd expect `clustering.max_clusters=0` not to produce any clustered data instead of acting as `None`. Might be easier way to effectively achieve `clustering.use=False` in some potential automated pipeline.

Baharis:
@ndevenish I absolutely agree with your more readable implementation of this fragment. However, I took the liberty to strictly look for None in said check. Otherwise, `clustering.max_clusters=0` acted as `None`, allowing as as many clusters as possible. I believe `clustering.max_clusters=0` should rather act similar as `clustering.use=False`.

Baharis:
@ndevenish As for `n_clusters`, it looks like a bug – I was fairly confused when my clusters were numbered 33 to 48 and not 0 to 15, but now I understand why. I am not sure why it is necessary in the first place, to effectively reverse the order of clusters and print them smallest to largest. Can I get rid of the reverse order there, @rjgildea, @irisdyoung?

irisdyoung:
The numbering is to be consistent with the naming of clusters as printed in the logs and propagated to cluster filenames -- I would actually much prefer they were ordered from largest to smallest as long as we keep this consistency.

Baharis:
@irisdyoung it looks like the naming of clusters as given in logs and filenames is defined based on the `exp_name` and `refl_name`, which are currently both defined together using `n_clusters - i_cluster` (lines 777–797), so removing the `n_clusters` should auto-update both log-line (`f"Saving combined experiments to {exp_name}"`; lines 821, 823) and file name (`experiments.as_file(exp_name)`; lines 822, 824). However, it still puzzles me why those clusters were defined in reverse to begin with though; maybe this is necessary somewhere - @rjgildea?

If there are no further suggestions to the solution of the issue raised here (Prevent default clustering.max_clusters = None from raising error), then I'll squash-merge this PR into main and open a new one to re-reverse the cluster order back to largest-to-smallest, with correct numbering.

ndevenish:
> However, I took the liberty to strictly look for None in said check. Otherwise, `clustering.max_clusters=0` acted as `None`, allowing as as many clusters as possible. I believe `clustering.max_clusters=0` should rather act similar as `clustering.use=False`

I guess the question is what we want `clustering.use=True max_clusters=0` to do. This way literally does do what you ask, which is writes out nothing (not even the non-clustered output), but it feels as though that should be an invalid state - I can't think of any situation in which wanting this would make sense?

In which case I've added an explicit check and error for this.

src/dials/command_line/combine_experiments.py
```diff
@@ -745,6 +745,8 @@ def save_in_batches(
 
         # cluster the resulting experiments if requested
         if params.clustering.use and len(experiments) > 1:
+            if params.clustering.max_clusters == 0:
+                sys.exit("Error: max_clusters must be None or >0")
```
Baharis, line 749: This seems reasonable. Is there any argument for using `sys.exit` instead of `ValueError`, which would give a clearer traceback? For example is there a chance the latter would be caught somewhere by mistake?

ndevenish: Yes! A stack trace here would be seen by the user. IMO a user should never be expected to have to figure out a stack trace (I say User STacktrace Always Bad), so we should aspire to avoid them in cases where it is specific feedback to the user about something that might be in their control. From the POV of the code it's an unrecoverable error (unless we want to guess what the user meant) and specifically _here_ we are in `dials.command_line.<something>.<function handling user invocation>` and not somewhere in e.g. the algorithms library (where a `ValueError` might be appropriate) so terminating is appropriate. This does admittedly get a little fuzzy from time-to-time as we don't have great separation between validating input and things just going plain wrong (and not a lot of Exception-level categorisation) but still I think something to aim towards.

In days of yore we used `Sorry` for this purpose of "We have an unrecoverable error and cannot fix and want to terminate with a specific message for the user" - however, we came to the conclusion that functionally this was identical to the `SystemExit` exception (including when and how we would want to catch) so try to use `sys.exit` directly nowadays.

I also noticed the other day that `dials.integrate` specifically does https://github.com/dials/dials/blob/995d6c2e2ffd4e43162961d2e309a9b0522e2bcc/src/dials/command_line/integrate.py#L728-L729

which should be applauded as the best kind of correct for being _technically_ correct in never showing the user a stack trace, but isn't really following the spirit of the aspiration (and is not very useful when a user does send us an actual problem report and we have to guess where it came from)

Baharis: Thank you, that was insightful. I learned that exceptions are generally preferable over system exit, but I can see why it might have its use in this case.

dwpaley: In principle I think the two choices `sys.exit()` and `raise Sorry` are supposed to be different when using mpi multiprocessing. This script `good.py`:

```
from mpi4py import MPI
from dials.util import Sorry

comm = MPI.COMM_WORLD
if comm.rank == 4: raise Sorry('rank 4 is unhappy')
MPI.COMM_WORLD.barrier()
print('ok')
```
should crash (as desired) when called with `mpirun -n 5 python good.py` whereas this `bad.py`:

```
from mpi4py import MPI
import sys

comm = MPI.COMM_WORLD
if comm.rank == 4: sys.exit('rank 4 is unhappy')
MPI.COMM_WORLD.barrier()
print('ok')
```
should confusingly hang forever.

In practice they currently both appear to hang forever due to faulty OpenMP signal handling as discussed here: https://github.com/chainer/chainermn/issues/236. A segfault on one rank is still handled correctly as in `actuallygood.py`:

```
from mpi4py import MPI

comm = MPI.COMM_WORLD
if comm.rank==4: import ctypes;ctypes.string_at(0)
MPI.COMM_WORLD.barrier()
print('ok')
```

...not that I'm advocating for showing segfaults to users. However maybe this is something to think about, I've definitely been bitten by this before when an MPI rank hits a `sys.exit`.

Baharis: @dwpaley This makes sense. Now that you mention that, I think that explains one of my bigger problemsI've encounter when processing current data. When calling `dials.refine` from cctbx's `ensemble_refinement` worker, if the number if ranks per node is too high, a `DivisionByZero` error is called due to an apparent lack of memory. If my memory does not fail me, this error is raised on multiple nodes and causes the whole process to hang instead of failing. This was especially painful to debug, as lack of memory incentified me to use more nodes, which just ate more NERSC resources without solving the issue.

phyy-nx: This is because dials.refine and dials.combine_experiments aren't "mpi safe".  Meaning since we invoke them using srun they are called by each mpi rank instead of by just 1 mpi rank.  That can be easily fixed by adding an mpi safe python script for ensemble refinement that would look something like this:

```
if rank == 0:
  call combine and refine
call mpi integrate
```

Something we should do before our next round of experiments.

dwpaley: Here's a little experiment to fix the faulty behavior I mentioned above: https://github.com/cctbx/cctbx_project/compare/master...mpi_excepthook If this is adopted then when running under MPI control it really will be better to use `Sorry` because it hits `sys.excepthook`.

ndevenish: MPI can't handle the underlying process just outright crashing?

phyy-nx: The behavior I've seen is that if a rank raises a traceback (or Sorry), then MPI shuts down all other ranks and the program exit.  If a rank calls sys.exit instead, that rank shuts down and the program indefinitely hangs while all the other ranks are still active.  Am I close @dwpaley?

phyy-nx: This may be MPI implementation dependent?  Like mpich vs. openmpi.

dwpaley: Tried that, but I don't think there's a mechanism for the registered function to know if we're exiting successfully or unsuccessfully. This seems better:

```
from mpi4py import MPI
import sys

def mpisafe(func):
  def wrapped_func(*args, **kwargs):
    try:
      return func(*args, **kwargs)
    except Exception:
      sys.excepthook(*sys.exc_info())
      MPI.COMM_WORLD.Abort(1)
    except SystemExit as e:
      if e.code:
        sys.stderr.write(e.code + '\n')
      MPI.COMM_WORLD.Abort(1)
  return wrapped_func

def do_work():
  comm = MPI.COMM_WORLD
  if comm.rank==4: sys.exit('something is wrong')
  return('ok')

@mpisafe
def run():
  print(do_work())
  MPI.COMM_WORLD.barrier()
  print('done')

if __name__ == '__main__':
  run()
```

and the decorator could be defined in `libtbx.mpi4py`.

ndevenish: Yep, catching to do cleanup generally seems the safest. Although if it's just cleanup (this presumably being cases where you actually do want a full abort) then
```python
try:
    call_function()
except BaseException:
    MPI.COMM_WORLD.Abort(1)
    raise
```
does the abort, but also propagates the unhandled exception, and doesn't swallow the SystemExit message and return code. I guess there also be a way to query if you closed the MPI session cleanly and to abort in an `atexit`. The documentation also does seems to distinguish between main and non-main threads, but that might only be if explicitly using thread support.

dwpaley: Unfortunately if you call Abort then you never hit raise. I don't immediately see a way to do both. Hence manually calling excepthook or printing the exit message.

That means this approach isn't compatible with propagating errors up to the caller (since we have to choose either raise or Abort). So the name `@mpisafe` is not really appropriate, it's more like `@mpi_abort_on_exception` to clarify this can only be applied in a place where exception means failure.

PR #2314, opened 2023-01-12
Add R-split (optionally) to reported merging stats

jbeilstenedmands:
The R-split metric is a quantity often used in serial crystallography. It is defined as the R-factor between merged half sets of intensities.
For `dials.scale` and `dials.merge`, if `additional_stats=True`, R-split is added to the reported merging stats "xia2-style" summary, and the resolution dependent values are reported in the `dials.merge.log` and plotted in the `dials.merge.html`.


src/dials/algorithms/merging/reporting.py
```diff
@@ -177,6 +177,26 @@ def make_dano_table(anomalous_amplitudes):
...
+    for i, i_bin in enumerate(list(stats.binner.range_used())):
+        d_max_bin, d_min_bin = stats.binner.bin_d_range(i_bin)
+        rows[i].append(f"{d_max_bin:.3f} - {d_min_bin:.3f}")
+    rows[-1].append("Overall")
+    header.append("r-split")
+    for (i, rsplit) in enumerate(stats.r_split_binned):
+        rows[i].append(f"{rsplit:.5f}" if rsplit is not None else "None")
+    rows[-1].append(f"{stats.r_split:.5f}")
+    output_ = "\n" + tabulate(rows, header)
+    return output_
```
rjgildea, line 197: Is there any particular reason for the `_` suffix in `output_` here?

src/dials/report/analysis.py
```diff
@@ -285,6 +286,16 @@ def table_1_stats(
...
+    extra_key_to_var = {}
+    if merging_statistics.r_split:
+        extra_key_to_var.update(
+            {
+                "Rsplit": {
+                    "overall": "r_split",
+                    "binned": "r_split_binned",
+                }
+            }
+        )
```
rjgildea, line 298: Am I missing something here why we need a separate key-var dictionary for the additional Rsplit statistic?

src/dials/command_line/merge.py
```diff
@@ -129,6 +129,8 @@
         .type = strings
         .help = "Dataset name to be used in MTZ file output (multiple names
             allowed for MAD datasets)"
+    additional_stats = False
+       .type = bool
```
rjgildea, line 133: Some help to explain what additional stats will be shown may be ... helpful? :)

src/dials/command_line/scale.py
```diff
@@ -97,6 +97,10 @@
       .type = int
       .help = "Number of bins to use for calculating and plotting merging stats."
       .expert_level = 1
+    additional_stats = False
+      .type = bool
+      .help = "Show additional merging statistics"
```
rjgildea, line 102: As above, it would be nice for the user to know _what_ additional statistics will be shown.

rjgildea (commented):
Thanks for this, it will be a useful addition (and the ability to relatively easily include further statistics). I have just tested the PR running `dials.merge addititional_stats=True` and have the following observations:
* I see Rsplit reported in the "Summary of merging statistics" table in the log file ✅ 
* I see a plot of Rsplit vs resolution in the html output ✅ 
* I was surprised that Rsplit wasn't included in the "Merging statistics by resolution bin" table in the log file - how hard would it be to add it here also?
* Similarly, it is in neither of the tables of statistics reported in the html file - how hard to include here also?

src/dials/report/analysis.py
jbeilstenedmands: Yes this is needed as the other dictionary is looped over when inspecting the attributes of the merging stats bin objects, but the Rsplit is stored separately so needs a different lookup mechanism.

jbeilstenedmands:
Thanks for the feedback. I've tried to add it to the log stats table and yes it's possible, just working on some formatting issues now.

jbeilstenedmands:
@rjgildea  The latest commits should address all the points raised in your first review

rjgildea:
Thanks for adding it to the log/html stats table. I've just tested with the latest state of the branch and it is now appearing in the "Merging statistics by resolution bin" log table and the "Merging statistics / Resolution shells" html table. However it seems to have disappeared from the "Summary of merging statistics" log table (and not in the "Merging statistics / Overall" html table either), is this to be expected?

jbeilstenedmands:
> However it seems to have disappeared from the "Summary of merging statistics" log table (and not in the "Merging statistics / Overall" html table either), is this to be expected?

Thanks for catching this, I had updated one dict key to Rsplit(I) but missed the other.

I've now added Rsplit to the merging statistics/overall html table as well.

src/dials/algorithms/scaling/scaling_library.py
```diff
@@ -372,8 +374,103 @@ def determine_best_unit_cell(experiments):
...
+            assert other.indices().size() == this.indices().size()
+            if this.data().size() == 0:
+                return 0.0
+
+            if assume_index_matching:
+                (o, c) = (this, other)
+            else:
+                (o, c) = this.common_sets(other=other, assert_no_singles=True)
+
+            den = flex.sum(o.data() + c.data())
```
rjgildea, line 448: Should the denominator have an extra factor 0.5 in? From equation (2) in [White et al. 2012](https://journals.iucr.org/j/issues/2012/02/00/db5097/index.html):
Screenshot 2023-01-17 at 16 05 21
Perhaps I'm missing something, but I don't see where the factor 0.5 is in your implementation...

src/dials/command_line/merge.py
```diff
@@ -129,6 +129,9 @@
...
+       .help = "Calculate and report the R-split statistic in the log"
```
rjgildea, line 134:
```suggestion
       .help = "Calculate and report the R-split statistic"
```
It's also in the html report not just the log.

src/dials/algorithms/scaling/scaling_library.py
```diff
@@ -372,8 +374,103 @@ def determine_best_unit_cell(experiments):
...
+        if not i_obs:
+            return
+        i_obs_copy = i_obs.customized_copy()
+        i_obs_copy.setup_binner(n_bins=n_bins)
+        i_obs = i_obs.map_to_asu()
+        i_obs = i_obs.sort("packed_indices")
+
+        seed = 0
+
+        split_datasets = split_unmerged(
```
rjgildea, line 399: As I understand it this is randomly assigning the observations to each half dataset (as per CC1/2 calculation) rather than randomly assigning whole images to each half dataset (as per the [published definition](https://journals.iucr.org/j/issues/2012/02/00/db5097/index.html)). I guess in practice it probably doesn't make much difference, but it's probably worth explicitly pointing out where we differ from the published definition, at least by a comment in the code.

src/dials/algorithms/scaling/scaling_library.py
jbeilstenedmands: This is confusing but I believe I'm correct: 2^(-0.5) / (1/2) = (1 / sqrt2) * 2 = sqrt2. So I just put 2^(+0.5) in the numerator and therefore don't need any factor in the denominator.

src/dials/algorithms/scaling/scaling_library.py
jbeilstenedmands: Thanks, I hadn't appreciated that the published definition is based on whole image splitting. I'm happy that using the random observation assignment, like in cc1/2, is a good way for measuring internal dataset consistency, but I will add a comment.

rjgildea (approved):

PR #2317, opened 2023-01-17
Kapton error handling

Baharis:
Previous implementation of `dials/command_line/stills_process.py Processor.integrate` and `dials/command_line/integrate.py run_integration` had an unexpected `UnboundLocalError` whenever `absorption_correction.apply=True`, but `absorption_correction.algorithm` was unspecified (former specifier 2/4 possible states, and later 1/4 possible state). Now, in both places whenever `absorption_correction.algorithm` is expected but unspecified, a `ValueError` is raised, unless user specifically requests `absorption_correction.algorithm=other`.'

This change should not affect any existing processes, but might save a few hours some other poor soul, who will inevitably  accidentally declare multiple absorption protocols and unexpectedly ends up with `{apply=True, algorithm=None}`, like me last week.

### DIFF
--- a/src/dials/command_line/integrate.py
+++ b/src/dials/command_line/integrate.py
@@ -609,19 +609,23 @@ def run_integration(params, experiments, reference=None):
     for abs_params in params.absorption_correction:
         if abs_params.apply:
             if abs_params.algorithm == "fuller_kapton":
-                from dials.algorithms.integration.kapton_correction \
-                    import multi_kapton_correction
+                from dials.algorithms.integration.kapton_correction import (
+                    multi_kapton_correction,
+                )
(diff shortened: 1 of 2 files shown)


ndevenish:
Do other parts of the code use "other" anywhere, or is it just worth removing "other" from the list of possibilities?

I also wonder if `apply=` could be ignored needing to specify it separately; the presence of `algorithm=` might be enough to know you want to turn it on?

ndevenish:
Okay, normally I'd push for the "other" handling to be removed from here (and the PHIL option) since it does nothing (anyone extending it would need to know that it went here anyway), but as a general policy don't touch anything in `dials.stills_process` - since it's hard to know if anything relies on it being set but doing nothing.

PR #2319, opened 2023-01-23
Use C++14, unpin boost, and update prebuilt CCTBX base

ndevenish:
Now that we can enable C++14 in the libtbx environment, enable this. This lets us unpin boost so that we can use boost::geometry in boost 1.78 (which only supports C++14).

Prebuilt pointer is updated here but currently doesn't work; The distributed libtbx environment is built against boost 1.78 but fixed to C++11, and it doesn't seem immediately easy to override. I'm not current sure how extensively we are actually using that mode now though; I think in most cases we now use CMake (which always supported C++14).


phyy-nx:
I'd say this closes #2291

ndevenish:
Cancelled because this won't work until https://github.com/cctbx/dxtbx/pull/605 is in

Issue #643, opened 2018-10-31
Additional tests needed in test_combine_experiments


asmit3:
test_combine_experiments.py needs to have tests for

- [x] n_subset_method
- [x]  n_refl_panel_list
- [x]  panel_filter

PR #2323, opened 2023-01-27
Update `dials.combine_experiments`

### LINKED ISSUE #1658: dials.combine_experiments requires reflections
At some point in the past, I remember `dials.combine_experiments` being able to work without providing reflections. This is no longer the case (or I imagined it). I think it should be possible to run without providing reflections as it is useful for shuttling models about between experiments. Alternatively, a separate `dials.copy_model` (or something) might be useful.

### LINKED ISSUE #1741: dials.combine_experiments should work on imported data
i.e. if I have imported two sweeps just how I want them I should be able to combine these without first finding spots etc. 

I cannot. Because no reflections.

jbeilstenedmands:
Start of refactoring work to make `dials.combine_experiments` into a command line interface + library code.

I have updated the code to handle the case of no input reflections (fixes #1741, fixes #1658). More tests will also be added here (to address e.g. #643).

Draft for now for publicity, I haven't changed any `sys.exit` calls yet, also more type hints needed and perhaps the phil scope should be moved to `dials.util.combine_experiments`. Perhaps some of the `sys.exit` calls should be `RunTimeErrors` but I haven't thought about this much yet.


src/dials/command_line/combine_experiments.py
```diff
@@ -187,649 +187,215 @@
...
+        if reflections:
+            output_reflections_list = [
+                reflections.select_on_experiment_identifiers(cluster.lattice_ids)
+                for cluster in clusters
+            ]
+            for table in output_reflections_list:
+                table.reset_ids()
+        for i_cluster, _ in enumerate(output_experiments_list):
+            exp_name = os.path.splitext(experiments_filename)[0] + (
+                "_cluster%d.expt" % (n_clusters - i_cluster)
```
Baharis, line 240: We had a discussion in #2311 about this line, line 219 and line 245. As of now, the clusters are numbered from smallest to largest, with the first few usually missing because they are rejected in lines 223–224. This appears to be a bug or, at least, an unexpected behavior. I started #2318 to tackle this issue (and to increase some readability), but this will be heavily superseded by this PR, so I'll pull out. Nevertheless, I still wonder whether this cluster indexing procedure could be changed to index clusters according to decreasing size and starting from 0.

jbeilstenedmands: Hi @Baharis , my intention here was not to override your work in #2318 in any way. I think it's work keeping the cluster order renaming as a separate piece of work for clarity, then I can just rebase these reorganisationalchanges on any work you do there.

Baharis: Alright, in this case I will re-test my PR and see if it can be quickly merged. However, if there are any problems or delays with my PR, feel free to merge first and I'll return to the issue of ordering whenever I'll find some time.

PR #2329, opened 2023-01-29
Fix masking failure case when direct beam is exactly in the centre of a pixel

dagewa:
In that case, take the pixel position from the edge of the pixel rather than the centre.


dagewa:
@huwjenkins just wondering if this fix works for you?

Issue #2330, opened 2023-01-30
dials.scale absorption=medium shows a stack trace


graeme-winter:
Probably to be avoided on the User Should Not See A Stack Trace principle:

```
ws448 work :) $ dials.scale symmetrized.* absorption=medium
Traceback (most recent call last):
  File "/dls_sw/apps/dials/dials-v3-12-1-2/build/../modules/dials/src/dials/command_line/scale.py", line 262, in <module>
    run()
  File "/dls_sw/apps/dials/dials-v3-12-1-2/conda_base/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/build/../modules/dials/src/dials/command_line/scale.py", line 217, in run
    params, options = parser.parse_args(args=args, show_diff_phil=False)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/dials/src/dials/util/options.py", line 867, in parse_args
    params, args = self._phil_parser.parse_args(
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/dials/src/dials/util/options.py", line 502, in parse_args
    params = self._phil.extract()
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 1782, in extract
    value = object.extract(parent=result)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 1782, in extract
    value = object.extract(parent=result)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 1782, in extract
    value = object.extract(parent=result)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 1145, in extract
    return self._type_from_words()(self.words, master=self)
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 192, in from_words
    return bool_from_words(words=words, path=master.full_path())
  File "/dls_sw/apps/dials/dials-v3-12-1-2/modules/cctbx_project/libtbx/phil/__init__.py", line 181, in bool_from_words
    raise RuntimeError(
[...]

PR #2331, opened 2023-01-30
Call parser.error() on phil exception

rjgildea:
If calling `phil.extract()` raises an exception (for example because the provided value is of the wrong type or out of range) then catch the error and call `parser.error()` which results in a call to `sys.exit()`.

Fixes https://github.com/dials/dials/issues/2330


rjgildea:
For the original error report in question (https://github.com/dials/dials/issues/2330) the resulting error message is now:
```
$ dials.scale absorption=medium
usage: Usage: dials.scale integrated.refl integrated.expt
[integrated.refl(2) integrated.expt(2) ....] [options]
scale.py: error: Invalid phil parameter: One True or False value expected, physical.share.absorption="medium" found (command line argument, line 1)
```

jbeilstenedmands (approved):
Thanks!

PR #2342, opened 2023-02-13
Handle blank ids

phyy-nx:
If any identifiers are of length zero, skip the sorting step. This makes dials.assign_experiment_identifiers work with an experiment list without identifiers.


jbeilstenedmands:
Hi, I don't see a problem with this but just trying to understand the use case, as `dials.assign_experiment_identifiers` does work on experiment lists where identifiers are not set, this is tested for in the first example in `tests/command_line/test_assign_experiment_identifiers.py::test_assign_identifiers`.
Is this issue triggered if you are using the code as a library?

src/dials/util/multi_dataset_handling.py
```diff
@@ -76,7 +76,8 @@ def sort_tables_to_experiments_order(reflection_tables, experiments):
     exp_id_to_table_idx = {}
     for i, table in enumerate(reflection_tables):
         id_values = table.experiment_identifiers().values()
-        if id_values:
+        any_blanks = any([len(id_value) == 0 for id_value in id_values])
```
graeme-winter, line 79: Suggest `all()` not `any()` because if you have _some_ id's this may not behave as you expect?

newsfragments/2342.bugfix
```diff
@@ -0,0 +1 @@
+Bugfix for dials.assign_experiment_identifiers to work with an experiment list without identifiers
```
graeme-winter, line 1: Suggest that this is a new feature not a bug fix?

graeme-winter (commented):
Makes sense - suggest you probably want `all` not `any` 

This is a comment as I don't know if we want to approve without test etc.

phyy-nx:
Hi, I was using the command line version.  The use case is that cctbx.small_cell doesn't assign experiment identifiers (it probably should), but it's a bit weirder than that.  Due to some micromanagement of identifiers in dials.stills_process, the output reflection tables have an experiment identifier dictionary that has N keys 0..N all matched to a blank string ('').  That causes [this comparison](https://github.com/dials/dials/blob/main/src/dials/util/multi_dataset_handling.py#L86) to return True, which causes a fail a few lines later.

The test in test_assign_experiment_identifiers.py loads a reflection table with an empty experiment identifier dictionary, which triggers [that same comparison](https://github.com/dials/dials/blob/main/src/dials/util/multi_dataset_handling.py#L86) to equal False.

I've added a test to test_assign_experiment_identifiers.py to cover this (admittedly weird but technically valid) use case.  Medium term i think dials.stills_process and cctbx.small_cell should both assign identifiers by default.  Also, the micromanagement of ids in dials.stills_process could be replaced to calls to either `reset_ids` or `concat`.

src/dials/util/multi_dataset_handling.py
phyy-nx: Thanks!

jbeilstenedmands:
Hi @phyy-nx 
Thanks for the explanation for clearing this up and for adding a test.

Your suggestion for the medium term plan sounds good, if there is any functionality missing from `reset_ids` and `concat` then we can look at improving those.

PR #2345, opened 2023-02-21
When importing with a reference, only require 1 beam model if use_beam_reference=True

jbeilstenedmands:
When importing with a reference for stills data, if we don't want to use the beam reference then we should not need to assert that the beam models have been combined.


rjgildea (approved):

jbeilstenedmands:
I made an extra change to rearrange the order of assertions. This means that you can in theory use a reference experiment with solely a detector model (if `use_beam_reference=False` and `use_gonio_reference=False`).

PR #2348, opened 2023-02-24
Use experiments instead of datablocks in tests

dagewa:
The last remaining `DataBlock` files used in tests are replaced by `ExperimentList` versions, created by a script like in https://github.com/cctbx/dxtbx/pull/570#issuecomment-1442162602.

This fixes the remaining test failures when https://github.com/cctbx/dxtbx/pull/570 is merged.


dagewa:
Given the policy of static `dials_regression` at DLS, this won't fly

https://github.com/cctbx/dxtbx/pull/570#issuecomment-1443575303

PR #2352, opened 2023-02-27
Use 3 macrocycles of refinement by default in dials.ssx_integrate

jbeilstenedmands:
At least a couple more macrocycles should be the default, so that the improved predictions from the first round of refinement can be used when calculating the reflection properties for subsequent refinement.

This makes the program have a higher runtime, but I have some WIP performance improvements that more than compensate.


newsfragments/2352.misc
```diff
@@ -0,0 +1 @@
+``dials.ssx_integrate``: Update default to 3 macrocycles of refinement.
```
rjgildea, line 1:
```suggestion
``dials.ssx_integrate``: Update default to 3 macrocycles of profile model refinement.
```

PR #2358, opened 2023-03-03
Platform-dependent path splitting

dagewa:
Split with the correct platform-dependent path separator. Should fix `test_ssx_index.py` on Windows.


dagewa:
Confirmed, with this change the test passes on Windows.

PR #2360, opened 2023-03-06
`dials.export_bitmaps` with resolution rings

dagewa:
This adds optional resolution rings to `dials.export_bitmaps`. Additional PHIL controls are:

```
show_resolution_rings = False
  .type = bool
resolution_rings {
  number = 5
    .type = int(value_min=1)
  fontsize = 30
    .type = int
}
```

Example with a P6M image:
```
dials.export_bitmaps grid_full_cbf_0005.cbf show_resolution_rings=True
```
image0001

This also works for multi-panel and non-coplanar detectors. For I23:
```
dials.export_bitmaps germ_13KeV_0001.cbf show_resolution_rings=True resolution_rings.number=10
```
image0001


dagewa:
I see there are some artefacts at the image edges for the single panel case

Baharis:
This might be more of an esthetic choice than anything, but what would you say about putting all phil parameters in one scope, i.e.:

```
resolution_rings {
  show = False
    .type = bool
  number = 5
    .type = int(value_min=1)
  fontsize = 30
    .type = int
}
```

dagewa:
Yeah, I'm inclined to agree that's better :+1:

dagewa:
Ah, I know why I did that. It's because `dials.image_viewer` has the parameter `show_resolution_rings` so I was going for consistency. But hardly anyone uses the PHIL parameters for `dials.image_viewer`, so that's not a very strong reason.

PR #2362, opened 2023-03-07
Multi-panel `dials.rs_mapper`

### LINKED ISSUE #2333: dials.rs_mapper does not work for multi-panel detectors
This was raised by I23 beamline scientists

```
$ dials.rs_mapper AUTOMATIC_DEFAULT_SAD_SWEEP1.expt

input {
  experiments = AUTOMATIC_DEFAULT_SAD_SWEEP1.expt
}

Setting nproc=4
Sorry: This program does not support multi-panel detectors.
```

This will also be relevant for the plan to use `rs_mapper` for the neutron diffraction data processing GUI.

dagewa:
Allow `dials.rs_mapper` to work for multi-panel detectors by looping over panels at the highest level and accumulating the results. So far I tested only for a very simple 4 panel, small molecule MicroED dataset and the results appear sensible.

image

This should close #2333.

### DIFF
--- a/newsfragments/xxx.feature
+++ b/newsfragments/xxx.feature
@@ -1,0 +1,1 @@
+Allow ``dials.rs_mapper`` to run for multi-panel detectors.
--- a/src/dials/command_line/rs_mapper.py
+++ b/src/dials/command_line/rs_mapper.py
@@ -77,7 +77,7 @@ output
 
 
 def process_block(
-    block, imageset, grid_size, reverse_phi, S, ignore_mask, xy, rec_range
+    block, imageset, i_panel, grid_size, reverse_phi, S, ignore_mask, xy, rec_range
 ):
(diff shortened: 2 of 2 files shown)


biochem-fan (approved):
Thank you very much for enhancement. The changes look OK to me.

biochem-fan:
> I think this restriction is artificial. Whether or not a multi-experiment calculation is meaningful depends on context, but we can leave that up to the user to decide.

How about adding an option to rotate the experiment according to the orientation matrix, when given
`indexed.expt`, not `imported.expt`? This might be useful when pathology (e.g. streaking) is suspected
in a serial or small-wedge crystallography dataset.

(Just an idea for future enhancement, not a requirement for this pull request.)

dagewa:
Thanks, that sounds useful. In fact, perhaps that should be default if multiple experiments are provided that have crystal models. I might still put it in here - I can't merge this yet because we're getting failures with I23 data and it isn't clear why.

PR #2363, opened 2023-03-08
Replace `imp` by `importlib`

dagewa:
[`imp`](https://docs.python.org/3/library/imp.html) is deprecated and will be removed in Python 3.12.

This is the suggested replacement using `importlib` from https://github.com/pyinvoke/invoke/issues/214#issuecomment-77428507. I don't know if we actually use this plugin mechanism at all though, so I have not been able to test this.


dagewa:
Plugins have filenames like `*_frame_plugin.py` apparently. @phyy-nx @dwpaley is this perhaps one of those? https://github.com/cctbx/cctbx_project/blob/master/xfel/util/kapton_2019_correction_frame_frame_plugin.py

phyy-nx:
> Plugins have filenames like `*_frame_plugin.py` apparently. @phyy-nx @dwpaley is this perhaps one of those? https://github.com/cctbx/cctbx_project/blob/master/xfel/util/kapton_2019_correction_frame_frame_plugin.py

Yup!

phyy-nx:
Works great, thanks

PR #2364, opened 2023-03-09
Calculate image size limits correctly for data access in `dials.rs_mapper`

dagewa:
This was spotted during testing of #2362 with I23 data. The program would segfault, ~~which was eventually tracked down to the indices used to access image data being reversed from expected given their accessor~~.

This was only noticed now, because I23 panels are highly elongated, with size (195, 2463). For previous uses with squarish single panel images it is likely that the pixels within the specified resolution limit (default 6 Å) had indices that were valid for either x or y.

The resultant output is clearly quite different. Here is a comparison of the resultant maps in Coot for the dials-data insulin dataset (`dials.data get insulin`), with the map from this branch in purple and that from main in green

image

It is not merely a case of transposing one to the other, because the pixel is also rotated to a particular position around an axis with fixed orientation wrt to the image.

This may have consequences for `dials.export format=json`.

I think this might need more testing to understand the issue before it is merged.


dagewa:
The latest change makes maps that align very closely. I need to try this on the multi-panel branch to see if it works for I23
image

src/dials/command_line/rs_mapper.py
```diff
@@ -188,12 +188,13 @@ def process_imageset(self, imageset):
         beam = imageset.get_beam()
         s0 = beam.get_s0()
         pixel_size = panel.get_pixel_size()
-        xlim, ylim = imageset.get_raw_data(0)[0].all()
+        nfast, nslow = panel.get_image_size()
+
```
dagewa, line 192: To avoid any further confusion, this is the critical change. The rest is just cosmetic.

PR #2366, opened 2023-03-10
Suppress Gtk terminal junk

dagewa:
You know those annoying messages like this from the viewers?
```
Gtk-Message: 14:31:36.498: Failed to load module "atk-bridge"
Gtk-Message: 14:31:36.499: Failed to load module "canberra-gtk-module"
```

This change makes them go away. At least, it does for me.


Baharis:
The "At least, it does for me" part intrigued me, so I checked out on this branch on my Perlmutter installation — which is otherwise a few weeks out of date. It seems the "canberra-gtk-module" warning is still there. Does it work out of the box for you, or should I do anything else?

dagewa:
Apparently it only works with glib 2.50 or later, so maybe that's the issue?

dagewa:
Well, that's what it says at https://docs.wxwidgets.org/trunk/classwx_app.html#a54432e69998e785a22838699ef8bf8f9, but that seems like nonsense as the current latest version is 2.37 (https://www.gnu.org/software/libc/). Maybe they mean 2.5.0. Anyway, according to `ldd --version`, I have 2.27.

dagewa:
Oops, wrong library. glib version, not glibc version 🙄

dagewa:
I just tried at Diamond and happily this does work there.

dagewa:
@bkpoon I noticed this affects `cctbx.multiplicity_viewer` too. Possibly you want something like this in `wxtbx.app.CCTBXApp`?

bkpoon:
Yeah, we can add it. Should it be in the constructor for the `CCTBXApp` class? We can add a keyword argument like `suppress_gtk_diagnostics` that defaults to `True` and runs that try/except block. And if someone needs to debug, they can   set that keyword to `False` in their code.

dagewa:
Yeah, as it seems to work when called ahead of the constructor, then putting it first thing in the `__init__` makes sense. Apparently you can mask specific message types, but I didn't look into that. I think making it a binary on/off under keyword control sounds much simpler :+1:

bkpoon:
I created the `gtk-error` branch in `cctbx_project`. @dagewa Can you test it? My default linux server is CentOS and it seems to have all the expected modules, so I don't get those error messages.

We can tweak the behavior. The keyword is `gtk_suppress_diagnostics` to match the function name, but we can change that as well.

And I agree that the binary choice is simplest.

dagewa:
Yep, works perfectly for me, thanks!

bkpoon:
I just committed the change, so it should be generally available now.

dagewa:
Thanks! I updated the `CCTBXApp`s in DIALS to use the new mechanism.

PR #2367, opened 2023-03-11
Increase robustness of constrained refinement test

dagewa:
This test occasionally fails in the CI. The tolerance was increased in #2343, but there are still some failures in PR test runs. This change switches off outlier rejection (which is irrelevant for what is being tested) in a further effort to ensure the test is deterministic.


dagewa:
A further, more fundamental, fix is to fix the `Tau2` and `Tau3` ("tilt" and "twist") parameters during refinement, because a constraint in the `Dist` parameter need not actually result in a constrained distance along the detector normal if that detector was also allowed to rotate out of plane.

PR #2371, opened 2023-03-15
Keep the experiment identifiers map when slicing a reflection table.

### LINKED ISSUE #2276: Slicing a reflection table drops the identifier map
Found while working on #1858, in `cctbx.xfel.merge` we have a reflection table [slice operation](https://github.com/cctbx/cctbx_project/blob/master/xfel/merging/application/reflection_table_utils.py#L23), but it turns out that drops the experiment identifier map.

Simple reproducer:
```
print(list(refls.experiment_identifiers().keys()))
refls = refls[0:10]
print(list(refls.experiment_identifiers().keys()))
```
Output
```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]
[]
```

I think this means that the slice code in [flex.reflection_table](https://github.com/dials/dials/blob/main/src/dials/array_family/boost_python/flex_table_suite.h#L455-L471) needs to include a pass through the experiment map.  It would like like this in python:

```
subset = refls[start:stop] # drops identifier map
for expt_id in set(subset['id']):
  subset.experiment_identifiers()[expt_id] = refls.experiment_identifiers()[expt_id]
```

jbeilstenedmands:
Fixes #2276


phyy-nx (approved):
Thanks!

PR #2372, opened 2023-03-15
Improve dials.ssx_index when unit cell values not known

jbeilstenedmands:
Currently, when indexing without a given unit cell, the max cell estimation across all images is used. This tends to be very high (~700A-1000A), and results in an overrepresentation of indexed cells with very large cell values in the final reported cell clustering. 

For example, on a dtpb dataset:
With input cell of a=85,b=120,c=197 : 69 images indexed, 71 crystals in main cluster
With no input cell, the max cell is set to 774A: 75 images indexed, 25 crystals in main cluster

By varying the max cell set, the trend is clear that overly high max cells reduce the number of correct cells found:
Max cell 200: 56 images indexed, main cluster 38
Max cell 300: 70 images indexed, main cluster 42
Max cell 400: 71 images indexed, main cluster 34
Max cell 500: 76 images indexed, main cluster 32
Max cell 800: 76 images indexed, main cluster 26
Max cell 1000: 74 images indexed, main cluster 26

This PR changes this to use the 95th centile value from the distribution of max unit cells.
**Note that this change only affects the use case when the cell values are not known, typically the cell values are known.**

In the example, this gives a max cell of 360, 73 images indexed, 39 in main cluster.

This approach was chosen over a Tukey approach, as for a typical low indexing rate e.g. 5-10%, most images do not contain many spots - so in general, the max cell distribution is not reliable for estimating max cells or the indexable images; the majority of the distribution can fall below the highest cell value.
This can be seen by plotting the max cell distributions:
image

Similar effects were seen in other test datasets.
The threshold parameter in a clustering function is also exposed, to enable a related change in xia2.


rjgildea:
I wonder how it would compare doing the nearest neighbour analysis across all images at the same time, rather than many independent max_cell estimations. I _think_ if you pass the entire reflection list to `find_max_cell` it will split up reflections by experiment for calculating the d-spacing of the nearest neighours before doing the max cell analysis on the combined list of d-spacings:
https://github.com/dials/dials/blob/085adbf96e2b3647d2a90e2372c909550c6c6ed7/src/dials/algorithms/indexing/nearest_neighbor.py#L37-L132

Issue #2161, opened 2022-07-06
What is our coordinate system, exactly?


graeme-winter:
In discussion with @benjaminhwilliams it turns out that what we _do_ deviates in detail from what we _say_, which may be very on message for 2022 but means our documentation at https://dials.github.io/documentation/conventions.html is misleading, and in particular:

"In the interests of standardisation, we choose to adopt the Image CIF (imgCIF) reference frame"

misrepresents the leaky abstraction that at times we adopt a different system if the input data state a vertical goniometer. In particular, we take the input system from imgCIF verbatim, but from others we rotate the coordinate systems to be imgCIF like (e.g. from NeXus, where we rotate by 180° about "up")

We should probably be (i) explicit and (ii) truthful about this

benjaminhwilliams:
The truth is closer to:
1. Take McStas (beam direction is _z_, _y_ is vertically up, _x_ follows).
2. Transform to a coordinate system that is related to McStas by a 180° rotation around _y_.
3. Wherever the rotation axis is, take that direction to be the cardinal direction, i.e. the one that is not refined.

This is different to McStas not just because of the 180° rotation, but also because in McStas, the cardinal direction is _z_, not the rotation axis.

It is also different to imgCIF because, although imgCIF has the same rule that the principal rotation axis is cardinal, the _x_ direction in imgCIF is defined to be the principal rotation axis.  This is not what we do, though for many synchrotron beamlines, the rotation axis points horizontally left, as seen when looking from the source to the sample, so that our coordinate system in such cases is exactly equivalent to imgCIF.

PR #2373, opened 2023-03-16
Clarify coordinate system conventions

dagewa:
We have a quite practical approach to convention. Make that clear in the documentation, for #2161


graeme-winter (approved):
Certainly an improvement on what was there before - which was false - but it does highlight an inconsistency. I am not sure we can fix that however, so thank you for the change set.

Issue #2285, opened 2022-11-23
exclude_images=0:1:200 not working in dials.integrate


cdavidowen:
DIALS 3.12.0-g13ced8fbc-release

Attempting to use format dataset_number:first_excluded_image:last_excluded_image. eg 0:1:200

dials.integrate crashes pretty quickly with a sample of the error message below. 

`Traceback (most recent call last):
  File "/dls_sw/apps/dials/dials-v3-12-0/build/../modules/dials/src/dials/command_line/integrate.py", line 736, in <module>
    run()
  File "/dls_sw/apps/dials/dials-v3-12-0/conda_base/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/dls_sw/apps/dials/dials-v3-12-0/build/../modules/dials/src/dials/command_line/integrate.py", line 677, in run
    params, options = parser.parse_args(args=args, show_diff_phil=False)
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/dials/src/dials/util/options.py", line 867, in parse_args
    params, args = self._phil_parser.parse_args(
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/dials/src/dials/util/options.py", line 502, in parse_args
    params = self._phil.extract()
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/cctbx_project/libtbx/phil/__init__.py", line 1782, in extract
    value = object.extract(parent=result)
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/cctbx_project/libtbx/phil/__init__.py", line 1145, in extract
    return self._type_from_words()(self.words, master=self)
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/cctbx_project/libtbx/phil/__init__.py", line 452, in from_words
    numbers = numbers_from_words(words=words, path=path)
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/cctbx_project/libtbx/phil/__init__.py", line 249, in numbers_from_words
    result.append(number_from_value_string(
  File "/dls_sw/apps/dials/dials-v3-12-0/modules/cctbx_project/libtbx/phil/__init__.py", line 223, in number_from_value_string
    raise RuntimeError(
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: Error interpreting exclude_images="0:1:100" as a numeric expression: SyntaxError: invalid syntax (line 1) (command line argument, line 1)
`
[...]

Issue #1227, opened 2020-04-07
dials.integrate scan_range broken


Anthchirp:
I have one continuous rotational dataset and want to integrate a subset of images.
```
dials.import *cbf
dials.find_spots imported.expt nproc=4
dials.index strong.refl imported.expt 
dials.refine indexed.*
dials.integrate refined.* scan_range=10,30
```
crashes with
```
+-----+---------+--------------+------------+--------------+------------+-----------------+
|   # |   Group |   Frame From |   Frame To |   Angle From |   Angle To |   # Reflections |
|-----+---------+--------------+------------+--------------+------------+-----------------|
|   0 |       0 |           10 |         30 |            1 |          3 |             137 |
+-----+---------+--------------+------------+--------------+------------+-----------------+

 Using multiprocessing with 1 parallel job(s)

Traceback (most recent call last):
  File "/dls_sw/apps/dials/dials-v2-2-1/modules/dials/algorithms/integration/processor.py", line 433, in __call__
    imageset = imageset[frame0:frame1]
  File "/dls_sw/apps/dials/dials-v2-2-1/modules/dxtbx/imageset.py", line 251, in __getitem__
    return self.partial_set(start - offset, stop - offset)
RuntimeError: dxtbx Internal Error: /dls_sw/apps/dials/dials-v2-2-1/modules/dxtbx/imageset.h(936): DXTBX_ASSERT(index < indices_.size()) failure.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/dls_sw/apps/dials/dials-v2-2-1/build/../modules/dials/command_line/integrate.py", line 665, in <module>
    script.run()
  File "/dls_sw/apps/dials/dials-v2-2-1/build/../modules/dials/command_line/integrate.py", line 353, in run
    reflections = integrator.integrate()
  File "/dls_sw/apps/dials/dials-v2-2-1/modules/dials/algorithms/integration/integrator.py", line 1083, in integrate
    reference, profile_fitter_list, time_info = processor.process()
  File "/dls_sw/apps/dials/dials-v2-2-1/modules/dials/algorithms/integration/processor.py", line 305, in process
    self.manager.accumulate(task())
  File "/dls_sw/apps/dials/dials-v2-2-1/modules/dials/algorithms/integration/processor.py", line 435, in __call__
    raise RuntimeError("Programmer Error: bad array range: %s" % str(e))
[...]

PR #2374, opened 2023-03-17
Use standard `exclude_images` syntax in `dials.find_spots` and `dials.integrate`

dagewa:
This PR extends the standard `exclude_images` interface to both `dials.find_spots` and `dials.integrate`.

Both cases make use of the `imageset.mark_for_rejection` feature as the core of these algorithms use `imageset`s rather than `experiments`. A new function, `set_invalid_images` was added to `exclude_images.py` and a few other fixes made in that file.

In the case of `dials.integrate`, there already _was_ an `exclude_images` option, but this used a different approach (#2285). That approach is deprecated by this PR. It was very user-unfriendly, as it required a list of image numbers using the 0-based array of `imageset` indices, no matter what image number the scan started from. It was also impossible to specify individual experiments using the old interface


dagewa:
This bug does not appear to affect `dials.scale`, but both are using the same `exclude_image_ranges_from_scans` function, so I'm not sure what is different here.

dagewa:
Ah, I see where it is going wrong. I made passing `reflections` into `_parse_exclude_images_commands` optional, as I don't have `reflections` where I call `set_invalid_images`. But then `ranges_to_remove` doesn't get updated :thinking: 
https://github.com/dials/dials/blob/9e1d8737ae2044e5c51e706cb5f95dca690537ad/src/dials/util/exclude_images.py#L122-L129

dagewa:
I've added what I _think_ should be the fix in [db8a4c5](https://github.com/dials/dials/pull/2374/commits/db8a4c5bab4105e0134b70d33161cff0a7a6e6ec), but I'd appreciate a careful check here as I'm not totally sure I'm using identifiers correctly.

jbeilstenedmands:
Hi @dagewa 
Thanks for addressing this long standing issue! Your fix looks correct, i'll attempt to run my own tests to validate the spotfinding and integration cases.
The reason the `if reflections` section uses the identifiers map is to allow use in the scaling case where you can do multiple cycles of filtering: in that case, you want to keep a constant mapping of experiments to reflection tables after you start removing some, so the dataset ids are kept constant and you need to go through the identifiers map to reliably match up to the experiment. For just working on experiments for spotfinding \& integration, locating based on position in the experiment list is valid.

dagewa:
Ah, that makes sense, thanks for the explanation!

jbeilstenedmands:
One thing that has come up in testing, is that if you exclude in spotfinding, integration doesn't 'know' about this and will integrate some of the full imageset anyway. E.g. beta-lactamase test dataset: import images 1-99, if I exclude images 51-99 in spotfinding, when I get to integration then all images get integrated by summation, and images 1-70 get integrated by profile-fitting (I guess the over-reach is to do with the block size for creating reference profiles).
Of course you can set the `exclude_images` again for integration, but I think the expectation would be that you maintain the exclusion.

The reason for this is that although the `valid_image_range` gets set during spotfinding, this is on the in-memory `imported.expt` which does not get written to disk. Then during indexing, there is no record of this from the `imported.expt`, so the `indexed.expt` has the default full valid image range set. Not sure yet what the best approach to solve this would be.

dagewa:
Is that not already true with scaling? That is, if I do `dials.scale integrated.expt integrated.refl exclude_images=20:30` then load the `scaled.expt`, I see e.g.
```
>>> scan.is_image_index_valid(35)
True
```

dagewa:
This is what I meant:
```
dials.scale integrated.expt integrated.refl exclude_images=20:30
```
then
```
>>> from dxtbx.model.experiment_list import ExperimentList
>>> el=ExperimentList.from_file("scaled.expt")
>>> el[0].scan.is_image_index_valid(25)
True
```

dagewa:
I was thinking about it a little longer and my feeling is that having `exclude_images` exclude only from _the current step_ is the least likely to cause surprise. So, I don't think it should accumulate from spot-finding, through integration and beyond in any case (including scaling). I think perhaps this is a difference between `exclude_images` and `scan_range=` type functionality (or `scan.image_range=` in the case of `dials.import`). In the latter case the user is modifying the shape of the experiment in a way that should be preserved through processing. As mentioned in the meeting today though I think a lot of this is still broken in DIALS (see #1227 and the image range importing issue mentioned by @toastisme)

jbeilstenedmands:
> Is that not already true with scaling? That is, if I do `dials.scale integrated.expt integrated.refl exclude_images=20:30` then load the `scaled.expt`, I see e.g.
> 
> ```
> >>> scan.is_image_index_valid(35)
> True
> ```

Yes and no. Scaling is using a slightly different property of the scan `valid_image_ranges`. This is a dict mapping experiment id to an image range, needed to handle the case where you have multiple lattices and want to trim one of them. So `dials.scale` will not try to create models for an image range that has been excluded in integration, which makes sense as there will be ~no reflections to scale there!

dagewa:
Ah, I see. Scaling is a complex use case. We might not want to overly enforce consistency, lest we invoke hobgoblins etc.

src/dials/util/exclude_images.py
```diff
@@ -173,6 +179,31 @@ def _remove_ranges_from_valid_image_ranges(experiments, ranges_to_remove):
...
+        valid_image_ranges_by_experiment, experiments
+    ):
+        if not experiment.scan:
+            raise ValueError("Trying to exclude a scanless experiment")
+        rejects = flex.bool(experiment.scan.get_num_images(), True)
+        first, last = experiment.scan.get_image_range()
+        for image_range in valid_image_ranges:
+            # Need to index into the imageset's 0-based array
+            accepts = (
+                flex.size_t(list(range(image_range[0], image_range[1] + 1))) - first
```
jbeilstenedmands, line 199:
```suggestion
                flex.size_t_range(image_range[0], image_range[1] + 1) - first
```
This lesser-known flex function exists.

jbeilstenedmands (approved):
Thanks for this. I'm satisfied that there could be use cases where you would not want the exclusion from spotfinding to carry through to integration, so let's leave as is.
It would be good to document the `scan.image_range`, `exclude_images` usage and effects throughout the different processing steps, but that's a separate issue and shouldn't preclude this work.

dagewa:
Thanks, I agree, I'll make an issue to document the various different approaches and to look for further improvements in the interfaces.

Issue #2379, opened 2023-03-31
`dials.export` does not add `HKL_base` to MTZ


dagewa:
According to 

```
gemmi mtz integrated.mtz
```

integrated MTZs from DIALS do not have the `HKL_base` dataset with id 0, which is [standard](https://www.ccp4.ac.uk/html/mtzformat.html#the-base-dataset). This often does not actually _matter_ because if the MTZ is touched by the CMTZ library, the base dataset will be added (this is also why it appears to be there on the output of `mtzdmp integrated.mtz` or `iotbx.mtz.dump integrated.mtz`).

However, this could be problematic if the MTZs are used without the CMTZ library (for example by software using gemmi).

jbeilstenedmands:
This should be straightforward, Richard did a similar thing for https://github.com/dials/dials/pull/23780, but only for merged MTZ I think:
```
mtz_writer.add_crystal(crystal_name="HKL_base", project_name="HKL_base")
mtz_writer.add_empty_dataset(wavelength=0, name="HKL_base")
mtz_writer.add_dataset(
    r_free_array=r_free_array,
)
 ```

dagewa:
Thanks. It was a difference I noticed compared to the MTZ files I was writing with gemmi, where the base dataset is set. If we decide to switch to gemmi output then this issue will be fixed (along with some others). I haven't got the PR quite ready for gemmi-fied MTZs yet though.

Issue #2057, opened 2022-04-04
MTZ output: sort order


dagewa:
Spun off from #1099 
> Could you have a check about that? And while you're at it: would it be
possible to write those MTZ files out in "standard" sort order, namely
"H K L M/ISYM BATCH" as before? That would make it easier for users
wanting to run MRFANA on those files.

dagewa:
Ok, not really a "bug", but non-standard. Columns `H`, `K`, `L`, `M/ISYM` and `BATCH` are set here:
https://github.com/dials/dials/blob/d8064546435e2db85f76e1d1675fedca6eb93ee5/util/export_mtz.py#L306-L312
After that it would be helpful to sort the MTZ data accordingly, but I couldn't see any function to do that...? We can't easily sort the reflection table prior to this because `M/ISYM` is set directly into the MTZ. We don't have those values earlier.

Issue #1100, opened 2020-01-17
Exported MTZ files don't contain meaningful data ranges


CV-GPhL:
Compared to previous versions of Dials (DIALS 1.14.13-g10ecfbb15-release), with DIALS 2.1.0-g40973e3ce we now get MTZ ranges always 1000000 ... -1000000. This requires an application to read the actual data once, while before we could just read these directly from the header.

I've never seen MTZ files without these items containing correct values - at least not with files written by the MTZ library itself.

graeme-winter:
@dagewa correctly highlights that this may have fallen off radar and needs attention

wojdyr:
I was about to report the same problem.
The minimum for all columns is set to 1000000 and maximum to -1000000.

```
VERS MTZ:V1.1
TITLE From dials.export
NCOL       17        93981      140
CELL    48.9148   77.3132   87.6687   89.9924   89.9558   89.9898
SORT    0   0   0   0   0
SYMINF   1  1 P     1                   'P1'     1
SYMM X,  Y,  Z
RESO 0.0001301100273849   0.3479580581188202
VALM NAN
COLUMN H                              H           1000000          -1000000    1
COLSRC H                              CREATED_16/04/2021_21:04:38              1
COLUMN K                              H           1000000          -1000000    1
COLSRC K                              CREATED_16/04/2021_21:04:38              1
COLUMN L                              H           1000000          -1000000    1
COLSRC L                              CREATED_16/04/2021_21:04:38              1
COLUMN M/ISYM                         Y           1000000          -1000000    1
COLSRC M/ISYM                         CREATED_16/04/2021_21:04:38              1
COLUMN BATCH                          B           1000000          -1000000    1
COLSRC BATCH                          CREATED_16/04/2021_21:04:38              1
COLUMN I                              J           1000000          -1000000    1
COLSRC I                              CREATED_16/04/2021_21:04:38              1
COLUMN SIGI                           Q           1000000          -1000000    1
COLSRC SIGI                           CREATED_16/04/2021_21:04:38              1
COLUMN SCALEUSED                      R           1000000          -1000000    1
COLSRC SCALEUSED                      CREATED_16/04/2021_21:04:38              1
COLUMN SIGSCALEUSED                   R           1000000          -1000000    1
COLSRC SIGSCALEUSED                   CREATED_16/04/2021_21:04:38              1
COLUMN BG                             R           1000000          -1000000    1
COLSRC BG                             CREATED_16/04/2021_21:04:38              1
[...]

Issue #1099, opened 2020-01-17
MTZ output: space group labels


graeme-winter:
Request from Clemens - 

we have problems running MRFANA on MTZ files exported from Dials:

From XIA2 0.6.354-gceb7eb98-dials-2.1, run on 15/01/2020 at 08:44:58
From DIALS 2.1.0-g40973e3ce, run on 2020-01-15 at 08:44:31 GMT

The problem seems to be that the spacegroup is written differently
into the MTZ file now: 'P1' instead of 'P 1' (the SG number is
obviously correct here):

* Space group = 'P1' (number      1)

This is different from previous versions of such MTZ files, ie. from

From XIA2 0.5.902-gffa11588-dials-1.14, run on 15/01/2020 at 10:10:25
From AIMLESS, version 0.7.4, run on 15/ 1/2020 at 10:09:52
From SORTMTZ 15/ 1/2020 10:09:05  using keys: H K L M/ISYM BATCH
From POINTLESS, version 1.11.21, run on 15/ 1/2020 at 10:09:04
From SORTMTZ 15/ 1/2020 10:08:36  using keys: H K L M/ISYM BATCH
From POINTLESS, version 1.11.21, run on 15/ 1/2020 at 10:08:33
From DIALS 1.14.13-g10ecfbb15-release, run on 2020-01-15 at 10:08:18 GMT

where we have

* Space group = 'P 1' (number      1)

Presumably, this will also affect other SGs, like "P 21 21 21" vs
"P212121" and "P 1 21 1" etc?

Although I agree that P1 and P21 is nicer than "P 1" and "P 1 21 1",
the normal CCP4/MTZ way would require the latter - and indeed, the
CCP4 libraries as called from (Fortran) MRFANA seem to get confused
with the space-less form.

Could you have a check about that? And while you're at it: would it be
possible to write those MTZ files out in "standard" sort order, namely
"H K L M/ISYM BATCH" as before? That would make it easier for users
wanting to run MRFANA on those files.

rjgildea:
Space group name is set here:
https://github.com/cctbx/cctbx_project/blob/42d9b7b7df5b988f2fb32f80102c1a4001b9f180/iotbx/mtz/__init__.py#L274

The name that goes into the file is determined by this function `ccp4_symbol`:
https://github.com/cctbx/cctbx_project/blob/42d9b7b7df5b988f2fb32f80102c1a4001b9f180/iotbx/mtz/extract_from_symmetry_lib.py#L25-L66

[...]

PR #2380, opened 2023-04-01
Export MTZ with gemmi

dagewa:
This is an experimental branch to rewrite MTZ export using gemmi. To explore how to do this while keeping like-for-like comparisons possible with the existing code, the approach taken here is to create a drop-in replacement for `iotbx.mtz.object`. The minimal interface required to write unmerged MTZs is created in a `GemmiMtzObject` (not all functions required for merged MTZ output are provided here). To use this rather than `iotbx.mtz.object`, set an environment variable, like this:

```
GEMMI_MTZ=1 dials.export integrated.{expt,refl
```

The gemmi-fied version of MTZ export fixes a number of longstanding issues:
- #1099 
- #1100 
- #2057 
- #2379 

However, reproducing the `iotbx.mtz.object` interface has proven to be an uncomfortable exercise with a shoehorn. In particular, gemmi does not really use an object hierarchy like MTZ/Crystal/Dataset/Column. It uses a simpler, [flattened hierarchy instead](https://gemmi.readthedocs.io/en/latest/hkl.html#mtz-format). Here the object hierarchy is recreated, but in a way that caches data until it is ready to be written. This adds some complication that would be unnecessary if `export_mtz` was rewritten with gemmi in mind, without attempting to preserve `iotbx.mtz.object` compatibility.

Gemmi has a very nice Python interface, which is completely ignored by this branch. It would be much cleaner if most of the work was done with a Python `gemmi.Mtz` object, which is then passed to a C++ function to add the batch headers. For example, the `show_summary` function would be so much easier to write in Python rather than using string formatting options available in C++11. However, `gemmi::Mtz`'s Python bindings are made with PyBind11. I don't know how to pass a `gemmi.Mtz` from Python into a DIALS C++ function. I think this would be worth finding out though. Also, it is useful to note that data can be set in a `gemmi.Mtz` by passing in NumPy arrays, which probably means we wouldn't need to wrap flex arrays up with PyBind11 in order to write an `add_dials_batches` function. We could `flumpy` them first.

Overall, I think this branch will be useful for testing, but it should not be merged in its current state. In particular, I have yet to test whether setting the space group does what we expect in all cases, including unusual settings. I have also not tested multi-wavelength MTZ output.

### DIFF
--- a/src/dials/util/boost_python/ext.cc
+++ b/src/dials/util/boost_python/ext.cc
@@ -79,7 +79,8 @@ namespace dials { namespace util { namespace boost_python {
       .def("set_title", &GemmiMtzObject::set_title)
       .def("add_history", &GemmiMtzObject::add_history)
       .def("set_space_group_by_name", &GemmiMtzObject::set_space_group_by_name)
-      .def("add_dataset", &GemmiMtzObject::add_dataset);
+      .def("add_dataset", &GemmiMtzObject::add_dataset)
+      .def("add_column", &GemmiMtzObject::add_column);
 
     class_<ResolutionMaskGenerator>("ResolutionMaskGenerator", no_init)
       .def(init<const BeamBase &, const Panel &>())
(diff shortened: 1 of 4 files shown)


dagewa:
About licensing: my understanding is that to include gemmi headers we also need to include the MPL 2.0 license file somewhere in the source tree. But it does not affect the licensing of our own software.

dagewa:
Superseded by #2549

PR #2381, opened 2023-04-04
Increase tolerance to fix Jenkins test failure

rjgildea:


rjgildea:
I think the test failure in `test_match_by_hkle ` is unrelated to the changes in this PR (see https://github.com/dials/dials/issues/2382)

PR #2385, opened 2023-04-11
Hide system/cctbx_project warnings when building DIALS

ndevenish:
This should make the build significantly clearer, and easier to see actual warnings.

dxtbx errors are left as-is, for now, as DIALS/dxtbx are often developed together and interactions between them are probably important.


ndevenish:
Much cleaner! I cancelled the testing as this just affects build output.

PR #2397, opened 2023-04-20
dials.export_bitmaps: display correct number of resolution rings

rjgildea:
The existing code meant that the nth ring was at the d_min of the image meaning that only n-1 rings were every actually visible on the image

`resolution_rings.number=5` (before):
rings5old0001
`resolution_rings.number=5` (after):
rings5new0001

`resolution_rings.number=10` (before):
rings10old0001
`resolution_rings.number=10` (after):
rings10new0001


dagewa (approved):
:+1: Thanks for yet another improvement

Issue #620, opened 2018-09-24
dials.index: Unexpected output from indexer_base.index_reflections?


ndevenish:
```
$ cd /dls/tmp/mep23677/tmp.iBLRwvoZUD/one
$ dials.index datablock_red.json reduced.pickle indexing.method=fft1d stills.indexer=sweeps max_lattices=5 reflections.outlier.algorithm=tukey   goniometer.fix=all   detector.fix=all   beam.fix=all auto_reduction.action=fix
```
Crashes inside refiner because of empty reflection table passed in.

To catch when it's added to the list to pass to the refiner, add or breakpoint at `indexer.py:1491` the test:
```python
    if len(refl.select(refl["id"]>=0)) == 0:
        import pdb
        pdb.set_trace()
    # The existing line 1492 is:
    args.append((params, refl, experiments))
```

The problem:

`choose_best_orientation_matrix` chooses experiments with id == -1:
https://github.com/dials/dials/blob/afa7ee9759b5f28825186f18262bd22aeefd661b/algorithms/indexing/indexer.py#L1413

It then runs the indexer:
https://github.com/dials/dials/blob/afa7ee9759b5f28825186f18262bd22aeefd661b/algorithms/indexing/indexer.py#L1434

and immediately validates that some of the reflections were indexed:
https://github.com/dials/dials/blob/afa7ee9759b5f28825186f18262bd22aeefd661b/algorithms/indexing/indexer.py#L1435-L1436

Later on, through `easy_mp`, `run_one_refinement` filters experiments by id > -1
https://github.com/dials/dials/blob/afa7ee9759b5f28825186f18262bd22aeefd661b/algorithms/indexing/indexer.py#L1327

And passes this into the refiner. So, the indexer is returning reflections that are unassigned for any experiments yet have the indexed flag set. I'm trying to understand this failure mode.

PR #2408, opened 2023-04-28
Test changes for add_properties_table_to_scan dxtbx PR

toastisme:
These are some minor changes required on the DIALS side when using [#620](https://github.com/cctbx/dxtbx/pull/620). The main one being imprecision when setting oscillation
```
>>> scan.set_oscillation(82.0, 0.15)
>>> scan.get_oscillation()
(82.0, 0.15000000000000338)
```
The loss of precision happens when converting oscillation into an array. Obviously it's negligible but it does actually change things downstream (e.g. one or two reflections right on the border of a slice are now included, leading to slightly different refinement and hence integration etc.)


toastisme:
> Not looked into why the tests are failing though

Sorry thought I had written it here somewhere: single test is failing due to [a dials-data](https://github.com/dials/data-files/pull/49) change needing to be merged first.

PR #2409, opened 2023-05-02
Handle no-conda environment, including msgpack

phyy-nx:
Relevant e.g. for virtualenv builds, using a facility provided python.  Uses the msgpack from the bootstrap 'hot' step.

Note in #2319, specifically fb95810459bb138b8160c0d569ee3ac7420485f0, the msgpack3-1-1 reference was removed.   The confusion here was that the comment said it was only needed for Python2.7.  Turns out that isn't strictly true, as the virtualenv build path uses this hot package.

Someone should make a CI that exercises the virtualenv build path :)


SConscript
```diff
@@ -13,7 +13,7 @@ if not env_etc.no_boost_python and hasattr(env_etc, "boost_adaptbx_include"):
     env = env_no_includes_boost_python_ext.Clone()
     env_etc.enable_more_warnings(env=env)
 
-    system_includes = [x for x in env_etc.conda_cpppath if x]
+    system_includes = [x for x in env_etc.conda_cpppath if x] if libtbx.env.build_options.use_conda else []
```
ndevenish, line 16: From this line, I assume that `env_etc.conda_cpppath` is literally missing, instead of empty, outside of a conda build?

ndevenish:
Wouldn't this be better handled by e.g. injecting the include path into the build environment at the same point that is responsible for installing the """"""hot"""""" dependency?

SConscript
phyy-nx: Ya, the error message I get is something like 'empty has no attribute conda_cpppath'

phyy-nx:
I think you are saying add the include directory in bootstrap if use_conda isn't set?  That could work too.  This PR kinda just puts us back where we were, so it was lower effort :)

FWIW, the old 'hot' mechanism in bootstrap was exactly for this purpose: distributing packages that are frozen due to some functionality being needed for a specific purpose.  Perhaps it would be better to call it a 'cold' mechanism :P 

If it helps, here's a list of pseudocode steps this build procedure uses (this one is for SACLA):

- bootstrap hot update
- module load python 3.8, openmpi, hdf5.  These are custom libraries from the facility.
- `python3 -m venv --system-site-packages env.` This sets up the virtual environment
- `source env/bin/activate`
- pip install just all the things
- bootstrap build

Issue #2405, opened 2023-04-27
`dials.model_background` fails for datasets that do not start at image 1


dagewa:
This was reported by a user, and I have reproduced using test data:

```
dials.import $(dials.data get -q insulin)/insulin_1_0{05..45}.img
dials.find_spots imported.expt 
dials.index imported.expt strong.refl 
dials.refine indexed.expt indexed.refl 
dials.integrate refined.expt refined.refl 
dials.model_background integrated.expt
```
fails with

```
Traceback (most recent call last):
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/integration/image_integrator.py", line 167, in __call__
    imageset = imageset[index0:index1]
  File "/home/fcx32934/sw/cctbx/modules/dxtbx/src/dxtbx/imageset.py", line 330, in __getitem__
    return self.partial_set(reader, start, stop)
OverflowError: can't convert negative value to unsigned int

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/model_background.py", line 307, in <module>
    run()
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/model_background.py", line 303, in run
    script.run(args)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/model_background.py", line 276, in run
    model = modeller.compute()
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/background/modeller.py", line 340, in compute
    _, time_info = processor.process()
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/integration/image_integrator.py", line 100, in process
    self.manager.accumulate(task())
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/integration/image_integrator.py", line 169, in __call__
    raise RuntimeError("Programmer Error: bad array range")
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: Programmer Error: bad array range
```
The block of code being executed is something to do with slicing an imageset
https://github.com/dials/dials/blob/f11792932d3cf67086b48098d91078ddb74f60c2/src/dials/algorithms/integration/image_integrator.py#L157-L169
[...]

PR #2411, opened 2023-05-05
Imageset slicing

dagewa:
This is the DIALS component of a PR pair with https://github.com/cctbx/dxtbx/pull/634.

Imageset slicing is reverted back to using zero-based indices, which is how the data is accessed. DIALS tests pass locally but some battle testing should be done with real data (scans, stills, image numbers starting from 1, 0, 42, etc.) to ensure the results of processing remain the same.


dagewa:
Oh, should add a regression test for #2405 here too really

dagewa:
To test whether image numbering matters for processing I used a simple script that renumbers the insulin data set from dials-data to start from image 0, 1 or 42, and processes through from import to scaling:

```bash
#!/bin/bash
TOP=$(pwd)
for START in 0 1 42
do
    cd $TOP
    mkdir -p "data_$START"
    cd "data_$START"
    i=$START
    for FILE in $(dials.data get -q insulin)/*.img
    do
        ln -sf "$FILE" $(printf "image_%04d.img" "$i")
        i=$((i+1))
    done
    dials.import *.img
    dials.scale integrated.expt integrated.refl
done
```
I first checked this with dxtbx and dials both on main branches. Comparing `dials.scale.log` for all three processing results I see there is no difference. That's good and as we expect. It means we don't currently have a dependence on image number (previously there have been issues reported when images start from number 0, for example).

Following this, I ran the same script with dials on https://github.com/dials/dials/tree/remove-batch-offset-from-imageset-slicing and dxtbx on https://github.com/dials/dxtbx/tree/633-imageset-slicing. Again, there is no difference between the 3 sets of results, and they are identical to the results on main. That means, I have not _introduced_ any dependence on image number by fixing the imageset slicing issue.

So, all looks good for single-file-per-image scans. I imagine it might be worth some tests with stills data sets (@jbeilstenedmands, @phyy-nx?) and possibly multi-image stack datasets like .nxs as there could be some different code paths.

jbeilstenedmands (approved):
I tested this out on some ssx data (cbf and h5) and the results are identical to before. I also ran the xia2 ssx tests which passed.
I'm glad to see that this simplifies the slicing to what it should always have been in `indexing/ssx/processing.py`
I added another test for `dials.import` as I realised that we didn't have an import test for ssx data in cbf form (and I thought this was failing at first, but I just had the wrong dxtbx branch checked out).

PR #2420, opened 2023-05-30
Implement reindexing against a reference file (pdb, mtz, cif) in dials.reindex

jbeilstenedmands:
Due to the series of cell transformations in cosym, when a reference dataset is given, the indexing mode can change, meaning that the final data are not reindexed into the correct mode for some use cases. This PR introduces a final test on the reindexed data, to calculate the CC with the reference dataset (in its true symmetry, without any cell reduction), and applying a final reindexing to the whole dataset to be in the most consistent indexing mode.

A few logging changes are required in dials.reindex as a side effect.


src/dials/command_line/reindex.py
```diff
@@ -70,6 +75,8 @@
   reflections = reindexed.refl
     .type = str
     .help = "The filename for reindexed reflections"
+  log = dials.reindex.log
+    .type = str
```
rjgildea, line 79:
```suggestion
    .type = path
```

rjgildea (approved):
Thanks for adding this, lgtm!

jbeilstenedmands:
After further thought and testing, I'm not sure if what I was trying to do is particularly robust. As the data are unscaled, it does not become reliable to calculate a cc against the reference (I was trying to do this on the expt and refl file data to avoid the symmetry transformations, but then the intensities haven't even had the normalisation applied). I think we can achieve what we want by a simple call to `dials.reindex` using a reference after scaling, so I think I will back out some of the changes to cosym and keep the recent commits to add the functionality to `dials.reindex`

Issue #2421, opened 2023-05-30
Unreliable test for stills in dials.find_spots


graeme-winter:
While trying to chase down #2416 

https://github.com/dials/dials/blob/45bc4e8460d680c1fbab980a0ba5c058318e16a3/src/dials/algorithms/spot_finding/factory.py#L429

This is not a good test for being still, given that `is_still()` exists as a function? E.g. `ImageSequence` can be a grid scan / SSX experiment

PR #2422, opened 2023-05-30
Remove force_2d option

### LINKED ISSUE #2416: dials.find_spots: force_2d does not
Running `dials.find_spots` with

```
DIALS 3.dev.891-gc51410c50

spotfinder {
  force_2d = True
  filter {
    max_separation = 100
  }
  threshold {
    algorithm = *dispersion dispersion_extended radial_profile
  }
}
input {
  experiments = imported.expt
}
```

I had imagined that spots on frame j, j+1 would be considered differently i.e. the sequence analysed as if a grid scan. They were not:

```
silver-surfer 2d :) $ dials.python
Python 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:23:20) 
[Clang 12.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dials.array_family import flex
>>> x = flex.reflection_table.from_file("strong.refl")
>>> bb = x["bbox"]
>>> bb = x["bbox"].parts()
>>> flex.max(bb[1] - bb[0])
7
>>> flex.max(bb[3] - bb[2])
16
>>> flex.max(bb[5] - bb[4])
79
```

graeme-winter:
Does not appear to work, not sure what it should do, so removing it in dials.find_spots makes most sense. We can work out a better way of doing what it should have done.

Fixes #2416, #2421

### DIFF
--- a/newsfragments/2422.removal
+++ b/newsfragments/2422.removal
@@ -1,0 +1,2 @@
+``dials.find_spots``: remove ``force_2d`` option which did not apparently do
+anything useful.
--- a/src/dials/algorithms/spot_finding/factory.py
+++ b/src/dials/algorithms/spot_finding/factory.py
@@ -7,7 +7,6 @@ import time
 
 import numpy as np
 
-from dxtbx.imageset import ImageSequence
 from iotbx.phil import parse
 
 import dials.extensions
@@ -35,10 +34,6 @@ def generate_phil_scope():
     hot_mask_prefix = 'hot_mask'
       .type = str
       .help = "Prefix for the hot mask pickle file"
-
-    force_2d = False
-      .type = bool
-      .help = "Do spot finding in 2D"
 
     scan_range = None
       .help = "The range of images to use in finding spots. The ranges are"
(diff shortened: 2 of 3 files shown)


rjgildea:
I would suggest going further and removing all references to the `no_shoeboxes_2d` parameter and the [`ExtractSpots._find_spots_2d_no_shoeboxes()` method ](https://github.com/dials/dials/blob/9cbc1e7e117c95a281b1a024762f3c1462a8ef40/src/dials/algorithms/spot_finding/finder.py#L535-L618). As far as I can tell neither of these are tested, and would be completely inaccessible if the `force_2d` is removed. Going further down the rabbit hole would suggest that [`ExtractPixelsFromImage2DNoShoeboxes` ](https://github.com/dials/dials/blob/9cbc1e7e117c95a281b1a024762f3c1462a8ef40/src/dials/algorithms/spot_finding/finder.py#L167-L239) could also then be removed as unused and untested.

graeme-winter:
Yeh, that feels sensible. Wilco. 

Will close other PR with prejudice. Did not see this sugestion.

graeme-winter:
@rjgildea removed, tests still pass, force pushed to the branch, would welcome review.

rjgildea:
Can the [ExtractPixelsFromImage2DNoShoeboxes](https://github.com/dials/dials/blob/9cbc1e7e117c95a281b1a024762f3c1462a8ef40/src/dials/algorithms/spot_finding/finder.py#L167-L239) be removed too?

PR #2423, opened 2023-05-30
Rename confusing variable

graeme-winter:
`no_shoeboxes_2d` -> `no_shoeboxes_and_2d` to indicate that this is the union of two conditions - this loads a different spot finder which is named more kin to the latter, which is how I figured this out


graeme-winter:
Will render useless

PR #2434, opened 2023-06-13
Testing: Prevent dials_regression having trailing slash

ndevenish:
If the environment DIALS_REGRESSION was set with a trailing slash, this could be passed through to tests. At least one test (test_export_mosflm) failed in this case.

Prevent this happening.


rjgildea:
The failing windows tests should be fixed by https://github.com/dials/dials/pull/2437

PR #2438, opened 2023-06-15
Mark xfel dependency on image grouping test

rjgildea:
See dials#2389 and dials#2404


rjgildea:
The failing windows tests should be fixed by dials/dials#2437

jbeilstenedmands (approved):

PR #2441, opened 2023-06-16
Ensure the `resolution_rings.fontsize` option of `dials.export_bitmaps` works on Macs

dagewa:
@huwjenkins I guess this should do it. Nested try is a bit yuck though


ndevenish:
Is there a reason that it has to be Arial instead of... whatever the default font is anyway?

dagewa:
It has to be a truetype font in order to scale to different sizes. Doesn't have to be arial, I just picked that as the most likely to be present

dagewa:
the default font is a fixed size (small) bitmap, so you get labels like this 
image

rather than being able to scale to whatever you want, like this
image

ndevenish:
Okay, this now just queries `matplotlib` to get it's default font, and only falls back if that fails. This should ensure that it a) works, and b) is the same font, across platforms, and even on headless machines without system fonts present (as we always package matplotlib and it appears to always package its own fonts).

PR #2445, opened 2023-06-27
Drop Python 3.8 support

rjgildea:
Upstream dependencies (e.g. numpy, scipy) have now dropped Python 3.8 support in latest releases.

See also https://dials.github.io/kb/proposals/dc3


rjgildea:
This looks to have been [fixed recently in h5py](https://github.com/h5py/h5py/commit/66f6158bbaf3095cc3ff0f3287328470785a82b2), but has perhaps not yet filtered through to the version of h5py we are currently pinned to.

ndevenish:
> This looks to have been [fixed recently in h5py](https://github.com/h5py/h5py/commit/66f6158bbaf3095cc3ff0f3287328470785a82b2), but has perhaps not yet filtered through to the version of h5py we are currently pinned to.

Ah, it looks like the conda-forge build is tied to HDF5 1.14+ only

Issue #2447, opened 2023-07-03
SSX (grid scan) spot finding not parallelized and using a lot of memory


JunAishima:
Using v3.14.2 (downloaded from website) on RHEL 8 computers. Datasets are collected as oscillations, though they should be considered stills. To convert the data to be seen as stills, we read datasets in as grid scans, 400 images (20x20). When we run the default spot finding algorithm, it insists on doing a couple of things that do not seem correct:

- running with only a single core (despite setting njobs or chunksize)
- constantly increasing in RAM usage until it crashes the process (when trying to spot find 4x 400 full Eiger 16M datasets - 256GB RAM). This actually works fine with a single dataset, but we wanted to test whether it can handle 4 full datasets. Note the output below is for 2x 400 image datasets, and seems to work ok.

> dials.import as_grid_scan=True grid_size=20,20 directory=.

DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 3.14.2-gf955030ae-release
The following parameters have been modified:

input {
  directory = "."
  as_grid_scan = True
  grid_size = 20 20
}

--------------------------------------------------------------------------------
  format: <class 'dxtbx.format.FormatNXmxEigerFilewriter.FormatNXmxEigerFilewriter'>
  num images: 800
  sequences:
    still:    0
    sweep:    0
  num stills: 800
--------------------------------------------------------------------------------
Writing experiments to imported.expt

> dials.find_spots imported.expt

DIALS (2018) Acta Cryst. D74, 85-97. https://doi.org/10.1107/S2059798317017235
DIALS 3.14.2-gf955030ae-release
The following parameters have been modified:

input {
  experiments = imported.expt
}

Setting spotfinder.filter.min_spot_size=3
Configuring spot finder from input parameters
--------------------------------------------------------------------------------
Finding strong spots in imageset 0
--------------------------------------------------------------------------------

Finding spots in image 1 to 1...
Setting nproc=128
Setting chunksize=1
Extracting strong pixels from images
 Using multiprocessing with 1 parallel job(s)

Found 1827 strong pixels on image 1

ndevenish:
Hi @JunAishima; as an initial guess, does the setting `dials.import convert_stills_to_sequences=true` help at all? It looks like the NXmx file is being picked up as the original "Stills" processing (which `dials.stills_process` uses), instead of the "Sequence of Stills" form (which is better for running the normal dials tooling on still images).

[...]

PR #2454, opened 2023-07-05
Remove as_grid_scan from dials.import

### LINKED ISSUE #2452: Remove as_grid_scan from dials.import
As seen in #2447 this does nothing useful and is misleading - either

 - remove `as_grid_scan`
 - modify it to be an alias which sets `scan.oscillation=0,0`

I think the second is superior, will look

graeme-winter:
Is not useful, and is misleading as it makes data into stills rather than image sequence. If we want to reinstate it, this can be reverted and fixed up appropriately.

Fixes #2452

Mentined in #2447


ndevenish:
Given the problem last time we removed one of these, I'd be more comfortable deprecating it first?

graeme-winter:
I take your point with this. It does not change the fact that it doesn't work and is misleading and useless, but deprecating for next release then removing after release does seem sane

Issue #2316, opened 2023-01-17
dials.scale: absorption_correction=auto -> disable if all scattering angles small


graeme-winter:
In dials.scale where we have `absorption_correction=auto` the default currently depends only on the total rotation angle. If however all the _scattering_ angles are small e.g. from µED (example I see has all cos(2ϑ) > 0.999) then all the paths are the same for all spots on one image so we cannot do absorption correction, it will just be another batch scale factor. 

Suggest if all cos(2ϑ) > 0.99 say then disable?

PR #2455, opened 2023-07-05
dials.scale: switch off absorption correction when all scattering angles are small

dagewa:
I don't really like magic numbers, but I don't think we have a better way of easily doing this. Chose 5° as the limit as a "large small number".


dagewa:
NB this is for https://github.com/dials/dials/issues/2316

graeme-winter:
> 5° feels like a large small number, but I appreciate the difficulty in selecting such a magic value. I could however suggest that cos(2ϑ) ~ 1 is a reasonable constraint? Certainly, if we select say 0.1% deviation from 1 we get something sensible-ish and a little more ... mathematical feeling?
> 
> ```
> >>> math.acos(0.999) * 180./math.pi
> 2.5625587331231388
> ```

Argument made on the basis that the path length through the crystal varies by 1/cos(2ϑ)

dagewa:
I take the point, but I was looking at one randomly chosen electron diffraction dataset and the highest $2 \theta$ angle in that case was 3.6°. I worried that a 2.56° limit would be switching behaviour a lot between different electron diffraction datasets, because they typically do have highest scattering angles around 2-3°.

I have generally been switching off absorption correction as a matter of course for ED. Perhaps I should not and just let `auto` do its job...

graeme-winter:
> I take the point, but I was looking at one randomly chosen electron diffraction dataset and the highest 2θ angle in that case was 3.6°. I worried that a 2.56° limit would be switching behaviour a lot between different electron diffraction datasets, because they typically do have highest scattering angles around 2-3°.
> 
> I have generally been switching off absorption correction as a matter of course for ED. Perhaps I should not and just let `auto` do its job...

If you have a ciunter example, that is stronger than my arbitrary cut off

5° also reasonable

dagewa:
I've followed your argument. If absorption correction _should_ always be switched off for ED, then it should be done via a different mechanism (https://github.com/cctbx/dxtbx/issues/575)

dagewa:
`test_model_background.py` currently seems unstable on Mac. The failure is not related to this PR

Issue #712, opened 2019-02-21
Show mask in dials.image_viewer is confusing


jmp1985:
In the image viewer there is a show mask toggle box on the main window and also in the mask tool. The toggle box in the mask tool does not appear to do anything. Also, when a new mask element is added, it does not appear on the image even if show mask is selected. The mask only appears when show mask in the main window is unselected and then reselected.

PR #2458, opened 2023-07-10
Remove useless checkbox

dagewa:
A 5 minute fix for #712 (after 4 and a half years)


dagewa:
I think the macos test failure has nothing to do with this PR

PR #2465, opened 2023-07-25
No datablocks

dagewa:
Two tests for `dials.search_beam_position` are changed to no longer use `datablock.json` files. After merging this and https://github.com/cctbx/dxtbx/pull/570, there should be no test failures.


newsfragments/2465.misc
```diff
@@ -0,0 +1 @@
+Tests for ``dials.search_beam_position`` are changed to no longer use ``DataBlock``
```
graeme-winter, line 1: Prefix this with "Mnt:"? Because this is a test thing not interesting-to-$USER

tests/command_line/test_search_beam_position.py
```diff
@@ -207,28 +205,20 @@ def test_search_small_molecule(dials_data, run_in_tmp_path):
...
-            os.path.join(data_dir, "SWEEP[1,2]", "index", "*_datablock_import.json")
-        )
-    )
+def test_multi_sweep_fixed_rotation(dials_data, run_in_tmp_path):
+    data = dials_data("l_cysteine_dials_output", pathlib=True)
+    experiments_path = data / "imported.expt"
+    refl_path = data / "strong.refl"
 
-    search_beam_position.run(reflection_files + experiment_files)
+    search_beam_position.run([os.fspath(experiments_path), os.fspath(refl_path)])
```
graeme-winter, line 213: Question: what is `os.fspath` doing here? I thought you can transparently use `path` objects as a string-like thing?

graeme-winter (approved):
Change set looks good, and as a happy side-effect also removes references to more pickle as well. Thanks you. Some questions inline for my own enlightenment.

tests/command_line/test_search_beam_position.py
dagewa: Without that I get
```
TypeError: Please report this error to dials-support@lists.sourceforge.net: 'PosixPath' object is not subscriptable
```
although it probably also works just to call `str` rather than `os.fspath`

ndevenish: Here it does nothing and is probably not hurting anything because we _know_ that we get `pathlib.Path` objects from dials_data - in general you might want to accept anything that has `__fspath__`.

That said, above in this PR _is_ explicitly using `str(Path)` so I've changed this to match.

ndevenish:
This looks ready. Will merge after release.

ndevenish:
Couple of remaining usages of datablock in xfail tests added by https://github.com/dials/dials/pull/1910 - [test_phi_scan.py](https://github.com/dials/dials/blob/e779efbf55bb6c9401a7e8ae1887adfa653093f1/tests/algorithms/indexing/test_phi_scan.py#L13-L25) and [test_index.py](https://github.com/dials/dials/blob/e779efbf55bb6c9401a7e8ae1887adfa653093f1/tests/algorithms/indexing/test_index.py#L573-L577)

These are probably fixed by https://github.com/dials/data/pull/312, which we can probably do now.

ndevenish:
Oh, and https://github.com/dials/dials/blob/main/doc/sphinx/documentation/tutorials/3DED/lysozyme_nanocrystals.rst mentions datablocks a whole bunch, how do we feel about that tutorial?

dagewa:
It's kind of frozen in time, that one. It mentions specific versions of DIALS and CCP4 at the start. IIRC it was kept around following a reviewer request that the instructions for processing those datasets be made public. However, it is now quite hard to find a version of DIALS for which those instructions actually work. Nevertheless, it was already de-linked from the tutorial index page, so people are unlikely to stumble across it.

PR #2469, opened 2023-07-26
Ensure `invert_rotation_axis=True` is only applied once to each goniometer model

dagewa:
More generally, ensure manual geometry changes are only applied once to each unique model.


dagewa:
@graeme-winter I realise I'm tagging you a lot for PR reviews lately, but you came up in the suggestions :wink:

newsfragments/2469.bugfix
```diff
@@ -0,0 +1 @@
+``dials.import``: ensure ``invert_rotation_axis=True`` is only applied once to each goniometer model
```
graeme-winter, line 1: Scope wise, this touches detector, scan and beam as well, though I note that goniometer is the only object where the reassignment is not idempotent since `invert_rotation_axis` is a toggle.

src/dials/command_line/dials_import.py
```diff
@@ -361,17 +362,27 @@ def __call__(self, imageset):
...
-            detector = DetectorFactory.from_phil(
-                self.params.geometry, imageset.get_detector(), beam
-            )
-            goniometer = GoniometerFactory.from_phil(
-                self.params.geometry, imageset.get_goniometer()
-            )
-            scan = ScanFactory.from_phil(
-                self.params.geometry, deepcopy(imageset.get_scan())
-            )
+            if beam and beam not in self.touched:
```
graeme-winter, line 372: I am not convinced that the `if beam and beam not in self.touched` logic here is consistent with the `beam = BeamFactory()` since every new beam will be a beam? To be precise, I think I would like a unit test which demonstrates that the logic works as advertised.

src/dials/command_line/dials_import.py
```diff
@@ -389,18 +400,35 @@ def __call__(self, imageset):
...
-                scan = ScanFactory.from_phil(self.params.geometry, imageset.get_scan(i))
+                if beam and beam not in self.touched:
+                    beam = BeamFactory.from_phil(
+                        self.params.geometry, imageset.get_beam(i)
+                    )
+                if detector and detector not in self.touched:
+                    detector = DetectorFactory.from_phil(
+                        self.params.geometry, imageset.get_detector(i), beam
+                    )
+                if goniometer and goniometer not in self.touched:
```
graeme-winter, line 411: Since the goniometer is what the PR is ultimately for, maybe should have test coverage? I note that the test added below may do this but not be included in the coverage analysis (I will check when I get to it)

tests/command_line/test_import.py
```diff
@@ -134,6 +134,31 @@ def test_can_import_multiple_sequences(dials_data, tmp_path):
...
 
 
+def test_invert_axis_with_two_sequences_sharing_a_goniometer(dials_data, tmp_path):
+    # Test for regression of https://github.com/dials/dials/issues/2467
+    image_files = sorted(
+        dials_data("centroid_test_data", pathlib=True).glob("centroid*.cbf")
+    )
+    del image_files[4]  # Delete filename to force two sequences
+
+    result = subprocess.run(
```
graeme-winter, line 144: The fact that this does not appear to be recognised by Codecov makes me 🤔

tests/command_line/test_import.py
```diff
@@ -134,6 +134,31 @@ def test_can_import_multiple_sequences(dials_data, tmp_path):
...
+        + image_files,
+        cwd=tmp_path,
+        capture_output=True,
+    )
+    assert not result.returncode and not result.stderr
+    assert (tmp_path / "experiments_multiple_sequences.expt").is_file()
+
+    experiments = load.experiment_list(tmp_path / "experiments_multiple_sequences.expt")
+    assert len(experiments.goniometers()) == 1
+    assert experiments.goniometers()[0].get_rotation_axis() == (-1.0, 0.0, 0.0)
```
graeme-winter, line 159: This test looks good / sane - what if however we also (manually) modify `experiments_multiple_sequences.expt` to have two turning forward, two turning backward and make sure they are all correctly inverted?

graeme-winter (approved):
On the balance I am fairly sure that the change set meets the brief of fixing the issue as the added test exercises the right code paths. Some concerns:

- the motif of `if scan and scan not in thing` `scan = ` worries me slightly as I am nervous that it does not do what we may intend
- I am left with the idea that `invert_rotation_axis` is fundamentally wrong, because the other overrides are idempotent but this one is a toggle (which gives rise to the fundamental issue)

On the latter - I note we do not offer `invert_beam` or `invert_detector_origin` etc. - however I also recognise that the inverted rotation axis is by far and away the most common issue. Just left with a slight sense of unease, which maybe I should document in a subsequent issue.

The change set I believe meets the brief and so is a good addition to the code base, just raises questions for me.

src/dials/command_line/dials_import.py
dagewa: This is more for the `goniometer` and `scan` models, which could be `None`, in which case I thought it should not try to override the geometry. No sane imageset would give `None` for `get_beam()` or `get_detector()`, but I figured the logic could just as well be consistent for each model. I'll think about a suitable test,,,

tests/command_line/test_import.py
dagewa: I'm not sure what to do about that. The file `experiments_multiple_sequences.expt` only exists _after_ import, so it can't be changed to provide different input to `dials.import`. It's possible a rotation axis can be manually specified _and_ `invert_rotation_axis=True` be set. I'm not sure what currently happens in that case or indeed what _should_ happen.

tests/command_line/test_import.py
dagewa: I suspect the relevant lines of code were already covered by tests. Adding this just tests behaviour in a particular case, but does not change the coverage.

newsfragments/2469.bugfix
dagewa: Updated to note the more general scope of the changeset in the newsfragment.

tests/command_line/test_import.py
dagewa: Ah should be able to do this with a direct test of `ManualGeometryUpdater` rather than testing `dials.import` as a command-line program. :+1:

Issue #2450, opened 2023-07-05
dials.index: if iterating through many experiments please give hint of experiment number


graeme-winter:
As in spot finding, where we say "finding spots on imageset N"

Indexing experiment N would seem polite (I am indexing a 200-sweep data set)

dagewa:
@graeme-winter can you please try the branch https://github.com/dials/dials/tree/2450-dialsindex-if-iterating-through-many-experiments-please-give-hint-of-experiment-number to see if this does what you want? I am not in a position to test at the moment.

graeme-winter:
> @graeme-winter can you please try the branch https://github.com/dials/dials/tree/2450-dialsindex-if-iterating-through-many-experiments-please-give-hint-of-experiment-number to see if this does what you want? I am not in a position to test at the moment.

Wilco, thank you (got bogged down in other issues)

graeme-winter:
Mechanism:

```
dials.import <many data>
dials.find_spots imported.expt
dials.index imported.expt strong.refl joint=0
```

dagewa:
I think the branch does _not_ do what you want. I imported 4 datasets, and in each case looked for multiple lattices. The printed experiment id resets after each imageset:
```
$ grep "Adding indexed" dials.index.log 
Adding indexed experiment 0
Adding indexed experiment 1
Adding indexed experiment 0
Adding indexed experiment 1
Adding indexed experiment 2
Adding indexed experiment 0
Adding indexed experiment 0
Adding indexed experiment 1
Adding indexed experiment 2

```

dagewa:
Had a look at this again but forgot what data I was looking at before. I'm not sure we have a suitable entry in `dials-data`, but I think the issue can be explored by splitting a dataset over 9 sweeps like this:
```
dials.import $(dials.data get -q x4wide)/X4_wide_M1S4_2_00{0..8}{1..8}.cbf 
dials.find_spots imported.expt
dials.index imported.expt strong.refl joint_indexing=False
```

ndevenish:
Closed by #2470

PR #2470, opened 2023-07-26
Log the experiment number when indexing multiple experiments

dagewa:
Second attempt at fixing #2450. This time I think it does the right thing.


graeme-winter:
@dagewa  something funny with the change set - many files touched - does this need to be rebased onto current `main` or similar? Am testing the code now to see whether the behaviour is as hoped

graeme-winter:
```
|          0 |          99 |            24 | 80.5%       |
Indexing experiment 3
Using max_cell: 87.1 Angstrom
Setting d_min: 1.70
FFT gridding: (256,256,256)
Number of centroids used: 463
```

OK, this now makes me happy - though a _slight_ improvement would be 

```
Indexing experiment 3 / N
```

If you have a chance. This is still much more useful than what was there before though thank you

dagewa:
> @dagewa something funny with the change set - many files touched - does this need to be rebased onto current `main` or similar? Am testing the code now to see whether the behaviour is as hoped

Could be... in my hands lately GitHub has been cursed. Will try merging `main` in

dagewa:
> ```
> Indexing experiment 3 / N
> ```
> 
> If you have a chance. This is still much more useful than what was there before though thank you

No problem, have done that. A consequence of doing so is that the numerator of the displayed fraction is now a 1-based index, not equal to the experimental `id`. This may be an improvement anyway.

phyy-nx:
Chiming in.  Off by ones plague us so much.  Maybe we could keep the original numbering?

`Indexing experiment 0 of 3 total`
`Indexing experiment 1 of 3 total`
`Indexing experiment 2 of 3 total`

Looks weird, but I use the logs to pull stuff out sometimes and correctness is more important than humanness to me.

(if I had my druthers, I'd change everything back to 0 indexed, including the image viewer... :)

dagewa:
I see your point, but I think seeing the message `Indexing experiment 2 of 3 total` last is going to confuse the user, who might then be looking for another one after that.

We have two pieces of information to transmit here, the `id` and the progress. Perhaps we should just keep them separate?

```
Indexing experiment id 2 (3/3)
```

phyy-nx:
👍-AaronOn Aug 6, 2023, at 5:19 AM, David Waterman ***@***.***> wrote:﻿
I see your point, but I think seeing the message Indexing experiment 2 of 3 total last is going to confuse the user, who might then be looking for another one after that.
We have two pieces of information to transmit here, the id and the progress. Perhaps we should just keep them separate?
Indexing experiment id 2 (3/3)

PR #2472, opened 2023-07-27
Slice using batch offsets fix

jbeilstenedmands:
Missed update in https://github.com/dials/dials/pull/2411 as highlighted by xia2 test failures


dagewa (approved):
Thanks for finding and fixing this. I did not spot it before as I didn't run xia2 tests on the PR pair.

I have checked this and agree that your change is correct. As noted, it essentially reverts e58d8fa4105, one of the fixes-upon-fixes that I have been trying to disentangle in this whole process.

Imagesets are accessed using 0-based indices, and sliced thusly. The scan `[` behaves in the same way, therefore calculating `beg` and `end` as 0-based and then using them for both slice operations is correct. To paraphrase the code, this is correct:

```
beg = sr[0] - 1 - arr_start
end = sr[1] - arr_start
exp.scan[beg:end]
exp.imageset = exp.imageset[beg:end]
```

PR #2474, opened 2023-07-27
Support CMake environments in bootstrap.py

ndevenish:
This adds:
```
modules/dials/bootstrap.py --cmake
```
which will install prebuilt cctbx, then configure and build dials, dxtbx and xia2.

DIALS and dxtbx and xia2 are installed (with --editable) into the `conda_base/` environment, so can be treated like any normal, installed python module. Any other (non-cctbx-only) modules can be added in the same way (`pip install -e`). build-time library files (`build/lib`) are inserted into the environment with a `.pth` file, so the conda environment can be used as a standard conda environment.

The sourcing script is only left in as a convenience for activating the conda environment.

Since sub-configured environments were always a bit hit and miss with prebuilt cctbx, no extra attempt is made to directly support them. I don't think this affects anything directly except `cctbx.xfel`, which I don't believe yet has a solution for building outside the direct libtbx build system.

Also:
  - Give libtbx.precommit a help string
  - Install mamba into fresh conda_base environments

I need a little more environment testing (linux, windows, python 2) before this is ready to merge.


ndevenish:
This now works on Windows. Probably a little more fragile, as more custom-paths. It doesn't fully support multi-config builds without manual configuration, but since that wasn't possible at all before (IIRC) that's probably okay.

ndevenish:
Oh, the windows `dials.bat` isn't written yet (I just used conda activate manually when testing, because the `.bat` is mostly redundant now)

ndevenish:
> Good stuff, I like. I wonder how much faster a trivial code change rebuild is now, since `SCons` used to take a while of "contemplation" before rebuilding things

tl;dr worst case nullop on my macbook a couple of seconds for `make`.

---

Unfortunately this answer goes slightly into details of CMake. You can customize the "Build Backend" because in the end it just generates e.g. by default a Makefile on linux or a MSBuild file on windows. The default is to generate Makefiles, which aren't super-fast:

```
% time make
...
[100%] Built target tst_collision_detection
make  1.93s user 1.47s system 95% cpu 3.557 total
```

This is one of the reasons (among other, overhead-related issues)  I use ninja:
```
% time ninja
ninja: no work to do.
ninja  0.01s user 0.05s system 98% cpu 0.060 total
```

This can be set as a systemwide preference -  installing `ninja` systemwide and settings `export CMAKE_GENERATOR=Ninja` means that bootstrap automatically uses your preferred generator (this would mean that you need to type `ninja` instead of `make` in the `build/` folder, or just use the generic `cmake --build .`. I wanted the defaults here to be as close to the normal bootstrap as possible, though.

Issue #2478, opened 2023-07-31
dials.image_viewer crashes when the detector swing angle is larger than 90


biochem-fan:
While testing my Rigaku dxtbx class, I discovered that `dials.image_viewer` cannot display images with a detector swing angle (THETA in Rigaku/Oxford's terminology but more adequately called $2 \theta$) is larger than 90 degrees.

```
...
  File "/home/user/prog/dials-dev/modules/dials/src/dials/util/image_viewer/slip_viewer/flex_image.py", line 113, in get_flex_image_multipanel
    beam_center /= npanels / 1e-3
  File "/home/user/prog/dials-dev/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in __truediv__
    return rec([e/other for e in self.elems], self.n)
  File "/home/user/prog/dials-dev/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in <listcomp>
    return rec([e/other for e in self.elems], self.n)
```

The cause is here:
https://github.com/dials/dials/blob/77db99a482231f271fc4cb6928653d067cbbfb39/src/dials/util/image_viewer/slip_viewer/flex_image.py#L103-L113

When the swing angle is larger than 90 degrees, the beam intersects with the detector "before" the crystal. This means `w` is negative in `get_ray_intersection()` and [raises an exception](https://github.com/dials/dxtbx/blob/0a068b56df0cc05cd5ea5e5ebe2f0c203e341e54/src/dxtbx/model/detector.h#L735-L737).

Thus, `npanels` becomes 0.

## How to reproduce

Download `frames/exp_864_16_1.rodhypix` from [XRDa-0094](https://xrda.pdbj.org/entry/94) and open it in `dials.image_viewer`. Note: this needs `FormatROD.py` merged into dxtbx last week.

PR #2479, opened 2023-07-31
Fix `dials.image_viewer` crash when the detector swing angle is larger than 90°

dagewa:
Two workarounds required to display images when `panel.get_beam_centre_lab` raises a `RuntimeError`, such as in the case of #2478


biochem-fan (approved):
Thank you very much!

I confirmed this fixed the problem on several HyPix datasets.

biochem-fan:
Wait. There is another related issue.
When spots are included, `_basis_vector_overlay_data()` in `spotfinder_frame.py` raises an exception.

biochem-fan:
```diff
diff --git a/src/dials/util/image_viewer/spotfinder_frame.py b/src/dials/util/image_viewer/spotfinder_frame.py
index 1937dcded..d05155e1e 100644
--- a/src/dials/util/image_viewer/spotfinder_frame.py
+++ b/src/dials/util/image_viewer/spotfinder_frame.py
@@ -1952,7 +1952,7 @@ def _basis_vector_overlay_data(self, i_expt, i_frame, experiment):
             if "DXTBX_ASSERT(w_max > 0)" in str(e):
                 # direct beam didn't hit a panel
                 panel = 0
-                beam_centre = detector[panel].get_ray_intersection(beam.get_s0())
+                beam_centre = detector[panel].get_bidirectional_ray_intersection(beam.get_s0())
             else:
                 raise
         beam_x, beam_y = detector[panel].millimeter_to_pixel(beam_centre)
```

This is a workaround but this shows very long vectors extending beyond the panel. Or should we hide the vectors?
I also don't know the intention of failing back to panel 0.

dagewa:
This reminds me that I think something is wrong with the display of vectors in general. In one case I struggled to generate a figure for a paper in which the basis vectors correctly lined up with the spots. In the end I removed them and used the "Show hkl" function instead.

With an additional change you get all 3 vectors rather than only the one that intersects the panel
```diff
$ git diff
diff --git a/src/dials/util/image_viewer/spotfinder_frame.py b/src/dials/util/image_viewer/spotfinder_frame.py
index 1937dcded..a81a88306 100644
--- a/src/dials/util/image_viewer/spotfinder_frame.py
+++ b/src/dials/util/image_viewer/spotfinder_frame.py
@@ -1952,7 +1952,7 @@ def _basis_vector_overlay_data(self, i_expt, i_frame, experiment):
             if "DXTBX_ASSERT(w_max > 0)" in str(e):
                 # direct beam didn't hit a panel
                 panel = 0
-                beam_centre = detector[panel].get_ray_intersection(beam.get_s0())
+                beam_centre = detector[panel].get_bidirectional_ray_intersection(beam.get_s0())
             else:
                 raise
         beam_x, beam_y = detector[panel].millimeter_to_pixel(beam_centre)
@@ -1970,8 +1970,8 @@ def _basis_vector_overlay_data(self, i_expt, i_frame, experiment):
                 s1 = matrix.col(beam.get_s0()) + r_phi
             panel = detector.get_panel_intersection(s1)
             if panel < 0:
-                continue
-            x, y = detector[panel].get_ray_intersection_px(s1)
+                panel = 0
+            x, y = detector[panel].get_bidirectional_ray_intersection_px(s1)
             x, y = self.map_coords(x, y, panel)
 
             vector_data.append(
```
Then setting "Basis scale" to 1 it should show the lengths of the first order reflections, but it clearly does not
image

dagewa:
I think the intention of the fall back to panel 0 is so that for cases where the direct beam falls between panels (CS-PAD for example) we can at least draw the blue cross in the panel gap by choosing one of the panel's coordinate systems (at it might as well be the first panel).

In this case the blue cross is well outside the image as a whole. I think it would be reasonable to simply avoid calculating the basis vectors in this case. However, I'm not sure how to test for this case.

biochem-fan:
> With an additional change you get all 3 vectors rather than only the one that intersects the panel

Perhaps this is because with a large $2 \theta$ we are looking at a very curved region of the Ewald sphere and a simple projection into a plane is a bad approximation.

> I think the intention of the fallback to panel 0 is so that for cases where the direct beam lands between panels (CS-PAD for example) we can at least draw the blue cross in the panel gap by choosing one of the panel's coordinate systems (at it might as well be the first panel).

I see.

> In this case the blue cross is well outside the image as a whole. I think it would be reasonable to simply avoid calculating the basis vectors in this case. However, I'm not sure how to test for this case.

How about calculating [the covex hull of all panels](https://github.com/dials/dxtbx/blob/main/src/dxtbx/model/detector.h#L691) and checking whether the direct beam is within it?

dagewa:
> How about calculating [the covex hull of all panels](https://github.com/dials/dxtbx/blob/main/src/dxtbx/model/detector.h#L691) and checking whether the direct beam is within it?

That sounds good. How about I add your change to this PR to get the image viewer working in this case, and then make a new issue regarding the basis vector display? I think there is more than one problem with it.

biochem-fan:
> That sounds good. How about I add your change to this PR to get the image viewer working in this case, and then make a new issue regarding the basis vector display? I think there is more than one problem with it.

Yes. That makes sense.

PR #2482, opened 2023-08-02
Beam model probe changes in `dials.show` tests.

dagewa:
This is the DIALS companion to https://github.com/cctbx/dxtbx/pull/647, to fix `test_show.py` for new output lines.


dagewa:
NB I have tested xia2 as well this time and a third PR over there is not necessary

Issue #2481, opened 2023-08-01
Cannot change settings in dials.recprocal_lattice_viewer when the detector swing angle is larger than 90


biochem-fan:
When the detector swing angle (THETA in Rigaku/Oxford's terminology but more adequately called $2\theta$) is larger than 90 degrees, `dials.reciprocal_lattice_viewer` cannot find the beam centre and raises exceptions.

```
Traceback (most recent call last):
  File "/home/prog/dials-dev/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 478, in OnChangeSettings
    self.parent.update_settings()
  File "/home/prog/dials-dev/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 179, in update_settings
    self.set_beam_centre(self.settings.beam_centre_panel, self.settings.beam_centre)
  File "/home/prog/dials-dev/modules/dials/src/dials/util/reciprocal_lattice/__init__.py", line 293, in set_beam_centre
    raise ValueError("No detector intersection")
ValueError: No detector intersection

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/prog/dials-dev/modules/dials/src/dials/util/reciprocal_lattice/viewer.py", line 483, in OnChangeSettings
    self.beam_fast_ctrl.SetValue(old_beam_centre[0])
TypeError: 'NoneType' object is not subscriptable
```

Because of this, we cannot toggle what to plot (e.g. `all`, `indexed`, `unindexed` etc). Other toggles (e.g. `Show reciprocal cell`) work but are reflected only after forcing redraw by dragging the image by mouse.

(Another issue: when there are multiple experiments, which experiment do `rotation axis`, `beam vector` etc refer to?)

This is conceptually related to #2478.

## How to reproduce

Download `frames/exp_864_16_*.rodhypix` from [XRDa-0094](https://xrda.pdbj.org/entry/94), zero pad file names, import, spot find and index. Start `dials.reciprocal_lattice_viewer indexed.{expt,refl}` and tweak toggles.

## Potential solution

How about just `pass`ing, instead of raising an exception?
https://github.com/dials/dials/blob/9e7f4dc20e4d12fce6ac5af2f381e87a0942e8fe/src/dials/util/reciprocal_lattice/__init__.py#L289-L296

dagewa:
I'm seeing different behaviour, but it is also bad (dials and dxtbx both on main).
```
dials.import renumbered/exp_864_16_0*
[...]

PR #2483, opened 2023-08-02
`dials.rlv` - handle cases where the beam centre cannot be set

dagewa:
Previously when a beam centre could not be calculated, `dials.reciprocal_lattice_viewer` would raise errors and may even distort the lattice (see https://github.com/dials/dials/issues/2481). The distortion was occurring because the beam centre controls defaulted to 0.0, 0.0. In some situations when setting the beam centre failed, the beam centre was then read from the control values and therefore set to 0.0, 0.0.

This PR works around by disabling the controls in situations when the beam centre cannot be set. Then, if the control is disabled, its value will not be passed back when updating the models.


biochem-fan (approved):
Thank you very much.

I confirmed that this fixed problems I reported.

PR #2489, opened 2023-08-11
Remove xfel-test-skipping for tests that no longer need xfel

ndevenish:
Since https://github.com/dials/dials/commit/867d1d6ff3537d06cb4a9dd8c39a288fa8a9ef4c, these tests no longer require the xfel module - so don't need to be skipped.


ndevenish:
Merging directly as these tests don't run on CI.

PR #2497, opened 2023-09-06
Possible bug fix for 2486

graeme-winter:
Pre-multiply the cb-op not post-multiply - I think this is correct since the `subgroup["cb_op_inp_best"]` is for the symmetry of the individual data set then the cb-op changes this to the right one for the collective.

### DIFF
--- a/newsfragments/XXX.bugfix
+++ b/newsfragments/XXX.bugfix
@@ -1,0 +1,1 @@
+``dials.cosym``: fix corner case in symmetry assignment for C2 case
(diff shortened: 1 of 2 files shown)


graeme-winter:
Yeh don't get excited. This worked on that one case but breaks other test cases. The other test cases look valid. So, much 🤔

TL;DR - I think we have a number of bugs

dwpaley:
Here's a very relevant example of ongoing confusion around this issue. https://github.com/cctbx/cctbx_project/commit/cfeb50e0cc97865feb240572533f1c7dce8ea5d5 That commit also turned out to fix some cases and break others. TLDR: Yes, bugs

ndevenish:
Other approaches being investigated now

PR #2499, opened 2023-09-19
Improve cc1/2 fit for resolution estimation on noisy data

jbeilstenedmands:
A user reported that `dials.estimate_resolution` consistently overestimates the resolution limit (as seen via `xia2-dials` and `xia2-3dii` processing), due to poor fitting of the cc1/2 curve e.g. 
https://ispyb.diamond.ac.uk/dc/visit/sw35712-2/id/11560169
https://ispyb.diamond.ac.uk/dc/visit/sw35712-2/id/11559926

The main reason is that the current code only fits to the data from resolution bins where the correlation is statistically significantly at the p=0.01 level, but this biases the fit upwards by removing the lower datapoints. Also there is no weighting to the resolution bins, so any noise at high resolution can cause issues with the fit.

The solution is to do a weighted fit on all data - this can be done by applying the fisher z-transformation to the cc1/2 values; now the standard error is symmetric and given by 1 / sqrt(n-3), where n is the number of observations used to calculate the cc1/2 value for a given resolution bin. So do the fitting of the curve in 'z-space' with proper weights and transform back to get the cc1/2 fit. Now one can include all the high res data which are appropriately weighted so that they don't skew the fit.

**Example 1 (link above):**
3dii-main
newplot(6)

dials-main
newplot(7)

3dii-after
newplot(5)

dials-after
newplot(8)

**Example 2 (link above):**
3dii-main
newplot(4)

dials-main
newplot(1)

3dii-PR
newplot(3)
dials-PR
newplot(2)


biochem-fan:
I think weighted fitting is a good idea but why raw CC1/2 oscillates so wildly in the first place?
Aren't shells too thin?

jbeilstenedmands:
Yes I feel they are too thin, 100 is the default (not sure why...), and xia2 keeps this unless it's small molecule where it gets set to 20.

jbeilstenedmands:
Note I have now reduced the default number of resolution bins from 100 to 50 as part of this PR.

jbeilstenedmands:
I was testing some high quality data and discovered that the initial change had made the fitting rather worse in those cases. For example:
Good data, this PR initially:
newplot(17)
after recent change:
newplot(18)


With the recent changes, the example fits for the noisy data now look like:

first example
3dii-PR
newplot(14)

dials-PR
newplot(15)

second example:
3dii-PR
newplot(13)
dials-PR
newplot(12)

So not perfect, but an improvement on the original. The main change is that we don't filter out datapoints below the p=0.01 threshold for the fit, while also weighting the tanh fit by a standard error.

PR #2503, opened 2023-09-20
Cosym: don't fill wij matrix where the overlap is less than min_pairs

dwpaley:
This was important when using cosym alignment for a serial dataset with a small-ish unit cell. In the current state, when a pair of experiments has 1 or 2 overlaps, that (nonzero) number goes in the weights matrix and a 0 goes in the correlations matrix. As a result the Brehm-Diederichs refinement is contaminated with bogus zero correlations. This pair of images shows the difference:
Before:
image
After:
image

This was previously flagged as a possible issue, but in practice it's only detectable when there are many barely-overlapping pairs of datasets.


dwpaley:
See cctbx_project PR https://github.com/cctbx/cctbx_project/pull/929

PR #2506, opened 2023-09-27
BugFix / Update viewer_low_level_util.py

luisodls:
resent version of WxPython requires integer scroll position


luisodls:
first fix to dials.reflection_viewer

PR #2507, opened 2023-09-27
Fix / Update bitmap_from_array.py

luisodls:
newer version of WxPython requires integer position of text


luisodls:
second fix in dials.reflection_viewer

luisodls:
Changeset merged into PR #2506

PR #2518, opened 2023-10-12
Add mosaicity_max_limit option to dials.ssx_integrate

jbeilstenedmands:
Allows control of what is considered unphysically large for the ellipsoid model mosaicity.


dagewa (approved):
Thanks! This helps for small molecule serial ED data with rather large spots.

PR #2521, opened 2023-10-16
dials.ssx_integrate: Add max_cell_volume_change_fraction parameter to catch highly divergent cell refinements

jbeilstenedmands:


dagewa (approved):
Thanks! This works for me by catching one image that otherwise got a crazy cell refinement and would then grind in processing. Now I get a useful message:
```
Processing failed due to error: Cell volume change fraction (8.9e+04) greater than 0.2 (max_cell_volume_change_fraction) in this cycle
  Initial cell volume 10611.12, final cell volume 946576257.54
```

PR #2528, opened 2023-10-18
Sourceforge really is no more

graeme-winter:
User fed back that "haven't heard of sourceforge for like 8 years"

And correctly so, propose some updates


dagewa:
The mailing list still works though, and or people without github accounts that provides an easier way to report issues. So unless we can replace dials-support@lists.sourceforge.net with some other address, my preference would be to keep that in the error messages.

phyy-nx:
Jiscmail would be fine, but keeping [dials-support@lists.sourceforge.net](mailto:dials-support@lists.sourceforge.net) is also fine

dagewa:
@phyy-nx I wanted to check, since funding situation is now different since this time last year, would you still be happy if we moved the support email address to jiscmail.ac.uk?

dagewa:
TBH I see no problem keeping the sourceforge mailing list. It works and there are [not a huge number of free alternatives](https://opensource.stackexchange.com/questions/1470/free-mailing-list-provider-for-open-source-projects) for open source projects (note xia2 uses gmail though). In that case, this PR would just need fixing up to keep the sourceforge address in error messages.

phyy-nx:
For now, let's keep it on sourceforge?  I'm interested in a larger conversation about user support spaces for DIALS, so I think even if the email should change, I'm not yet ready to say what it should change to.

phyy-nx (approved):
Approving since I have edits I need to do to add the DIALS R24

phyy-nx:
I think the test failures here are unrelated, so merging.

PR #2535, opened 2023-10-24
Ellipsoid profile model scale factor in mask calculation

dagewa:
We (@jbeilstenedmands, @jmp1985, @aimeon and myself) have been comparing the behaviour of `dials.ssx_integrate` for SSX data with SerialED data, which has a much shorter wavelength.

Here is some test data consisting of a two simulated lysozyme still images, one with ED geometry (0.02508 Å) and the other with MX geometry (1.0332 Å): [sim.zip](https://github.com/dials/dials/files/13114796/sim.zip). The images can be processed in their respective directories by running the `process.sh` script.

The output of `dials.show integrated_1.refl` for the ED case shows that no pixels are modelled as background. Dumping out reflections with shoeboxes and viewing in the `dials.image_viewer` shows this:

ED-without-fix

It appears that bounding boxes are appropriately calculated, but the mask determination has gone wrong and everything is considered foreground. After some debugging we found that the change in this PR resolves the issue. Following that change, background masking appears correct:

ED-with-fix

For the MX geometry data, this scale factor is $1/1.0332=0.9678668$, which has only a minor effect on the mask calculation. Here is an image of spot masks before the change:

MX-without-fix

and here is the image after the change. A few extra pixels are marked as peak in this case:

MX-with-fix


dagewa:
Converting to a draft PR because we don't yet fully understand this change - we just see that it appears empirically correct.

dagewa:
Thanks @jbeilstenedmands for figuring this out!

PR #2537, opened 2023-10-24
Remove duplicate debug parameter in ssx_integrate. Fixes #2536

### LINKED ISSUE #2536: Dumping shoeboxes from `dials.ssx_integrate`
During investigations for #2535 we found that the `debug.output=true` parameter for `dials.integrate` appears to have no effect for `dials.ssx_integrate`. We worked around this by hardcoding an `as_file` call on the reflections just after mask calculation, where the shoeboxes are present. However, it would be nice if the shoeboxes could be preserved using this phil parameter. On a brief look through I couldn't see where the shoeboxes were being deleted.

jbeilstenedmands:
Removes ellipsoid.debug.output.shoeboxes, which has no effect, as the program uses
debug.output.shoeboxes to control whether the shoeboxes are output into the reflection table.


dagewa (approved):
Ah, totally missed that. Thanks! Can confirm, `debug.output.shoeboxes=true` got me what I want :+1:

PR #2538, opened 2023-10-26
dials.index: fix for multi-imageset + joint_indexing=False

jbeilstenedmands:
The combination of these reveals logical errors in part of the command line script.
In particular, the imageset id is incorrectly written into the output. Then if there are multiple lattices per imageset, or you try indexing on the indexed output (therefore triggering known-orientation indexing), additional problems are revealed.

I will shortly add a comprehensive test for multi-imageset + multi-lattice to demonstrate the issues and fix.


jbeilstenedmands:
Requires https://github.com/dials/data/pull/445 for test data

PR #2539, opened 2023-10-26
Correctly count crystals and experiments when >1 sweep and >1 lattice in indexing

### LINKED ISSUE #1821: dials.index with > 1 sweep and > 1 lattice
dials.index with multple sweeps and multiple lattices - if you say max_lattices=2 will not try for more lattices (presumably it is counting _experiments_ in the code) 

Had to set max_lattices=8 to get 

```
|          0 |       22175 |           895 | 96.1%       |
|          1 |       10894 |           548 | 95.2%       |
|          2 |        9108 |           622 | 93.6%       |
|          3 |        7260 |           443 | 94.2%       |
```

Probably annoying small book keep error

jbeilstenedmands:
Fixes #1821

Thanks to those who looked into this in the issue above, I realised that for a multi-sweep (`joint_indexing=True`) case, the test for `max_lattices` should be comparing to `len(experiments.crystals())`, which returns unique crystal models, as each crystal model is shared between the different sweep. So after the first crystal is assigned, `len(experiments) == 4` but `len(experiments.crystals()) == 1`. This then also revelead a couple of places where the counting of experiments needed changing too.

I made the equivalent changes to the stills_indexer where the same code is used. I don't suspect this is usually triggered, unless there is the case of a shared crystal model between multiple images. I recall somthing also those lines when you introduced `known_orientations` as an option for stills processing, perhaps when you had data collected with multiple detectors simulatenously? I would appreciate it if someone from @phyy-nx 's  team could confirm or test this change with the xfel regression test suite, thanks!


jbeilstenedmands:
Note that this particular issue/fix is only relevant for `joint_indexing=True`. When `joint_indexing=False,` each imageset/sweep is processed in independent jobs, and therefore in the check against max_lattices within each job, `len(experiments) == len(experiments.crystals())` i.e. you are testing for up to `max_lattices` per independent sweep, which I what I understand the parameter to mean.

phyy-nx (approved):
Hi @jbeilstenedmands, when we have done shared crystal models between experiments from known_orientations, we've done that manually.  We did one pass using regular dials.stills_process to get known_orientations from the first detector, then a second pass using the known_orientations on the second detector.  The result is two separate experiment lists, which we then stitched together using a manual script, so we could run dials.refine with the shared models.

Long story short, this PR wouldn't affect that other use case.

jbeilstenedmands:
@phyy-nx  super, thanks for confirming. I will still go ahead with the changes to the stills_indexer.py file, for consistency of code with the standard indexer.py

jbeilstenedmands:
Note that this PR now needs to wait until #2545 is merged and will need be updated to account for changes therein.

PR #2545, opened 2023-11-01
Don't lose unindexed reflections in indexers

jbeilstenedmands:
In order to address #1029 , we need to ensure that all unindexed reflections are retained in the indexer classes, which is not always the case at present (as demonstrated by the added test).

The assertion I make is that we expect `unindexed_reflections.size() + refined_reflections.size()` always equals the size of the input reflection table. In our existing code, we always take the `refined_reflections` as the indexing solution, so any reflections which were indexed at first, but then are determined to be outliers/belong to a crystal that cannot be successfully refined, should be transferred to the `unindexed_reflections` table, which is what this PR does, as well as updating the indexed flag.


jbeilstenedmands:
Description of changes to `stills_indexer.py`:

The key output, `idxr.refined_reflections`, that is used in processing (`ssx_index`, `stills_process`), is unchanged.

Currently within the indexer, `self.unindexed_reflections` is set before refinement is performed, and `self.refined_reflections` is set after refinement. In the multi-lattice case, if refinement fails on the nth lattice (n>1), then neither `self.unindexed_reflections` or `self.refined_reflections` can contain the spots that were indexed in the nth lattice.
Additionally when refinement succeeds, any indexed reflections that are rejected in refinement do not get saved into `self.unindexed_reflections` or `self.refined_reflections`. Therefore by only updating `self.unindexed_reflections` after refinement, and making sure we extend the unindexed table with any reflections rejected, we can make sure that any input reflections that do not make it to `self.refined_reflections` are retained in `self.unindexed_reflections`.

dwpaley (approved):

PR #2546, opened 2023-11-04
bootstrap: Always use environment variables for build

### LINKED ISSUE #2532: Developer bootstrap.py install fails on MacOS Sonoma
More precisely, it's the C++ compiler that fails, since an identical error message pops up when running `make reconf` in the `./build` folder. (The error message is at the end.)

I have tried the following to no avail:
- installed XCode 15.0.1 (the only version compatible with Sonoma, I'm told) with command-line tools
- updated conda
- installed new MacPorts (something-something could affect compilers)
- looked through issues here to see if similar things have happened before, but searches pulled up nothing and looking item-by-item made me feel dizzy.

I'm out of ideas. Any insights? Here's the error:

```
dyld[62862]: Library not loaded: @rpath/libc++.1.dylib
  Referenced from: <AB207B6D-AE1F-3D61-BB57-4D5E628E9457> /Users/art/sw/xfel/build/.sconf_temp/conftest_e73942b1e04111dec1918a18e43ee693_0_d00030b344d40c9aae8103f4d5202a4e
  Reason: no LC_RPATH's found
sh: line 1: 62862 Abort trap: 6           .sconf_temp/conftest_e73942b1e04111dec1918a18e43ee693_0_d00030b344d40c9aae8103f4d5202a4e > ".sconf_temp/conftest_e73942b1e04111dec1918a18e43ee693_0_d00030b344d40c9aae8103f4d5202a4e.out"
RuntimeError: C++14 is not available, see config.log:
  File "/Users/art/sw/xfel/build/SConstruct", line 6:
    SConscript("libtbx/SConscript")
  File "/Users/art/sw/xfel/conda_base/lib/python3.10/site-packages/SCons/Script/SConscript.py", line 662:
    return method(*args, **kw)
  File "/Users/art/sw/xfel/conda_base/lib/python3.10/site-packages/SCons/Script/SConscript.py", line 598:
    return _SConscript(self.fs, *files, **subst_kw)
  File "/Users/art/sw/xfel/conda_base/lib/python3.10/site-packages/SCons/Script/SConscript.py", line 285:
    exec(compile(scriptdata, scriptname, 'exec'), call_stack[-1].globals)
  File "/Users/art/sw/xfel/modules/cctbx_project/libtbx/SConscript", line 1197:
    raise RuntimeError("{} is not available, see config.log".format(
```

ndevenish:
This fixes a case where libtbx failed to configure with conda compilers on MacOS 14 (Sonoma). Since the build system doesn't attempt to work out what fundamental flags are needed to do basic compiler checks itself, defer this decision to the configured environment directly.

In most cases using DIALS bootstrap, this will end up being just the flags from the configured conda environment.

This also has the side-effect of better aligning the built DIALS objects between CMake and the libtbx SCons variant builds.

Fixes #2532.


ndevenish:
This could potentially cause problems if users have so thoroughly misconfigured systems that even after loading the conda environment builds are broken.

I think we'll only be able to validate this by releasing into the wild.

PR #2548, opened 2023-11-06
Attempted fix for cosym reindex bug #2486

### LINKED ISSUE #2486: dials.cosym finds the right operator but puts it wrongly
Reprocessing raw data for 5AUI (https://xrda.pdbj.org/entry/5aui/path/TeFd),
```
dials.import /data/xrda/publicdir/5aui/TeFd/collect_1_00*.cbf
dials.find_spots nproc=24 imported.expt 
dials.refine indexed.{expt,refl} scan_varying=True
# This gave
#    Unit cell: 32.156(2), 38.719(3), 38.680(3), 86.8902(16), 88.4272(15), 88.2662(14)
#    Space group: P 1
dials.integrate refined.{expt,refl} nproc=16
dials.cosym integrated.{expt,refl}
```
`dials.cosym` rightly found the C-centering and the 2-fold.

```
+--------------+--------+------+-----+-----------------+
|   likelihood |   Z-CC |   CC |     | Operator        |
|--------------+--------+------+-----+-----------------|
|        0.092 |   2.28 | 0.23 |     | 4 |(1, 0, 0)    |
|        0.087 |   1.99 | 0.2  |     | 4^-1 |(1, 0, 0) |
|        0.092 |   2.3  | 0.23 |     | 2 |(1, 0, 0)    |
|        0.088 |   2.07 | 0.21 |     | 2 |(0, 1, 0)    |
|        0.09  |   2.18 | 0.22 |     | 2 |(0, 0, 1)    |
|        0.092 |   2.29 | 0.23 |     | 2 |(0, 1, 1)    |
|        0.946 |   9.9  | 0.99 | *** | 2 |(0, -1, 1)   |
+--------------+--------+------+-----+-----------------+
Scoring all possible sub-groups
+-------------------+-----+--------------+----------+--------+--------+---------+--------------------+
| Patterson group   |     |   Likelihood |   NetZcc |   Zcc+ |   Zcc- |   delta | Reindex operator   |
|-------------------+-----+--------------+----------+--------+--------+---------+--------------------|
| C 1 2/m 1         | *** |        0.917 |     7.72 |   9.9  |   2.19 |     0.1 | -b-c,b-c,a         |
| P -1              |     |        0.053 |    -4.26 |   0    |   4.26 |     0   | a,b,c              |
| C m m m           |     |        0.009 |     3.88 |   6.02 |   2.13 |     2.3 | b-c,b+c,a          |
| P 1 2/m 1         |     |        0.005 |    -2.2  |   2.3  |   4.5  |     2.3 | -b,a,c             |
| C 1 2/m 1         |     |        0.005 |    -2.22 |   2.29 |   4.5  |     2.3 | b-c,b+c,a          |
| P 1 2/m 1         |     |        0.005 |    -2.33 |   2.18 |   4.51 |     3.5 | -a,c,b             |
| P 1 2/m 1         |     |        0.005 |    -2.45 |   2.07 |   4.52 |     3.4 | -a,-b,c            |
| P 4/m             |     |        0     |    -3.1  |   2.2  |   5.3  |     3.5 | b,c,a              |
| P m m m           |     |        0     |    -3.12 |   2.19 |   5.3  |     3.5 | a,b,c              |
| P 4/m m m         |     |        0     |     4.26 |   4.26 |   0    |     3.5 | b,c,a              |
+-------------------+-----+--------------+----------+--------+--------+---------+--------------------+
Best solution: C 1 2/m 1
Unit cell: (56.1942, 53.224, 32.1564, 90, 92.2774, 90)
Reindex operator: -b-c,b-c,a
```

However, the output was not properly reindexed and a wrong Bravais lattice constraint was applied:
```
$ dials.show symmetrized.expt

Crystal:
    Unit cell: 53.224, 56.194, 32.156, 90.000, 90.118, 90.000
    Space group: C 1 2 1
```

The b axis (where the 2-fold is) must be the one with 53.224 Å and the beta angle must be 92.2774.
But the output has the 2-fold along the a axis. This would be OK if the space group was "C 2 1 1" and
the angles were 92.2774 90 90 (i.e. non-conventional a-unique setting), but not.

I have encountered this problem on several other unpublished datasets (not necessarily this protein; some are small molecule datasets). This happened at least on DIALS 3.8 from CCP4, `DIALS 3.15.1-g9dd0ab42c-release` and `DIALS 3.dev.1014-g3ded4afe5`.

For this particular dataset, running `dials.refine_bravais_settings` right after indexing gives the right solution, but for other cases (e.g. small wedge datasets), we need `dials.cosym`. So this bug is critical.

jbeilstenedmands:
I've been giving this some thought, and I believe there is no issue with the fundamental cosym symmetry analysis, rather we are just not ensuring that we are outputting in the standard setting.
In some of the example cases highlighted, where the lattice symmetry is higher than the best cell (within tolerance of `lattice_symmetry_max_delta`), the cosym procedure can reindex into a valid but non-conventional cell. So we have to then recalculate what is the input-to-best cell transformation after the cosym procedure, in order to apply it correctly and recover the standard setting. I don't believe it is possible to reliably do this when choosing the seeds for the cosym reindexing.

This change correctly processes the two examples - the 5AUI example and Dan's failing-input-single, and all the cosym command line tests pass. Note I have temporarily commented out the recent change that stops cosym being run on single datasets, just for testing purposes.
I plan to add some full tests to this PR to demonstrate the fix and stop regressions. I still need to think about a few of the FIXME's in the `change_of_basis_ops_to_best_cell` function, but wanted to get this out so people can test and comment.

This has highlighted to me that we probably also want to reduce `lattice_symmetry_max_delta` (defaut 5). I think we can accurately index to a much higher tolerance, as we only want to catch datasets here which really could be misindexed. I note the default in dials.symmetry is 2, and I recently changed the default down from 2 to 0.5 in xia2.ssx for example because it was catching too many cases that didn’t need accidental-ambiguity assessment.
We could definitely do with better documentation on this confusing/complicated program too.

Fixes #2486


jbeilstenedmands:
I would appreciate @biochem-fan and @dwpaley testing this change on any interesting/tricky datasets that you have. In particular any multi-crystal monoclinic datasets with some distribution in the integrated unit cell values, maybe also increase the `lattice_symmetry_max_delta` for extra fun.
To confirm the fix on your single-dataset examples, you will need to comment out L470-L473 in `command_line/cosym.py`.

biochem-fan:
Wonderful! I tested this on another C-centered monoclinic dataset (unpublished) that caused problems before.
It seems working fine. Tomorrow I will test more. If I don't report further issues, please consider I approved your fix.

dwpaley:
This looks great. 

One comment about the procedure to determine `reindex_changes_cell`. I believe in the typical use case, this function `_apply_reindexing_operators` is working on a list of experiments that were indexed in P1, i.e. no symmetry constraints applied. The median of a bunch of non-symmetrized cells is also a non-symmetrized cell of course. That means any non-identity reindexing operator will change the cell. For clarity is it better to just proceed as if `reindex_changes_cell` is true?

biochem-fan:
> I believe in the typical use case, this function _apply_reindexing_operators is working on a list of experiments that were indexed in P1, i.e. no symmetry constraints applied.

Not necessarily. I often use cosym to sort out indexing ambiguities in small wedge datasets.

jbeilstenedmands:
@dwpaley you make an astute observation. 
At this point in `_apply_reindexing_operators`, the `input_cell` can still be symmetrized depending on the input, which can be seen by printing out the `input_cell` values for some of the test examples in `tests/command_line/test_cosym.p::test_synthetic`.
"P422", (79, 79, 37, 90, 90, 90): `input_cell` = `37.0, 79.0, 79.0, 90.0, 90.0, 90.0`
"P321", (59.39, 59.39, 28.35, 90, 90, 120): `input_cell` = `8.35, 59.39, 59.39, 60.0, 90.0, 90.0`
"C2", (56.194, 53.224, 32.156, 90.000, 92.277, 90.000): `input_cell` : `32.156, 38.699, 38.699, 86.89, 88.35, 88.35`

The original cells have been mapped to a reduced cell, which is symmetrized. The tests only use one cell value, but if there were a distribution in cell parameters but with symmetry enforced, the first two cases would still have a symmetrized median cell.
For high symmetry cells like P422 and P321, `reindex_changes_cell` is always False. For P321 and other rhombohedral symmetries, "cb_op_inp_best" is not "a,b,c", so we enter the 'elif' branch on L305.

If the input were indexed in P1 but the true cell is high symmetry....  I need to think a bit more about exactly what is going to happen....

Issue #2505, opened 2023-09-26
MTZ export with an unsupported space group setting


dagewa:
The test `test_export.py::test_mtz_primitive_cell` reindexes an I23 insulin cell into a primitive setting and then exports the MTZ. This can be reproduced at the command line like this:

```
dials.reindex $(dials.data get -q insulin_processed)/scaled.{expt,refl} change_of_basis_op='y+z,x+z,x+y'
dials.export reindexed.expt reindexed.refl
```

Using `iotbx.mtz.dump` the space group written into the MTZ is shown as follows:

```
Space group number from file: 197
Space group from matrices: I 2 3 (y+z,x+z,x+y) (No. 197)
```

While `iotbx` can read that MTZ back into memory, CCP4 software cannot understand this space group. For example, from `mtzdmp`:

```
 Failed to find spacegroup in SYMINFO! 
 MTZDUMP:  Fatal error in ccp4spg_register_by_symops
 MTZDUMP:  Fatal error in ccp4spg_register_by_symops
```

and from `pointless`:
```
Spacegroup_registry: ASU warning, LGhash=0x6eefd48e
Spacegroup_registry: ASU fail
```

Meanwhile, the `gemmi` library can read the MTZ file, but `mtz.spacegroup == None`.

My feeling is that we should be allowed to make these arbitrary basis changes during DIALS processing, but at the point of MTZ export, any unsupported setting should be automatically reindexed back to something conventional. Any thoughts?

ndevenish:
So, looking at the MTZ file directly, it seems to write:
``
SYMINF  12 12 P   197               'No.197'    23
```
So, I guess CCP4/GEMMI is getting confused by the combination of Primitive and the number for I23? If this is non-standard (which it looks like?) then we probably shouldn't be doing this, and I'm surprised we never saw it before.

wojdyr:
Gemmi tries to parse 'No.197' as space group name and fails.
CCP4 library might be smarter and it tries to determine space group from SYMM records, but it also fails because there is no such entry in `syminfo.lib`.
Pointless seems to get a bit further, these error messages come from Clipper not being able to find reciprocal space ASU.

wojdyr:
[...]

PR #2549, opened 2023-11-07
Export unmerged mtz with gemmi python

dagewa:
This supersedes #2380. With gemmi 0.6.3 everything needed for unmerged MTZ export is available using the python interface. Also, the old C++ method `add_dials_batches` that was required for performance with the iotbx interface is no longer necessary and is removed.

The gemmi interface to MTZ handling is somewhat different to iotbx, which is a low level wrapper around CMTZ, and constructs an object hierarchy that mirrors the Project -> Crystal -> Dataset -> Column hierarchy of the MTZ format. Gemmi has a much simpler flattened interface that is considerably easier to work with. It is also very well documented (https://gemmi.readthedocs.io/en/latest/hkl.html).

This PR solves a number of longstanding issues with unmerged MTZ output:
- #1099 
- #1100 
- #2057 
- #2379 
- #2505 

There are differences between the gemmi MTZ output and the old iotbx-based one. For example, with a particular comparison performed during testing:
- gemmi output writes space group symbol as `P 4` instead of `P4`
- gemmi writes point group symbol as `PG4` instead of `4`
- gemmi has `Id: 1` for the `HKL_base` `Crystal`, and puts `HKL` in there, instead of in the dataset containing all the column data
- Both versions have a single dataset with 0 columns with `Wavelength: 1.03936`
- Both versions have a second dataset with `Wavelength: 1`. This has 14 columns in the gemmi version, and 17 in the iotbx version, because here it also has H, K and L.
- The batches are identical apart from `refinement flags for cell: [0, 0, 0, 0, 0, 0]` from gemmi and `refinement flags for cell: [-1, -1, -1, -1, -1, -1]` from iotbx.

I have tested output using all 530 tabulated settings from `cctbx.sgtbx.space_group_symbol_iterator`. One concession on this branch is that we no longer allow MTZ output in non-tabulated settings. This was added in #1279 to work around the fact that the previous methods to perform a change of basis to reference setting using the iotbx MTZ interface could leave the data in an inconsistent state (#1278). This is no longer the case with this branch, so files in a non-tabulated setting are again reindexed. I consider this an improvement, because CCP4 software could anyway not read MTZ files output with non-tabulated settings.

I have run DIALS, dxtbx and xia2 tests with `pytest --regression -n auto`. I could not remember how to run full xia2 regression tests, but I think that should be done too, and if there is any other testing we can do I would welcome it.

This only changes unmerged MTZ output so far, but if this is accepted a follow up could target all other forms of MTZ output in DIALS.


dagewa:
I ran more xia2 tests with
```
pytest -n auto --regression-full
```
Three tests failed, but the same three failed on the main branch too.

PR #2550, opened 2023-11-07
Correct spelling mistakes using codespell

dagewa:
A few new ones since #1807. Also, this PR changes a couple of variable names in code, not only comments. These appear to be local in scope so the change should be harmless (and tests should pass)


dagewa:
Ignoring failure of `test_integrate.py::test_basic_threaded_integrate`

Issue #349, opened 2017-05-22
dials.predict does not take shadowing into consideration (as well as I can tell)


graeme-winter:
Should be straightforward to add, though will slow the code down (or else, I am not using it right, also possible! no obvious option to use it though)

graeme-winter:
Test:

import with

```
format {
  dynamic_shadowing = True
}
```

passed as input to import; find spots; index; predict - verified no shadowed regions visible

graeme-winter:
Ah, possibly a _thing_, handling the shadows requires that at run-time dials.predict has access to the images, and needs to have check_format=true...

graeme-winter:
With these changes in mind, 

```
Graemes-MBP-5:shadow graeme$ dials.predict experiments.json ignore=false output=noignore.pickle
The following parameters have been modified:

output = "noignore.pickle"
ignore_shadows = False
input {
  experiments = experiments.json
}

Saved 446274 reflections to noignore.pickle................................0.26s
Graemes-MBP-5:shadow graeme$ dials.predict experiments.json output=ignore.pickle
The following parameters have been modified:

output = "ignore.pickle"
input {
  experiments = experiments.json
}

Saved 471804 reflections to ignore.pickle..................................0.26s

```

now have shadows respected...

graeme-winter:
<img width="618" alt="screen shot 2017-05-23 at 10 02 09" src="https://cloud.githubusercontent.com/assets/1241716/26346690/f06e3072-3f9e-11e7-9c4a-c9bfe513df81.png">

PR #2553, opened 2023-11-13
Allow `dials.predict` to work without images

dagewa:
If we just want to predict for some geometry, without having image data, we need `check_format=False`


dagewa:
Actually setting `check_format=False` breaks the (untested) `ignore_shadows=False` option, added for #349. In 863044f3be it was noted that we really need a way to set `check_format=auto` (and the relevant issue is tracked here https://github.com/cctbx/dxtbx/issues/187).

There is currently no good way to do this other than having a fallback parser

rjgildea:
There’s a pattern/workaround used in `dials.show` to reinstantiate the experiment object with `check_format=True` when required for options that need access to the image:
https://github.com/dials/dials/blob/189fe5a7d33446dce926cb2b0f4e355869c35c15/src/dials/command_line/show.py#L322

dagewa:
Oh my, there's a feeling of déjà vu... Yes, I'll try this. Thanks!

dagewa:
@graeme-winter added you for review as this touches on the shadow filtering stuff that you added. I think this needs a test case. Can you suggest a good one? (otherwise I guess I can use almost anything just to exercise the mechanics of it)

dagewa:
@graeme-winter I tried testing this with any old data set:
```
dials.predict $(dials.data get -q centroid_test_data)/indexed.expt ignore_shadows=False

ignore_shadows = False
input {
  experiments = /home/fcx32934/sw/cctbx/build/dials_data/centroid_test_data/indexed.expt
}

  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/predict.py", line 141, in <module>
    run()
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/predict.py", line 137, in run
    script.run(args)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/predict.py", line 118, in run
    shadowed = filter_shadowed_reflections(
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/shadowing/filter.py", line 19, in filter_shadowed_reflections
    shadow = masker.project_extrema(
AttributeError: Please report this error to dials-support@lists.sourceforge.net: 'NoneType' object has no attribute 'project_extrema'
```

Seems to me that the shadowing stuff you added is broken? This error is the same on main.

src/dials/command_line/predict.py
```diff
@@ -112,9 +115,17 @@ def run(self, args=None):
             predicted_all.extend(predicted)
 
         # if we are not ignoring shadows, look for reflections in the masked
-        # region, see https://github.com/dials/dials/issues/349
-
+        # region, see https://github.com/dials/dials/issues/349. In this case
+        # we need access to the image data
         if not params.ignore_shadows:
+            try:
+                experiments = ExperimentListFactory.from_json(
```
graeme-winter, line 122: This makes me think we need a better way in the API of making this change

Not suggesting that we change it _here_ just that the `ExperimentList` needs to have an `add_format()` or something

graeme-winter (approved):
Looks like a sensible change set, thank you, highlighted one area for API improvement for `ExperimentList` but that can wait for another day

dagewa:
Thanks @graeme-winter. I don't need this merged urgently, so I wonder if we should look at the failures with `ignore_shadows=False` (https://github.com/dials/dials/pull/2553#issuecomment-1920888724) in this PR, or spin it out as a separate issue?

PR #2567, opened 2023-12-14
dials.index: extend input experiment list

jbeilstenedmands:
This PR addresses the ideas discussed in  https://github.com/dials/dials/issues/1029. It adds an option to `dials.index` to allow retaining of an experiment to which the unindexed reflections can belong for each imageset (`retain_unindexed_experiments=True/False`).
This is managed through bookkeeping in `src/command_line/index.py` before calling the `Indexer` classes, so avoids any changes to the core indexing code.

To mitigate the impact of the change in output on the rest of the command-line programs, 
I have taken the approach of defining an experiment editor (`Expeditor`) class to manage the bookkeeping at the start and end of a command-line process, which allows us to use the same function calls as before, regardless of the structure of the experiment lists and reflection tables.

e.g. in `dials.refine`:
```
try:
    experiments, reflections, refiner, history = run_dials_refine(
    )
except (DialsRefineConfigError, DialsRefineRuntimeError) as e:
    sys.exit(str(e))
```
becomes
``` 
expeditor = Expeditor(experiments, [reflections])
experiments, reflections = expeditor.filter_experiments_with_crystals()
reflections = flex.reflection_table.concat(reflections)
try:
    experiments, reflections, refiner, history = run_dials_refine(
    )
except (DialsRefineConfigError, DialsRefineRuntimeError) as e:
    sys.exit(str(e))
else:
    experiments, reflections = expeditor.combine_experiments_for_output(
        experiments, [reflections]
    )
```
The first method returns the subset of experiments and reflections with crystals, and the last call recombines at the end to allow all models to be saved and therefore propagate the unindexed experiments.

This kind of change will be made to all programs which can run before data reduction. At the point of data reduction, it no longer makes sense to keep unindexed data in output files.

Draft as I have not finished updating all peripheral command line programs or added test datasets in the new structure, but I wanted to make this available for discussion.


graeme-winter:
Just for info: I suspect that this would have the side-effect of fixing some weird reciprocal lattice viewer errors about "coverage" etc. which result from imagesets being dropped.

jbeilstenedmands:
All dials.* command line programs have been updated to work with new/old style indexed experiment lists, so this is ready for testing. Unindexed experiment models are not retained after integration, so data reduction programs are unaffected.

dagewa:
I've been testing this with some moderately complicated data processing jobs involving things like multiple lattices, `dials.split_experiments`, `dials.sequence_to_stills` and so on. Mostly it looks like the results are unchanged by the PR, apart from some `dials.refine` jobs that failed. @jbeilstenedmands is looking at those.

I have a few other comments though. The most important one is that I am absolutely certain we are going to introduce some bookkeeping confusion due to the changing experiment ids without and within DIALS programs. For example, in one `dials.refine.log` I see
```
WARN: The reflections for experiment 1 do not fill the scan range. The scan will be trimmed to images {1,24} to match the range of observed data
```
But looking at `dials.show refined.expt`, the modified scan is here: 
```
...
Experiment 2:
...
Scan:
    number of images:   24
    image range:   {1,24}
```
This is just one example (and perhaps not a particularly important one), but this kind of thing happens everywhere. I can't think of a good solution other than going through every time we log an experiment and log it by identifier rather than id. Otherwise I am certain we are going to get into off-by-one hell at some point.

dagewa:
Another comment, perhaps just an overly-cautious worry: one of the motivations for this PR is that one can go back and revisit unindexed reflections using updated geometry models. We have to be careful about how to go about this, because the geometry models are shared with the already-indexed experiments. So we can't just index them alone, as refinement could then mess up the geometry for the existing models

dagewa:
I found that running `dials.split_experiments` ignores the unindexed experiments - that is they are lost by a `split`+`combine`. That makes sense, I think. Splitting necessarily loses knowledge of the relationship between experiments, so keeping them around might be dangerous. Worth a thought though - would you expect the unindexed experiment to be silently lost? Of course we also suffer from some off-by-one numbering here:
```
Saving experiment 0 to split_0.expt
Saving reflections for experiment 0 to split_0.refl
Saving experiment 1 to split_1.expt
Saving reflections for experiment 1 to split_1.refl
```
while `dials.show indexed.expt indexed.refl` shows the crystals are in experiments 1 and 2, not 0 and 1

dagewa:
Here's a bug, not caused by the PR but uncovered by it. If I run `dials.show show_shared_models=True indexed.expt` for a 2-lattice case I get this:
```
Crystal:
              0  1  2
Experiment 0  x  .  .
Experiment 1  .  x  .
Experiment 2  .  .  x
```
It looks like `Experiment 0` has a crystal, but actually it doesn't.

dagewa:
Finally, I also noticed that `dials.refine` does not print out the final crystal model (for a single lattice case) on this branch whereas it does on main. That's because of this code:
```python
    # For the usual case of refinement of one crystal, print that model for information
    crystals = experiments.crystals()
    if len(crystals) == 1:
        logger.info("")
        logger.info("Final refined crystal model:")
        logger.info(crystals[0])
```
but in the PR, `experiments.crystals()` returns something like
```
[None, <dxtbx_model_ext.Crystal object at 0x7f19d76717b0>]
```
Is this what we would expect? Perhaps there are other instances of `experiments.crystals()` in the codebase that would be similarly affected.

graeme-winter:
> This is just one example (and perhaps not a particularly important one), but this kind of thing happens everywhere. I can't think of a good solution other than going through every time we log an experiment and log it by identifier rather than id. Otherwise I am certain we are going to get into off-by-one hell at some point.

I think this is an important issue to raise thank you: maybe we need to set aside some time to review the logs etc. as a team (even, a code-camp based code review and PR merge?)

jbeilstenedmands:
> 
> Is this what we would expect? Perhaps there are other instances of `experiments.crystals()` in the codebase that would be similarly affected.

I think we only need to worry about instances in `\command_line` that could be acting on pre-integrated data (your example is valid). Deeper within the code we are still only passing experiments solely with or without crystals, for now. I am doing a review now as this was overlooked, thanks for pointing this out.

Issue #2098, opened 2022-04-26
dials.slice_sequence failure: "Input experiment list and image_ranges are not of the same length"


dagewa:
As reported on helpdesk, this fails with an error:
```
$ dials.slice_sequence integrated.expt integrated.refl image_range="1 35" image_range="36 112"
The following parameters have been modified:

image_range = 1 35
image_range = 36 112
input {
  experiments = integrated.expt
  reflections = integrated.refl
}

Traceback (most recent call last):
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/slice_sequence.py", line 250, in <module>
    run()
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/slice_sequence.py", line 246, in run
    script.run(args)
  File "/home/fcx32934/sw/cctbx/build/../modules/dials/src/dials/command_line/slice_sequence.py", line 200, in run
    sliced_experiments = slice_experiments(experiments, params.image_range)
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/util/slice.py", line 20, in slice_experiments
    raise ValueError(
ValueError: Please report this error to dials-support@lists.sourceforge.net: Input experiment list and image_ranges are not of the same length
```
The help message gives this example:
```
  # two experiments and reflections with IDs '0' and '1'
  dials.slice_sequence models.expt observations.refl     "image_range=1 20" "image_range=5 30"
```
I suppose this implies that `integrated.expt` and `integrated.refl` should have two experiments, whereas what the user was probably trying to do is produce two separate slices from a single experiment. I found the help text quite confusing, so I'm not totally sure of the intention here.

jbeilstenedmands:
Just stumbled across this, I think the best thing here would be to change to a Sorry (to avoid the traceback) and update the message to say how many experiments there are, and to suggest either multiple runs of the program, or to use dials.split_experiments.

PR #2570, opened 2023-12-19
Print helpful error message in dials.slice_sequence in case of invalid input.

jbeilstenedmands:
Also allow multi-expt files for exclude_images_multiple

FIxes https://github.com/dials/dials/issues/2098


jbeilstenedmands:
In an example case, it now prints:
```
$ dials.slice_sequence split_0.* image_range=1,20 image_range=25,40

image_range = 1 20
image_range = 25 40
input {
  experiments = split_0.expt
  reflections = split_0.refl
}

Sorry: The input experiment list and image_ranges are not of the same length (1 != 2).
To achieve multiple slices from a single sweep, the dials.slice_sequence can be run multiple times with different image_range values.
Alternatively, multi-experiment file pairs can be split with dials.split_experiments to help manage multiple-experiment workflows.
```

I also found the traceback could be triggered in the new `exclude_images_multiple` for multi-experiment input, but that case can be handled so the code has been updated.

src/dials/command_line/slice_sequence.py
```diff
@@ -275,6 +276,13 @@ def run(self, args=None):
         else:
             # slice each dataset into the requested subset
             if slice_exps:
+                if len(experiments) != len(params.image_range):
+                    raise Sorry(
+                        "The input experiment list and image_ranges are not of the same length"
+                        + f" ({len(experiments)} != {len(params.image_range)})."
+                        + "\nTo achieve multiple slices from a single sweep, the dials.slice_sequence can be run multiple times with different image_range values."
```
dagewa, line 283:
```suggestion
                        + "\nTo achieve multiple slices from a single sweep, dials.slice_sequence can be run multiple times with different image_range values."
```

dagewa (approved):
Happy that this avoids the user seeing a traceback and the error message provided is informative :+1:

Issue #2571, opened 2024-01-08
dials.reciprocal_lattice_viewer broken geometry for I23 beamline


dagewa:
As reported by @co2e14, the reciprocal space mapping is correct when files are first opened

![image](https://github.com/dials/dials/assets/11502083/2825df7b-161f-4584-a554-b4989d69dca6)

but then click on `indexed` and the geometry is distorted into a big ring

![image](https://github.com/dials/dials/assets/11502083/34935fe8-41e0-434d-be2c-13831d6279d0)

and clicking back on `all` does not recover the correct mapping

![image](https://github.com/dials/dials/assets/11502083/348e425a-2c52-4d11-8d6f-8acdc3ee0037)

PR #2572, opened 2024-01-08
dials.reciprocal_lattice_viewer: correctly record beam centre panel for multi-panel detectors

dagewa:
When models are loaded, the beam centre is calculated, and if this is succesful, the result is written back to the beam centre controls in the GUI. Unfortunately, the panel number was not written back to its control, which defaulted to `0`. This meant that whenever the relp positions were next calculated, it would do so under the assumption of the beam centre being in panel `0`. This is the cause of the errors reported with data from I23 (https://github.com/dials/dials/issues/2571).

This PR correctly updates the control when models are loaded, so that the geometry is correct when next recalculated.


dagewa:
Windows test failures are something unrelated. Merging this now to make it available for beamline scientists.

PR #2578, opened 2024-01-11
Pin Gemmi to 0.6.3

ndevenish:
We have encountered windows issues with Gemmi 0.6.4. Unfortunately, despite the version change this looks like quite a lot of things changed, so it seems easier to pin for now.


wojdyr:
Hi, what are the issues?

jbeilstenedmands:
Hi @wojdyr .
We're seeing some of our windows CI tests failing (e.g. on https://github.com/dials/dials/pull/2576)

with the error:
```Error: __setitem__(): incompatible function arguments. The following argument types are supported:\r\n    1. (self: gemmi.BatchInts, arg0: int, arg1: int) -> None\r\n\r\nInvoked with: <gemmi.BatchInts object at 0x000002192C241470>, 12, 1```

which is coming from this line in dials:
https://github.com/dials/dials/blob/82957eea3427c60f531022fc92d63604a629e993/src/dials/util/export_mtz.py#L318

This line looks very uncontroversial, so it looks like a potential issue with the python-cpp binding on the windows build? We haven't got a lot of time/resources for a full investigation of this before we would like to make a release build, hence the temporary pin, but if you have time to look into this on your windows builds then that would be really helpful, it shouldn't be too difficult to test this line of code in isolation?

**Comment by dagewa** [comment]
@wojdyr Have just confirmed that this fails on 0.6.4 from conda-forge, but works ok with 0.6.3
```
import gemmi
b=gemmi.Mtz.Batch()
b.ints[12]=1
```

dagewa (approved):
Makes sense to pin to lift the block on release :+1:

wojdyr:
OK, I'll add it to my tests and try to find out why it happens.
Quite puzzling. There were no changes to these functions after 0.6.3.

ndevenish:
(we also need to pin because we're trying to get a release out and this is blocking CI, happy to unpin later)

wojdyr:
FTR it's fixed now. When the next release (0.6.5) is out, try it out.

wojdyr:
gemmi 0.6.5 is now in conda-forge

PR #2583, opened 2024-01-22
Move the common function to a separate cmake file

namannimmo10:


namannimmo10:
(@ndevenish prefers to keep it in separate files, as it is private, hence closing)

PR #2586, opened 2024-01-23
Warn about unused `indexing.refinement_protocol` params

### LINKED ISSUE #2584: Unexpected behavior of `d_min_start` in `dials.stills_process`
I am trying to index simulated XFEL data. Each of my images features more than 1000 strong spots, some well visible up to the corner resolution of 2.5Å. In these conditions, the default orientation refinement protocol fails to refine the initial orientation matrix ~20% of the times. After some suggestions from @phyy-nx, I figured that the refinery must struggle with starting from a large amount of high-resolution data, and thus decided to set `indexing.refinement_protocol.d_min_start = 8.0`. In this resolution range I still have > 200 spots per image. After this change the indexing rate increased from 79% to 99%.

According to the `dials.stills_process` phil string, the desired behavior controlled by `indexing.refinement_protocol` for `mode=refine_shells` is to "refine in increasing resolution cutoffs after indexing". In other words, start from `d_min = d_min_start` and linearly subtract from `d_min`, either `n_macro_cycles` times or in `d_min_step` steps, until all reflections are included or `d_min_final` is reached. In this way, the orientation can be gradually improved by including more and more reflections. The final experimental geometry should still rely on all found spots:

```
  refinement_protocol {
    mode = *refine_shells repredict_only None
      .help = "refine_shells: refine in increasing resolution cutoffs after"
              "indexing. repredict_only: do not refine after indexing, just"
              "update spot predictions. None: do not refine and do not update"
              "spot predictions."
      .type = choice
      .expert_level = 1
    n_macro_cycles = 5
      .help = "Maximum number of macro cycles of refinement, reindexing all"
              "reflections using updated geometry at the beginning of each"
              "cycle. Does not apply to stills.indexer=stills."
      .type = int(value_min=1, allow_none=True)
    d_min_step = Auto
      .help = "Reduction per step in d_min for reflections to include in"
              "refinement."
      .type = float(value_min=0, allow_none=True)
    d_min_start = None
      .type = float(value_min=0, allow_none=True)
    d_min_final = None
      .help = "Do not ever include reflections below this value in refinement."
      .type = float(value_min=0, allow_none=True)
    disable_unit_cell_volume_sanity_check = False
      .help = "Disable sanity check on unrealistic increases in unit cell"
              "volume during refinement."
      .type = bool
      .expert_level = 1
  }
```

I am fairly sure that this isn't what happens at all when using `dials.stills_process`. In fact, most of mentioned parameters are completely unused. The only parameter affecting refinement is `d_min_start`, and its meaning is different than described. It is simply treated as a lower limit of d-spacing ([link](https://github.com/dials/dials/blob/7b19b5f69eb12270e3e3e3569b5dc3e5beb94d56/src/dials/algorithms/indexing/stills_indexer.py#L110)). Consequently, the output files `*indexed.refl` and `*refined.expt` rely only on the information up to `d_min_start`:

```
expt = ExperimentList.from_file('idx-0000_refined.expt', check_format=False)
refl = flex.reflection_table.from_file('idx-0000_refined.expt')
ds = expt[0].crystal.get_unit_cell().d(refl['miller_index']))
print(min(ds), max(ds))
7.986007295763752 127.12173710745984
```

Given these circumstances, I believe that the `indexing.refinement_protocol` for `dials.still_process` needs to be revised. The parameters should at least raise warnings when set, though it would be best if the `refine_shells` mode actually worked. I don't feel competent enough to offer rewriting the code myself, but I would like to start a discussion and help with finding solutions.

Baharis:
Some phil parameters remain unused when provided to `StillsIndexer`. Since there are no imminent plans to expand the capabilities of this indexer, this PR suggests increasing the verbosity of indexing phil string and logging subtle warnings when unused phil parameters are being set. Resolves #2584.

### DIFF
--- a/src/dials/algorithms/indexing/stills_indexer.py
+++ b/src/dials/algorithms/indexing/stills_indexer.py
@@ -97,7 +97,7 @@ class StillsIndexer(Indexer):
             # The stills_indexer provides its own outlier rejection
             params.refinement.reflections.outlier.algorithm = "null"
         super().__init__(reflections, experiments, params)
-        self.warn_if_setting_unused_params(params)
+        self.warn_if_setting_unused_params()
 
     def index(self):
         # most of this is the same as dials.algorithms.indexing.indexer.indexer_base.index(), with some stills
@@ -820,8 +820,7 @@ class StillsIndexer(Indexer):
 
         return ref_experiments, reflections
 
-    @staticmethod
-    def warn_if_setting_unused_params(params):
+    def warn_if_setting_unused_params(self):
         warning_message = (
             "Warning: the value of indexing.refinement_protocol.{} has been"
             " changed to {}, but this parameter is unused by stills indexer."
@@ -833,7 +832,8 @@ class StillsIndexer(Indexer):
             "disable_unit_cell_volume_sanity_check": False,
         }
         for param, default in unused_refinement_protocol_defaults.items():
-            if value := getattr(params.refinement_protocol, param) != default:
+            value = getattr(self.params.refinement_protocol, param)
+            if value != default:
                 logger.info(warning_message.format(param, str(value)))
 
 
--- a/src/dials/algorithms/indexing/stills_indexer.py
+++ b/src/dials/algorithms/indexing/stills_indexer.py
@@ -822,17 +822,19 @@ class StillsIndexer(Indexer):
 
     @staticmethod
     def warn_if_setting_unused_params(params):
-        msg = "Warning: the value of indexing.refinement_protocol.{} has been"\
-              " changed to {}, but this parameter is unused by stills indexer."
+        warning_message = (
+            "Warning: the value of indexing.refinement_protocol.{} has been"
+            " changed to {}, but this parameter is unused by stills indexer."
+        )
         unused_refinement_protocol_defaults = {
             "n_macro_cycles": 5,
             "d_min_step": libtbx.Auto,
             "d_min_final": None,
-            "disable_unit_cell_volume_sanity_check": False
+            "disable_unit_cell_volume_sanity_check": False,
         }
         for param, default in unused_refinement_protocol_defaults.items():
             if value := getattr(params.refinement_protocol, param) != default:
-                logger.info(msg.format(param, str(value)))
+                logger.info(warning_message.format(param, str(value)))
 
 
 """Mixin class definitions that override the dials indexing class methods specific to stills"""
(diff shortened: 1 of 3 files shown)


Baharis:
This rabbit hole goes deeper than expected... It looks like `dials.stills_process` does not even use `StillsIndexer`. Investigating.

Baharis:
First things first, I discovered the inconsistency with logging is not an error, but rather comes from changes to phil parameters discussed in #2587. After enabling logging in `stills_indexer.py`, I confirmed that the log messages display correctly whenever requested.

Personally, I would like to merge this PR because I believe it introduces much-needed verbosity to distinguish between the refinement protocol of both indexers. The warning messages add a layer of safety, but they can be probably seen as clutter, so I would appreciate everyone's opinion on how/if this should be included.

src/dials/algorithms/indexing/stills_indexer.py
```diff
@@ -819,6 +820,21 @@ def refine(self, experiments, reflections):
...
+    def warn_if_setting_unused_params(self):
+        warning_message = (
+            "Warning: the value of indexing.refinement_protocol.{} has been"
+            " changed to {}, but this parameter is unused by stills indexer."
+        )
+        unused_refinement_protocol_defaults = {
+            "n_macro_cycles": 5,
+            "d_min_step": libtbx.Auto,
+            "d_min_final": None,
+        }
```
phyy-nx, line 832: If the defaults change in the future, they'd need to be changed here too.  Instead of hard-coding here, could you import the un-fetched params and check them against the user params?

Baharis: I happily would if I knew how 😄

phyy-nx: It would be something like:
```
from dials.algorithms.indexing.indexer import phil_scope as defaults_phil_scope
defaults = defaults_phil_scope.extract
if self.params.refinement_protocol != defaults.indexing.refinement_protocol:
  ...
```

Baharis: Makes sense. I was looking for defaults somewhere in the parsed phil scope, but this is straightforward. Will do.

Baharis: Fixed and tested, warnings are the same, but defaults are now pulled automatically:
```
Warning: the value of indexing.refinement_protocol.n_macro_cycles has been changed to 4, but this parameter is unused by stills indexer.
Warning: the value of indexing.refinement_protocol.d_min_step has been changed to 2.0, but this parameter is unused by stills indexer.
Warning: the value of indexing.refinement_protocol.d_min_final has been changed to 2.0, but this parameter is unused by stills indexer.
```

Baharis:
It's been 2 weeks and I don't want to leave this PR hanging while I move to work on something else right now, so I'll take the liberty to squash-merge this change into main. It only increases verbosity and should not affect other functionality.

PR #2590, opened 2024-01-28
update for changes in fmodel

dcliebschner:


phyy-nx:
Merging, as this will be part of the imminent cctbx release.

PR #2592, opened 2024-01-29
Hide gemmi imports so `ImportError` is not raised unless gemmi is used

Baharis:
Ever since recent PR #2549, gemmi has been used to export unmerged mtz files. While this feels like a great move, I would like to suggest a slight change to implementation. Program `cctbx.xfel.merge`, central to the `xfel` module in `cctbx`, imports classes `MADMergedMTZWriter` and `MergedMTZWriter` for writing merged mtz files from `from dials.util.export_mtz`. However, file `export_mtz.py` includes an exposed `import gemmi` statement at the top. Consequently, running any processing within `cctbx.xfel` requires the user to install gemmi, but does not use it. *Nota bene*, gemmi 0.6.3 can not be installed in `cctbx.xfel` environment due to version clash.

I would like to suggest moving the `import gemmi` from the main scope to two functions that utilize it. In this way, no functionality is being affected, but one can still import objects from the `export_mtz` submodule.


Baharis:
@nksauter I assume you got the same error as me – this solution solves the issue for `cctbx.xfel.merge`.

ndevenish:
Sensible in principle, but we generally consider inline/hidden imports a bit of an antipattern.

Could you make it "Up-front but nullable (and possibly asserted against later)" instead? Example: https://github.com/dials/dials/blob/7b19b5f69eb12270e3e3e3569b5dc3e5beb94d56/src/dials/util/log.py#L9-L12

Baharis:
I don't think I need to prove anything here, but I find it somewhat hilarious that the only test failing (across all branches) is precisely xfel regression, as it can't solve the environment with gemmi (see [my previous comment](https://github.com/dials/dials/pull/2592#issue-2106345134)).

dagewa:
I was intending to follow up #2549 with a PR that also replaces `MADMergedMTZWriter` and `MergedMTZWriter` (and any other MTZ handling in DIALS) with gemmi-fied versions. Whereas #2549 was necessary to fix a set of longstanding bugs, this second PR would be for consistency and simplicity, to help reduce the ongoing maintenance burden of DIALS. I think this is important, but non-urgent, so I can't say how long it will be before I get on to it.

If this future PR gets merged, I suppose you would then have two choices: either accept gemmi as a dependency of cctbx.xfel.merge, or move the `iotbx` versions of those classes into the xfel code. I think the former would be preferable, because I find gemmi to be easy to use, well-documented and it is actively developed. The interface for MTZ handling is much simpler than the `iotbx` version, which IIRC is based on a frozen version of the `CMTZ` library, which is no longer developed. Most new developments within CCP4 are using gemmi (which does a whole lot more than just MTZ handling, of course).

So, I'm not against making this change as a temporary workaround, but I would like to have the conversation now about solidifying gemmi as a dependency of DIALS. What is the version clash that stops gemmi 0.6.3 from being installed in cctbx.xfel?

Baharis:
@dagewa I am not able to pinpoint the clash exactly; I captured the log from `conda install gemmi`, which mentions all the dependencies and clashes, but there is too much information for me to make anything out of it.

There is always a chance that a version of gemmi other than 0.6.3 not cause the clash. I guess there is also an argument to make here: the previous merging tool was problematic because it used a frozen version of `CMTZ`, yet the current merging essentially relies on a frozen version of `gemmi`...

I will raise the subject of mtz export tools and gemmi at the cctbx group meeting tomorrow.

ndevenish:
FWIW the mamba resolver is normally much better at telling you what the problem is

Baharis:
A peek into the azure shows that the version clash includes:

```
PackagesNotFoundError: The following packages are not available from current channels:

  - hdf5[version='<1.13.*']
  - gemmi=0.6.3.*
```

Overnight I prepared a fresh `cctbx.xfel` installation with `gemmi` which worked correctly, but pulled `gemmi` version 0.6.4, which is undesirable due to #2578. Maybe allowing version 0.6.5 will fix this issue for dials' regression tests. Cctbx's xfel regression tests do not fail at the moment due to differences in the environment and version restrictions.

Baharis:
As for the issue with gemmi, I have just modified the environment variables in `alcc-recipes` which we use to install `cctbx.xfel`. Consequently, new installations of `cctbx.xfel` should include `gemmi` (probably in version 0.6.4), so this particular issue will gradually disappear whenever folks will recreate their installation. Nevertheless I would like to push this PR to prevent any issues now, and potentially revert it later when everyone have switched to a "gemmified" installation.

dagewa (approved):
Happy with the change as a stopgap measure. I don't understand the failure, which suggests gemmi 0.6.3 is not available

PR #2603, opened 2024-02-07
Remove % from dials.index % indexed column.

toastisme:
Changes `dials.index` output from 

```
|          0 |         204 |             1 | 99.5%       |

Saving refined experiments to indexed.expt
Saving refined reflections to indexed.refl
```
to
```
|   Imageset |   # indexed |   # unindexed |   % indexed |
|          0 |         204 |             1 |        99.5 |

Saving refined experiments to indexed.expt
Saving refined reflections to indexed.refl
```

### DIFF
--- a/newsfragments/XXX.misc
+++ b/newsfragments/XXX.misc
@@ -1,0 +1,1 @@
+Remove % from value in `% indexed` column in `dials.index` output.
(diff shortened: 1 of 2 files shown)


dagewa (approved):
In the example the heading (`% indexed`) also went from left- to right-justified, but I don't see a change that does that? Could that also be added? Otherwise, yes, this is nicer.

toastisme:
> In the example the heading (`% indexed`) also went from left- to right-justified, but I don't see a change that does that? Could that also be added? Otherwise, yes, this is nicer.

From the looks of it the columns are by default right-justified and the added `%` undid that.

PR #2604, opened 2024-02-07
Pinkindexer

kmdalton:
This PR adds a fairly simple implementation of the PinkIndexer algorithm in pure numpy / scipy. This is a WIP. I'm attaching an example with strong spots and an experiment file for a small gfp data set of monochromatic still images. @phyy-nx @dagewa , it'd be helpful if you can have a look at the example and try to run the included script on this branch. 

[gfp_example.tar.gz](https://github.com/dials/dials/files/14202256/gfp_example.tar.gz)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Features
 - Circular dependencies have been removed between DIALS and ``cctbx.xfel`` by using the new ``serialtbx``. (`#2404 <https://github.com/dials/dials/issues/2404>`_)
 - ``dials.import``: Enable use of an image range selection when importing a still sequence. (`#2490 <https://github.com/dials/dials/issues/2490>`_)
 - ``dials.estimate_resolution``: Improved CC½ fitting by using a weighted tanh fit. (`#2499 <https://github.com/dials/dials/issues/2499>`_)
-- ``dials.find_spots`` and ``dials.integrate`` now accept ``exclude_images_multiple=N``, which will be expanded to the appropriate ``exclude_images=`` definition that excludes every Nth image from the data set. This is intended for the sitation where a scan of diffraction images is regularly interrupted by calibration images. (`#2511 <https://github.com/dials/dials/issues/2511>`_)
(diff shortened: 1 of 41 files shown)


dagewa:
Thanks @kmdalton! I can confirm the script runs for me and produces 94 indexed crystals from 115 input experiments. I'll experiment with some of my serial ED test cases and report back.

kmdalton:
Okay great. Full disclosure: the geometry I used to import my data was not fantastic. I would think it should be possible to index all the patterns, but I also had some fail. I'm not sure whether they fail because 
 - I should improve the PinkIndexer implementation more
 - the stills index assigner needs different hyperparameters
 - the initial geometry is too far from optimal

I think we need more examples to tease out what is going on. I appreciate your help, @dagewa.

kmdalton:
@PrinceWalnut , I made the phil string more verbose. 

@dagewa , @phyy-nx : Can I / Should I write a test for this? How do I do it?

graeme-winter:
If you do, we can add the test data to the dials data system 

Tests are always good

dagewa:
For comparison, changing `method=pink_indexer` to `method=real_space_grid_search` in your script indexes 109 images. It will be interesting to dig deeper on the cases that where one algorithm works and the other doesn't

dagewa:
There's a basic test for `low_res_spot_match` here: https://github.com/dials/dials/blob/f22597724a513c55a204567bd5a22991d82ea40c/tests/algorithms/indexing/test_index.py#L808

I guess we could start with something similar, though there might be a more appropriate still image in dials-data somewhere

kmdalton:
Note that these parameters will index 111 images on the example data:

```bash
dials.index imported.expt strong.refl \
    joint_indexing=False \
    bandwidth=2 \
    min_lattices=5 \
    method=pink_indexer \
    known_symmetry.unit_cell=51.1,62.2,69.6,90,90,90 \
    known_symmetry.space_group=19  \
    indexing.nproc=6
```

kmdalton:
I have added a small test based on this ssx dataset ([cunir_serial_processed](https://github.com/dials/data/blob/master/dials_data/definitions/cunir_serial_processed.yml))

dagewa (approved):
Thanks for adding a test! I have not yet managed to try this with serial ED data, but I don't think that should block merging this. The algorithm clearly works. I think we should get it into DIALS and take it from there

kmdalton:
@phyy-nx , do you want to have a look before I merge this?

PR #2605, opened 2024-02-08
`joint indexing=False` for stills experiments

dagewa:


src/dials/command_line/index.py
```diff
@@ -167,6 +168,12 @@ def index(experiments, reflections, params):
     if params.indexing.image_range:
         reflections = slice_reflections(reflections, params.indexing.image_range)
 
+    if params.indexing.joint_indexing is Auto:
+        if any(e.is_still() for e in experiments):
```
graeme-winter, line 172: `any` or `all`?

Assert: if `any` but `not all` and `Auto` throw an error as expected behaviour not well defined here e.g. if someone tries to index a grid and rotation scan together (which I could imagine doing)

src/dials/command_line/index.py
```diff
@@ -167,6 +168,12 @@ def index(experiments, reflections, params):
...
+            params.indexing.joint_indexing = False
```
graeme-winter, line 173: Do we alert the user anywhere that this choice is made, either way?

newsfragments/2605.bugfix
```diff
@@ -0,0 +1 @@
+``dials.index``: ``joint_indexing=False`` is set automatically for stills.
```
graeme-winter, line 1: Set joint indexing to be automatically defined by default: true for rotation data, false for still shots. This default can be overridden by the user by explicit use of `joint_indexing=`

graeme-winter (approved):
LGTM - comments I made are more about the existing ambiguity than changes you have added, but feels reasonable while we are here to do a little camsite tidying at minimal additional cost. 

I'm happy to add a commit to the stack with proposed changes, but then we need a new reviewer :)

dagewa:
Thanks, I agree with your suggestions. Happy for you to add commits :+1: Logging when `Auto` is resolved is a good idea, and a better newsfragment - I was putting this together in a spare 5 minutes in the evening :wink: 

NB there's a second check for stills experiments later where a mixture of stills and scans is disallowed. I think this is the place that @phyy-nx was referring to. However, I couldn't put the `Auto` resolution there because `joint_indexing` is only used in the command line program `dials.index` https://github.com/dials/dials/blob/fc4c37a26d28c1855aa440913177269a31ccf691/src/dials/algorithms/indexing/indexer.py#L386-L397)

kmdalton:
bump @dagewa , @graeme-winter

dagewa:
Thanks for the reminder, @kmdalton. I've added commits to address the review comments.

dagewa:
NB this adds `is_stills()` calls that should also be changed in https://github.com/dials/dials/pull/2627, once https://github.com/cctbx/dxtbx/pull/702 is merged.

dagewa:
@graeme-winter I see you already approved, but I tried to address your earlier comments. If you get a chance for a quick second look please do, otherwise I'll merge later today.

dagewa:
Test failure seems unrelated. Not sure what is going on with the CI

src/dials/command_line/index.py
graeme-winter: Now we do 👍

graeme-winter (approved):
Looks good, thanks for the additions

graeme-winter:
> @graeme-winter I see you already approved, but I tried to address your earlier comments. If you get a chance for a quick second look please do, otherwise I'll merge later today.

Thanks for picking up the comments - re-reviewed

PR #2612, opened 2024-02-21
dials.stills_process: write the diff to disk

dermen:
always leaves a debug trail


dermen:
@phyy-nx , any chance we might merge this in to master ?

phyy-nx:
Hi, the phil params are shown in the log file for `dials.stills_process`, named `dials.process.log`.  That happens [here](https://github.com/dials/dials/blob/main/src/dials/command_line/stills_process.py#L578-L581).  Is this sufficient for the phil params?

The other bits you've added are os.cwd and the command line arguments.  I'd argue os.cwd isn't that relevant because output.output_dir is more important, which is in the diff phil.  The command line arguments are folded up into the diff phil as well, except for the image file arguments (in the variable `all_paths`).  Those I could see printing out using logger.info either right before or right after the diff_phil is shown.  Probably useful to do that.

Thoughts?  My opinions aren't super strong here :)

dermen:
I can checkout the logger.info functionality (for the different logging options , e.g. disabled)

I personally like the idea of having a single file that I can use to re-launch the job, thats always there.. It kind of sucks if you forgot to write the log, and want to reproduce results.

jbeilstenedmands:
@phyy-nx @dermen Is there still appetite for this feature?

dwmoreau:
I would like this feature. Many times, I go back to reprocessing data and have to guess at what the phil parameters were. This would aid in reproducibility.

phyy-nx:
I like this change in general for the non XFEL GUI case.  The XFEL GUI already does this.  It writes params_1.phil and submit.sh, which contain the command line and the parameters used, so you always get a full reproducable record of what happened.  Maybe provide a parameter to turn this off that the XFEL GUI could use?

jbeilstenedmands:
Ok great, sounds like we should get this PR finalised, perhaps it just wants the final "o.write" line to cleanly write the parameters into a reusable phil?

phyy-nx:
I'd say add the parameter output.write_diff_phil=True, expert_level=2, then check the parameter in the new bit of code.  I'll set it to false for the XFEL GUI since it's redundant.

PR #2622, opened 2024-02-29
Use Mamba to install CI build dependencies

benjaminhwilliams:
Use Mamba in place of Conda.

### DIFF
--- a/.azure-pipelines/unix-build.yml
+++ b/.azure-pipelines/unix-build.yml
@@ -55,7 +55,7 @@ steps:
 - bash: |
     set -e
     . dials
-    conda install -y dials-data pytest-azurepipelines pytest-cov pytest-timeout
+    mamba install -y dials-data pytest-azurepipelines pytest-cov pytest-timeout
     set -ux
     dials.data info -v
     echo "##vso[task.setvariable variable=DIALS_DATA_VERSION_FULL]$(dials.data info -v | grep version.full)"
@@ -111,7 +111,7 @@ steps:
     set -e
     . dials
     rm -rf tests &  # this will be overwritten with xfel tests
-    conda install -y bash distro git-lfs mpi4py openmpi pandas
+    mamba install -y bash distro git-lfs mpi4py openmpi pandas
     cd modules
     git clone https://gitlab.com/cctbx/xfel_regression.git
     cd xfel_regression


benjaminhwilliams:
@phyy-nx, this was an attempt to get around the CI issue with running the `xfel_regression` tests in the Python 3.9 bootstrap build.  Originally, when using Conda to install the dependencies, we got [this error](https://dev.azure.com/azure-dials/dials/_build/results?buildId=13674&view=logs&j=f1969cb6-27fc-501b-a53f-82c2e07f83cb&t=048ccf09-5125-5889-fdf2-d105242652ef&l=19):
```
PackagesNotFoundError: The following packages are not available from current channels:

  - hdf5[version='<1.13.*']
  - gemmi=0.6.3.*
```
Now, using Mamba instead, we get [a different error](https://dev.azure.com/azure-dials/dials/_build/results?buildId=13676&view=logs&j=f1969cb6-27fc-501b-a53f-82c2e07f83cb&t=048ccf09-5125-5889-fdf2-d105242652ef&l=13):
```
Looking for: ['bash', 'distro', 'git-lfs', 'mpi4py', 'openmpi', 'pandas']

Pinned packages:
  - python 3.9.*

Could not solve for environment specs
The following packages are incompatible
├─ mpich is installable and it requires
│  └─ mpi 1.0 mpich, which can be installed;
└─ openmpi is not installable because it requires
   └─ mpi 1.0 openmpi, which conflicts with any installable versions previously reported.
```

I can reproduce this more simply on Ubuntu 22.04 and RHEL8 when running
```
$ python bootstrap.py update base --python 3.9
$ mamba install -dp ./conda_base openmpi
```

So it's a problem with installing Open MPI, as required by `xfel_regression`, specifically on a Python 3.9 build.  Python 3.10 and 3.11 are both OK and Python 3.12 isn't yet supported for DIALS installs.

@phyy-nx, is Open MPI actually used in the `xfel_regression` tests?

Baharis:
How does it relate to the fact that the gemmi version is soon to be unpinned and required to be >0.6.5, see #2618 ?

ndevenish:
It merely removes it from the complaint:
```
PackagesNotFoundError: The following packages are not available from current channels:

  - hdf5[version='<1.13.*']
```

Gemmi being mentioned is a red herring, it’s a side-effect of the conda resolver giving unsatisfactory results (admittedly conda is much better now, than the hundreds of pages it used to dump out).

phyy-nx:
Hi, mpi is used in lots of xfel_regression tests.  

https://github.com/dials/dials/blob/d7062646c8c999ba4bf68eaa14e2e37f901e0f8e/.azure-pipelines/unix-build.yml#L114

Could openmpi be dropped from this line?  mpi4py will bring in either openmpi or mpich, depending on which satisfies the dependency chain, and either are fine for the xfel_regression tests.

That would also sync up with how we do regular [cctbx.xfel builds](https://github.com/cctbx/cctbx_project/tree/master/xfel/conda_envs), where neither openmpi or mpich is specified:
https://github.com/cctbx/cctbx_project/blob/master/xfel/conda_envs/psana_environment.yml#L44

Issue #2608, opened 2024-02-12
Small molecule tutorial should use `dials.export format=shelx`


dagewa:
The [small molecule processing tutorial](https://dials.github.io/documentation/tutorials/small_molecule_tutorial.html#exporting) describes export to `.hkl` by going via MTZ then running
```
xia2.to_shelx scaled.mtz lcys CHNOS
```

This _should_ not be necessary now because we have `dials.export format=shelx`

TODO

- [ ] Check the tutorial works with `dials.export format=shelx`
- [ ] If so, update the text

Another problem with the tutorial is that it does not explain that `dials.symmetry` may not suggest the correct space group if there are mirrors and glides, and as a result the `dials.scale` job may not be using all the symmetry that it could.

huwjenkins:
This was discussed last year. One comment is that, unlike `xia2.to_shelx`, `format=shelx` writes an `.ins` file that only contains information known to `dials.export` i.e. the space group, unit cell parameters and their esds. 

Is adding an option to add information about the composition of the crystal desired? FWIW for solution by SHELXT I add:

```
SFAC C H etc etc
UNIT 0
HKLF 4
END
```
For SHELXD the `UNIT` line has to be correct which also means the space group must be too. For SHELXT only the Laue group matters and I almost never actually run `dials.scale` in the correct (usually non-MX) space group.

dagewa:
Good point. I think it would be useful to optionally add a composition. I usually use `edtools.make_shelx` for this (as all my small molecule work is 3D ED) but as I have discovered the electron scattering `SFAC` lines apparently hardly matter for structure solution with `shelxt`. So in that case just setting minimal things like your example, and then updating the instructions in e.g. Olex2 would do.

huwjenkins:
> as I have discovered the electron scattering SFAC lines apparently hardly matter for structure solution with shelxt.

I tested this a while ago but repeated just now. For one of our structures using the following `SFAC` in the ins file:

```
SFAC C H B N O F P Mn
UNIT 0
```
vs 
```
SFAC C 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 =
 0 0 0 0 1.000 1.000
SFAC H 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 =
 0 0 0 0 1.000 1.000
SFAC B 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 1.0000 =
[...]

PR #2623, opened 2024-02-29
dials.export: Add option to specify composition of asymmetric unit so a complete SHELX instruction file can be written.

huwjenkins:
As discussed in  #2608

This PR adds the option to specify the composition of the asymmetric unit when using `format=shelx` in `dials.export`. This will allow the `.ins` file written to be used directly for SHELXT etc. The command line option is `composition`. The string from the user is parsed using a variant of `parse_compound()` from `xia2.to_shelx`

Depending on the users input one of four things happen:

1) No input. Minimal `SFAC` and `UNIT` instructions are written:
```
SFAC C
UNIT 0
```
This is hard coded and does not call the added `_parse_compound()` function. Personally I see this as the most useful option - users can always edit the `.ins` file in Olex2/ShelXle/vi etc. 

2) A list of elements is given but each element occurs only once. The list of elements is written to the `SFAC` instruction and all zeros to the `UNIT` instruction. For SHELXT this is all that is required.

3) A formula. The list of elements is written to the `SFAC` instruction and Z*the number of times each element in the formula `UNIT` instruction. For SHELXT this is not required but I think for SHELXD it is. Obviously this requires that the space group is correct before export. 

4) A formula with unknown element - an error is raised as it seems better to correct here rather than write a potentially broken `.ins` file.

### DIFF
--- a/src/dials/util/export_shelx.py
+++ b/src/dials/util/export_shelx.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import string
 from math import isclose
 
 from cctbx import crystal, miller
@@ -96,12 +97,13 @@ def export_shelx(scaled_data, experiment_list, params):
     _write_ins(
         experiment_list,
         best_unit_cell=params.mtz.best_unit_cell,
+        composition=params.shelx.composition,
         ins_file=params.shelx.ins,
     )
     logger.info(f"Written {params.shelx.ins}")
 
 
-def _write_ins(experiment_list, best_unit_cell, ins_file):
+def _write_ins(experiment_list, best_unit_cell, composition, ins_file):
     sg = experiment_list[0].crystal.get_space_group()
     unit_cells = []
     wavelengths = []
@@ -167,3 +169,161 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
                 )
             )
         LATT_SYMM(f, sg)
+        if composition:
+            logger.info(
+                f"Using composition {composition} to write SFAC & UNIT instructions"
+            )
+            sorted_composition = _parse_compound(composition)
+            f.write("SFAC %s\n" % " ".join(sorted_composition))
+            f.write(
+                "UNIT %s\n"
+                % " ".join(
+                    [
+                        str(sorted_composition[e] * sg.order_z())
+                        for e in sorted_composition
+                    ]
+                )
+            )
+        else:
+            f.write("SFAC C\n")
+            f.write("UNIT 0\n")
+        f.write("HKLF 4\n")  # Intensities
+        f.write("END\n")
+
+
+def _parse_compound(composition):
+    """
+    Modified from parse_compund() in xia2/src/xia2/cli/to_shelx.py
+    """
+    elements = [
+        "H",
+        "He",
+        "Li",
+        "Be",
+        "B",
+        "C",
+        "N",
+        "O",
+        "F",
+        "Ne",
+        "Na",
+        "Mg",
+        "Al",
+        "Si",
+        "P",
+        "S",
+        "Cl",
+        "Ar",
+        "K",
+        "Ca",
+        "Sc",
+        "Ti",
+        "V",
+        "Cr",
+        "Mn",
+        "Fe",
+        "Co",
+        "Ni",
+        "Cu",
+        "Zn",
+        "Ga",
+        "Ge",
+        "As",
+        "Se",
+        "Br",
+        "Kr",
+        "Rb",
+        "Sr",
+        "Y",
+        "Zr",
+        "Nb",
+        "Mo",
+        "Tc",
+        "Ru",
+        "Rh",
+        "Pd",
+        "Ag",
+        "Cd",
+        "In",
+        "Sn",
+        "Sb",
+        "Te",
+        "I",
+        "Xe",
+        "Cs",
+        "Ba",
+        "La",
+        "Ce",
+        "Pr",
+        "Nd",
+        "Pm",
+        "Sm",
+        "Eu",
+        "Gd",
+        "Tb",
+        "Dy",
+        "Ho",
+        "Er",
+        "Tm",
+        "Yb",
+        "Lu",
+        "Hf",
+        "Ta",
+        "W",
+        "Re",
+        "Os",
+        "Ir",
+        "Pt",
+        "Au",
+        "Hg",
+        "Tl",
+        "Pb",
+        "Bi",
+        "Po",
+        "At",
+        "Rn",
+        "Fr",
+        "Ra",
+        "Ac",
+        "Th",
+        "Pa",
+        "U",
+        "Np",
+        "Pu",
+    ]
+    result = {}
+    element = ""
+    number = ""
+    composition += "Q"  # Not in any element symbol
+    for c in composition:
+        if c in string.ascii_uppercase:
+            if not element:
+                element += c
+                continue
+            if number == "":
+                count = 1
+            else:
+                count = int(number)
+            if element not in result:
+                if element in elements:
+                    result[element] = 0
+                else:
+                    raise Sorry(f"Element {element} in {composition} not recognised")
+            if element in elements:
+                result[element] += count
+            element = "" + c
+            number = ""
+            if c == "Q":
+                break
+        elif c in string.ascii_lowercase:
+            element += c
+        elif c in string.digits:
+            number += c
+    if "C" in result:
+        # move C to start of list
+        elements.insert(0, elements.pop(5))
+    if all(n == 1 for n in result.values()):
+        # Assume user has just supplied list of elements so UNIT instruction cannot be caluclated
+        result = result.fromkeys(result, 0)
+    sorted_result = {e: result[e] for e in elements if e in result}
+    return sorted_result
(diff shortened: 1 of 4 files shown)


dagewa (approved):
Thanks @huwjenkins, this looks great! CI failures seem unrelated.

src/dials/util/export_shelx.py
```diff
@@ -167,3 +169,161 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
...
+            f.write("SFAC C\n")
+            f.write("UNIT 0\n")
+        f.write("HKLF 4\n")  # Intensities
+        f.write("END\n")
+
+
+def _parse_compound(composition):
+    """
+    Modified from parse_compund() in xia2/src/xia2/cli/to_shelx.py
+    """
```
graeme-winter, line 197: This I am sure is already available in `cctbx` somewhere else?

graeme-winter: yes - `cctbx/eltbx/chemical_elements.h` which I assume is exported somewhere

src/dials/util/export_shelx.py
```diff
@@ -167,3 +169,161 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
...
+        "Th",
+        "Pa",
+        "U",
+        "Np",
+        "Pu",
+    ]
+    result = {}
+    element = ""
+    number = ""
+    composition += "Q"  # Not in any element symbol
```
graeme-winter, line 297: @benjaminhwilliams has suggestions here, I would not write this this way today

tests/command_line/test_export.py
```diff
@@ -584,6 +584,39 @@ def test_shelx_ins_best_unit_cell(dials_data, tmp_path):
                 assert result == pytest.approx(cell_esds[instruction], abs=0.001)
 
 
+def test_shelx_ins_composition(dials_data, tmp_path):
```
graeme-winter, line 587: Is it worth adding a test which actually executes `shelxt` if available to verify?

graeme-winter (commented):
Some suggestions which are not for this PR _per se_ since they are improvements which are needed in the source material: the implementation you copied has much room for improvement (said the author thereof)

dagewa:
@huwjenkins, I think @benjaminhwilliams might have some concrete suggestions coming soon, so hold on for now...

src/dials/util/export_shelx.py
huwjenkins: Ahh yes - thanks!
```
from cctbx.eltbx.chemical_elements import proper_caps_list
proper_caps_list()[:94]
```
Can replace the list of elements with scattering factors encoded in SHELXL included here.

dagewa:
I thikn @benjaminhwilliams was going to provide a regex-based solution? Getting string parsing right might be a bit of a pain, but I think it will be a much more satisfactory solution than using PHIL choices

benjaminhwilliams (approved):
Thanks @huwjenkins for resurrecting this old code from xia2 and tidying it up somewhat.  You'll see I have some non-blocking suggestions, but it may be that they really belong in a subsequent PR, which I'm happy to provide.  If you prefer to roll them into this PR, I'm equally happy to add some commits to this branch.  Just let me know your preference.

src/dials/util/export_shelx.py
```diff
@@ -167,3 +169,161 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
...
+            element += c
+        elif c in string.digits:
+            number += c
+    if "C" in result:
+        # move C to start of list
+        elements.insert(0, elements.pop(5))
+    if all(n == 1 for n in result.values()):
+        # Assume user has just supplied list of elements so UNIT instruction cannot be caluclated
+        result = result.fromkeys(result, 0)
+    sorted_result = {e: result[e] for e in elements if e in result}
```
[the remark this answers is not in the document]
benjaminhwilliams: Obviously, if I was being good and not just scribbling things down, I'd be using `re.compile`.  Examples are serving suggestions only.

src/dials/util/export_shelx.py
```diff
@@ -167,3 +169,161 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
...
+        "Rn",
+        "Fr",
+        "Ra",
+        "Ac",
+        "Th",
+        "Pa",
+        "U",
+        "Np",
+        "Pu",
+    ]
```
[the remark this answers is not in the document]
huwjenkins: yes - see above:
```
from cctbx.eltbx.chemical_elements import proper_caps_list
proper_caps_list()[:94]
```

dagewa:
Just a note to say it might also be worth evaluating what's already out there for handling chemical formulae. Some quick googling led me to https://pypi.org/project/chemformula/

benjaminhwilliams:
> Just a note to say it might also be worth evaluating what's already out there for handling chemical formulae. Some quick googling led me to https://pypi.org/project/chemformula/
> 
> Edit, also: https://pypi.org/project/chemparse/

Ooh, ChemFormula is nice.  I hadn't seen that before.  Scratch my review!

src/dials/util/export_shelx.py
```diff
@@ -169,161 +170,36 @@ def _write_ins(experiment_list, best_unit_cell, composition, ins_file):
...
-            element += c
-        elif c in string.digits:
-            number += c
+    elements = chemical_elements.proper_caps_list()[:94]
+    m = re.findall(r"([A-Z][a-z]?)(\d*)\s*", composition)
+    if all(e[1] == "" for e in m):
+        # Assume user has just supplied list of elements so UNIT instruction cannot be calculated
+        result = {element: 0 for element, count in m}
+    else:
+        result = {element: int(count or 1) for element, count in m}
```
benjaminhwilliams, line 195: A thought occurs to me.  Does SHELX input support non-stoichiometric formulae?  Should `int` be `float`?

src/dials/util/export_shelx.py
```diff
@@ -167,3 +170,36 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
...
+                [str(sorted_composition[e] * sg.order_z()) for e in sorted_composition]
+            )
+        )
+        f.write("HKLF 4\n")  # Intensities
+        f.write("END\n")
+
+
+def _parse_compound(composition):
+    elements = chemical_elements.proper_caps_list()[:94]
+    m = re.findall(r"([A-Z][a-z]?)(\d*)\s*", composition)
```
benjaminhwilliams, line 190: The trailing `\s*` in the regular expression is redundant in a `re.findall`.  This suggestion will parse `C6 H10 N4 O2` and `C6H10N4O2` equally well.
```suggestion
    m = re.findall(r"([A-Z][a-z]?)(\d*)", composition)
```

src/dials/util/export_shelx.py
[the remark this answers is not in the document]
benjaminhwilliams: Ah yes, sorry, I think I started drafting my review before @graeme-winter suggested that.

src/dials/util/export_shelx.py
dagewa: It seems it is not an error for `shelxl`, but it would be unconventional. I haven't found any examples of non-integer values after `UNIT` online. Wouldn't this just be represented by the occupancy column for the atom?

As for `shelxt`, it ignores what's in `UNIT` entirely.

huwjenkins: `UNIT` element count in unit cell not asu. I was multiplying the formula entered by Z. Quite common to have fractions of the formula in the asu but fractions of an atom in the _unit cell_ ?

dagewa: I think we should keep this as int. Convenient to have the composition multiplied by Z :+1:

src/dials/util/export_shelx.py
dagewa: @huwjenkins what do you think? I was wondering if you actually want to _enforce_ a space between items. Probably not, right?

huwjenkins: no that wasn't the intent. Happy for someone with better understanding like @benjaminhwilliams to push a commit to fix this :grin:

src/dials/util/export_shelx.py
huwjenkins: I think I now understand @benjaminhwilliams's comment. Take an example like CCDC 1880252. In that case I think with the current code you'd want to write 'C6 H5 B0.25 F P0.25' if you scaled the data in I-4. So maybe we should have `float` here and convert to int when writing the `UNIT` card? Or allow users to set Z (as in `_cell_formula_units_Z`) to override `space_group().order_z()`?

dagewa: Ah yes, gotcha. Allow float input and convert to int makes sense to me.

huwjenkins: Thinking some more about this probably the best approach is to *not* allow non-integer values but add a parameter `shelx.zprime` so for the example above the user would enter 'C24 H20 B F4 P' and `zprime=0.25` and get Z = 2. 

In that case should `shelx.composition` should be renamed `shelx.formula`? or should we have both - where `shelx.composition` is just element types and `shelx.formula` is always interpreted as a formula even when there no numbers.

src/dials/util/export_shelx.py
dagewa: I've committed @benjaminhwilliams's suggestion

dagewa:
The CI :x: seems unrelated. Merging this now, so any further work on "rubbish detection" and non-stoichiometric forumulae could go in a future PR

PR #2624, opened 2024-02-29
dials.export format=shelx: Increase number of decimal places in .ins file

### LINKED ISSUE #2621: dials.export format=shelx insufficient precision in esds for small molecule tutorial dataset
There is a similar issue to #2620 in `format=shelx` in `dials.export`

The `.ins` file has:

```
CELL  0.68890   5.4217   8.1333  12.0225   90.000   90.000   90.000
ZERR    4.000   0.0000   0.0000   0.0001    0.000    0.000    0.000
```
i.e. esd of a and b are lost. 

I will fix this in separate a PR to the one adding the feature request from #2608

huwjenkins:
Fixes #2621

Format of `.ins` file taken from screenshots in `XPREP` tutorials and matches the 3DED Biotin tutorial.


dagewa (approved):
:+1:

src/dials/util/export_shelx.py
```diff
@@ -156,13 +156,13 @@ def _write_ins(experiment_list, best_unit_cell, ins_file):
             f"TITL {sg.type().number()} in {sg.type().lookup_symbol().replace(' ','')}\n"
         )
         f.write(
-            "CELL {:8.5f} {:8.4f} {:8.4f} {:8.4f} {:8.3f} {:8.3f} {:8.3f}\n".format(
+            "CELL {:7.5f} {:9.5f} {:9.5f} {:9.5f} {:8.4f} {:8.4f} {:8.4f}\n".format(
```
graeme-winter, line 159: I wonder if we should check the number of digits needed to express the unit cell and derive the format accordingly?

i.e. just saying if it _does_ have a crazy unit cell like 10, 10, 1001, 90, 90, 90 we could probably drop the precision by 1dp.

graeme-winter (approved):
Looks sensible, made one suggestion which I would be happy for you to ignore, thanks for the contribution :-)

src/dials/util/export_shelx.py
huwjenkins: Probably the correct behaviour is to check the esds to determine the precision required. I don't believe any of the SHELXx programs *require* fixed-width here?

dagewa: From what I can find online it seems that `.ins` files are read in free format by most (all?) shelx programs

dagewa: Personally I think matching the XPREP output would be fine rather than adding extra complexity

dagewa:
There's a spurious :x:. Merging as-is, any future work on adaptive precisions could go in a new PR

PR #2642, opened 2024-04-10
Fix error in propagation of partiality variance for scaling still shot data

jbeilstenedmands:
The formula was missing a squaring of a prefactor.

I think the correct maths is (assuming no covariance):

CodeCogsEqn

Here, `c` is what is called `conv` in the code (comprises LP and dqe correction) and `p` is partiality.
Currently we are missing the squaring of `Iraw` and `c` in the second term in line 3, so the uncertainty was being underestimated.


src/dials/algorithms/scaling/scaling_library.py
```diff
@@ -105,10 +105,8 @@ def choose_initial_scaling_intensities(reflection_table, intensity_choice="profi
             ] * flex.pow2(conv * inverse_partiality)
             if "partiality.inv.variance" in reflection_table:
                 reflection_table["variance"] += (
-                    reflection_table["intensity.sum.value"]
-                    * conv
-                    * reflection_table["partiality.inv.variance"]
-                )
+                    (reflection_table["intensity.sum.value"] * conv) ** 2
```
benjaminhwilliams, line 108: I see that `flex.pow2` is used elsewhere.  Does that make sense here too?
```suggestion
                    flex.pow2(reflection_table["intensity.sum.value"] * conv)
```

benjaminhwilliams (approved):
Looks good to me.  One minor query but it's not a big deal.

newsfragments/2642.bugfix
```diff
@@ -0,0 +1 @@
+Fix error in propagation of partiality variance when scaling still data
```
graeme-winter, line 1: Probably I would put in the wikipedia reference you showed me

src/dials/algorithms/scaling/scaling_library.py
graeme-winter: Agree on `flex.pow2()` because it's idiomatic

graeme-winter (approved):
Thanks for the in person discussion, I agree with the motivation for the change: I ask that you reference the wikipedia page on propogation of errors in this commit somewhere

I also note, but not strongly, that this makes me feel we should have some program documentation somewhere which has equations for what the program is actually doing.

PR #2644, opened 2024-04-11
Fix RHEL8+ gltbx compilation

ndevenish:
On RHEL8 (and presumably other modern systems), gltbx compilation fails due to a interaction between the conda-forge sysroot libc++ and the local system libc++. This happens because gltbx adds the system include paths to the build (multiple times).

This code detects this situation, and resolves it. Other tasks to make this complete:

- [ ] Build DIALS bootstrap builds with the OpenGL platform-appropriate CDT packages enabled.
- [ ] Don't do this patching if not running with the DIALS bootstrap, or if the CDT isn't otherwise available.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -5,11 +5,55 @@ from libtbx.env_config import get_boost_library_with_python_version
 from pathlib import Path
 
 Import("env_etc")
+Import("env_no_includes_boost_python_ext")
+
+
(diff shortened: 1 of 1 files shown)


ndevenish:
This may only be necessary to fix existing installations

ndevenish:
yes!

PR #2649, opened 2024-04-18
XFEL regression: try using the conda-forge channel

phyy-nx:


dagewa:
This seems dumb, but might this be failing because hdf5 1.13.* does not exist on conda-forge? I see 1.14 and 1.12, which are maintenance releases, but not 1.13, which is an experimental release. I mean naively I'd think the `version='<1.13.*'` specifier should just drop down to 1.12, but perhaps it does not because 1.13 is actually a different release series?

ndevenish:
This... looks like it solved itself, again?

Issue #2609, opened 2024-02-12
Corner case where indexing loses all reflections but still writes out files


dagewa:
This is a corner case in which indexing is attempted on a small molecule data set using only two images. The first image has only very few reflections. Initial indexing is successful, but indexes only 1 reflection on the first image, and 42 on the second. After some macrocycles of refinement, eventually the model no longer indexes any reflections on the first image. At this point, refinement fails, as per #2607.

At this point we end up here and remove the reflections: https://github.com/dials/dials/blob/6ec457a50ce1f7ffabe473190e6a53f322c2b1bd/src/dials/algorithms/indexing/indexer.py#L668-L692

Because the crystal model is shared between the imagesets, this removes _all_ reflections from both. Surprisingly, `dials.index` still writes out an `indexed.expt` containing no models, and `indexed.refl` containing only unindexed reflections.

I'm not sure what the best behaviour would be in this situation, but I have quite a few examples of this occurring with this 2-image indexing test case on sparse data so it isn't completely rare. One possibility is that indexing should continue with data just from the good image. More conservatively, we could continue to fail, but make this obvious and avoid writing out `{expt,refl}` files.

Here are the files for this specific case: [2image-indexing.zip](https://github.com/dials/dials/files/14248022/2image-indexing.zip)

To run, use:
```
dials.index imported_2images.expt strong_2images.refl detector.fix=distance space_group="P21/n" unit_cell="7.72 11.40 14.23 90.00 92.85 90.00" restraint.phil
```

dagewa:
The use case here is a form of serial crystallography in which multiple images (maybe just 2, as in this case) are recorded from each crystal at different angles. This should help with indexing. However, with the current behaviour if one image is bad then indexing fails, not matter how good the other image is. I think my previous suggestions were not strong enough: we should not fail in this case, but should continue refining the solution against the single image.

This situation is detectable in `dials.index` because the refiner is being called with an experiment that has zero reflections (the reflections are not lost in outlier rejection, they are already not there right at the setup of the refiner). So, the indexer has the opportunity to remove this bad experiment.

PR #2653, opened 2024-04-23
dials.index: do not write out reflections for removed experiments

dagewa:
If a new crystal model was found to be too similar to an existing model, but only _after_ some refinement was done, then indexed reflections associated with that experiment made it into the output, and crashed dials.integrate.

This PR fixes that corner case by resetting such reflections to unindexed straight after the experiment was removed.


src/dials/algorithms/indexing/indexer.py
```diff
@@ -613,6 +613,11 @@ def index(self):
 
                 if self._check_have_similar_crystal_models(experiments):
                     have_similar_crystal_models = True
+                    try:
+                        # Unset ids for the last experiment that is now deleted
+                        self._remove_id_from_reflections(len(experiments))
+                    except AttributeError:
```
jbeilstenedmands, line 619: What is the scenario in which the above call would raise an `AttributeError`?

dagewa: I wondered if there is a circumstance in which `self.refined_reflections` does not exist because it's not actually defined until later. However I think the condition in this block is only true if there are already two crystals, in which case `self.refined_reflections` must have been created during the handling of the first one.

Maybe it would be better if `self.refined_reflections` was always created at the start of the function, even as an empty reflection table.

src/dials/algorithms/indexing/indexer.py
```diff
@@ -613,6 +613,11 @@ def index(self):
 
                 if self._check_have_similar_crystal_models(experiments):
                     have_similar_crystal_models = True
+                    try:
+                        # Unset ids for the last experiment that is now deleted
+                        self._remove_id_from_reflections(len(experiments))
```
jbeilstenedmands, line 618: I think this may be overlooking the case of shared crystal models.  A possibility is that you have four experiments and two crystals (e.g. multi-sweep indexing). I think case you need to match experiments that share the crystal to be removed, which is what is done in the other case of `_remove_id_from_reflections` below.

src/dials/algorithms/indexing/indexer.py
jbeilstenedmands: In that case i think adding in the `except AttributeError` is misleading, and we should either define self.refined_reflections at the start or add a comment to explain that this is only called once `self.refined_reflections` exists.

src/dials/algorithms/indexing/indexer.py
dagewa: I see. I think the additional shared crystal matching could be incorporated into the function so there's one place that does the removal properly. However, this is an area where I've run into trouble before (https://github.com/dials/dials/issues/2609). Maybe there are use-cases where we don't want to throw out all experiments if the crystal model is bad for just one of them. This might need more thought...

dagewa:
I attempted to include the deletion of shared crystal models when removing a crystal and reflections that are too similar to a previously-found crystal. However, this now causes a test failure in `test_multi_lattice_multi_sweep_joint`. This can also be investigated by running the command
```
dials.index $(dials.data get -q l_cysteine_dials_output)/indexed.* max_lattices=2
```
where it is seen that all experiments and reflections are removed. I have not been able to figure out why this is happening yet.

dagewa:
Ah, I think I see the problem. I did not need to include the removal of the crystal models, because this is actually done in `_check_have_similar_crystal_models` (it does more than just check...)

dagewa:
I think this addresses the comments and is ready for re-review

ndevenish:
@jbeilstenedmands do you think you can look over this again?

dagewa:
Tests pass locally

PR #2661, opened 2024-05-01
Add option to apply partiality correction for ellipsoid integration.

jbeilstenedmands:
Enables use of ellipsoid integrated output for use in cctbx.xfel.merge (if run with post-refinement disabled).

This corrects the 'intensity.sum.value' and 'intensity.sum.variance' columns by partiality, and renames the partiality column to 'partiality_applied'. By default this is not done.


dermen:
Excited to try this!

jbeilstenedmands:
I no longer think this is the right approach and will cause confusion by being inconsistent between methods, so closing to consider better approaches.

PR #2671, opened 2024-05-31
With clustering tables

amyjaynethompson:
Added clustering tables into the html output and migrated over the rest of the clustering code in xia2.multiplex. 

Also fixes a bug found in xia2.multiplex where the tables in the html listing all clusters were found to have incorrect multiplicity and completeness statistics (indices were shifted by 1, so while the listed cluster ids were correct, the statistics were calculated based on different dataset ids).

### DIFF
--- a/src/dials/algorithms/correlation/analysis.py
+++ b/src/dials/algorithms/correlation/analysis.py
@@ -125,9 +125,6 @@ class CorrelationMatrix:
 
         self.cosym_analysis = CosymAnalysis(self.datasets, self.params)
 
-        # QUERY THIS BIT WITH JAMES
-        # self.cosym_analysis.run()
-
     def _merge_intensities(self, datasets: list) -> list:
         """
         Merge intensities and elimate systematically absent reflections.
@@ -324,7 +321,7 @@ class CorrelationMatrix:
                 if merged is None:
                     merged = copy.deepcopy(d)
                 else:
-                    merged = merged.concatenate(d)
+                    merged = merged.concatenate(d, assert_is_similar_symmetry=False)
             merging = merged.merge_equivalents()
             merged_intensities = merging.array()
             multiplicities = merging.redundancies()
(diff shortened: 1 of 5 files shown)


src/dials/algorithms/correlation/analysis.py
```diff
@@ -87,9 +93,16 @@ def __init__(
...
+        # Also to retain dataset ids when used in multiplex
+        if self.ids_to_identifiers_map is None:
+            self.ids_to_identifiers_map = {}
+            for table in self._reflections:
+                self.ids_to_identifiers_map.update(table.experiment_identifiers())
+
+        self.labels = list(dict.fromkeys(self.ids_to_identifiers_map))
+        self._labels_all = flex.size_t()
+        for i in self.labels:
+            self._labels_all.extend(flex.size_t(1, i))
```
jbeilstenedmands, line 105: I think there is a slightly more efficient way to do this than n extends, perhaps `self._labels_all = flex.size_t(self.labels)` ?

jbeilstenedmands (changes requested):
Thanks for adding this, it looks good except for the couple of small changes to address. This will reduce the size of the overall code base once the xia2 PR is merged 👍

src/dials/algorithms/correlation/analysis.py
```diff
@@ -109,6 +124,8 @@ def __init__(
         self.params.lattice_group = self.datasets[0].space_group_info()
         self.params.space_group = self.datasets[0].space_group_info()
 
+        self.params.cc_weights = "sigma"
```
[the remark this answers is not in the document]
amyjaynethompson: True! This line has been removed :)

src/dials/algorithms/correlation/analysis.py
amyjaynethompson: Have changed to the suggested code - thanks for the suggestion :)

jbeilstenedmands (approved):
Thanks!

Issue #424, opened 2017-07-25
Outlier rejection in stills indexer


biochem-fan:
I often observe Sauter, Poon (2010) in `stills_indexer.py` rejects lots of innocent reflections.

The problem in the algorithm is that it assumes a single spot RMSD for all panels and all resolutions. In reality, the metrology of outer panels tends to be less accurate, and high resolution spots are more strongly affected by radial streaking due to parallax and bandwidth. Thus, high resolution spots are more likely to be rejected than low resolution spots. Once a spot is flagged as an outlier in the indexer, it will never be used, for example, in the joint refinement. Thus, panels with poor metrology receive less surviving spots and they will not be refined properly. Another problem is that it leads to a poor mosaicity model and under-prediction.

However, completely disabling the outlier rejection led to unstable refinement. Do you have any suggestions?

# Example
This is Proteinase K collected at 10 keV (similar to CXIDB 48).
Actually, this is not a good example, as there are many spots from the beginning.

## strong.pickle
![image](https://user-images.githubusercontent.com/1611483/28564987-c3e8d130-7165-11e7-9264-cd08d1ed409a.png)

## indexed.pickle
![image](https://user-images.githubusercontent.com/1611483/28565015-d54c3b74-7165-11e7-8d01-0f15c5a3fd36.png)

## integrated.pickle
![image](https://user-images.githubusercontent.com/1611483/28565070-0c858dca-7166-11e7-91f6-2ecdb1117741.png)

NOTE: pixels at edges of each panel are actually masked (not shown).

biochem-fan:
This image also shows the effects of bandpass and parallax.

The bandpass is about 1.5 % of 10 keV = 150 eV. The sensor thickness is 300 um.
The incident angle is about 45 deg at the edges.

The spot finder sometimes splits radial streaks into multiple spots.
![image](https://user-images.githubusercontent.com/1611483/28565828-4f04dc9e-7168-11e7-8276-690f9bd4aede.png)
(rotated 90 deg)

phyy-nx:
Hi, the real issue here is that the stills indexer always does a round of
Sauter Poon outlier rejection algorithm while refining candidate basis
sets.  Specifically, after refinement of a candidate, we re-run outlier
rejection using Sauter Poon.  It sounds like you need a way to disable this
behavior, so that you can
specify refinement.reflections.outlier.algorithm=None and have it be
respected by the stills indexer.  I agree.  That makes a lot of sense for
[...]

PR #2676, opened 2024-06-07
Add MCD outlier rejection to stills

phyy-nx:
This is a reworking of outlier rejection for indexing and refinement for stills, including in dials.stills_process and the stills indexer.  The stills indexer has used sauter_poon since its inception, but was never really the right answer for the problem at it assumes it is removing non-systematic error.  However, still images from e.g. XFEL data come from a wider bandpass and therefore have a systematic radial displacement compared to transverse.  This is visible in detector residuals in the deltaXY/deltaPsi plot:

image

```
export TEST_DATA=`dials.data get -q iterative_cspad_refinement`
cctbx.xfel.detector_residuals $TEST_DATA/cspad_ref* colormap=RdYlBu_r tag=cspad_refined delta_scalar=25 
```

In this plot the displacement of the prediction from the observation of every spot is shown relative to the center of the panel, exaggerated by a factor of 25, with the color being the deltaPsi value.  This is typical of XFEL data.

sauter_poon works great for monochromatic systems, but here it truncates out the caps of the data, because those spots are too widely displaced from their predictions:

```
dials.refine $TEST_DATA/refine.phil $TEST_DATA/cspad_ref*
cctbx.xfel.detector_residuals refined.* colormap=RdYlBu_r tag=cspad_refined2 delta_scalar=25 
```

Output:
```
Detecting centroid outliers using the SauterPoon algorithm
16036 reflections have been flagged as outliers
90465 reflections remain in the manager

```

image

The new code allows refining using MCD, adding new coordinates for fitting an ellipsoid that better fits serial data with a band pass and a flat detector:

```
dials.refine $TEST_DATA/refine.phil $TEST_DATA/cspad_ref* refinement.reflections.outlier.algorithm=mcd mcd.threshold_probability=0.999 mcd.coordinates=deltatt_transverse
cctbx.xfel.detector_residuals refined.* colormap=RdYlBu_r tag=cspad_refined3 delta_scalar=25 
```

Output:
```
Detecting centroid outliers using the MCD algorithm
10374 reflections have been flagged as outliers
96127 reflections remain in the manager
```

image

So fewer outliers are rejected, and the remaining reflections better fit the expected distribution.

The choice of MCD coordinates is now available through the new parameter: `coordinates = *xy radial_transverse deltatt_transverse`:
- xy: this is the current default: x_resid and y_resid.
- radial_transverse: this uses the radial displacement and transverse displacements of predictions from the observations, as computed in lab space relative to the beam vector.  This puts the data in a common frame, independent of the azimuthal angle of the reflection around the beam, as the radial component is always pointed at the beam center and the transverse is orthogonal to that vector.  The columns are named r_resid and t_resid.
- deltatt_transverse: this uses delta two theta instead of the radial displacement for one of the coordinates, keeping the transverse displacement.  This compacts the ellipsoid at high resolution, eliminating a resolution dependent effect from parallax that arises from using a flat detector.  The column is named twotheta_resid.

Additionally, deltaPsi is added as a coordinate for stills using mcd, formally it was only x_resid and y_resid.  This is the only change that could affect existing stills processing using mcd already without using the new coordinates parameter.

This PR also eliminates the assertions that sauter_poon is being used in the stills indexer, which means the default for dials.stills_process and for dials.ssx_index needs to be changed from null to sauter_poon.  Updating them further to be mcd should likely happen as well, after further review of these changes.

More detail on the changes:

- Add helper function compute_radial_and_transverse_residuals, adapted from cctbx.xfel.detector_residuals
- Add delpsical.rad as a data dimension for mcd for stills
- Add parameter mcd.coordinates = *xy radial_transverse, which chooses between x_resid/y_resid and r_resid/t_resid
- Use is_still when picking outlier algorithm from auto
- Allow stills_process and ssx_index to chose non-sauter_poon options, which means also updating their defaults from null to sauter_poon.
- Add alternate mcd coordinate axis: delta two theta

### DIFF
--- a/src/dials/algorithms/indexing/stills_indexer.py
+++ b/src/dials/algorithms/indexing/stills_indexer.py
@@ -64,13 +64,11 @@ def plot_displacements(reflections, predictions, experiments):
 
 
 def e_refine(params, experiments, reflections, graph_verbose=False):
-    # Stills-specific parameters we always want
-    assert params.refinement.reflections.outlier.algorithm in (
-        None,
-        "null",
-    ), "Cannot index, set refinement.reflections.outlier.algorithm=null"  # we do our own outlier rejection
-
     from dials.algorithms.refinement.refiner import RefinerFactory
+
+    outlier_algorithm = params.refinement.reflections.outlier.algorithm
+    if outlier_algorithm == "sauter_poon":
+        params.refinement.reflections.outlier.algorithm = "null"
 
     refiner = RefinerFactory.from_parameters_data_experiments(
         params, reflections, experiments
@@ -78,8 +76,7 @@ def e_refine(params, experiments, reflections, graph_verbose=False):
 
     refiner.run()
 
-    ref_sel = refiner.selection_used_for_refinement()
-    assert ref_sel.count(True) == len(reflections)
+    params.refinement.reflections.outlier.algorithm = outlier_algorithm
 
     if not graph_verbose:
         return refiner
@@ -95,8 +92,7 @@ class StillsIndexer(Indexer):
 
     def __init__(self, reflections, experiments, params=None):
         if params.refinement.reflections.outlier.algorithm in ("auto", libtbx.Auto):
-            # The stills_indexer provides its own outlier rejection
-            params.refinement.reflections.outlier.algorithm = "null"
+            params.refinement.reflections.outlier.algorithm = "sauter_poon"
         super().__init__(reflections, experiments, params)
         self.warn_if_setting_unused_refinement_protocol_params()
 
@@ -200,7 +196,6 @@ class StillsIndexer(Indexer):
                 isoform_experiments = ExperimentList()
                 isoform_reflections = flex.reflection_table()
                 # Note, changes to params after initial indexing. Cannot use tie to target when fixing the unit cell.
-                self.all_params.refinement.reflections.outlier.algorithm = "null"
                 self.all_params.refinement.parameterisation.crystal.fix = "cell"
                 self.all_params.refinement.parameterisation.crystal.unit_cell.restraints.tie_to_target = (
                     []
@@ -702,6 +697,9 @@ class StillsIndexer(Indexer):
         refiner = e_refine(params, experiments, indexed, graph_verbose=False)
 
         RR = refiner.predict_for_reflection_table(indexed)
+
+        if params.refinement.reflections.outlier.algorithm != "sauter_poon":
+            return refiner.selection_used_for_refinement()
 
         px_sz = experiments[0].detector[0].get_pixel_size()
 
--- a/src/dials/algorithms/indexing/indexer.py
+++ b/src/dials/algorithms/indexing/indexer.py
@@ -330,7 +330,11 @@ class Indexer:
             "auto",
             libtbx.Auto,
         ):
-            if self.experiments[0].goniometer is None:
+            if (
+                self.experiments[0].goniometer is None
+                or self.experiments[0].scan is None
+                or self.experiments[0].scan.is_still()
+            ):
                 self.all_params.refinement.reflections.outlier.algorithm = "sauter_poon"
             else:
                 # different default to dials.refine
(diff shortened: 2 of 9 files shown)


phyy-nx:
@jbeilstenedmands there are some test failures that I will be looking at, so this isn't quite ready to merge, but I wanted to get it out there.

biochem-fan:
@phyy-nx
Thank you very much for this work.
This resolves my long-standing concern https://github.com/dials/dials/issues/424 about the Sauter-Poon algorithm (initially posted in 2017).

phyy-nx:
> @phyy-nx Thank you very much for this work. This resolves my long-standing concern #424 about the Sauter-Poon algorithm (initially posted in 2017).

Thanks @biochem-fan!  We have been testing it over the last few days at the beamline.  I think the deltapsi component needs to come out as it is affecting the mosaic estimation.  When separate_experiments=True, MCD is removing low resolution reflections that have high deltapsi.  So not including deltapsi as a coordinate in the MCD ellipsoid might be better for processing individual stills.  

More as it develops :)

dagewa:
@nksauter That's an interesting idea. I think that's the quantity we call $\epsilon$, the excitation error. That reminds me, if you pass `spherical_relp_model=True` to `dials.refine`, then minimisation is done using $\epsilon$ as the third dimension rather than $\Delta \psi$. In such a case, the MCD algorithm ought also to be using that value for outlier rejection.

I don't think we ever did a proper assessment to work out whether `spherical_relp_model=True` gives better or worse results than using $\Delta \psi$ in refinement. That would be interesting to do.

dagewa:
Looking back, I realised what I wrote there is not correct. With `spherical_relp_model=True` we don't actually minimise against $\epsilon$. We still use $\Delta psi$ as the closeness-to-Ewald sphere restraint. But what we do is predict the reflection positions assuming a spherical relp ("Nave 3") model. So the predicted reflection positions are different, and the gradients going into refinement are different, but otherwise refinement proceeds similarly.

However, I'm generally interested in considering excitation error as data for things like outlier rejection and potentially for minimisation in some circumstances. The concept is used in other software (like PETS) for things like a cutoff to determine which reflections should be considered to be simultaneously excited. It may be worth calculating $\epsilon$ anyway whenever we do stills prediction, and returning it to the reflection table alongside $\Delta \psi$.

src/dials/algorithms/indexing/stills_indexer.py
```diff
@@ -64,27 +64,19 @@ def plot_displacements(reflections, predictions, experiments):
...
-            "null",
-        )
-    ), (
-        "Cannot index, set refinement.reflections.outlier.algorithm=null"
-    )  # we do our own outlier rejection
-
     from dials.algorithms.refinement.refiner import RefinerFactory
 
+    outlier_algorithm = params.refinement.reflections.outlier.algorithm
+    if outlier_algorithm == "sauter_poon":
```
jbeilstenedmands, line 70: I notice that this means with mcd we can now have rejection during refinement, good, however this, along with the new L696, seems to cause the final returned `refined_reflections` table to include these rejected reflections which have xyzcal.mm set to (0,0,0), which then results in some very big rmsds being reported in `dials.ssx_index` for those crystals where reflections were rejected in the refiner. I think we some extra flags or filtering needs to be done, but I haven't worked out exactly where yet, any suggestions?

src/dials/algorithms/indexing/stills_indexer.py
```diff
@@ -702,6 +692,9 @@ def identify_outliers(self, params, experiments, indexed):
 
         RR = refiner.predict_for_reflection_table(indexed)
 
+        if params.refinement.reflections.outlier.algorithm != "sauter_poon":
+            return refiner.selection_used_for_refinement()
```
jbeilstenedmands, line 696: Does this selection include any reflections that were subsequently filtered during refinement?

DHekstra:
I'd love to see this be part of DIALS -- this could be quite helpful in improving outlier rejection in pink-beam data analysis where the same challenge raises its head strongly!  Let us know if we can be of help.

phyy-nx:
Test failure due to also needing cctbx/cctbx_project#1106

I'm almost ready to merge this.  While I've been staring at distributions of indexed reflections in radial, transverse, and deltapsi space for the last week, I've not really seen convincing improvement from using mcd yet in terms of merged data set improvements, but it definitely improves the shape of those distributions.  So since this PR doesn't change existing behavior (at least according to my testing), and since it just adds options, I think it's ok as is.

PR #2677, opened 2024-06-13
Add option to save reflection table to a HDF5 file format

jbeilstenedmands:
This PR adds the means to save a reflection table into a working H5 format.

The motivation is that this enables data-files that could be written (or read) by programs outside of DIALS, e.g. a spot-finding service. HDF5 data files would also allow us to implement more flexible file reading in future (e.g. only load data being used), whereas currently the full reflection table needs to be loaded into memory at the start of a DIALS process.

This PR is not changing the default to use a h5 format, saving to h5 is enabled by setting the environment variable `DIALS_USE_H5`

The data are saved and loaded in the usual way:
`reflections.as_file("data.refl")`
`reflections = flex.reflection_table.from_file("data.refl")`

Each reflection table column is stored as a 1D or multidimensional array, converted to `numpy` using `dxtbx.flumpy`.
If present, shoeboxes are stored as a set of arrays in a subgroup, e.g. the shoebox data, background and mask are stored as a length N array, where N is the number of pixels in all the shoeboxes.

Example file contents:
h5refl

I will add some comparison read/write times shortly.


jbeilstenedmands:
Example timings (averaged over 30 trials):

1.35GB `scaled.refl` file (no shoeboxes):
```
Time to save msgpack = 1.305s
Time to load msgpack = 1.524s
Time to save h5 = 0.679s
Time to load h5 = 0.824s
```

38MB `strong.refl` file (containing shoeboxes):
```
Time to save msgpack = 0.067s
Time to load msgpack = 0.050s
Time to save h5 = 0.066s
Time to load h5 = 0.117s
```

29MB `indexed.refl` (no shoeboxes):
```
Time to save msgpack = 0.039s
Time to load msgpack = 0.009s
Time to save h5 = 0.031s
Time to load h5 = 0.027s
```

420MB `integrated.refl` (containing shoeboxes):
```
Time to save msgpack = 0.662s
Time to load msgpack = 0.262s
Time to save h5 = 0.316s
Time to load h5 = 0.531s
```

jbeilstenedmands:
While not immediately relevant, the encoders/decoders support reading data that has been stored into a hdf5 group per experiment, which could be a useful model when working on large multi-crystal datasets, where we do calculations on individual datasets but have to join/split the data into a single table when saving/loading.

As a demonstration, I profiled the memory usage for "example" multi-crystal operations on a 3-sweep dataset (1.35GB on disk). The operations are load, split_by_experiment_id, join (with extend) and save, e.g. this process:

```
data = flex.reflection_table.from_file("scaled.refl")
split = data.split_by_experiment_id()
joint = split[0]
for d in split[1:]:
    joint.extend(d)
joint.as_file("test.refl")
```
Then, as a second test, instead of joining for save, save the tables directly as separate groups in the h5 file:
```
with HDF5TableFile("test.h5", "w") as f:
    f.add_tables(split)
```
And as a third test, replace the read with the loader too:
```
with HDF5TableFile("test.h5", "r") as f:
    split = f.get_tables()
```

Here are the three memory traces:
memory
I added a couple of 5 second gaps between load, split and join/save for clarity. There is a spike in memory at the start when loading from msgpack (red & blue), in the middle when calling split_by_experiment_id (red & blue), and at the end when joining for save (red). Extracting and saving the already-separated tables in h5 (black) avoids these spikes.

src/dials/util/table_as_hdf5_file.py
```diff
@@ -0,0 +1,271 @@
...
+
+class ReflectionListDecoder(object):
+    """Decoder for the reflection data."""
+
+    @staticmethod
+    def decode(handle: h5py.File) -> List[flex.reflection_table]:
+        """Decode the data to a list of reflection tables."""
+
+        # Get the group containing the reflection data
+        g = handle["entry/data_processing"]
```
graeme-winter, line 129: I don't think this should go under `entry/data_processing` since that is a distraction - this is _our_ reflection format not something compliant with e.g. NeXus. We _do_ care about what steps this came from as that impacts the provenance of the data so:

```
/dials/find
/dlals/index
```

_may_ be a more useful group to put things under. We can then include links so that all the "columns" you would expect to find under e.g. `index` are present even if they refer back to results from spot finding.

src/dials/util/table_as_hdf5_file.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Create the list of reflection tables
+        tables = []
+        for dataset in g.values():
+            if "num_reflections" not in dataset.attrs:
```
graeme-winter, line 134: What if the `num_reflection` differs from the dataset shape? Can we not just get this from the shape of the HDF5 data sets?

tests/array_family/test_reflection_table.py
```diff
@@ -1149,6 +1150,40 @@ def compare(a, b):
     table["col9"] = flex.int6(c9)
     table["col10"] = flex.miller_index(c10)
     table["col11"] = flex.shoebox(c11)
+    return table, [c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11]
+
+
+def test_to_from_h5(tmp_path):
```
graeme-winter, line 1156: `test_to_from_hdf5`

graeme-winter (approved):
Ok, initial comments:

- we don't need the `/entry/data_processing` because this is dials formats
- we should use `/dials/index` or `/dials/refine` etc.
- will allow us in future to use links internally to update files with new data
- change h5 to hdf5 in file signatures

More generally this looks sane, thank you, but I would get the data model right from _now_ i.e. recognising that we may want to have file be both the input and output to dials steps and have it be updated with new values.

src/dials/array_family/flex_ext.py
```diff
@@ -231,6 +232,8 @@ def as_file(self, filename):
         """
         if os.getenv("DIALS_USE_PICKLE"):
             self.as_pickle(filename)
+        elif os.getenv("DIALS_USE_H5"):
```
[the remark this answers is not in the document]
jbeilstenedmands: This was so we can merge into the code base but not use as the default option straight away.

src/dials/util/table_as_hdf5_file.py
jbeilstenedmands: Yes to /dials/ , to include `find` or `index` would not really be possible in the current form of the function call `flex.reflection_table.as_file(filename)` without some complex introspection logic. So maybe `dials/processing` at the moment. Or would you like an extra optional argument for `as_file` which would be a string to use for the name of the second level in the tree, but is only used when saving in hdf5 format?

dagewa:
What does this do with the reflection table flags? I could not immediately see from the diff

jbeilstenedmands:
> What does this do with the reflection table flags? I could not immediately see from the diff

@dagewa The reflection table flags don't need any special treatment, the flags column is stored as a 32-bit int array.

ndevenish:
@phyy-nx with this PR the code is present, but not default (needing the `DIALS_USE_H5` environment variable set). Would you object to us merging this still WIP so that we can use internally and start looking for problems?

We won't further turn on by default without extended discussions and consensus.

src/dials/array_family/flex_ext.py
```diff
@@ -244,7 +247,12 @@ def from_file(filename):
                 filename
             )
         except RuntimeError:
-            return dials_array_family_flex_ext.reflection_table.from_pickle(filename)
+            try:
+                return dials_array_family_flex_ext.reflection_table.from_h5(filename)
+            except OSError:
```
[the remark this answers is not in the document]
jbeilstenedmands: If you try to open non-hdf5 file with h5py, you get:
```OSError: Unable to open file (file signature not found)```

File: src/dials/util/table_as_hdf5_file.py (hunk 2 above)
Comment (jbeilstenedmands) [reply src/dials/util/table_as_hdf5_file.py]: Agreed, removed this attribute

File: src/dials/util/table_as_hdf5_file.py (hunk 1 above)
Comment (jbeilstenedmands) [reply src/dials/util/table_as_hdf5_file.py]: I have changed the top level name to `/dials`. I have left the second level as a generic `/processing` for now, but removed any requirements on the name at the second level, so we could add naming based on process in future.

**Comment by jbeilstenedmands** [comment]
Following a suggestion from @ndevenish , I have added another group to the file where the data is named according to the nxreflections definition, which are linked to the data in the dials working format through references and region-references in the case of multidimensional data like miller_index -> h,k,l

**Comment by phyy-nx** [comment]
@ndevenish can we wait a bit more?  Not longer than DIALS core this week.

@jbeilstenedmands The NXreflections bit is really what I was looking for, and using references should add essentially no overhead.  Can you add the attribute `NX_class = NXreflections` to the `nx_reflections` group?

**Comment by ndevenish** [comment]
> @ndevenish can we wait a bit more? Not longer than DIALS core this week.

Of course; I might not make it, but others should.

**Comment by jbeilstenedmands** [comment]
I have added the `NX_class = NXreflections` attribute now.

The general structure in the file looks like this:
```
/dials
  /processing
    /group_0
      /{dials data arrays}
/nx_reflections
  /group_0 (this group has NX_class = NXreflections attribute)
    /{nx reflections data arrays (references to dials data arrays)}
```
Hierarchy structured this way to allow multiple data 'groups' in a single file (like in my profiling example above), and to allow multiple 'processing' group entries in future (which could represent different processing steps linked together through references/VDS).

For now, saving a single reflection table to file gives this kind of hdf5 file:
hdf5example

PR #2678, opened 2024-06-14
Change logger statements in spot finding to use f-strings 

toastisme:
This was noticed when using [laue-dials](https://github.com/rs-station/laue-dials/tree/main).
There are some instances when the logger statements in spot finding are not evaluated correctly (in this case when using `easy_mp` with `nproc > 1`). The user then just sees `Extracted %d spots` etc. I can't see the actual cause for this, but using f-strings seems to fix it.


ndevenish:
f-strings fixes this because it evaluates the string in-place, before passing to logging; this pays any cost for stringification even if the log message is not emitted. The logging module %-based system avoids this, by carrying the logging arguments through the handler stack and only calling `__str__` if the message is actually emitted.

We have code that is supposed to pass the message over the process barrier, this definitely worked in the past and has been fixed before.

I have mostly given up trying to get people to not use f-strings on logging calls though, so am somewhat minded to merge this; but it is probably a sign that something else is also wrong/broken.

PR #2680, opened 2024-06-19
Fix Pinkindexer Bug

kmdalton:
My own code was using the real part first quaternion format (w, x, y, z) whereas scipy uses real last (x, y, z, w). I fixed my own functions to be compatible with scipy's. I would have prefered to use scipy's code exclusively, but it doesn't broadcast past one leading dimension. From my testing earlier this year, I seem to recall it being relatively slow for this application as well. 

@dagewa, I was able to index your 10 simulated ED patterns now using the following script based on the version you shared. 

```bash
dials.import noiseimage_000??.img
dials.find_spots imported.expt min_spot_size=3
  stills.indexer=sequences\
  unit_cell=78.840,78.840,38.290,90.000,90.000,90.000\
  space_group=P43212\
  n_macro_cycles=2\
  indexing.method=pink_indexer\
  percent_bandwidth=2.0 \
  rotogram_grid_points=180\
  voxel_grid_points=180\
  min_lattices=5 \
  max_refls=50
```

I think the only salient change is upping the bandwidth to 2 percent. The bandwidth probably shouldn't be thought of as literal bandwidth in this case. It also soaks up some of the expected unit cell error. 

All-in-all it is remarkable that this worked at all before.


kmdalton:
@dagewa, the CI was complaining about an unused import in an unrelated test. i fixed that. hopefully it should run now. either way, i think this should not be a contentious patch.

PR #2681, opened 2024-06-20
Data id fix

amyjaynethompson:
Found a bug when outputting the json file from dials.correlation_matrix that was causing the multiplex tests to fail when called by xia2. Correct dataset ids now output in json file.


jbeilstenedmands (approved):
Thanks for adding the test, looks good!

PR #2687, opened 2024-07-01
In the scans indexer, do not convert mm to px when `indexing.refinement_protocol.mode==None`

### LINKED ISSUE #2686: Setting refinement_protocol.mode=None for dials.index can result in failure
In [this line](https://github.com/dials/dials/blob/97e2b7327236666bd025bbbbc98eb383683fce17/src/dials/algorithms/indexing/indexer.py#L658) we have the code block for `refinement_protocol.mode=None`. We have that `refined_reflections` becomes `reflections_for_refinement`, which are the indexed reflections. There does not seem to be any column of the variety `xyzcal.XX`, so when this set of reflections proceeds to [this line](https://github.com/dials/dials/blob/97e2b7327236666bd025bbbbc98eb383683fce17/src/dials/algorithms/indexing/indexer.py#L786) then DIALS is unable to find the column to convert mm to pixels. I'm not certain, but I would guess that the refinement engine is assumed to have supplied the `xyzcal.mm` column, but since no refinement was run there is no such column.

dagewa:
Copy what is done by the `StillsIndexer` :

https://github.com/dials/dials/blob/97e2b7327236666bd025bbbbc98eb383683fce17/src/dials/algorithms/indexing/stills_indexer.py#L474-L476

Should fix #2686 

@PrinceWalnut can you check this does what you want?


PrinceWalnut:
The code looks like what I needed. I suspect this should work fine, but will build a dev version of DIALS and test this later today.

PrinceWalnut:
I've tested on local data and it works, so should be good to merge!

dagewa:
Very good, thanks :)

PR #2688, opened 2024-07-02
Fix matplotlib v3.9 colormaps

dagewa:
Change previously deprecated `matplotlib.cm.get_cmap(cmap)` calls to `matplotlib.colormaps[cmap]`


Baharis:
Relevant cctbx PR: https://github.com/cctbx/cctbx_project/issues/894. It's 2 am at the west coast right now so I might be wrong, but IIRC I finally settled on the other accessor as syntax `matplotlib.colormaps[cmap]` is absent in old-ish versions of mpl.

dagewa:
Ah, thanks

dagewa:
Looks like `plt.get_cmp` was kept because it is "too widely used to be removed" https://github.com/matplotlib/matplotlib/issues/20853. That implies removal was considered. I'll stick with the `matplotlib.colormaps[cmap]` version for DIALS as I don't think we are aiming to support older matplotlibs.

Baharis:
@dagewa I share your sentiment and, personally, dislike the need to support old installations (cctbx being an extreme case of sticking to Python 2); I just say that because I feel the need to note that in case someone is working on a late 2021 / early 2022 installation and updates codebase only (or some other stuff that people do), this will break their code since this syntax was introduced in [November 2021](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.5.0.html#colormap-registry-experimental). Other than that I am all in and happy for you that you can use this new smooth syntax :).

PR #2689, opened 2024-07-02
Build dials image from rockylinux:8 fixes #2690

### LINKED ISSUE #2690: Docker build fails as CentOS 7 is now EOL
Dockerfile references `centos:7` image as reference. As of 30/06/2024 CentOS 7 has reached EOL and the relevant repositories are taken off line.

isikhar:
Centos7 has reached EOF on 30/06/2024 with the corresponding distribution servers taken offline. Changing reference image to Rocky Linux 8 in line with changes deployed by Scientific Computing at Diamond and STFC.


ndevenish:
I don't think there is a reason not to just go to 9?

Fixes #2690

PR #2692, opened 2024-07-03
Updates for the small molecule tutorial

dagewa:
- Use `dials.export format=shelx`
- A couple of other minor tweaks


dagewa:
The Windows test failure cannot be related as this changes a documentation file only.

PR #2700, opened 2024-07-15
Convert merged MTZ export to use gemmi

### LINKED ISSUE #2684: dials.merge does not write a space group name P2111
I have a pseudo-orthorhombic dataset that can be *roughly* solved in P212121 but is not quite right. Trying to find out which axis has the crystallographic symmetry, I tried merging in P2111, P1211 and P1121 by running e.g. `dials.cosym space_group=P2111` followed by `dials.scale symmetrized.{expt,refl}`.

```
$ dials.show scaled.expt | grep Space
    Space group: P 21 1 1
    Space group: P 21 1 1
... # several crystals were merged. All P 21 1 1 as expected

$ dials.export scaled.{expt,refl}
 Space Group: P 21 1 1
 Space Group Number: 0

$ echo GO | mtzdump hklin scaled.mtz | grep Space
 * Space group = 'P 21 1 1' (number     0)
```

So far, everything is fine.
However, `dials.merge` does something weird.

```
$ dials.merge scaled.{expt,refl}

 Writing reflections to merged.mtz
 Title: From dials.merge
 Space group symbol from file: No.4 # This is not a SYMBOL!
 Space group number from file: 4
 Space group from matrices: P 21 1 1 (No. 4)
 Point group symbol from file: 2

$ echo GO| mtzdump hklin merged.mtz|grep Space
 * Space group = 'No.4' (number     4) # !?
```

Servalcat fails because "No.4" is not a space group name.
Why cannot `dials.merge` write `P 21 1 1`, while `dials.export` can?

I cannot upload this dataset because it is unpublished, but I guess you can reproduce this issue on other orthorhombic or tetragonal datasets (e.g. lysozyme). If you cannot, I will share the dataset privately.

### LINKED ISSUE #2691: Convert merged MTZ export to use gemmi
> My first guess is that this might be because `dials.export` of an unmerged MTZ uses gemmi, whereas creation of merged MTZs still uses `iotbx.mtz`. This reminds me that I wanted to follow up https://github.com/dials/dials/pull/2549 by changing _all_ MTZ handling in DIALS to use gemmi. I just haven't found time to do that yet.

_Originally posted by @dagewa in https://github.com/dials/dials/issues/2684#issuecomment-2194108745_

dagewa:
Following #2549, in which export of unmerged MTZs was changed to use gemmi, this PR also changes output of merged MTZ to use gemmi. This ensures consistency with things like exact format of space group names, and also makes it easier to maintain. The `gemmi.Mtz` interface is simple, and [well-documented](https://gemmi.readthedocs.io/en/latest/hkl.html). It is also fairly low level - you don't get too much for free, hence matching of indices between columns to add and the overall set of merged indices has to be done here, not in the library code.

Fixes #2691 
Fixes #2684


dagewa:
@Baharis, previously you [commented](https://github.com/dials/dials/pull/2592) that `MADMergedMTZWriter` and `MergedMTZWriter` were imported by `cctbx.xfel.merge`, and `gemmi` imports should be made optional. This PR does not remove the older MTZ writer classes, but it does deprecate them. However, I could not find this import anyway, so perhaps this is not relevant now? If that is the case, can we `import gemmi` with impunity now?

dagewa:
This change necessitates some changes in xia2 - I'm currently preparing a PR for those.

Baharis:
@dagewa Indeed, I can no longer find the offending `from dials.util.export_mtz` in the `cctbx` code. Unfortunately, I failed to include in my note the precise location of the import statement, so I can't say with certainty what happened, but this might be no longer an issue. The revealed gemmi import used to cause one of the `xfel_regression` tests to fail, so I'd suggest doing things the Python way: reveal the import and see if anything breaks :).

jbeilstenedmands (approved):
I gave this a quick test on some ssx data (plus the associated xia2 changes), and everything worked as expected, including a subsequent run of dimple 👍

Issue #2695, opened 2024-07-09
Adjust tolerances for sys abs check in dials.symmetry


dagewa:
Users with electron diffraction data sometimes find that dials.symmetry fails to identify screw axes, apparently because systematic absences are _not_ exactly absent. Here is an example of a small molecule case where the correct space group is $P2_1 2_1 2_1$, but dials.symmetry keeps it in $P222$.

From the table it is clear that there is significant intensity in the "absences", however it is also clear that the intensity of absences is quite a bit lower than the present reflections.
```
+--------------+---------+---------------+--------------+---------------+--------------+-------------------+------------------+
| Screw axis   |   Score |   No. present |   No. absent |   <I> present |   <I> absent |   <I/sig> present |   <I/sig> absent |
|--------------+---------+---------------+--------------+---------------+--------------+-------------------+------------------|
| 21a          |       0 |             3 |            3 |        33.797 |        8.859 |            35.244 |           10.699 |
| 21b          |       0 |             3 |            4 |       103.125 |       12.033 |           125.073 |           18.319 |
| 21c          |       0 |             0 |            0 |         0     |        0     |             0     |            0     |
+--------------+---------+---------------+--------------+---------------+--------------+-------------------+------------------+
```
![image](https://github.com/dials/dials/assets/11502083/59c80554-9e1c-4881-88df-71673cede291)

It seems it should be possible to separate the two classes but I've not found how to make dials.symmetry more tolerant of intensity in the absences. There is this `significance_level = *0.95 0.975 0.99` choice, but none of them allow finding the correct space group in this case. I would like to find a way to tune the tolerance over a wider range.

jbeilstenedmands:
Hi @dagewa , did you try the option `systematic_absences.method=fourier`? This was implemented by @kmdalton to try to help with these kinds of cases. Perhaps we should also change the `significance_level` to be settable to any floating point value between 0 and 1.

dagewa:
Thanks, yes, I did try `systematic_absences.method=fourier` but I think this case is tricky because the intensity in the "absences" is quite high. I like the idea of having `significance_level` be a float in [0,1]. That would also help with another weirdness with the PHIL choice, which is that `significance_level=.95` is not accepted, it has to be `0.95`

dagewa:
Investigating this: changing `significance_level` to take a float, then on the data in question, this command successfully found the two screw axes that were observed:
```
[...]

PR #2701, opened 2024-07-16
Oversample the FFT in dials.symmetry method=fourier

kmdalton:
This PR uses zero-padding of the direct space measurements to oversample the FFT in `dials.symmetry systematic_absences.method=fourier integrated.{expt,refl}`. This smooths out the Fourier transform and provides more points to condition the null model used to assign a probability to the screw axis. 

This may partially address #2695.


dagewa (approved):
This is great, thanks! Now to get the algorithm to _fail_ to find the $2_1$ around $a$ I have to set `significance_level=0.99`

dagewa:
The Fourier method works so much better for this data than the default. Are there reasons to prefer the default method in some circumstances, or should we consider making this the default?

kmdalton:
I think it is worth considering making the fourier method the default. My hesitation is that I have only really validated it on a handful of test cases.

dagewa:
Yes, this is one of the things it would be nice to test with a large number of representative data sets. It would be very useful to define such a sample of diffraction data space for tests like this, or even just to be able to run test jobs as part of autoprocessing at beamlines and wait until enough results are accumulated...

kmdalton:
Yeah, maybe we can ask one of the beamlines at Diamond to test it on some historical data? For what it is worth, I think the Fourier method is the default in recent versions of Pointless (see section A3 of [Evans 2011](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3069743/)). However the version used there has a couple of important differences including a 
 1) more sophisticated model of the null hypothesis and
 2) more sophisticated peak scoring. 

I think it would require a significant refactor to implement the exact algorithm in DIALS, because it requires access to non-axial reflections to form the null model. Currently the screw axis scoring methods only get passed the axial reflections. I propose we first figure out how we're doing with the simple score function I implemented, and put a more extensive refactor on the to-do list.

jbeilstenedmands (approved):
Thanks Kevin @kmdalton for implementing this, it looks better on the datasets I tested it on 👍 .
Regarding making this the default, I discussed with David and we are keen on this idea but we need to make sure we have robustly tested on a range of datasets from Diamond before I would be comfortable changing over.

kmdalton:
I'm 100% on board with thorough testing before changing the default. I'm going to work on this PR a bit more before merging as well. I'm a little unsatisfied with it still.

dagewa:
@kmdalton did you manage to find time to look at this? I certainly found this branch to be an improvement over the current approach. Would it be reasonable to merge this as-is and post an issue regarding the further work you planned?

kmdalton:
@dagewa, this totally fell off my radar. Let me get back to you on Monday. I recall having some concerns, but I need to refresh my memory.

dagewa:
Ok, no problem!

dagewa:
Thanks @kmdalton!

dagewa:
Ok, looks like we get a couple of test failures. I'll try to look into it soon

dagewa:
Hi @kmdalton these tests fail with the new oversampling method

```
FAILED tests/algorithms/symmetry/absences/test_screw_axes.py::test_screw_axes_example_data[fourier] - assert 0.0 > 0.99
FAILED tests/algorithms/symmetry/absences/test_screw_axes.py::test_not_screw_axes[fourier] - assert 0.9979400892500825 < 0.01
```

If I set the default to `oversample=None` then the tests pass, but at least one fails with any value I tested for `oversample` over a large range. I am not sure what is going on here.

dagewa:
Ah I think it might be as simple as the value of `n` is not updated for the new length of the vector `fourier_space`.  I'm guessing based on the other changes that `n` should actually be `d` here:

https://github.com/dials/dials/blob/05d311fca8bc3df149e88d803d260032f06766f2/src/dials/algorithms/symmetry/absences/screw_axes.py#L190-L191

dagewa:
(if I make that change, then the tests pass)

kmdalton:
I concur that it looks wrong. Let me have a look ...

kmdalton:
I will have a fix shortly. Thanks for looking into this.

kmdalton:
@dagewa, I believe the Fourier screw axis detection is fixed now. Let me double check that I did not break the diagnostic plot.

Issue #2703, opened 2024-07-19
OpenGL missing from DIALS from conda on Windows


dagewa:
I've been testing a conda install of DIALS on Windows. Most things seem to work, but `dials.reciprocal_lattice_viewer` fails with 
```
(dials-conda-test) C:\Users\fcx32934\data\dials-test>dials.reciprocal_lattice_viewer indexed.expt indexed.refl
Traceback (most recent call last):
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\gltbx\glu.py", line 3, in <module>
    import OpenGL  # implicit import
    ^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'OpenGL'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\boost_adaptbx\boost\python.py", line 22, in import_ext
    try: mod = __import__(name)
               ^^^^^^^^^^^^^^^^
ImportError: DLL load failed while importing gltbx_glu_ext: A dynamic link library (DLL) initialization routine failed.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Scripts\dials.reciprocal_lattice_viewer-script.py", line 6, in <module>
    from dials.command_line.reciprocal_lattice_viewer import run
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\dials\command_line\reciprocal_lattice_viewer.py", line 16, in <module>
    from dials.util.reciprocal_lattice.viewer import ReciprocalLatticeViewer, phil_scope
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\dials\util\reciprocal_lattice\viewer.py", line 22, in <module>
    from dials.util import wx_viewer
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\dials\util\wx_viewer.py", line 14, in <module>
    import gltbx.glu as glu
  File "C:\Users\fcx32934\AppData\Local\miniconda3\envs\dials-conda-test\Lib\site-packages\gltbx\glu.py", line 7, in <module>
    ext = bp.import_ext("gltbx_glu_ext")
[...]

PR #2707, opened 2024-07-24
Fix `dials.reciprocal_lattice_viewer` middle mouse drag to translate

dagewa:
I could not find any documentation on this, but this appears to be the fix. It looks like the `pyopengl` bindings choose (sensibly) not to match the openGL function signatures exactly, so instead of passing empty lists to emulate pointers to values that will be returned (see `winX`, `winY`, `winZ` in https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/gluProject.xml), these are simply the return value of the call instead.


dagewa:
I think we currently have the issue that `pyopengl` is not actually a dependency of DIALS, because it is not a dependency listed for cctbx-base. See https://github.com/dials/dials/issues/2703

dagewa:
Added `pyopengl` dependency here. See further discussion at https://github.com/dials/dials/issues/2703

dagewa:
Adding the `pyopengl` dependency was discussed in dials-core. Seems acceptable, so I'll merge this. If not re-bootstrapping, developers might need to `conda install pyopengl`

Issue #1340, opened 2020-07-15
Spot finding does not use format get_mask()


graeme-winter:
imageset get_mask() is delegated to 

```
  /**
   * Compute the mask
   * @param index The image index
   * @returns The image mask
   */
  Image<bool> get_mask(std::size_t index) {
    return get_dynamic_mask(index);
  }
```

which means if you have a static mask, it is lost. Could explain a lot of the issues I have observed in recent history.

graeme-winter:
This could be either delegated to 

```
  virtual Image<bool> get_dynamic_mask(std::size_t index) {
    return get_trusted_range_mask(get_static_mask(), index);
  }
```

which would kinda sorta make sense or 

```
  /**
   * Get the dynamic mask for the requested image
   * @param index The image index
   * @returns The image mask
   */
  virtual Image<bool> get_dynamic_mask(std::size_t index) {
    // Get the masker
    ImageSetData::masker_ptr masker = data_.masker();

    // Create return buffer
    Image<bool> dyn_mask;

    // Get the image data object
    if (masker != NULL) {
      DXTBX_ASSERT(scan_ != NULL);
      DXTBX_ASSERT(detector_ != NULL);
      double scan_angle = rad_as_deg(
        scan_->get_angle_from_image_index(index + scan_->get_image_range()[0]));
      dyn_mask = masker->get_mask(*detector_, scan_angle);
    }

    // Return the dynamic mask
    return get_trusted_range_mask(get_static_mask(dyn_mask), index);
  }
```

which in turn loops in ```
  /**
   * Get the static mask common to all images
   * @param mask The input mask
   * @returns The mask
   */
  Image<bool> get_static_mask(Image<bool> mask) {
    return get_untrusted_rectangle_mask(
      get_external_mask(mask.empty() ? get_empty_mask() : mask));
  }

  /**
   * Get the static mask common to all images
   * @returns The mask
   */
  Image<bool> get_static_mask() {
    return get_static_mask(Image<bool>());
  }
```

and finally 

```
  /**
   * Apply the external mask
   * @param mask The input mask
   * @returns The external mask
   */
  Image<bool> get_external_mask(Image<bool> mask) {
    Image<bool> external_mask = external_lookup().mask().get_data();
    if (!external_mask.empty()) {
      DXTBX_ASSERT(external_mask.n_tiles() == mask.n_tiles());
      for (std::size_t i = 0; i < mask.n_tiles(); ++i) {
[...]

Issue #236, opened 2016-10-18
Pixel masking in DIALS


biochem-fan:
Could someone please summarize how pixel masking is done in DIALS (spot finding and integration)?
1.  Is panel's `trusted_range` set in a dxtbx class automatically respected? What if `get_mask` is defined? Do I have to implement `get_mask()` in the format class, where the mask is generated by panel's `get_trusted_range_mask` and (possibly) merged with other pixel masks?
   
   In my MPCCD class, setting `trusted_range` alone seems sufficient; but some dxtbx classes like `FormatPYunspecified` and `FormatCBFMiniPilatus` implement `get_mask`. So I wanted to make sure.
2.  When `mask=mask.pickle` is supplied to `dials.spot_find` or `dials.integrate`, does it replace the mask from the dxtbx class, or is it merged?
3. It seems that I have to specify `mask=mask.pickle` to `dials.spot_find` and `dials.integrate`, not only to `dials.import`. Is this intended behaviour?

jmp1985:
So pixel masking is done in the following way:

1. Imageset has a function get_mask; this function will call the format class get_mask function. The default return value for this is "None"; however it can be overriden to provide more sophisticated masking (e.g. dynamic masking etc). If the format get_mask function returns a value that is not "None", then this is used. Otherwise, if "None" is returned then a mask is created on the fly from the trusted_range values given by the panel class. So to summarize, if you implement get_mask then you are responsible for making sure that it masks things correctly; if you don't implement get_mask then the trusted_range is used.

2. Is mask is set then this is merged with the mask from the format get_mask function or the trusted range mask.

3. I may have to look into this. But I think that you can serialize a mask into the JSON file through dials.import and this will be used during spot finding and integration but there is also the option of specifying a mask in spot finding and integration that will not be serialized into the imageset itself. This sounds a bit complicated so maybe we need to tidying up the way this is done.

I hope that answers your questions?

biochem-fan:
Hi James,

Thank you very much for clarification. (1) and (2) are very clear.

(3) Specifying a mask to `dials.import` does not seem to be sufficient. I have an EIGER serial (still) dataset with ice rings. So I created `mask.pickle` by `dials.generate_mask`.

> But I think that you can serialize a mask into the JSON file through dials.import

Yes. `dials.import dataset_master.h5 mask=mask.pickle` creates `datablock.json`, where `mask` is correctly specified. 

[...]

PR #2711, opened 2024-07-31
Allow combining masks using dials.generate_mask

phyy-nx:
eg dials.generate_mask imported.expt beamcenter.mask border=1

or dials.generate_mask beamcenter.mask border.mask

Requires cctbx/dxtbx#744

- [x] TODO: update docs and add a test


biochem-fan:
Hi Aaron, this is a great functionality I also need.

> TODO: update docs and add a test

Yes, documenting how masking works in DIALS would be very nice.

It is not clear how the trusted range mask, the mask from a dxtbx class and the user supplied mask work. Are they combined, or one replaces the other?

There are some old discussions (e.g. https://github.com/dials/dials/issues/236 and https://github.com/dials/dials/issues/1340) but I am not sure if the current implementation works in the same way.

phyy-nx:
> It is not clear how the trusted range mask, the mask from a dxtbx class and the user supplied mask work. Are they combined, or one replaces the other?

My understanding is that the final mask for spotfinding and integration is the static mask from dxtbx & the dynamic mask from dxtbx & the trusted range mask & the user supplied mask

src/dials/command_line/generate_mask.py
```diff
@@ -15,8 +17,11 @@
     untrusted.circle=200,200,100
 
   dials.generate_mask models.expt d_max=2.00
-"""
 
+  dials.generate_mask models.expt d_max=2.00 border.mask
```
graeme-winter, line 21: ~border.mask~ existing.mask

src/dials/command_line/generate_mask.py
```diff
@@ -15,8 +17,11 @@
...
+
+  dials.generate_mask beamcenter.mask border.mask
```
graeme-winter, line 23: Suggest for most of our users `backstop.mask` and `shadow.mask` may be more useful example names

src/dials/command_line/generate_mask.py
```diff
@@ -64,6 +69,7 @@
 def generate_mask(
     experiments: ExperimentList,
     params: phil.scope_extract,
+    starting_masks: list = None,
```
graeme-winter, line 72: can you constrain `list = None` to be `None | list(Mask)`?

src/dials/command_line/generate_mask.py
```diff
@@ -83,13 +89,31 @@ def generate_mask(
         experiments: An experiment list containing only one imageset.
         params: Masking parameters, having the structure defined in
             :data:`phil_scope`.
+        starting_masks: list of masks to combine together with the mask being
```
graeme-winter, line 92: `starting` feels awkward here - `exsting`? Also combine together is redundant 🙃

src/dials/command_line/generate_mask.py
```diff
@@ -83,13 +89,31 @@ def generate_mask(
...
+            generated
 
     Returns:
         A list of masks, one for each imageset.
 
         A copy of :param:`experiments` with the masks applied (optional,
         only returned if :attr:`params.output.experiments` is set).
     """
+
+    if starting_masks:
```
graeme-winter, line 102: Again `starting` awkward here, `existing` or `prior` or `input`?

src/dials/command_line/generate_mask.py
```diff
@@ -83,13 +89,31 @@ def generate_mask(
...
+        starting_mask = list(starting_masks[0])
+        for mask in starting_masks[1:]:
```
graeme-winter, line 104: should check here that the masks are explicitly all the same shape?

src/dials/command_line/generate_mask.py
```diff
@@ -83,13 +89,31 @@ def generate_mask(
...
+            for panel_idx in range(len(starting_mask)):
+                starting_mask[panel_idx] &= mask[panel_idx]
+        starting_mask = tuple(starting_mask)
+
+    # Check if only combining masks
+    if not experiments and starting_masks:
+        # Save the mask to file
+        log.info("Writing mask to %s", params.output.mask)
+        with open(params.output.mask, "wb") as fh:
+            pickle.dump(starting_mask, fh)
```
graeme-winter, line 114: Not for here but parenthetic comment: we really should move away from using `pickle` for masks

src/dials/command_line/generate_mask.py
```diff
@@ -143,7 +172,7 @@ def run(args: List[str] = None, phil: phil.scope = phil_scope) -> None:
         args: Arguments to parse. If None, :data:`sys.argv[1:]` will be used.
     """
     # Create the parser
-    usage = "usage: dials.generate_mask [options] models.expt"
+    usage = "usage: dials.generate_mask [options] [models.expt] [masks]"
```
graeme-winter, line 175: `[masks]` => `[mask, mask, ...]` to make it clear that > 1 is allowed?

src/dials/command_line/generate_mask.py
```diff
@@ -153,19 +182,43 @@ def run(args: List[str] = None, phil: phil.scope = phil_scope) -> None:
...
+    )
     experiments = flatten_experiments(params.input.experiments)
 
     # Configure logging
     dials.util.log.config(verbosity=options.verbose, logfile=params.output.log)
 
+    # Read in any starting masks
+    remains_unhandled = []
+    starting_masks = []
+    for arg in unhandled:
```
graeme-winter, line 196: Just because it will unpickle does not make it a mask? Need to check that the unpicked object is the right type?

src/dials/command_line/generate_mask.py
```diff
@@ -153,19 +182,43 @@ def run(args: List[str] = None, phil: phil.scope = phil_scope) -> None:
...
+        if os.path.isfile(arg):
+            try:
+                with open(arg, "rb") as fh:
+                    mask = pickle.load(fh)
+            except Exception:
+                remains_unhandled.append(arg)
+            else:
+                if isinstance(mask, tuple):
```
graeme-winter, line 204: So the `if isinstance()` here gets merged in with the exception above?

src/dials/command_line/generate_mask.py
```diff
@@ -153,19 +182,43 @@ def run(args: List[str] = None, phil: phil.scope = phil_scope) -> None:
...
+                    starting_masks.append(mask)
+                else:
+                    remains_unhandled.append(arg)
+    if remains_unhandled:
+        print(
+            "Couldn't recognize the following arguments:", ",".join(remains_unhandled)
```
graeme-winter, line 210: Needs to include pickled things which are not masks

graeme-winter (commented):
Mostly cosmetic though some comments around handling files gracefully which are pickled but not pickled _masks_

`starting` is a nasty prefix, suggested improvements

src/dials/command_line/generate_mask.py
phyy-nx: The `|` is only python 3.10, and I assume we are still supporting 3.9? At least the end of life for 3.9 is next year.  But Union works:

```
    starting_masks: list = Union[None, Masks],
```

phyy-nx:
> Mostly cosmetic though some comments around handling files gracefully which are pickled but not pickled _masks_
> 
> `starting` is a nasty prefix, suggested improvements

Done!

phyy-nx:
@graeme-winter this is ready for your re-review.  Tests are passing.  Thanks!

PR #2714, opened 2024-07-31
Tidy warning around joint_indexing

### LINKED ISSUE #2713: Only print alert if > 1 scan
`dials.index` now prints

```
joint_indexing=True has been set for scans experiments
```

but this is only pertinent for > 1 scan, and also I will rewrite the phrasing

graeme-winter:
Fixes #2713


dagewa (approved):
As the author of the original log and error messages here, I agree yours are much nicer :+1:

PR #2718, opened 2024-08-07
Reflection table performance improvement

dwpaley:
For `cctbx.xfel.merge` runs with very large (100k image) datasets, the slowest step is creating selections from the huge reflection tables. We traced the performance hit to a double loop, outer over `id` values in the new table and inner over `experiment_identifiers` in the old table. Since `experiment_identifiers` is a map, we can replace the inner loop with a lookup.


graeme-winter:
I am pulling some test failures

```
Results (149.13s (0:02:29)):
      22 passed
       2 failed
         - tests/algorithms/indexing/test_index.py:804 test_pink_indexer
         - tests/command_line/test_ssx_index.py:94 test_ssx_index_input_unit_cell
      13 skipped
```

but those failures are also present on `main` so can be ignored 

```
Results (151.13s (0:02:31)):
      22 passed
       2 failed
         - tests/algorithms/indexing/test_index.py:804 test_pink_indexer
         - tests/command_line/test_ssx_index.py:94 test_ssx_index_input_unit_cell
      13 skipped
```

newsfragments/2718.bugfix
```diff
@@ -0,0 +1,2 @@
+Big (1000x?) performance improvement for selections from reflection tables
```
graeme-winter, line 1: Please qualify the 1000x for very large numbers of serial experiments so as to not mislead more general users

graeme-winter: i.e. O(100,000) experiments, say

graeme-winter (approved):
Change set looks good and I can believe it makes a huge difference for large N experiments as it reduces a problem from O(N^2) to O(N)

Good spot. Thank you.

newsfragments/2718.bugfix
dwpaley: Thanks @graeme-winter. Newsfragment updated and button pushed.

Issue #2724, opened 2024-08-21
Image stacking does not work with stills


biochem-fan:
Image stacking in `dials.image_viewer` does not work for multi-frame containers (e.g. MPCCD from SACLA [dials-data/image_examples/SACLA-MPCCD-Phase3-21528-5images.h5](https://github.com/dials/data-files/blob/master/image_examples/SACLA-MPCCD-Phase3-21528-5images.h5)). This is probably because frames are interpreted as multiple experiments, not a single image set.

Indeed, importing the image with `dials.import convert_stills_to_sequence=True` allows image stacking. But this comes with a side effect that per-shot wavelengths are ignored (#2725).

graeme-winter:
@biochem-fan 

> But this comes with a side effect that per-shot wavelengths are ignored.

If this is an image sequence you can still have one experiment per image, but with them using the same underlying image sequence, with a `scan` to handle the indexing. 

This can be made to work

biochem-fan:
@graeme-winter
I see. Indeed I manually edited the `imported.expt` and confirmed that `dials.show` displayed per-shot wavelengths.
This is worth doing.

But this alone does not solve the problem with `dials.image_viewer XXXX.h5`, where importing happens implicitly.
Potential options are:

1. fix image stacking for multiple experiments
2. make `convert_stills_to_sequence=True` the default

I prefer 2, because this also solves #2447.

biochem-fan:
> 1. fix image stacking for multiple experiments

The patch below implements this. I will send a PR together with a fix for #1512.

```diff
diff --git a/src/dials/util/image_viewer/spotfinder_frame.py b/src/dials/util/image_viewer/spotfinder_frame.py
index 56e7304e1..b46978460 100644
--- a/src/dials/util/image_viewer/spotfinder_frame.py
+++ b/src/dials/util/image_viewer/spotfinder_frame.py
@@ -1087,15 +1087,12 @@ def stack_images(self):
             if not isinstance(image_data, tuple):
                 image_data = (image_data,)
 
-            i_frame = self.image_chooser.GetClientData(
-                self.image_chooser.GetSelection()
-            ).index
-            imageset = self.images.selected.image_set
+            i_frame = self.image_chooser.GetSelection()
 
             for i in range(1, self.params.stack_images):
-                if (i_frame + i) >= len(imageset):
+                if (i_frame + i) >= len(self.images):
                     break
-                image_data_i = imageset[i_frame + i]
[...]

PR #2730, opened 2024-08-27
dials.image_viewer: fix stacking of masks and stacking of multiple experiments

### LINKED ISSUE #1512: dials.image_viewer show mask does not work with stacked images
Just makes a mess
Screenshot from 2020-12-04 15-10-30

biochem-fan:
This fixes #1512 (dials.image_viewer show mask does not work with stacked images) and #2724 (Image stacking does not work with multiple experiments, e.g. stills).

For #1512, I also implemented OR-merging of masks ("pixels which are masked in any of the images") as suggested there but I do not have such test cases. Do you have any?

This addresses only image views; stacking of other elements (spots, shoeboxes, etc) from multiple experiments are beyond the scope of this PR and not tested. The current fix at least helps finding beam stop shadows etc in SFX/SMX experiments.


biochem-fan:
> For https://github.com/dials/dials/issues/1512, I also implemented OR-merging of masks ("pixels which are masked in any of the images") as suggested there but I do not have such test cases. Do you have any?

I realized that I could test this by setting the maximum trusted value artificially low so that many pixels got masked. By this test, I noticed my first implementation had a problem and repaired it in 5c5e8c1. Mask merging should be OK now.

Issue #2712, opened 2024-07-31
Mask generation for multi-panel detectors


biochem-fan:
How should masks for multi-panel detectors be generated?

Our new XFEL detector (CITIUS) has 72 panels. My pipeline masks module borders and stores it in the NXmx file but users need to add masks for the beam stopper, the injector nozzle etc. Because they vary from experiment to experiment,  they cannot be hard-coded in the preprocessing pipeline.

`dials.image_viewer`'s masking tool complains "Point must be within a panel" (circle tool), "All points must be contained within a single panel" (polygon tool).

`dials.generate_mask imported.expt circle=-23,5,100` (coordinates were read from the image viewer's toolbar) made a mask on the first panel (the top left panel in the image), not on the image center.

(The beam stopper for our previous detector, MPCCD, was glued to the capton window and lacked the "arm". Thus, we could use the `d_max` option for masking. This does not apply to CITIUS.)

@phyy-nx do you have any experiences with CSPAD?

<img width="566" alt="image" src="https://github.com/user-attachments/assets/18b588a8-4cac-4157-9802-239d285bf141">

graeme-winter:
For reference: is some data available? The errors you report feel like unnecessarily tight constraints

biochem-fan:
The CITIUS image above is not available yet because I haven't finalized my NXmx writer but the same issue can be observed in any multi-panel detectors. For example, an MPCCD example is in [dials-data/image_examples/SACLA-MPCCD-Phase3-21528-5images.h5](https://github.com/dials/data-files/blob/master/image_examples/SACLA-MPCCD-Phase3-21528-5images.h5).

> The errors you report feel like unnecessarily tight constraints

Having a global 2D coordinate system is non-trivial for segmented detectors with non-parallel sensors (e.g. PILATUS 12M-DLS). But for the rest, removing this constraint is very useful.

phyy-nx:
For this kind of thing (like on the cspad, jungfrau, or mpccd detectors), I either mask each panel by hand using the polygon tool or I use a threshold.  For thresholding, I use dxtbx.image_average with a water only run and hopefully the beamstop is the lowest counts on the average, standard deviation, or max image.  Then I run a script like this one:

```
import dxtbx 
from libtbx import easy_pickle 
image=dxtbx.load('avg.cbf') 
m=tuple([d>=3 for d in image.get_raw_data()]) # keep all pixels with at least 3 counts on the average
easy_pickle.dump('threshold_3.mask', m)
```

[...]

Issue #2260, opened 2022-10-24
Exception handling with print()


graeme-winter:
Following looking at cctbx/dials#567...

In a few places we have the motif

```
        try:
            something()
        except Exception as e:
            print(e)
        else:
```

We probably want to use logging in such cases?

`print(e)` is found a these places - worth fixing?

- [ ] ./src/dials/util/image_viewer/mask_frame.py
- [x] ./src/dials/command_line/index.py
- [x] ./src/dials/command_line/geometry_viewer.py
- [x] ./src/dials/command_line/report.py

Issue #1964, opened 2022-01-11
Polygon mask mode cannot be deactivated once you click outside a panel


biochem-fan:
In `dials.image_viewer`, one usually completes a polygon mask by clicking the `Polygon` button again to exit the mode. However, if a pixel outside a panel has been included in the polygon, the mode cannot be deactivated and the only way out is to restart the image viewer.

This is because `AddUntrustedPolygon` raises an exception `Point must be within a panel` https://github.com/dials/dials/blob/main/util/image_viewer/mask_frame.py#L849 and this cancels the `OnUpdate` event handler midway before setting `self._mode_polygon_layer = None` https://github.com/dials/dials/blob/main/util/image_viewer/mask_frame.py#L512.

A dirty fix is the following but this is not user-friendly; all the points in the current polygon are lost. Because we often want to make a mask reaching an edge of the detector (e.g. masking a beam stopper shadow), probably a better way is to "snap" a pixel outside a panel to the nearest panel edge.

```diff
diff --git a/util/image_viewer/mask_frame.py b/util/image_viewer/mask_frame.py
index bca3ad4ca..f0a2a1669 100644
--- a/util/image_viewer/mask_frame.py
+++ b/util/image_viewer/mask_frame.py
@@ -509,7 +509,9 @@ def OnUpdate(self, event):
             or self.mode_circle_button.GetValue()
             or self.mode_rectangle_button.GetValue()
         ):
-            self.AddUntrustedPolygon(self._mode_polygon_points)
+            try:
+               self.AddUntrustedPolygon(self._mode_polygon_points)
+            except: pass
             self._mode_polygon_points = []
             self._pyslip.DeleteLayer(self._mode_polygon_layer)
             self._mode_polygon_layer = None
```

biochem-fan:
> A dirty fix is the following but this is not user-friendly; all the points in the current polygon are lost

This is a better patch, which blocks addition of vertices outside sensors.

```diff
diff --git a/src/dials/util/image_viewer/mask_frame.py b/src/dials/util/image_viewer/mask_frame.py
index 2be91a16e..a4a2740df 100644
--- a/src/dials/util/image_viewer/mask_frame.py
+++ b/src/dials/util/image_viewer/mask_frame.py
@@ -687,8 +687,13 @@ def OnLeftDown(self, event):
                 self._circle_radius = None
                 return
             elif self._mode_polygon:
-                self._mode_polygon_points.append(click_posn)
[...]

Issue #1824, opened 2021-07-29
`dials.image_viewer`: improvements to masking UI


dagewa:
The masking UI could do with some improvements. One is that it is possible to save a mask before the polygon mode has finalised, which can end up with an empty mask. Perhaps clicking "save" should deselect/complete the polygon mode.

biochem-fan:
Also: #539

biochem-fan:
Another issue is that when one accidentally double-clicks on the image while drawing a polygon mask, the zoom level changes and the polygon becomes a mess.

biochem-fan:
Also #1964: I made a separate issue for this one because there is no user-side workaround.

biochem-fan:
For completeness: other issues include https://github.com/dials/dials/issues/1512 and https://github.com/dials/dials/issues/2712.

biochem-fan:
> Another issue is that when one accidentally double-clicks on the image while drawing a polygon mask, the zoom level changes and the polygon becomes a mess.

This issue can be resolved by storing vertices (`self._mode_polygon_points`) in the PySlip Geo coordinate system, not in the View coordinate system. I will send a PR together with other masking fixes.

PR #2731, opened 2024-08-28
Image viewer masking UI fix

biochem-fan:
This PR fixes various glitches in `dials.image_viewer` masking tools.

[Panning or changing the zoom level used to mess up a polygon being drawn](https://github.com/dials/dials/issues/1824#issuecomment-1009743928), because vertex coordinates were stored in the screen coordinates, not in the logical (geometry) coordinates. This problem was fixed.

As reported in #1964, once an inappropriate polygon (i.e. going outside a single panel) was drawn, the tool used to get stuck in the polygon mode and one had to restart the viewer (and lose the mask) to recover from the inconsistent state. This problem was addressed by making sure a new vertex is added only when it is within a single panel. This check can be removed once we implement multi-panel polygon masking (#2712).

Similarly, once an inappropriate rectangle was drawn, the bad rectangle used to remain on the screen and the viewer used to stay in the rectangle mode even though the "Rectangle" button looked unset. This problem was resolved in the same way as the circle tool.

Error messages about single-panel constraints used to be displayed on the console as assertion exceptions and can be missed by the user. This patch displays the messages in a dialog box (#2260).


ndevenish:
Image viewer fixes are always welcome! Does this address everything in #1964 ?

biochem-fan:
@ndevenish Yes. Thank you very much for merging.

PR #2732, opened 2024-08-28
DIALS dependency unification

ndevenish:
Work applying the same principle in https://github.com/cctbx/dxtbx/commit/853bbb76b7595f9b21e557aa882a66ccad0abb09 to DIALS and via bootstrap.

This consolidates the requirements for specifying DIALS dependencies for development, testing, conda-forge, mac, linux, windows, prebuilt-cctbx and bootstrap into a single `dependencies.yaml` in the DIALS root. They are now split up into categories matching the [conda-build](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section) definitions:
- `build` is for build tooling only required when you build the packages (the user would not install these)
- `host` is.. slightly harder to pin down. It's effectively the binary packages that you need present at build time, to e.g. link against.
- `run` is any dependencies only required to actually run dxtbx, separate from building it
- `test` is any dependencies used for running tests (e.g. dials-data, pytest...)

Platform/target selection of dependencies is handled with _selectors_. This is a very small subset of [the concept used by conda-build](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocess-selectors). Effectively, you can mark lines as only included when a condition is met. For example:
```
  - pycbf #[prebuilt_cctbx]
  - scipy
  - wxpython >=4.2.0 #[not osx]
  - wxpython >=4.2.0=*_5 #[osx]
  - pytest
  - pytest-nunit # [win]
```
- `pycbf` will only be included when using prebuilt (e.g. from conda-forge) cctbx. `scipy` and `pytest` will always be included, `wxpython` will have different constraints depending on whether you are running on macOS or not, and `pytest-nunit` will only be included on windows.

Valid selector combinations are currently specified in the header of `dependencies.yaml`.

The driver for this is continually having to update dependencies in several places whenever something needs to be changed. As well as separate dependency lists for mac, linux and windows, we have some custom heuristics for removing unnecessary packages from release bundles, we have the dependencies entirely separately specified as part of the conda-forge package, and it's otherwise pretty difficult to work out what is needed for build vs runtime (which lets us deliberately include less in the bundles).


dagewa:
Seems like a very good idea to me :+1: Editing three files is annoying and the selectors look neat.

ndevenish:
The remaining xfel stuff is due to ordered_set changes

ndevenish:
Okay, so now, the fixed `macos.txt`, `linux.txt`, and `windows.txt` files are still present, with an "autogenerated" header. When the `dependencies.yaml` file is changed, these will be regenerated to match.

I believe this should allow anything downstream to continue working?

ndevenish:
@phyy-nx do you have any specific objections to this approach? I believe with keeping the `.conda-envs` files it should be fully compatible with the way you are picking up dependencies in xfel now...

phyy-nx:
We discussed this in a DIALS telco, but for completeness here, no objections from me on this :)

graeme-winter:
In 2025-07-31 DIALS telco agreement is that this is ready to merge

ndevenish:
test reindex comment to make appear

PR #2735, opened 2024-09-13
Multi-panel polygon masking in dials.image_viewer

### LINKED ISSUE #2712: Mask generation for multi-panel detectors

biochem-fan:
This resolves #2712.

A polygon can be drawn over multiple panels. Vertices in panel gaps are also allowed. The resulting polygon is divided into sub-masks, each of which is contained within a single panel so that `dials.generate_mask` can handle it.

An example:
image

As pointed out by @dagewa in #2712, cyan lines run through the center of pixels. This is probably related to the way the coordinate system works in `dials.image_viewer` (#2190).

image

This does not pose problems; the resulting mask extends to the very edges of panels.

image


phyy-nx:
Hi Takanori!  Very excited about this change, this comes up all the time.

I checked out the branch and tried it on a multipanel dataset.  I found a case where it throws an exception:
image

Here you can see several successful masks crossing several panels, but the central one, attempting to draw a mask on the four inner tiles, produced this exception:

```
  File "/dev/shm/aaron/modules/dials/src/dials/util/image_viewer/mask_frame.py", line 506, in OnUpdate
    self.AddUntrustedPolygon(self._mode_polygon_points)
  File "/dev/shm/aaron/modules/dials/src/dials/util/image_viewer/mask_frame.py", line 904, in AddUntrustedPolygon
    assert p_id == i
AssertionError
```

The image I used is in dials_data:
```
dials.image_viewer `dials.data get -q 4fluoro_cxi`/lcls_2022_smSFX_workshop_data/ten_cbfs/cxily6520_r0164_02081.cbf
```

What do you think?

biochem-fan:
@phyy-nx Thank you very much for your report.
I reproduced your problem and is investigating.

biochem-fan:
The difficulty is in the numerical fudges.

https://github.com/dials/dials/blob/2214232dd528799d3a8ac801fefa5fcd82a69570/src/dials/util/image_viewer/mask_frame.py#L894-L896

> This does not pose problems; the resulting mask extends to the very edges of panels.

I found that this does not hold always, depending on the panel configuration. I have to find a more robust approach.

biochem-fan:
Investigating this, I found two issues in the underlying (single-panel) polygon masking code.

- It does not allow a triangle mask
- Some pixels are inclusive, while other pixels are exclusive

I will investigate and report them as separate issue(s). Meanwhile, I switch this PR to a draft state.

biochem-fan:
> It does not allow a triangle mask

Reported to https://github.com/cctbx/dxtbx/issues/757.

biochem-fan:
@phyy-nx I fixed the code. Could you please test again?

>> It does not allow a triangle mask

>  Reported to https://github.com/cctbx/dxtbx/issues/757.

I handled triangles in the image viewer code without changing dxtbx (for the moment).

> Some pixels are inclusive, while other pixels are exclusive

In [the polygon masking code in dxtbx](https://github.com/cctbx/dxtbx/blob/9e2674712261458c7828cf873d4582361e6c8165/src/dxtbx/masking/masking.h#L145-L147), an integer coordinate (i, j) means a corner of a pixel, while a pixel's center (i + 0.5, j + 0.5) must be within the polygon for the pixel to be masked. Thus, to fully mask a panel edge, we must go **one pixel beyond** the panel limit. That is, if a panel is 512 px wide, valid coordinates are from 0 to 511 (inclusive), but we have to specify 512 as the polygon vertex coordinate.

I'm not sure if this is user-friendly (especially with the 0.5 px error in the image viewer #2190), but I did not to change it in dxtbx for backward compatibility.

phyy-nx:
Very nice

image

I agree about the half pixel thing, it's annoying, but doesn't need to be fixed here.

src/dials/util/image_viewer/mask_frame.py
```diff
@@ -691,11 +694,11 @@ def OnLeftDown(self, event):
                 )
                 p1, p0, p_id = self._pyslip.tiles.flex_image.picture_to_readout(yc, xc)
 
-                if p_id < 0:
-                    return
+                # if p_id < 0:
+                #    return
```
phyy-nx, line 698: Little bit of leftover clutter here, but otherwise LGTM

phyy-nx (approved):

src/dials/util/image_viewer/mask_frame.py
biochem-fan: Thanks! These lines and the next `if False` block are unnecessary. I will fix it.

PR #2741, opened 2024-09-24
Handle the case of very few reflections for a given dataset in weighted cosym analysis

jbeilstenedmands:
Likely only an issue for ssx data. Apply the min_reflections filter again, after the initial cosym filtering, so that at least min_reflections are present for the cosym analysis. Cleanly handle the zero curvatures assertion and check that no weights columns are all zero during construction of the wij matrix (which leads to the zero curvatures issue).


jbeilstenedmands:
Hi @phyy-nx , I have asked for your review as I realized that this will affect cctbx.xfel.merge through the inheritance of the CosymAnalysis class. 
The main issue I am trying to address is that even though we currently filter on `min_reflections`, this occurs at the start of cosym processing - there is then further filtering, merging and the setup of the `symmetry_base` class - all of which can easily reduce the number of reflections for some serial datasets to below `min_reflections`. The issue is that when doing weighted cosym analysis, if the number of reflections is very low (like 2 or 3), if may not be possible to calculate the correlation between that and any other dataset, resulting in a weights column of zero, which then causes zero curvatures and the lbfgs minimizer falls over (maybe you have seen this before?). So the solution proposed here is to recheck and filter after the setup of the `symmetry_base` class, to at least guarantee that there are `min_reflections` of each dataset when going into the cosym analysis. This is a real bug we have noticed after turning on weighting for ssx analysis.

ndevenish:
@phyy-nx, are you able to have a look at this at all? We're planning to make a release within the next few (working) days, and are pretty keen to get this in as it solves some problems we have seen internally.

Otherwise (and I do think you have been away/unavailable), barring objection, we plan to go ahead and merge this anyway and address any cctbx.xfel.merge issues after.

phyy-nx:
Apologies, I have been on leave.  This PR broke a [test](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/util/tst_TargetWithFastRij.py?ref_type=heads) in xfel_regression:

```
libtbx.python ../modules/xfel_regression/test/util/tst_TargetWithFastRij.py
  File "/dev/shm/aaron/test/../modules/xfel_regression/test/util/tst_TargetWithFastRij.py", line 52, in <module>
    test_rij_wij_computer().run()
  File "/dev/shm/aaron/test/../modules/xfel_regression/test/util/tst_TargetWithFastRij.py", line 28, in run
    rij_dials, wij_dials = ref_instance._compute_rij_wij(use_super=True)
  File "/dev/shm/aaron/modules/cctbx_project/xfel/merging/application/modify/aux_cosym.py", line 384, in _compute_rij_wij
    return super()._compute_rij_wij(use_cache=use_cache)
  File "/dev/shm/aaron/modules/dials/src/dials/algorithms/symmetry/cosym/target.py", line 459, in _compute_rij_wij
    n_refl = sum(self._lattice_ids == i)
AttributeError: 'TargetWithFastRij' object has no attribute '_lattice_ids'
```

It is not a test I worked on myself.  I think I'd like to tag @dwpaley?

phyy-nx:
Note that our cosym test in xfel.merge is passing:

libtbx.python ../modules/xfel_regression/test/command_line/tst_xfel_merge.py modify_cosym

jbeilstenedmands:
> Apologies, I have been on leave. This PR broke a [test](https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/util/tst_TargetWithFastRij.py?ref_type=heads) in xfel_regression:
> 
> ```
> libtbx.python ../modules/xfel_regression/test/util/tst_TargetWithFastRij.py
>   File "/dev/shm/aaron/test/../modules/xfel_regression/test/util/tst_TargetWithFastRij.py", line 52, in <module>
>     test_rij_wij_computer().run()
>   File "/dev/shm/aaron/test/../modules/xfel_regression/test/util/tst_TargetWithFastRij.py", line 28, in run
>     rij_dials, wij_dials = ref_instance._compute_rij_wij(use_super=True)
>   File "/dev/shm/aaron/modules/cctbx_project/xfel/merging/application/modify/aux_cosym.py", line 384, in _compute_rij_wij
>     return super()._compute_rij_wij(use_cache=use_cache)
>   File "/dev/shm/aaron/modules/dials/src/dials/algorithms/symmetry/cosym/target.py", line 459, in _compute_rij_wij
>     n_refl = sum(self._lattice_ids == i)
> AttributeError: 'TargetWithFastRij' object has no attribute '_lattice_ids'
> ```
> 
> It is not a test I worked on myself. I think I'd like to tag @dwpaley?

This looks like quite a simple fix for you, the `aux_cosym` code follows a separate branch for testing which doesn't initialize the parent class and is currently missing the definition of the `._lattice_ids` member:
https://github.com/cctbx/cctbx_project/blob/4dacb50b2514377404d2d5dd52b76714de6090ab/xfel/merging/application/modify/aux_cosym.py#L349

jbeilstenedmands:
Looking further into the failing xfel test, I found that the issue is that the class data is not correctly set in the testing part of the code in cctbx, this branch is what is needed to correctly load the first 10 datasets into the target:
https://github.com/cctbx/cctbx_project/compare/master...fix_xfel_cosym_test_input
Then it doesn't matter that `._lattice_ids` are not set as the error is never caught by the newly added dials code.
Unfortunately the xfel regression test now fails on comparison to the reference test result, which incorrectly contains all zero entries for the correlation calculations of the final dataset, so either the reference needs updating or the tests against this reference removing from the `test/util/tst_TargetWithFastRij.py` test in xfel_regression.

phyy-nx:
Thanks @jbeilstenedmands, I've merged in your cctbx_project change and updated the reference data in xfel_regression.  The test passes for me locally and should start passing on XFEL CI

PR #2743, opened 2024-09-25
Clustering dimensions

amyjaynethompson:
Add in the capability for cosym to optimise the number of dimensions required to best perform cosine-angle clustering on datasets. 

This has been done in a way that (hopefully) doesn't alter behaviour for symmetry analysis (though perhaps could be compressed further). 

An additional outlier rejection step has been added as an option when computing the functional as suggested by Graeme - currently has only been tested for dataset clustering dimension optimisation

There is also a hard turning on of cc_weights=sigma in dials.correlation_matrix if the dimension optimisation has been selected to be run. This is because it was found to completely break down without this option and thus should not be used in isolation (also documented in the code). 

Currently dimension optimisation is optional and would need to be selected by the user


src/dials/algorithms/symmetry/cosym/__init__.py
```diff
@@ -370,6 +380,7 @@ def _principal_component_analysis(self):
             "Explained variance: "
             + ", ".join(["%.2g" % v for v in pca.explained_variance_])
         )
+        logger.info(pca.components_)
```
jbeilstenedmands, line 383: Was this addition intentional?

src/dials/algorithms/correlation/analysis.py
```diff
@@ -182,7 +198,11 @@ def calculate_matrices(self):
         self.cosym_analysis._intialise_target()
 
         # Cosym proceedures to calculate the cos-angle matrix
-        self.cosym_analysis._determine_dimensions()
+        if self.params.dimensions is Auto:
+            self.cosym_analysis._determine_dimensions(
+                self.cosym_analysis.number_of_datasets,
```
jbeilstenedmands, line 203:
```suggestion
                len(self.unmerged_datasets),
```
My changeset removed the `number_of_datasets` variable, so this line needs updating

jbeilstenedmands: Further, we probably want to have a phil parameter of the maximum number of dimensions to test (default 20?, 50?) otherwise it will become very slow for processing >>100s of datasets

src/dials/command_line/correlation_matrix.py
```diff
@@ -23,11 +23,14 @@
 
 phil_scope = iotbx.phil.parse(
     """\
-include scope dials.algorithms.correlation.analysis.phil_scope
+include scope dials.algorithms.correlation.analysis.working_phil
 
 seed = 42
   .type = int(value_min=0)
 
+outlier_rejection = True
```
jbeilstenedmands, line 31: For clarity it might be better to nest this parameter, alongside a parameter for the maximum number of dimensions to test, into a scope with a name such as 'dimensionality assessment'.

src/dials/algorithms/symmetry/cosym/__init__.py
```diff
@@ -265,65 +265,74 @@ def _intialise_target(self):
             nproc=self.params.nproc,
         )
 
-    def _determine_dimensions(self):
-        if self.params.dimensions is Auto and self.target.dim == 2:
-            self.params.dimensions = 2
```
jbeilstenedmands, line 270: We are now not setting this value back into the params object, but I don't think this is used after this point so that is ok?

src/dials/command_line/correlation_matrix.py
```diff
@@ -23,11 +23,14 @@
 
 phil_scope = iotbx.phil.parse(
     """\
-include scope dials.algorithms.correlation.analysis.phil_scope
+include scope dials.algorithms.correlation.analysis.working_phil
 
 seed = 42
```
jbeilstenedmands, line 28: This seed, as well as the cosym seed, should be moved into the cosym common phil scope in algorithms/symmetry/cosym/__init__.py. We should use the common value of 230 that is used for the symmetry programs.

newsfragments/2743.feature
```diff
@@ -0,0 +1 @@
+``dials.cosym``: Add dimension optimisation for intensity-based dataset clustering
```
jbeilstenedmands, line 1: This should be dials.correlation_matrix, as we are only making this available through correlation_matrix.

src/dials/algorithms/symmetry/cosym/__init__.py
amyjaynethompson: Definitely not, it has been removed :)

src/dials/algorithms/correlation/analysis.py
amyjaynethompson: Added in a phil parameter for max dimensions - set default as 50 as important to have good trend for analysis, and put the smaller number into self.cosym_analysis._determine_dimensions

src/dials/command_line/correlation_matrix.py
amyjaynethompson: Have nested this as well as maximum_dimensions into dimensionality_assessment parameter

src/dials/algorithms/symmetry/cosym/__init__.py
amyjaynethompson: I think so too? :)

src/dials/command_line/correlation_matrix.py
amyjaynethompson: seed removed from correlation_matrix and cosym commandline programs and put into the cosym __init__ phil scope

newsfragments/2743.feature
amyjaynethompson: True - have made this change :)

jbeilstenedmands (approved):
Apart from the small suggestion to relocate the addition to the phil scope, this PR looks good to go, thanks!

src/dials/command_line/correlation_matrix.py
```diff
@@ -23,10 +23,16 @@
 
 phil_scope = iotbx.phil.parse(
     """\
-include scope dials.algorithms.correlation.analysis.phil_scope
-
-seed = 42
-  .type = int(value_min=0)
+include scope dials.algorithms.correlation.analysis.working_phil
+
+dimensionality_assessment {
```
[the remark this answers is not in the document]
amyjaynethompson: Has been moved :)

Issue #2744, opened 2024-09-25
Distortion maps are no longer applied


dagewa:
Distortion maps calculated by `dials.generate_distortion_maps` no longer have any effect because the pixel to millimeter strategy is not updated.

Compare
```bash
dials.import $(dials.data get -q x4wide)/*.cbf
dials.find_spots imported.expt
dials.index imported.expt strong.refl
```

with

```bash
dials.generate_distortion_maps $(dials.data get -q x4wide)/X4_wide_M1S4_2_0001.cbf dx=10 dy=20
dials.import $(dials.data get -q x4wide)/*.cbf lookup.dx=dx.pickle lookup.dy=dy.pickle
dials.find_spots imported.expt
dials.index imported.expt strong.refl
```

The results are identical, despite the hefty 10 and 20 pixel shifts.

The issue is probably because

```
dials.show imported.expt | grep "pixel to"
```
gives
```
pixel to millimeter strategy: ParallaxCorrectedPxMmStrategy
```

whereas what we want here is `OffsetPxMmStrategy`

dagewa:
Incomplete testing :disappointed: 

https://github.com/dials/dials/blob/7e48a28a193f87a184cadaa7d5a7d3dc79bed5a0/tests/command_line/test_generate_distortion_maps.py#L86-L87

dagewa:
Weird. While `dials.show` says the panel has a `ParallaxCorrectedPxMmStrategy`, the actual `imported.expt` says differently:

```
$ cat imported.expt | grep -A2 strategy
          "px_mm_strategy": {
            "type": "OffsetParallaxCorrectedPxMmStrategy"
          }
--
        "px_mm_strategy": {
          "type": "SimplePxMmStrategy"
        },
```
NB the first entry is for the panel, the second is for the hierarchy :roll_eyes:

dagewa:
The `dials.show` problem might be a symptom of the underlying issue. When the experiments are loaded using `reflections_and_experiments_from_files` it appears the `px_mm_strategy` is reset. If instead the experiments are loaded with `ExperimentList.from_file("imported.expt")`, then the strategy is correctly retained.

dagewa:
Actually it's before that - it happens in the options parser :scream: 

```
params, options = parser.parse_args(args=args, show_diff_phil=True)
print(params.input.experiments[0].data[0].detector[0])
```
gives
```
Panel:
  name: Panel
  type: SENSOR_PAD
  identifier: 
[...]

PR #2745, opened 2024-09-25
`dials.generate_distortion_maps`: much faster generation of elliptical distortion maps

dagewa:
Move the loop over pixels from Python to C++. For a large detector (Gatan K3, $4092 \times 5760$ pixels) this reduced the whole program run time from 2m56.584s to just 0m3.203s (i.e. 55 times speed improvement).

In addition, set sensible defaults for ellipse map generation, including automatic ellipse centring at the beam centre.


dagewa:
Nevertheless, the maps remain useless because of #2744

PR #2749, opened 2024-09-26
Replace (unmaintained) orderedset with ordered_set

rolandmas:
Hi, it seems that the orderedset module is abandoned by its author, and it started failing; in particular with Python 3.13 as discovered in Debian, but see also https://github.com/simonpercivall/orderedset/issues/36

There's an API-compatible replacement called ordered_set, see https://github.com/rspeer/ordered-set

This patch replaces the one with the other. It's already been used as part of the Debian packaging of dials with no influence on the build or the testsuite. I suggest applying it to upstream dials.


dagewa:
We should check performance. The author [comments](https://github.com/simonpercivall/orderedset/issues/36#issuecomment-1022578854) that `orderedset` is much faster. Most instances look likely to be small sets, but an exception might be [Ih_table.py](https://github.com/dials/dials/blob/main/src/dials/algorithms/scaling/Ih_table.py), used by scaling.

jbeilstenedmands:
Thanks @dagewa , I will review any potential performance impact in scaling

dagewa:
I checked this with 259 Eiger data sets from proteindiffraction.org, timing `dials.scale` jobs on main and on the pr. Here is a QQ-plot of the results, which show quite similar run times in either case:
qqplot
In fact, surprisingly, the pr branch jobs tended to finish a bit _quicker_ than the main branch jobs! That is shown by the data points tending to fall under the y=x line.

jbeilstenedmands (approved):
Thanks @dagewa for running that test, assuming the tests pass this looks good to go

dagewa:
Tests are failing because the change is needed in dxtbx as well. I'll make a PR to change that one first. When I ran my comparison I didn't see this failure because `ordered_set` and `orderedset` could both still be imported.

dagewa:
Hmm unfortunately this change fails here
https://github.com/dials/dials/blob/d5a455cbdbf661dd1072228a0f3b759f594dbcb7/src/dials/command_line/integrate.py#L519
with
```
In [7]: isets.index(experiments[0].imageset)
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[7], line 1
----> 1 isets.index(experiments[0].imageset)

File ~/sw/cctbx/conda_base/lib/python3.12/site-packages/ordered_set/__init__.py:246, in OrderedSet.index(self, key)
    233 """
    234 Get the index of a given entry, raising an IndexError if it's not
    235 present.
   (...)
    243     1
    244 """
    245 if isinstance(key, Iterable) and not _is_atomic(key):
--> 246     return [self.index(subkey) for subkey in key]
    247 return self.map[key]

File ~/sw/cctbx/conda_base/lib/python3.12/site-packages/ordered_set/__init__.py:247, in OrderedSet.index(self, key)
    245 if isinstance(key, Iterable) and not _is_atomic(key):
    246     return [self.index(subkey) for subkey in key]
--> 247 return self.map[key]

KeyError: (<scitbx_array_family_flex_ext.double object at 0x796e68faa2c0>,)
```

dagewa:
Ok, it's a little concerning that `ordered_set.OrderedSet` appears not to work with `ImageSet` objects, which is a use case that appears in `dials.integrate` and `dials.split_experiments`. I've made changes that work around that, but it is a shame that `ordered_set` is not a drop-in replacement for `orderedset` for us.

dagewa:
Merging now, after discussion with @phyy-nx

Issue #2751, opened 2024-09-28
`dials.rlv` ugly buttons and spacing


dagewa:
Did a fresh build, ended up with GTK 3, and now the buttons look nasty on the reciprocal lattice viewer:
![image](https://github.com/user-attachments/assets/c5564b85-b0a6-4a28-97d2-f6cd65eee83b)

This is related to the issue with CCP4 9.0, though in that case the spacing is even worse:
![image](https://github.com/user-attachments/assets/ab2acfe9-6795-47a0-9c6f-9a5a0b55af29)

dagewa:
Maybe it's not _really_ a bug, but I don't like it

PR #2753, opened 2024-09-30
Viewer widget sizing tweaks

dagewa:
With GTK3 there seem to be some issues with widget sizing and margins. This PR attempts to fix the worst of these issues. I would be interested to see how this looks on other platforms.

1. `dials.image_viewer`

The margins seem larger with GTK3. As a result the settings window filled the height of my screen. I reduced margins from 5 to 3 to make the window more compact. In addition I increased the width of the brightness slider.

imviewer-settings

2. `dials.reciprocal_lattice_viewer`

The `FloatSpin` control default size truncated the text entry box. This is even worse (for whatever reason) on CCP4 9.0's distribution of DIALS (see #2751). I set explicit sizes to get wider boxes.

rlv

This is still far from perfect. In particular the `rlv` controls are now quite wide. The second beam centre `FloatSpin` could go on the line below to improve that. But before doing that I'd be interested to know how this looks for other people.


dagewa:
The conversation [here](https://discuss.wxpython.org/t/my-personal-crusade-against-gtk-3-spinctrls-size/30506) covers the ugly spin control buttons, probably designed for touchscreens. If we're bothered to change the custom code linked in the discussion might recover GTK2-like controls, which looked prettier.

ndevenish:
This probably doesn't need to sync main, as I don't think any of this code is tested anyhow.

Have you tested this on older machines, or do you think that doesn't matter any more?

dagewa:
I've not tested on any other machine yet. I'm not too bothered about going back to check how it looks with GTK 2, but I would like to check if the changes look ok on Windows and Mac. The first I should be able to check myself.

dagewa:
The image viewer settings window looks quite compact, but acceptable I think on Windows
image

The RLV panel is a bit over-wide, but I think it would be okay if I got the second beam centre control on the line below
image

dagewa:
Yeah, that's better
image

dagewa:
Could someone please check this branch doesn't make the viewers look excessively horrible on Mac? I'm happy that it's an improvement on Linux and Windows.

dagewa:
Would like this in the release, if anybody wants to check Mac?

ndevenish:
image

PR #2754, opened 2024-09-30
Do polarization correction only for X-rays

### LINKED ISSUE #2734: Is the polarization correction relevant for electrons and neutrons?
The `lp` column in the reflection table contains contributions from both the Lorentz correction and the polarization correction. When we process electron diffraction datasets **kinematically**, we need to apply the Lorentz correction $\zeta \sin(2\theta)$. But I am not sure if the polarization correction is relevant for electrons; the electric field $E$ for X-ray is a vector field but electron's wave function is a scalar field (we ignore spins).

In the code below, `L` corresponds to the Lorentz correction while `P` is the polarization correction. Although "a value of 0.5 implies an unpolarized beam", only `P2` vanishes and `P4` remains non-zero as $(1+\cos^2 2\theta)/2$.

https://github.com/dials/dials/blob/1747513625b726af062658048dfb57db593539f8/src/dials/algorithms/integration/corrections.h#L30-L60

Admittedly, this is merely an resolution dependent scale factor, which can be partially absorbed into atomic B factors. Because $2\theta$ is very small in electron diffraction, the effect is not large, but if the right physics is without the polarization corection, we should disable it using the new probe type field in dxtbx (https://github.com/cctbx/dxtbx/issues/575).

I am not an expert on the physics. Does anyone know if the polarization correction is necessary or not?

dagewa:
Intended to fix #2734. Currently untested


biochem-fan:
> Compute the X-ray LP correction for a single reflection

I found this confusing. LP correction is Lorentz correction + polarization correction.
For neutron diffraction and electron diffraction, we don't apply polarization correction but apply the Lorentz correction.

Perhaps we should divide the LP correction into two functions or at least improve the phrasing in the comment.

e.g. "Compute the Lorentz correction for a single reflection. For X-ray, polarization correction is also included."

dagewa:
Yes, makes sense to split the functions to reduce confusion. At a higher level, for processing the corrections are always called via an `lp` method of a `CorrectionsMulti` object. But this should do the right thing for each case now.

biochem-fan:
@dagewa This looks OK to me. There are controversies regarding Lorentz correction for stills and/or Laue but that is beyond the scope of this PR.

e.g. 
- nXDS: https://journals.iucr.org/d/issues/2014/08/00/dz5332/index.html (Eq. 41 to 43)
- EVAL: https://journals.iucr.org/d/issues/2015/09/00/dz5376/index.html (Eq. 4)

dagewa:
Thanks, yes I remember this idea of a Lorentz correction for stills. I'm pretty sure @jmp1985 was aware of these papers when he wrote the correction code, so I was surprised to find that DIALS apparently sets $L=1.0$ for still shots. I haven't changed that in this PR, but maybe it is something we should look into as a separate issue. Likewise for Laue. I'm not sure what the current DIALS Laue processing routes are doing with the Lorentz correction ( :eyes: @toastisme @PrinceWalnut)

toastisme:
> Thanks, yes I remember this idea of a Lorentz correction for stills. I'm pretty sure @jmp1985 was aware of these papers when he wrote the correction code, so I was surprised to find that DIALS apparently sets L = 1.0 for still shots. I haven't changed that in this PR, but maybe it is something we should look into as a separate issue. Likewise for Laue. I'm not sure what the current DIALS Laue processing routes are doing with the Lorentz correction ( 👀 @toastisme @PrinceWalnut)

The Lorentz correction has to be done for every pixel at every wavelength for ToF data. To make it fast it's done a[s part of shoebox extraction](https://github.com/dials/dials/blob/main/src/dials/algorithms/scaling/tof/tof_scaling.h#L354) during integration.

PrinceWalnut:
Similarly, Laue-DIALS has a custom integration routine independent of the standard DIALS pipeline. From what I can tell, this PR only affects integration code, so it should not impact Laue-DIALS. Our integrator does not apply a Lorentz correction, and Polarization correction is handled by `careless` during scaling/merging, so we simply work with raw pixel intensities.

PR #2755, opened 2024-09-30
Make CMake the default build mode for dials

ndevenish:
By replacing the `--cmake` flag with a `--libtbx` flag and invert the CI flags so the full pipeline build passes it.

Also, change the way we call pytest to be compatible with the new libtbx approach; remove vestigial references to generating the pytest dispatchers.


ndevenish:
I think this needs real PR to run

ndevenish:
Okay, I think the remaining error is the wxpython windows error that we didn't track down yet

ndevenish:
And that has been fixed by updating to the newly release wxpython package! Everything is working now.

Issue #2757, opened 2024-10-02
Inconsistent experiment IDs after indexing


dagewa:
I'm trying to move more indexing tests to use `dials.data` and in doing so found a funny thing. Straightforward indexing of this data set:

```
dials.index $(dials.data get -q semisynthetic_multilattice)/{ag_strong_1_50.refl,ag_imported_1_50.expt}
```
leads to `indexed.{expt,refl}` with consistent experiment IDs:
```
>>> from dxtbx.model.experiment_list import ExperimentList
>>> from dials.array_family import flex
>>> el=ExperimentList.from_file("indexed.expt", check_format=False)
>>> print(el[0].identifier)
cd6e2f9b-23c7-4062-cea6-128303d81cf4
>>> refl=flex.reflection_table.from_file("indexed.refl")
>>> print(refl.experiment_identifiers()[0])
cd6e2f9b-23c7-4062-cea6-128303d81cf4
```

However, the indexing command used in the test:
```
dials.index $(dials.data get -q semisynthetic_multilattice)/{ag_strong_1_50.refl,ag_imported_1_50.expt}\
  indexing.method=real_space_grid_search\
  reflections_per_degree=10\
  n_macro_cycles=5\
  known_symmetry.unit_cell=54.3,58.3,66.5,90,90,90\
  known_symmetry.space_group=P212121\
  image_range=0,10\
  beam.fix=all\
  detector.fix=all\
  max_cell=70\
  index_assignment.method=local\
  nearest_neighbours=50
```
completes fine but gives inconsistent IDs:
```
>>> from dxtbx.model.experiment_list import ExperimentList
>>> from dials.array_family import flex
>>> el=ExperimentList.from_file("indexed.expt", check_format=False)
>>> print(el[0].identifier)
158302f1-2a08-c939-59ab-da9855947b34
>>> refl=flex.reflection_table.from_file("indexed.refl")
>>> print(refl.experiment_identifiers()[0])
12cac2cb-d7a1-8a19-740d-4f0f395e4b7c
```

dagewa:
The second ID, `12cac2cb-d7a1-8a19-740d-4f0f395e4b7c` is the one listed in the input experiments
```
$ head $(dials.data get -q semisynthetic_multilattice)/ag_imported_1_50.expt
{
  "__id__": "ExperimentList",
  "experiment": [
    {
      "__id__": "Experiment",
      "identifier": "12cac2cb-d7a1-8a19-740d-4f0f395e4b7c",
[...]

PR #2758, opened 2024-10-03
Move tests to dials data

dagewa:
Various tests are edited to use files on `dials-data` rather than `dials_regression`. In some cases the files are identical, in others the tests have been edited to use different files.

`dials_regression` still appears in 26 files under `dials/tests/`, but it is progress.


dagewa:
The two test failures are expected: one because of #2757, and the other because of the regression introduced by #2736.

dagewa:
Not sure why there are `xfel_regression` test failures and I don't have this checked out locally to look into it @phyy-nx @dwpaley any ideas?

phyy-nx:
Hi, so XFEL CI is passing on cctbx and I don't see how this PR would have affected it.  I started to try and reproduce the problem by running the kind of dials bootstrap action I see in dials/.github/workflows/build_test.yml, but I get this:

```
python modules/dials/installer/bootstrap.py update base --python 3.9 --libtbx
<snip>
Downloading  (5) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╸━ 393.1MB / 395.6MB @  34.5MB/s dbus                       3.0s
Extracting   (6) ━━━━━━━━━━━━╸━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━      60 /     298             at-spi2-core               2.6sterminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
Process failed with return code -6
```

So I'm not really sure how to help.  If @ndevenish could look that would be great, but we can chat about it in DIALS core tomorrow too.

phyy-nx:
Made a bit more progress.  I reset my terminal, sourced what I think is a clean python, and ran these commands, which let me reproduce the cosym error (using a precloned xfel_regression):

```
echo "pytest-md
          dials-data
          flexparser<0.4
          pytest-cov
          pytest-timeout" >> modules/dials/.conda-envs/linux.txt
python modules/dials/installer/bootstrap.py update base --python 3.9 --libtbx
python modules/dials/installer/bootstrap.py build --libtbx
source dials
conda install -yp ./conda_base bash distro git-lfs mpi4py openmpi pandas
libtbx.configure xfel_regression
mkdir tests
cd tests/
libtbx.python ../modules/xfel_regression/test/command_line/tst_xfel_merge.py modify_cosym
```

The problem is the program is producing `cosym_embedding_0.png` when it should be producing `cosym_embedding_1.png`.  The XFEL CI version of the build does produce the correct file.  I haven't yet examined the difference between the two builds.

All that said, I see the same problem with dials on the main branch, so none of this is related to this PR.

phyy-nx:
The MPI that is being picked up seems bad:

```
$ libtbx.python -c "from mpi4py import MPI"
  File "<string>", line 1, in <module>
ImportError: libcudart.so.11.0: cannot open shared object file: No such file or directory
```

For reference:
```
$ conda list | grep -i mpi
c-compiler                1.8.0                h2b85faf_1    conda-forge
cxx-compiler              1.8.0                h1a2810e_1    conda-forge
h5py                      3.12.1          nompi_py39h30a5a8d_102    conda-forge
hdf5                      1.14.3          nompi_hdf9ad27_105    conda-forge
mpi                       1.0                     openmpi    conda-forge
mpi4py                    4.0.1            py39h86f8d4e_0    conda-forge
openmpi                   5.0.5              hd45feaf_104    conda-forge
```

This explains why tst_mpi_abort.py is failing.

phyy-nx:
Switching from openmpi to mpich fixes the cosym error:

```
conda remove mpi openmpi mpi4py
conda install -c conda-forge mpich mpi4py
```

But gives a different error for tst_mpi_abort...

phyy-nx:
Ok, impi_rt fixes both tests:

```
conda remove mpi mpich mpi4py
conda install -c conda-forge mpi4py impi_rt
```

I think this isn't great news but it at least gets us to a solution.  I'll put in a PR for it separately.

dagewa:
Thanks @phyy-nx ! I'm at a workshop today so won't be at the core meeting, but I'll merge this one

ndevenish:
> The MPI that is being picked up seems bad:
> 
> ```
> $ libtbx.python -c "from mpi4py import MPI"
>   File "<string>", line 1, in <module>
> ImportError: libcudart.so.11.0: cannot open shared object file: No such file or directory
> ```

Was this via tweaking CI or on your own machine? I'm having reproducing this independently

phyy-nx:
For completeness, this appears to be a bug upstream in openmpi.  Will monitor to see if the test failures go away on their own.

Issue #767, opened 2019-04-26
Profile xia2/dials on NP high res ProtK dataset:


rjgildea:
- https://ispyb.diamond.ac.uk/dc/visit/cm23003-2/id/3474639
- /dls/i03/data/2019/cm23003-2/proteinasek/protk_1/protk_1_1_master.h5
- ProtK, 3600 images @ 0.1 deg, 0.99 Å at corners

rjgildea:
Using dials/latest dials.integrate fails with the following error:

/dls/mx-scratch/rjgildea/dials_13/protk_1_1_/001/DEFAULT/SAD/SWEEP1/integrate/13_dials.integrate_INTEGRATE.log

```
 Using multiprocessing with 20 parallel job(s)

easy_mp crash detected; subprocess trace: ----
Stacktrace:
exit code = -9
----------------------------------------------

Traceback (most recent call last):
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/build/../modules/dials/command_line/integrate.py", line 700, in <module>
    halraiser(e)
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/build/../modules/dials/command_line/integrate.py", line 698, in <module>
    script.run()
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/build/../modules/dials/command_line/integrate.py", line 370, in run
    reflections = integrator.integrate()
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/modules/dials/algorithms/integration/integrator.py", line 1254, in integrate
    self.reflections, _, time_info = processor.process()
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/modules/dials/algorithms/integration/processor.py", line 295, in process
    preserve_exception_message=True,
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/modules/dials/util/mp.py", line 172, in multi_node_parallel_map
    preserve_exception_message=preserve_exception_message,
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/modules/dials/util/mp.py", line 45, in parallel_map
    preserve_exception_message=preserve_exception_message,
  File "/dls/science/groups/scisoft/DIALS/CD/latest/dials-dev20190426/modules/cctbx_project/libtbx/easy_mp.py", line 628, in parallel_map
    result = res()
[...]

PR #2765, opened 2024-10-10
Fix beam.get_s0() errors in image viewer for polychromatic data.

toastisme:
This fixes `dials.image_viewer` failing for polychromatic data due to several calls to `beam.get_s0()`, which `Polychromatic` beams do not have. In each of these cases this is only to calculate angles, and so `beam.get_unit_s0()` gives the same result.


phyy-nx:
Hello, `get_unit_s0` doesn't seem to give the same result, as I discovered today working with MicroED geometry:

```
from dxtbx.model.detector import DetectorFactory
from dxtbx.model.beam import BeamFactory

detector = DetectorFactory.simple('UNKNOWN', 1000, (28,28), '+x', '-y', (0.014,0.014), (4000,4000))
panel = detector[0]
beam = BeamFactory.simple(0.0251)

print("MicroED")
print("Unit s0", panel.get_resolution_at_pixel(beam.get_unit_s0(), (0,0)))
print("s0", panel.get_resolution_at_pixel(beam.get_s0(), (0,0)))

detector = DetectorFactory.simple('UNKNOWN', 100, (50,50), '+x', '-y', (0.050,0.050), (1000,1000))
panel = detector[0]
beam = BeamFactory.simple(1.2)

print("Xray")
print("Unit s0", panel.get_resolution_at_pixel(beam.get_unit_s0(), (0,0)))
print("s0", panel.get_resolution_at_pixel(beam.get_s0(), (0,0)))
```

Output
```
MicroED
Unit s0 25.2686565554563
s0 0.6342432795419082
Xray
Unit s0 1.6506801238857844
s0 1.9808161486629412
```

Because of this the image image viewer's resolution display is off.  Is it too late to fix this for the current release?  @toastisme @ndevenish 
Thanks!

phyy-nx:
@toastisme would this work for tof?
```
$ git diff slip_viewer/tile_generation.py
diff --git a/src/dials/util/image_viewer/slip_viewer/tile_generation.py b/src/dials/util/image_viewer/slip_viewer/tile_generation.py
index 6850f8f95..e1f0ba029 100644
--- a/src/dials/util/image_viewer/slip_viewer/tile_generation.py
+++ b/src/dials/util/image_viewer/slip_viewer/tile_generation.py
@@ -438,7 +438,10 @@ def get_resolution(self, x, y, readout=None):
         beam = self.raw_image.get_beam()
         if detector is None or beam is None:
             return None
-        beam = beam.get_unit_s0()
+        try:
+            beam = beam.get_s0()
+        except AttributeError:
+            beam = beam.get_unit_s0()

         if readout is None:
             return None
```

dagewa:
Perhaps methods like `get_resolution_at_pixel` should take a `Beam` not merely a vector, so they can validate their data

toastisme:
Sorry this slipped my mind! @phyy-nx the s0 vector is still relevant for Polychromatic case for getting the max resolution, but here I think it makes most sense to take the minimum wavelength. Agree with @dagewa that modifying get_resolution_at_pixel might be the best approach. PR for this [#767](https://github.com/cctbx/dxtbx/pull/767)

PR #2769, opened 2024-10-10
Switch CI to Github Actions

ndevenish:
Since CI required an overhaul anyway. This should make managing/viewing a little easier, and help solve the parallel job issues we have on Azure.

Not yet done:
- [x] XFEL Regression
- [ ] Coverage


ndevenish:
So, it's short-circuited to download the release rather than build every time (for speed, to test), but all except the breakages from https://github.com/dials/dials/pull/2741 are passing now (with a patch for the `ubuntu == 20.04` test)

PR #2770, opened 2024-10-10
Cluster cosym coords

amyjaynethompson:
Addition of plots and cluster output to dials.correlation_matrix 

New Plots:
 - dimensionality assessment plot added to .html 
 - when data assessed to be of higher dimension than 2, additional plots for the coordinates are given for each possible combination of 2-axes 

Clustering and Output:
 - Use OPTICS algorithm to identify clusters and noise in the optimised cosym coordinates 
 - Colour code these in the graphs (note that colour scheme changed because otherwise ended up with white data points)
 - When turned on via phil parameter, output separate clusters as dials .expt/.refl files


src/dials/algorithms/correlation/analysis.py
```diff
@@ -216,7 +230,7 @@ def calculate_matrices(self):
             dims_to_test = self.params.dimensionality_assessment.maximum_dimensions
 
         if self.params.dimensions is Auto:
-            self.cosym_analysis._determine_dimensions(
+            self._dim, self._func = self.cosym_analysis._determine_dimensions(
```
jbeilstenedmands, line 233: If dimensions is not Auto, then we have a failure in L476 when calling the `plot_dims` function as `self._dim` and `self._func` are not defined.

src/dials/algorithms/correlation/plots.py
```diff
@@ -47,6 +47,41 @@ def add_node(node):
...
 
 
+def plot_dims(dims: list, funcs: list) -> dict:
+    """
+    Prepares a plotly-style plot of the dimension optimisation procedure.
+    Args:
+        dims(list): list of dimensions tested
+        funcs(list): resulting functional when the tested number of dimensions is used in the cos-angle clustering procedure
+    Returns:
+        d(dict): dimensionallity analysis plot saved as a dictionary for future plotting
```
jbeilstenedmands, line 57: Typo for `dimensionality`

amyjaynethompson: typo fixed :)

jbeilstenedmands (approved):
Thanks for implementing this and addressing the requested changes, looks good to go in pending the relevant tests passing 🙂

PR #2778, opened 2024-10-23
Reachability plot

amyjaynethompson:
Add reachability plot to dials.correlation_matrix which is very helpful at visualising outcome of OPTICS clustering algorithm on cosym coordinates. Include addition of 'xi' parameter which can be used to tailor how OPTICS interprets the reachability plot to define clusters.


src/dials/algorithms/correlation/plots.py
```diff
@@ -82,6 +82,69 @@ def plot_dims(dims: list, funcs: list) -> dict:
...
+    d = {
+        "reachability": {
+            "data": data,
+            "layout": {
+                "title": "OPTICS Reachability",
+                "xaxis": {"title": "Cluster-order of Datasets"},
+                "yaxis": {"title": "Reachability Distance Between Datasets"},
+            },
+            "help": """\
+Reachability plot from OPTICS analysis (M. Ankerst et al, 1999, ACM SIGMOD). As this is a density-
```
jbeilstenedmands, line 134: I think it would be good for us to also reference the scikit learn optics page here, so that people can read that to understand more about the parameters?

jbeilstenedmands (approved):
Looks good thanks, just the one minor point regarding docs which i think would be helpful but this is good to go as is.

src/dials/algorithms/correlation/plots.py
amyjaynethompson: Have added a link to the documentation page

PR #2781, opened 2024-10-28
Fix beam.get_s0() errors in image viewer for polychromatic data

toastisme:
Reopening [#2765](https://github.com/dials/dials/pull/2765) with the resolution methods changed to using the Beam model directly [#767](https://github.com/cctbx/dxtbx/pull/767). @phyy-nx could you check this?

This fixes `dials.image_viewer` failing for polychromatic data due to several calls to `beam.get_s0()`, which Polychromatic beams do not have.


phyy-nx:
Sorry for the belated response.  This works.  I tested on a monochromatic image and the resolution mouseover is the same with and without this code.

PR #2786, opened 2024-10-30
Fix symmetry not being applied consistently for multiple crystals post indexing 

toastisme:
This was discovered when processing time-of-flight data with multiple orientations and multiple crystals.

When indexing a dataset with two orientations with two crystals using 

`dials.index imported.expt strong.refl max_lattices=2`

Both crystals are initially indexed as `P-1`. The first is then reindexed to `P1` using `Indexer._apply_symmetry_post_indexing()`. But the second remains in `P-1`. Shown here in the crystal frame:

Screenshot from 2024-10-30 14-36-23

This is due to a bookkeeping error in `Indexer.index` where `n_lattices_previous_cycle` is set to the number of experiments, rather than the number of crystals. Changing to the latter fixes this case

Screenshot from 2024-10-30 14-35-32


src/dials/algorithms/indexing/indexer.py
```diff
@@ -540,7 +540,7 @@ def index(self):
                     )
                     break
 
-            n_lattices_previous_cycle = len(experiments)
+            n_lattices_previous_cycle = len(experiments.crystals())
```
jbeilstenedmands, line 543: Looking at where `n_lattices_previous_cycle` is used, should L577 also be updated to `elif len(experiments.crystals())` ?

jbeilstenedmands (approved):
Looks good thanks, assuming the tests all pass

PR #2788, opened 2024-10-31
Experiment identifiers added to dials.predict

yash4karan:
Previously, dials.predict did not assign experiment identifiers to the output reflection file when running dials.predict. This raise an error when trying to output HDF5-formatted files (using James Beilsten-Edmands's "h5_table_format" branch). This commit fixes the issue, and includes an updated _test_predict.py_ file to check for it in the future.


src/dials/command_line/predict.py
```diff
@@ -111,6 +111,9 @@ def run(self, args=None):
             predicted = flex.reflection_table.from_predictions(
                 expt, force_static=params.force_static, dmin=params.d_min
             )
+
+            # Assign an experiment identifier to reflections (this is required for saving HDF5 reflection files)
```
ndevenish, line 115:
```suggestion
            # Copy the experiment identifiers verbatim to the new table
```

ndevenish:
I can't see how to view the old runs while it's still running, but it looked like network errors fetching the data e.g. nothing to do with the PR

PR #2798, opened 2024-11-11
Line tool

dagewa:
3D ED users often want to set the beam centre by finding the mid point between Friedel pairs. This PR supports that use case by adding a line tool. While I was at it, I also threw in a line profile graph for free :wink: 

image


dagewa:
Unlikely to do any more on this now so will merge

PR #2799, opened 2024-11-12
Local indexing exp ids fix

dagewa:
This makes the `__call__` method of `AssignIndicesLocal` more like that in `AssignIndicesGlobal`. In particular, it ensures experiment IDs are set consistently, which `AssignIndicesGlobal` did, but `AssignIndicesLocal` did not.

There is also now a loop over imagesets in `AssignIndicesLocal`, which was not there before. I don't know if that was intentional.

There is plenty of code duplication between these classes. Perhaps this is due a refactor. It may also be worth adding a test for #2757. I want to see if CI passes first...

### DIFF
--- a/newsfragments/xxx.bugfix
+++ b/newsfragments/xxx.bugfix
@@ -1,0 +1,1 @@
+``dials.index``: fix inconsistent experiment IDs after indexing with ``index_assignment.method=local``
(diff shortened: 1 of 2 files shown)


dagewa:
Ok, it's more complicated than I thought. The loop over imagesets causes some test failures. Not sure why

dagewa:
Back in business now. Seems like a deficiency of the tests, not the code being tested

jbeilstenedmands (approved):
Thanks for the fix, sorry I had not seen the original issue that had been assigned to me.

I agree a refactor would be good, but let's get this in first.

dagewa:
Thanks, that should unlock #2758 ...

PR #2803, opened 2024-11-14
Switch XFEL CI from openmpi to impi_rt

phyy-nx:
Fixes XFEL CI failures

While found in #2758, probably not related to that PR


phyy-nx:
This didn't work and this appears to be a bug upstream in openmpi.  Will monitor to see if the test failures go away on their own.

PR #2804, opened 2024-11-14
`dials.ssx integrate`: do not write out zero length reflection list

dagewa:
If no reflections were integrated, we do not want `.{expt,refl}` output.

### DIFF
--- a/newsfragments/xxx.bugfix
+++ b/newsfragments/xxx.bugfix
@@ -1,0 +1,1 @@
+``dials.ssx_integrate``: do not output a reflection file if no reflections were integrated.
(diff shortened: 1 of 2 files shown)


jbeilstenedmands (commented):
Please do not merge this until I have chance to test a few things. Various workflows for xia2.ssx_reduce rely on being able to do things like xia2.ssx_reduce batch_*/integrated*.{expt,refl}, which I think was the motivation behind this behaviour in the first place. But there are probably other workarounds.

jbeilstenedmands (approved):
Thanks, this looks good. I also added the equivalent change so that `dials.ssx_index `does not output empty files, and I am satisfied that this won't now cause issues with `xia2.ssx`.

PR #2807, opened 2024-11-14
Null PR so we can test XFEL

ndevenish:
A load of our CI got hit by https://github.com/conda-forge/openmpi-feedstock/issues/185

It looks like this should be blocked from installation in https://github.com/conda-forge/admin-requests/pull/1164, so let's test that we don't need to manually pin here.

**Do not merge this.**


ndevenish:
@phyy-nx it looks like them marking it as broken solved the problem!

phyy-nx:
Excellent!

PR #2808, opened 2024-11-14
use fft and robust test for screw axis

kmdalton:
This PR updates the FFT-based screw axis detection to be more robust. The major changes are to
1) Exclude all possible peak locations from the null model
2) Test each peak location independently and combine the probabilities for each by taking the product
3) Use Wilcoxon rank-sum to test whether each peak is greater than the null distribution.


dagewa (approved):
This is great @kmdalton, thank you. It seems to work very nicely for my test case. We really should test over many data sets to determine whether to change the default to the Fourier method. In the meantime, let's just get this in.

ndevenish:
You merged instead of squashed, so I flattened as https://github.com/dials/dials/commit/ebb50ee4954e152d45c9dd9227ddc0fe896896ff

kmdalton:
Thanks @ndevenish. I'll squash next time!

PR #2813, opened 2024-11-21
Weight the cosym objective function

jbeilstenedmands:
The cosym objective function should always have some form of weights matrix, in order to discard the diagonal element residuals (where `r_ii` is set to 0) and elements where it was not possible to calculate a pairwise correlation (for sparse cases). This PR adds a constant weights matrix of 0's and 1's for the current default case where `weights=None` (i.e. not `weights=count` or `weights=standard_error`).
However, for sparse cases, including some of the test cases, reliability weights are preferable, so this PR also changes the default to `weights=count`, where count is the number of pairs of observations used in a pairwise correlation calculation, to improve the results for sparse cases while having minimal impact on non-sparse cases.


jbeilstenedmands:
Note this PR causes a failure in xfel_regression due to the change of scale of the weights matrix, which can be handled by updating `test/util/tst_TargetWithFastRij.py`:
```
     test_instance = TargetWithFastRij(test_data_path=self.test_data_path)
     rij_test, wij_test = test_instance._compute_rij_wij()
+    # dials now normalises the wij_matrix: https://github.com/dials/dials/pull/2813
+    # so rescale in order to compare.
+    import numpy as np
+    scale = np.sum(wij_test) / np.count_nonzero(wij_test)
+    wij_dials *= scale
```
But I don't yet have access to make a corresponding PR to that repository.

phyy-nx:
Fixed in xfel_regression!  Thanks @jbeilstenedmands.  I've also added you so you can make commits there.

PR #2816, opened 2024-11-22
Feature: add split= to dials.import

### LINKED ISSUE #2815: dials.import: add simple mechanism to import data set as 3 chunks
Screenshot 2024-11-22 at 06 51 10

Vector data collection at 3 points, 3600 at each => 10800 in total

```
dials.import (data) image_range=1,3600 image_range=3601,7200 image_range=7201,10800
```

only imports the last range, I ended up with

```
export DATA=/dls/i04/data/2024/mx39148-1/TestThermolysin/Thermolysin_12/Thermolysin_12_3_master.h5

dials.import $DATA image_range=1,3600 output.experiments=imp0.expt
dials.import $DATA image_range=3601,7200 output.experiments=imp1.expt
dials.import $DATA image_range=7201,10800 output.experiments=imp2.expt

dials.combine_experiments imp0.expt imp1.expt imp2.expt
dials.find_spots combined.expt
```

which is _horrible_ 🤣

graeme-winter:
Syntax of split=start,end,size to split imported data into given chunk sizes. Also allows subset of the data to be used i.e. if start > 1 or end < end.

Fixes #2815


biochem-fan:
Perhaps adding a single argument shorthand of `split=size` == `split=1,last,size` would be useful for scripting because each pin usually contains different number of crystals and we do not one to check the number with `h5dump`.

newsfragments/2816.feature
```diff
@@ -0,0 +1 @@
+``dials.import``: add syntax for splitting a long scan into chunks e.g. for sequential data collections on small crystals. New syntax is split=start,end,size for example split=1,9000,90 for 100 x 90 image scans
```
biochem-fan, line 1: "100 x 90 image" is ambiguous: 90 sweeps of 100 frames each, or vice versa?

src/dials/command_line/dials_import.py
```diff
@@ -115,6 +115,10 @@ def _pickle_load(fh):
       .help = "A directory with images"
       .multiple = True
 
+    split = None
+      .type = ints(size=3)
+      .help = "Scan split instructions"
```
biochem-fan, line 120: Let's make this a bit more concrete: "Scan split instructions of three numbers: start, end, frames_per_chunk. start and end are 1-indexed."

graeme-winter:
> Perhaps adding a single argument shorthand of `split=size` == `split=1,last,size` would be useful for scripting because each pin usually contains different number of crystals and we do not one to check the number with `h5dump`.

This was a good suggestion, I have implemented it

src/dials/command_line/dials_import.py
```diff
@@ -885,6 +889,37 @@ def do_import(
...
+    if params.input.split:
+        if len(params.input.split) == 3:
+            split_start, split_end, step = params.input.split
+            split_start -= 1
+        elif len(params.input.split) == 1:
+            split_start = 0
+            split_end = 0
+            step = params.input.split[0]
+        else:
+            sys.exit("split=blocks or split=start,end,blocks")
```
biochem-fan, line 903: blocks => frames_per_block

biochem-fan:
I tested this on my small-wedge dataset from SPring-8 BL41XU.
It seems working but I found a corner case.

`split=1,5,3` gives two sweeps, one with frames 1 to 3 and the other with frames 4 to 6 (1-indexed). That is, when the frames_per_chunk does not divide the number of frames in the image range, the last sweep goes beyond the second argument. Shouldn't we truncate the last sweep at the specified frame, perhaps with a warning message?

graeme-winter:
> I tested this on my small-wedge dataset from SPring-8 BL41XU. It seems working but I found a corner case.
> 
> `split=1,5,3` gives two sweeps, one with frames 1 to 3 and the other with frames 4 to **6** (not 5!), 1-indexed. That is, when the `frames_per_chunk` does not divide the number of frames in the image range, the last sweep goes beyond the second argument. Shouldn't we truncate the last sweep at the specified frame, perhaps with a warning message?

Added failing test, will add fix later

biochem-fan:
@graeme-winter Thanks for the update. I confirmed the truncation worked.
The PR looks good to me now.

graeme-winter:
Current failing test in xfel-regression unrelated to this change set => will merge when other tests are ✅

jbeilstenedmands (approved):
Thanks, I think it makes sense to add this as a separate parameter rather than overload an existing parameter with more logic, I haven't tested on example datasets but the logic looks correct here.
As an aside, the fact that you can't specify multiple 'image_range=' feels like a bug, but not worth addressing here as it probably has other implications.

PR #2838, opened 2024-12-19
Avoid `FutureWarning` in test

dagewa:
```
pytest --regression -s -W error::FutureWarning tests/algorithms/scaling/test_scaler.py::test_update_error_model
```
fails with
```
FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value '<scitbx_array_family_flex_ext.double object at 0x7f33849ee440>' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
```

It seems this is an artefact of a `MagicMock` used in the tests. Changing this to return a `numpy` array fixes the issue.

### DIFF
--- a/newsfragments/xxx.misc
+++ b/newsfragments/xxx.misc
@@ -1,0 +1,1 @@
+Avoid a ``FutureWarning`` in ``test_scaler.py``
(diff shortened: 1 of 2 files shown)


jbeilstenedmands (approved):
Unrelated to this change, I note the windows tests are failing due to gemmi issues, is this revealing an issue with https://github.com/dials/dials/pull/2834 or have the changes just not been picked up here for some reason?

dagewa:
Hmm, it does look like it pulled in gemmi 0.7.0. I'm away from my Windows PC so I think I'll just re-pin gemmi to <0.7 then pick this up again in the new year.

PR #2842, opened 2025-01-15
Copy create_installer.py from libtbx and update for Python 3.12

ndevenish:
The version of create_installer.py in libtbx doesn't work with python 3.12, which is now the default python chosen by bootstrap. Move it here, so that we have a script to generate the installer more inside DIALS.


ndevenish:
This was causing our internal latest build/deploy pipelines to fail.

Issue #150, opened 2016-06-07
dials.find_spots should use a chunksize commensurate with #cores and #images


graeme-winter:
i.e. chunksize=auto by default such that nframes / (ncores*chunksize) is integer (i.e. no remainder)

graeme-winter:
Agree with RG that this should be fixed in DIALS principally, will copy results in but it can make a big difference to the time taken to find spots....

graeme-winter:
```
10 Time Taken: 80.592925
20 Time Taken: 77.165398
30 Time Taken: 68.305521
40 Time Taken: 82.535501
50 Time Taken: 70.923461
60 Time Taken: 82.675676
70 Time Taken: 91.532685
80 Time Taken: 100.463441
90 Time Taken: 65.609781
```

chunk size for 1800 frames / 20 cores

graeme-winter:
```
10 Time Taken: 67.411552
20 Time Taken: 67.646297
30 Time Taken: 61.639138
40 Time Taken: 76.718679
50 Time Taken: 66.580639
60 Time Taken: 78.578976
70 Time Taken: 87.189984
80 Time Taken: 95.329265
90 Time Taken: 60.818384
```

same for CBF files - some scope for usefulness here

graeme-winter:
![cs](https://cloud.githubusercontent.com/assets/1241716/15863673/d9d5ce42-2ccb-11e6-8407-95bced8ccaba.png)

graeme-winter:
30, 90 chunk size up to 30% faster than suboptimal combinations.

rjgildea:
The current solution has the side effect of removing the intermediate printing of progress:

```
Found 788 strong pixels on image 1
Found 731 strong pixels on image 2
Found 712 strong pixels on image 3
Found 716 strong pixels on image 4
Found 743 strong pixels on image 5
Found 885 strong pixels on image 6
Found 871 strong pixels on image 7
Found 862 strong pixels on image 8
Found 862 strong pixels on image 9
etc...
```

E.g. spotfinding on 4000 image sweep with nproc=8, chunksize=auto sets chunksize=500 -> no output for a very long time!

```
$ dials.find_spots datablock.json nproc=8
DIALS 1.dev.343-g71311a5
The following parameters have been modified:

spotfinder {
  mp {
    nproc = 8
  }
}
input {
  datablock = datablock.json
}

Setting spotfinder.filter.min_spot_size=3
Configuring spot finder from input parameters
[...]

PR #2848, opened 2025-02-10
Fix chunk size calculation

graeme-winter:
Before would always leave a little remainder to work on one CPU once all the others are finished: now works as one would expect


graeme-winter:
Before:

```console
Finding spots in image 1 to 36000...
Setting nproc=112
Setting chunksize=107
Extracting strong pixels from images
 Using multiprocessing with 112 parallel job(s)
```

>>> 36000 % (112 * 107)
48

This leaves a remainder of 48 images which are processed in a single core job at the end.

After:

```
Finding spots in image 1 to 36000...
Setting nproc=112
Setting chunksize=322
Extracting strong pixels from images
 Using multiprocessing with 112 parallel job(s)
```

>>> 112 * 322
36064

So this will _just run_ all the jobs without hanging around for the last one to complete.

jbeilstenedmands:
I agree that leaving a little remainder is not clever and needs addressing. Doing some archaeology, the current logic was implemented to address this ticket: https://github.com/dials/dials/issues/150
Are some of the concerns there still valid? If you processed your dataset with nproc=10 and the proposed changeset, the chunksize would be 3600 and it would be a while until you got any feedback. Maybe there is a logic error in the 'while' statement that should be addressed?

graeme-winter:
> I agree that leaving a little remainder is not clever and needs addressing. Doing some archaeology, the current logic was implemented to address this ticket: #150 Are some of the concerns there still valid? If you processed your dataset with nproc=10 and the proposed changeset, the chunksize would be 3600 and it would be a while until you got any feedback. Maybe there is a logic error in the 'while' statement that should be addressed?

I'll have a look later to see if I can make a more minimal change to make the logic correct here

jbeilstenedmands:
Ok, I think this is what is needed to add the right amount extra to the chunk size, to avoid small leftovers but otherwise keep the behaviour. Would welcome someone else to verify @graeme-winter . Easiest to verify by setting nproc to a number that n_images won't fully divide by, e.g. nproc=7.

graeme-winter:
> Ok, I think this is what is needed to add the right amount extra to the chunk size, to avoid small leftovers but otherwise keep the behaviour. Would welcome someone else to verify @graeme-winter . Easiest to verify by setting nproc to a number that n_images won't fully divide by, e.g. nproc=7.

I have just run a case with 1800 images / 16 cores: the main version runs blocks of 28 so a remainder of 8 images, on branch 29 per block so we do not have an annoying remainder. The time difference is minute because the machine is fast but I can see this being valuable.

graeme-winter:
Change set looks good to me

PR #2854, opened 2025-02-14
Allow output of excluded data sets from `dials.cosym`

dagewa:
For cases with multiple crystal isoforms with different unit cells, it may be useful to set aside the excluded data sets from a `cosym` job so that these can be investigated separately (i.e. with another `cosym` job). This PR allows that optional functionality and provides a test for it.

### DIFF
--- a/newsfragments/xxx.feature
+++ b/newsfragments/xxx.feature
@@ -1,0 +1,3 @@
+``dials.cosym``: Add the option ``output.excluded=True`` to set aside
+the data sets that cosym rejected, so that these can more easily be
+investigated separately.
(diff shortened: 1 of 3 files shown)


dagewa:
Now I realise that `dials.cluster_unit_cell` does things correctly, by putting all data into consistent minimum cells for clustering, then my immediate need for this change to `cosym` is reduced. Nevertheless, I think it could be useful to optionally not throw away excluded data, so I'll leave this open.

ndevenish:
Any reason not to merge in?

jbeilstenedmands (approved):

PR #2855, opened 2025-02-17
Deprecated `pkg_resources`

dagewa:
`pkg_resources` is deprecated. This draft PR contains one change that fixes a `DeprecationWarning` raised during dxtbx tests (see https://github.com/cctbx/dxtbx/issues/784). Other commits that retire `pkg_resources` can be added here.


dagewa:
It looks like the other uses of `pkg_resources` are also mainly about iterating overt entry points, so could be fixed similarly

dagewa:
Ok, following the lead of https://github.com/cctbx/dxtbx/pull/727 this seems to be the wrong fix for Python > 3.10

dagewa:
It looks like this is only for when Python 3.10 is minimum, but I'm not sure when that is, @ndevenish?

ndevenish:
That is now, as of early Jan

dagewa:
Ok, I guess this is ready then

PR #2867, opened 2025-03-07
Support iso-resolution lines (circles, ellipses and hyperbola) on non-coplanar and/or two_theta != 0 detectors

biochem-fan:
This fully fixes an ancient issue #66 (from 2016!) by supporting iso-resolution lines (circles, ellipses and hyperbola) on non-coplanar and/or two_theta != 0 detectors.

Previously, the main, analytical code path supported circles and ellipses but not hyperbola (large two_theta) and non-coplanar detectors. In #1899, @dagewa introduced a more general code path but it was invoked only on a very few detectors like PILATUS 12M. In this commit, the analytical but limited code path was removed in favor of the general code path.

I also fixed a minor bug in the general code path, where the curves were not closed.

## Examples

- Hyperbola on RIGAKU HyPix, two_theta >> 0:  
Image

- Discontinuous rings on CITIUS, where panel rows are at different depths:  
Image

## Potential drawbacks

- Curves are less smooth than analytical circles/ellipses; [can be tuned by the `n_rays` parameter](https://github.com/dials/dials/issues/66#issuecomment-2703328612).
- Computationally slightly more expensive (performs `n_rays * ring` ray intersections) but this should be negligible


biochem-fan:
Also confirmed on `image_examples/DLS_I23_P12M_thau_00001.cbf.gz`, where `projection_2d` is defined.

dagewa (approved):
Thanks! This is great. Nice to resolve an issue near the front of the list :smile:

biochem-fan:
@dagewa Thank you very much for approval.

PR #2876, opened 2025-03-24
Hierarchical clustering linkage updates

amyjaynethompson:
Linkage parameter has now been added as a tailorable phil parameter 

Ward linkages have been found to be more robust and thus have been made the new default behaviour (previously used only average linkages) 

Average linkages have been retained as an option


jbeilstenedmands (approved):
Thanks for adding this, I checked the relevant xia2 tests and they pass too 👍

Issue #2886, opened 2025-04-03
`dials.rlv` middle mouse drag not working


dagewa:
On an updated build on Ubuntu:
```
dict_keys([<class 'list'>, <class 'OpenGL.arrays.lists.c_double_Array_16'>, <class 'ctypes.c_double'>])
ArgumentError: argument 7: TypeError: No array-type handler for type _ctypes.CArgObject (value: <cparam 'P' (0x75c8c29d5420)>) registered
Traceback (most recent call last):
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/util/wx_viewer.py", line 292, in OnMouseMotion
    self.OnMiddleDrag(event)
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/util/wx_viewer.py", line 279, in OnMiddleDrag
    self.OnTranslate(event)
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/util/wx_viewer.py", line 607, in OnTranslate
    objx, objy, objz = glu.gluUnProject(
                       ^^^^^^^^^^^^^^^^^
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.12/site-packages/OpenGL/latebind.py", line 63, in __call__
    return self.wrapperFunction( self.baseFunction, *args, **named )
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcx32934/sw/cctbx/conda_base/lib/python3.12/site-packages/OpenGL/GLU/projection.py", line 55, in gluUnProject
    result = baseFunction(
            ^^^^^^^^^^^^^
ctypes.ArgumentError: argument 7: TypeError: No array-type handler for type _ctypes.CArgObject (value: <cparam 'P' (0x75c8c29d5420)>) registered
```

dagewa:
Probably this? https://github.com/mcfletch/pyopengl/commit/b49af26c615236ebc29cf125a8315091482a4a2a

dagewa:
This might be useful https://github.com/silx-kit/silx/pull/3982

PR #2888, opened 2025-04-03
Fix `dials.rlv` translate on Python 3.12

dagewa:
Fix for older versions of pyopengl, as per https://github.com/silx-kit/silx/pull/3982


ndevenish:
I'm unclear whether we can avoid this by pinning to a newer pyopengl? It looks like conda-forge only has 3.1.7 but this has some backports?

dagewa:
That was unclear to me to, and I admit I didn't try

dagewa:
No, it seems that with `pyopengl==3.1.7` and Python 3.12, I still get the error as per #2886

ndevenish:
This seems like a reasonably proportionate response then

dagewa:
Looks like the fix [here](https://github.com/mcfletch/pyopengl/commit/b49af26c615236ebc29cf125a8315091482a4a2a) is tagged to the 3.1.8 release. Maybe that would get to conda-forge at some point soon. I guess we can merge this and try to remember to revisit this?

ndevenish:
Sounds good

PR #2890, opened 2025-04-04
dials.integrate - add a specialised inflight integrator to overcome high memory usage

jbeilstenedmands:
Status: Proof of principle draft, for visibility.
Todo:
- [ ] Create separate classes for new behaviour rather than hacking onto existing data structures
- [ ] Add profile fitting (just summation at the moment).

This PR is an attempt to overcome the very high memory usage in dials for datasets with high mosaicity (see e.g. https://github.com/dials/dials/issues/659), for 'standard' rotation experiments.
It defines a new integrator class (`integrator=inflight`) which is a specialisation for data from counting detectors (i.e. what are used almost exclusively at synchrotrons) and constant 3d background model (our default choice). Rather than storing three arrays in the shoebox (data, background, flags) which are the size of the shoebox volume, the intensity summation and foreground/background/mask calculations are done as the data are loaded, while the background data is stored as a histogram, thus avoiding the need to allocate any of the three arrays. I plan to add the profile fitting in here as well.

To compare to the existing integrator, the two commands are:
`dials.integrate integrator=inflight refined.*`
which gives equivalent results to
`dials.integrate profile.fitting=False refined.*`

As an example, I tested on a mosaic dataset with sigma_m ~ 0.4, for the standard integrator I was getting a memory usage of ~13GB just running with nproc=1, whereas for the inflight integrator, I could use all the cores of my machine and each process was taking ~400MB.  When using the same nproc for both integrators, the calculation time is approximately the same.

### DIFF
--- a/src/dials/algorithms/background/glm/creator.h
+++ b/src/dials/algorithms/background/glm/creator.h
@@ -20,7 +20,6 @@
 #include <dials/model/data/image_volume.h>
 #include <dials/error.h>
 #include <unordered_map>
-#include <iostream>
 
 namespace dials { namespace algorithms {
 
@@ -276,7 +275,6 @@ namespace dials { namespace algorithms {
       // Compute the background
       double mean_background = result.mean();
       sbox.mean_background = mean_background;
-      std::cout << "Mean bg : " << mean_background << std::endl;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


dagewa:
This seems like an excellent step in the right direction :smile:

PR #2904, opened 2025-04-25
Add manual dispatch of build_test workflow

dwpaley:
Run the workflow manually on a branch with: Actions->select the action->Run workflow->select the branch.


ndevenish:
No issues with this, and normally I'd be happy for this sort of stuff to be pushed as a `MNT:` commit directly

ndevenish:
Yes! Obviously I was remembering the case where the manual dispatch was only present in the workflow on a branch, in which case there wasn't a UI way to trigger it.

Issue #2902, opened 2025-04-25
Freshly bootstrapped libtbx builds failing tests


ndevenish:
For a few days now, format registry and all other information stored in entry-points seems to be broken when running tests e.g.:

```
        # Starting at "Format" and using any potential prioritisation information
        # look for any path through the DAG of formats, stopping at the first
        # accepting leaf node
>       for format in sorted(_format_dag["Format"], key=format_sort):
E       KeyError: 'Format'

../dials-dev20250425/modules/dxtbx/src/dxtbx/format/Registry.py:127: KeyError
```

Initial investigation looks likely correlated to new setuptools release e.g. https://github.com/dials/dials/pull/2901 to temporarily fix.

Suspect method of injecting package info into libtbx-dispatchers without touching conda-base has broken.

ndevenish:
For info @dwpaley I believe it is this workaround that has broken (and the corresponding [dxtbx](https://github.com/cctbx/dxtbx/blob/d4daf99574c914b9e008a812504b5900144063a9/libtbx_refresh.py#L22-L80):

https://github.com/dials/dials/blob/9bc589b29ec40b073705f923cb6da767e89d9e55/libtbx_refresh.py#L73-L124

ndevenish:
https://github.com/dials/dials/pull/2901 should fix temporarily, by pinning setuptools. This should work now that I added the missed _linux_ dependency change 🤦.

dwpaley:
Thanks for making this issue. Looks like the libtbx_refresh script assumes that the package metadata (entrypoints.txt and so on) will be in e.g. `modules/dxtbx/src/dxtbx.egg-info` as in a "legacy editable" installation. I'm trying to track down the new location. For a DIALS build the answer is a place like `conda_base/lib/python3.12/site-packages/dxtbx-3.24.dev0.dist-info`, but I'm not yet sure about a libtbx build with new setuptools. Will update again soon.

ndevenish:
Yes, I think the solution will involve working out why it was working in "legacy" mode now and how editable is supposed to work... my understanding was that as long as we used pip to make it, it should have been doing the "current" method.

dwpaley:
Hmm, maybe if the libtbx build was sneaking something into the environment... https://github.com/cctbx/cctbx_project/blob/6538acd62ef391ae61756e76ff427fe3ab2c3b3b/libtbx/env_config.py#L2188-L2200

[...]

PR #2906, opened 2025-04-29
libtbx_refresh.py: find dist-info in the install directory instead of src

dwpaley:
Fixes dials/dials#2902


ndevenish:
Ah, I looked at the failures this evening but forgot about that nuance

dwpaley:
right, me too :) any thoughts on https://github.com/dials/dials/pull/2906/commits/6172520ef381adb7e3d2e84aada53ac7733793d2? Is it sufficient to put the installed location in the python path, or does src also need to be there explicitly? (we'll see if the tests pass...)

ndevenish:
`install -e` for non-legacy should be dropping a .pth to point to the sources. But If remain unsure exactly which mix of behaviours we are getting here.

dwpaley:
Yes, I have a file inside `build/dials/lib/python3.10/site-packages` named `__editable__.dials-3.22.dev0.pth` and pointing to `modules/dials/src` so it should be sufficient to add just the site-packages inside build.

libtbx_refresh.py
```diff
@@ -101,14 +120,15 @@ def _install_setup_readonly_fallback(package_name: str):
...
     env = _get_real_env_hack_hack_hack()
 
-    # Update the libtbx environment pythonpaths to point to the source
-    # location which now has an .egg-info folder; this will mean that
-    # the PYTHONPATH is written into the libtbx dispatchers
-    rel_path = libtbx.env.as_relocatable_path(import_path)
-    if rel_path not in env.pythonpath:
-        env.pythonpath.insert(0, rel_path)
+    # As of PEP 660, the package metadata (dist-info) goes in the install dir,
+    # not the source dir. Add this location to the python path too.
```
ndevenish, line 124: I think with that latest change this is now not "too"
```suggestion
    # not the source dir. Add this location to the python path.
```

ndevenish:
Otherwise, I think this looks good to merge dxtbx, then revert the branch changing here for final testing?

ndevenish:
Thank you very much for your work on this!

Issue #2647, opened 2024-04-16
dials.import_xds shows no signs of working


graeme-winter:
I am not sure how it should work so I check

```
ws448 xds :( $ dials.import_xds --help
usage: usage: dials.import_xds [options] (SPOT.XDS|INTEGRATE.HKL)
```

OK then:

```
ws448 xds :) $ dials.import_xds INTEGRATE.HKL
Traceback (most recent call last):
  File "/dls_sw/apps/dials/dials-v3-18-1/build/../modules/dials/src/dials/command_line/import_xds.py", line 463, in <module>
    run()
  File "/dls_sw/apps/dials/dials-v3-18-1/conda_base/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/dls_sw/apps/dials/dials-v3-18-1/build/../modules/dials/src/dials/command_line/import_xds.py", line 459, in run
    script.run(args)
  File "/dls_sw/apps/dials/dials-v3-18-1/build/../modules/dials/src/dials/command_line/import_xds.py", line 441, in run
    importer(params, options)
  File "/dls_sw/apps/dials/dials-v3-18-1/build/../modules/dials/src/dials/command_line/import_xds.py", line 185, in __call__
    raise RuntimeError("No XDS file found")
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: No XDS file found
```

YOU ARE HOLDING IT WRONG!!1!

OK I will try again

```
ws448 xds :( $ dials.import_xds xds_file=INTEGRATE.HKL
The following parameters have been modified:

input {
  xds_file = "INTEGRATE.HKL"
}

usage: usage: dials.import_xds [options] (SPOT.XDS|INTEGRATE.HKL)

options:
  -c, --show-config     Show the configuration parameters.
  -a ATTRIBUTES_LEVEL, --attributes-level ATTRIBUTES_LEVEL
                        Set the attributes level for showing configuration
                        parameters
  -e EXPERT_LEVEL, --expert-level EXPERT_LEVEL
                        Set the expert level for showing configuration
                        parameters
  -h, --help            Show this help message and exit. Can be specified
                        multiple times to increase verbosity.
  -v                    Increase verbosity (can be specified multiple times)
  --phil FILE           PHIL files to read. Pass '-' for STDIN. Can be
                        specified multiple times, but duplicates ignored.
```

Nope... 

OK so the other example was SPOT.XDS, let's test that 

```
[...]

Issue #2205, opened 2022-08-17
dials.import_xds needs a better interface and documentation


dagewa:
A user is having trouble with this. I looked, and now I am having trouble with it.

The documentation is minimal, consisting of a single usage string (or a "usage: usage:" string? So helpful, it was named twice):
```
usage: usage: dials.import_xds [options] (SPOT.XDS|INTEGRATE.HKL)
```

Things that work

- `dials.import_xds $(dials.data get -q centroid_test_data)` - creates `xds_models.expt` from `INTEGRATE.HKL`
- `dials.import_xds $(dials.data get -q centroid_test_data) xds_file=INTEGRATE.HKL` - creates `xds_models.expt` from `INTEGRATE.HKL`
- `dials.import_xds $(dials.data get -q centroid_test_data)/SPOT.XDS method=reflections` - creates `spot_xds.refl` from `SPOT.XDS`

Things that fail with tracebacks

- `dials.import_xds $(dials.data get -q centroid_test_data) method=reflections`
- `dials.import_xds $(dials.data get -q centroid_test_data) xds_file=SPOT.XDS`
- `dials.import_xds $(dials.data get -q centroid_test_data)/SPOT.XDS`
- ...

dagewa:
BTW the user reports that doing things the other way round - that is spot-finding and indexing in DIALS, then _exporting_ to XDS - works very well and has been helpful to work around `dials.integrate` memory errors...

graeme-winter:
I also find this annoying, as noted I made a duplicte issue

erikssod:
I had a play with this, and I could get the following to work: 



```
$ dials.data get -q centroid_test_data
$ cd path/centroid_test_data
$ centroid_test_data]$ dials.import_xds .
--------------------------------------------------------------------------------
Read 1 experiments from ./INTEGRATE.HKL
--------------------------------------------------------------------------------
Experiment 0
  format: <class 'dxtbx.format.Format.Format'>
  type: <class 'dxtbx_imageset_ext.ImageSequence'>
  num images: 900
--------------------------------------------------------------------------------
Writing experiments to xds_models.expt

$ centroid_test_data]$ dials.import_xds INTEGRATE.HKL xds_models.expt method=reflections
The following parameters have been modified:

input {
  method = experiment *reflections
}

Read 174911 reflections from INTEGRATE.HKL file............................1.64s
Reindex matrix:
1 0 0
0 1 0
0 0 1
Reindexed 174911 reflections...............................................0.01s
[...]

PR #2913, opened 2025-05-07
Better import of integrate.hkl from xds

CV-GPhL:
I've tested that quite  a bit over a large number of examples ... and am fairly confident it is doing "the right thing" (or: doing it at least better than the previous version). Using the resulting reflection list with dials.scale seems to work well now.


dagewa:
Thanks! Linking #2647 and #2205 here, as they may be relevant

CV-GPhL:
Sorry - something went wrong with the pull request here (still finding my way with the correct procedure). I'll rework it ... thanks.

PR #2914, opened 2025-05-07
Improvements on XDS/INTEGRATE.HKL import

CV-GPhL:
I've tested that quite a bit over a large number of examples ... and am fairly confident it is doing "the right thing" (or: doing it at least better than the previous version). Using the resulting reflection list with dials.scale seems to work well now.

The command-line working for me now is

  % dials.xds_import /where/ever/directory/with/XDS/output/.


dagewa:
Thanks! Linking https://github.com/dials/dials/issues/2647 and https://github.com/dials/dials/issues/2205 here, as they may be relevant

src/dials/command_line/import_xds.py
```diff
@@ -125,14 +154,43 @@ def __call__(self, params, options):
...
+        table["partiality"] = peak
+        table["profile.correlation"] = corr
+
+        table.centroid_px_to_mm(self._experiments)
+
+        # same as prf?
+        table["intensity.sum.value"] = iobs * peak / rlp
+        table["intensity.sum.variance"] = flex.pow2(sigma * peak / rlp)
+        table.map_centroids_to_reciprocal_space(self._experiments)
+        table["intensity.scale.value"] = flex.double(table.size(), 1.0)
```
jbeilstenedmands, line 176: In testing this PR, I found that having the `"intensity.scale.value"` column present caused some issues, as some programs (e.g. `dials.symmetry`) use this to assume that the data was scaled, which then caused issues due to the zero "intensity.scale.variance". It would be safer to remove the three columns `"intensity.scale.value"`, `"intensity.scale.variance"` and `"inverse_scale_factor"`, to make the output conform to the expected form of an `integrated.refl` file.

jbeilstenedmands (changes requested):
Thanks for making these improvements, it looks to be doing the right thing and `dials.scale` looks to work nicely on this, please just make the suggested change regarding the scaling columns and then I will merge this in.

dagewa:
@CV-GPhL we list all contributors in the [AUTHORS](https://github.com/dials/dials/blob/main/AUTHORS) file. Please add yourself here too (#2912 went in first, so it should have gone there)

src/dials/command_line/import_xds.py
CV-GPhL: I'm not 100% sure, but I might have added those because they are needed further down in the computation of ZETA and LP?

Should we zap them afterwards once everything is done? Or not create them at all in the first place?

jbeilstenedmands: I have just tested your branch without those three columns being created and it seems to work fine, I wouldn't expect any further functions to need them in this script, so I think they can just be left out, unless you can reproduce a failure due to their absence?

graeme-winter:
Just to add my thanks to this, and also recognise that this will give a potential route for the much asked for bridge between XDS and `xia2.multiplex`

If we can get this into a release soon that would be even more awesome 😁

jbeilstenedmands (approved):
Thanks!

Note to others, I have worked on addressing the linked issues to make the program usage easier to use and generally more robust, so I will make a PR for that once this one is merged in.

Issue #2909, opened 2025-05-02
`test_filter_overlaps.py` does not work properly


dagewa:
As mentioned in [#2880](https://github.com/dials/dials/issues/2880#issuecomment-2845116168), I tried to change `test_filter_overlaps.py` to use the `trypsin_multi_lattice` data set on `dials-data`, which contains many overlaps.

However, even selecting only the first two experiments, the test fails thus
```
FAILED tests/algorithms/integration/test_filter_overlaps.py::test_for_overlaps - AssertionError: Overlapping foreground found at indexed position (1207, 276), observed centroid (1207, 275)
```

going to that location, I see this 

![Image](https://github.com/user-attachments/assets/bebdc4bf-1eb9-41e7-8084-c47e05a1cc80)

which is clearly not overlapped 🤔 

I suspect the test does not do what it is supposed do, and indeed overlap handling is known to be broken in DIALS for other reasons (#1890).

dagewa:
Tempted to change the test to use this data set anyway, then xfail it as being either a broken test, or broken `OverlapsFilterMultiExpt`

dagewa:
Worth noting that the code path is not usually taken in production, as `dials.integrate` defaults disable it
```
  overlaps_filter {
    foreground_foreground {
      enable = False
        .help = "Remove all spots in which neighbors' foreground impinges on"
                "the spot's foreground"
    }
    foreground_background {
      enable = False
        .help = "Remove all spots in which neighbors' foreground impinges on"
                "the spot's background"
    }
```

Issue #2880, opened 2025-03-26
Fully retire `dials_regression`


dagewa:
CCP4 core would like to be able to run all of the DIALS and dxtbx tests, but are still missing many due to a lack of access to `dials_regression`. We should _finally_ retire this private repository and make all test data public. I've done this in the past for refinement and some of the indexing tests, so I am happy to pick this up.

Much of the image data in `dials_regression` has unknown authorship. I do not believe any of this data to be sensitive. Nevertheless it is reasonable at least to seek permission from the LBL, being the owners of the source repository at boa.lbl.gov and therefore effectively the curators of this material with unknown authorship. @phyy-nx @bkpoon can you confirm that it is okay if I make selected files (those touched by DIALS and dxtbx tests) from the `dials_regression` SVN repository public?

phyy-nx:
This seems reasonable.  Would it be possible to make a list of the files in quesiton?  I can help doing a bit of investigation.

phyy-nx:
Gotcha, will review

phyy-nx:
Chatting with @nksauter, it seems to us these files are ancient enough that it's ok to post them to dials-data at the point.

dagewa:
That's great, thanks. I take it as a green light and I will go ahead and do it for all the files listed above

dagewa:
Although, I'll explore compressing image files that don't have decent native compression, so it won't be an exact copy

dagewa:
Doing a bit of tidying too in `image_examples/`, and documenting my decisions:
- The files in `XDS/` should not be in `image_examples/`, so remove
- `ED_From_TIFF/170112330001.cbf` does not import in DIALS: remove
- Just keep 1 of 10 images in `APS_24IDE_test/`
- Images in `dials-190/` should go in their own `dials-data` data set
- Various examples have images at 0 and 90° or 0 and 45°. For purely read testing this is unnecessary (and we have enough indexing tests elsewhere). So keep only one of each pair.
- Just keep 1 of 4 images in `LCLS_cspad_nexus/`. `idx-20130301060858401.cbf` and `idx-20130301060858801.cbf` are specifically used in tests, but I think either will do.
- Some files cannot be read if compressed:
  - `ALS_733/200mMNaCl5pcGlyc_400.edf`
  - `APS_22ID/mar300.0001`
  - `DESY_BW7B/mar345_01_001.mar2300`
  - `DLS_I19/I19_P300k_00001.cbf` (imports but prints warnings)
[...]

PR #2915, opened 2025-05-07
Convert remaining tests to use `dials-data` rather than `dials_regression`

dagewa:
This PR completes the conversion of DIALS tests to use `dials-data` (for #2880). In some cases tests were pointed to close copies of the  original files (generally updated to modern `.{expt,refl}` format, renamed, and path records edited). In other cases the tests were changed to use files already on `dials-data` to avoid excessive increases in the size of data downloaded by `dials-data`.

Mostly this was straightforward. One exception is `test_filter_overlaps.py`, which does not appear to function as expected when the test data set was changed. Seeing as we believe this functionality to be broken anyway (#2909) it seems reasonable to `xfail` the test at this point.

This PR will be followed by a similar one for `dxtbx`, and then a further pair of PRs to remove the `dials_regression` fixtures from `conftest.py` files.


dagewa:
The test errors are a dumb thing fixed by https://github.com/dials/data/pull/506. I'll try again tomorrow when a new version of `dials-data` makes it to `conda-forge`

PR #2924, opened 2025-05-14
Write integrate and scale history to MTZ

dagewa:
This is the DIALS companion to https://github.com/cctbx/dxtbx/pull/816, and requires that to be merged first. Changes here:

- History is explicitly copied from the imported experiments in `dials.index`, fixing the issue that this got lost because of the way `dials.index` constructs experiments
- When `dials.integrate` and `dials.scale` write `.expt` files they flag that these lines are interesting for `dials.export`
- `dials.export` then takes these interesting lines and adds appropriate MTZ history items

As an example, these lines are written to the MTZ history for standard single crystal rotation processing:
```
From dials.integrate v3.22.dev0, run on 2025-05-14 at 10:53:02 UTC
From dials.scale v3.22.dev0, run on 2025-05-14 at 10:53:18 UTC
```

The version string looks a  bit odd. When I run `dials.version` on the same computer I see `3.dev.1314-g6a06be331`. The way version information is extracted for experiment history is by using `importlib.metadata.version`, which for DIALS gives:

```
>>> importlib.metadata.version("dials")
'3.22.dev0'
```

@ndevenish is this `importlib.metadata.version` call for DIALS correct? For the history functionality in dxtbx I could not rely on `dials.version`, because I can't be sure the software writing the `.expt` is actually DIALS. Indeed for `xia2.multiplex` I would get `'3.9.dev184+g8e8dffbc'` as the version string here on my computer.

### DIFF
--- a/newsfragments/xxx.bugfix
+++ b/newsfragments/xxx.bugfix
@@ -1,0 +1,3 @@
+``dials.index``: Do not lose the
+``dials.export``: Write MTZ history lines for ``dials.integrate`` and ``dials.scale``
+
--- a/src/dials/command_line/index.py
+++ b/src/dials/command_line/index.py
@@ -234,6 +234,7 @@ def index(experiments, reflections, params):
                     tables_list.append(idx_refl)
                     indexed_experiments.extend(idx_expts)
         indexed_reflections = flex.reflection_table.concat(tables_list)
+
     return indexed_experiments, indexed_reflections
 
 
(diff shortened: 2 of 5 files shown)


dagewa:
I note that for the release version of DIALS v3.22.1, `importlib.metadata.version("dials")` also gives `'3.22.dev0'`

dagewa:
Possibly rethinking this to use the MTZ appendix rather than MTZ history.

dagewa:
@yangha7 I changed this PR so that it does not write the integration and scale history to the MTZ history, but it just captures and logs these for later addition to the MTZ Appendix, if that's what we decide on.

I think we should merge this in preparation for whatever we decide to do as it has useful functionality such as changing the datestamps to UTC, fixing `dials.index` so it does not lose history, and ensure `dials.integrate` and `dials.scale` mark their history entries as `integrated` and `scaled`, so those lines can be easily extracted from the full processing history.

PR #2929, opened 2025-05-15
Ensure mmcif export contains the same reflections as mtz export.

### LINKED ISSUE #2921: mmcif export should output exactly the same reflections as mtz output
Currently, there are a number of differences in reflection handling in the two programs that cause a different in the number of unmerged reflections that end up in the two output files, however we should probably output the same by default as that it what would be expected.

Places of divergence include different parameters used in the `filter_reflection_table` function (specifically `partiality_threshold`, `combine_partials` and `min_isigi`). Some of these are settable in the mtz phil scope but not the mmcif phil scope, so there may need to be some updates there.

For mmcif export, we are only exporting both profile and summation integrated intensities due to this line:
https://github.com/dials/dials/blob/f1dfe2325d5b404b1935eb2360dc8473e478508f/src/dials/util/export_mmcif.py#L96
I think the original logic behind this was that for the `v5_next` pdb dictionary version, we output columns for both profile and summation intensities, with no way to indicate a missing value (it is supported in the file standard but not the library we use to create cifs), so we could only safely output when we have both. However, for the default dictionary version (`v5`), we only output the intensity used in scaling (which could be prf, sum or a combination), so this is not an issue. Also, even if using `v5_next`, we still export the intensity used in scaling as the main intensity (`_pdbx_diffrn_unmerged_refln.intensity_meas`), so again I think having a few reflections where both sum and profile are not both valid is less of a concern. Maybe we can add a quick pass over the cif block to replace any invalid values with a `.` or `?` based on the values of the `intensity_sum` and `intensity_prf` flags from the reflection table.

jbeilstenedmands:
Fixes #2921


jbeilstenedmands:
Note that the filtering parameters come from the mtz phil scope, this is not ideal but this pattern is used in lots of the other export programs too. Changing the phil scope could break lots of things, so I am leaving that refactor for a separate time.

PR #2934, opened 2025-05-19
Stills process profiling

dwmoreau:
Profiling was performed on dials.stills_process identifying redundancies and efficiencies. The edits in this pull request should speed up the code without alter its functionality code.


jbeilstenedmands:
This looks like it touches quite a few key bits of code across a number of modules, I think it would be worth someone from Diamond performing a review before this is merged.

dwmoreau:
Yes, I agree. I have performed some testing on a lysozyme and a cytochrome data set and obtained identical results. I cannot say with certainty if this will be true for all options within dials.stills_process or if it will be free of unintended consequences.

biochem-fan:
Is this caching compatible with per-image masks? For examle, resolution masks must be re-created per image in XFEL, whose wavelength varies shot-by-shot. Trusted-range masks (e.g. to exclude hot pixels) also vary every frame, I guess.

That being said, such shot-by-shot variations are not big (affecting a very small fraction of pixels), so it might be fine to use a common mask across frames in practice. Did you discuss this with @phyy-nx?

dwmoreau:
@biochem-fan This is a good point. It is not compatible with per-image masks. The one place in this PR where the results do differ is when there is a resolution mask and a variable wavelength. The resolution mask is determined using the wavelength from the first imageset and is not updated for future imagesets. Caching this mask is a non-trivial amount of execution time per frame. I have always assumed that "dynamic" masking was a desired, but non-implemented feature in dials.stills_process. I could create an option where the masking is updated with each imageset to allow for dynamic masking.

phyy-nx:
Agree that in practice we never use d_min as a resolution filter in dials.stills_process because of the performance hit.  We instead make a static mask and pass it in on the command line.  However, @dwmoreau how about trusted range?  Is that still respected?

ndevenish:
> It is not compatible with per-image masks

this seems a blocker for the cctbx.xfel use case?

> @phyy-nx, could you clarify what is the trusted range?

This is normally the trusted range of pixel intensities

dwmoreau:
4/23/26 DIALS Core meeting discussion points:
- The per-image dynamic masking was flagged in the previous comments on this PR. This behavior was reverted to the original code.
- The code edits have been verified by counting the number of integrated lattices before and after edits. The verification should be more rigorous.
- This PR is quite long to review. It should be broken down into smaller PRs that are easier to read.

dwmoreau:
This PR has been replaced with five smaller PR's

Issue #2541, opened 2023-10-27
`dials.index` with known crystal, but new experiments


dagewa:
`dials.index` allows to index a spot list with a previous solution, by passing it in as input:

```
dials.index indexed.expt strong.refl
```

However, in some circumstances it might be useful to index an `imported.expt` with just a crystal model. For example, I have found myself wanting to do this today. After figuring out various different reindexing issues, I want to create a simple script that indexes straight off with the basis I finally chose, but without having to construct a full reference `indexed.expt`. Another reason might be trialling the effect of various different parameters at import but guaranteeing a particular indexing solution.

I suggest to add a parameter to `dials.index` that would provide 9 floats that could be either the UB matrix elements, or the real space vectors elements (though probably the former, as that is what is output by `dials.show`). This could be multiple, in which case multiple cells could be indexed in turn. Something like:

```
dials.index imported.expt strong.refl known_crystal_model=-0.0036,0.0464,-0.0282,0.0005,-0.0192,-0.0682,-0.0499,-0.0036,0.0014
```

Thoughts about whether this is a good idea, naming of the parameter and so on?

Some previous issues have touched on this area: #1577, #416

Issue #2927, opened 2025-05-15
`dials.modify_geometry` does not change image range


dagewa:
```
$ dials.show $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.expt | grep "image range"
    image range:   {1,90}

$ dials.modify_geometry $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.expt geometry.scan.image_range=1,10
The following parameters have been modified:

geometry {
  scan {
    image_range = 1 10
  }
}
input {
  experiments = /home/fcx32934/.cache/dials_data/x4wide_processed/AUTOMATIC_DEFAULT_scaled.expt
}

Saving modified experiments to modified.expt

$ dials.show modified.expt | grep "image range"
    image range:   {1,90}
```

No complaints when `geometry.scan.image_range` is passed, but it does not take effect in the output.

dagewa:
I'm down with that

dagewa:
https://github.com/dials/dials/issues/1577 😀

dagewa:
Doing some investigation. We get here https://github.com/dials/dials/blob/d83339e822d49309b234bb0c8b4d24a37cf23dbc/src/dials/command_line/modify_geometry.py#L41-L43

with an inconsistent state somehow

```
In [6]: experiments[0].imageset.get_scan().get_image_range()
Out[6]: (1, 10)

In [7]: experiments[0].scan.get_image_range()
Out[7]: (1, 90)
```

This points to some deeper issue with dxtbx. However, I will probably work around this by moving from imageset manipulation to experiment manipulation in `dials.modify_experiments`

Issue #1577, opened 2021-02-03
dials.modify_geometry -> dials.modify_experiment


graeme-winter:
Proposal:

Extend `dials.modify_geometry` to become `dials.modify_experiment` which does exactly what the former does but also allows updating (or setting) the image data pointers e.g. directory, filenames, templates. etc. 

Options: 

- make new tool, deprecate old one
- duplicate tool and add extension to API

Suggesting @elena-pascal  to look at this as as a good learning task

rjgildea:
**Alternative proposal**

You can already do e.g.:

```
$ dials.import $(dials.data get -q x4wide)/*.cbf reference_geometry=$(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled.expt
```
which does half of the job of what you are asking of a putative `dials.modify_experiment`. All that is missing is we need to extend `dials.import` to optionally copy across the crystal/profile/scaling models in addition to the beam/gonio/scan/detector models.

I think this option would be by far the easiest, as the handling of image filenames/templates is already implemented within `dials.import`, and we would essentially have to duplicate much of this functionality within `dials.modify_experiment` to allow you to replace the imageset with one pointing to new data.

graeme-winter:
I have to say that is a very reasonable proposition, if slightly obscure. But this is a long way off the beaten track, so completely reasonable. 

I like the idea thank you.

graeme-winter:
Some time has passed but the need has not gone away.

elena-pascal:
More time passed, still on the to do list.

elena-pascal:
This is actually the same issue as David's #416

Issue #416, opened 2017-07-13
dials.modify_geometry for crystal models


dagewa:
A `dials.edit_experiments` or `dials.modify_experiments` (or other similarly-named) command would be useful to override models _after_ import. The interface should be similar to `dials.import`'s geometry scope, but in addition allow changes to the crystal model.

The context in which this request arose is so that the model state panels in DUI can be made editable (by running this command when changes are made) but generally this seems a useful addition to the command-line toolbox too.

rjgildea:
You mean like `dials.modify_geometry`?

```
$ dials.modify_geometry 
Usage: dials.modify_geometry [options] datablock.json | experiments.json

Options:
  -h, --help            show this help message and exit
  -c, --show-config     Show the configuration parameters.
  -a ATTRIBUTES_LEVEL, --attributes-level=ATTRIBUTES_LEVEL
                        Set the attributes level for showing configuration
                        parameters
  -e EXPERT_LEVEL, --expert-level=EXPERT_LEVEL
                        Set the expert level for showing configuration
                        parameters
  -v                    Increase verbosity

$ dials.modify_geometry -c -e2
Showing configuration parameters with:
  attributes_level = 0
  expert_level = 2

geometry {
  beam {
    wavelength = None
    direction = None
  }
  detector {
    panel {
      id = 0
      name = None
      type = None
      pixel_size = None
      image_size = None
      trusted_range = None
      thickness = None
      material = None
      fast_axis = None
      slow_axis = None
      origin = None
      parallax_correction = None
    }
    hierarchy {
      name = None
      fast_axis = None
      slow_axis = None
      origin = None
      group {
        id = None
        name = None
        fast_axis = None
        slow_axis = None
        origin = None
        panel = None
      }
    }
    mosflm_beam_centre = None
    distance = None
    slow_fast_beam_centre = None
  }
  goniometer {
    axes = None
    angles = None
    names = None
    scan_axis = None
    fixed_rotation = None
    setting_rotation = None
  }
  scan {
    image_range = None
    extrapolate_scan = False
    oscillation = None
  }
  convert_stills_to_sweeps = False
  convert_sweeps_to_stills = False
}
output {
  datablock = modified_datablock.json
  experiments = modified_experiments.json
}
input {
  datablock = None
  experiments = None
}
```

rjgildea:
https://github.com/dials/dials/commit/1ec6e6718e777d834d26d66e8d6f7efc0bc0dd8a
It's even got a test :)

[...]

PR #2936, opened 2025-05-21
`dials.modify_experiments`

dagewa:
This is a fix for #416 and #2927, and a partial fix for #1577. What remains from the second issue is allowing

> updating (or setting) the image data pointers e.g. directory, filenames, templates. etc.

Hmm, have just noticed that #1577 asked for `modify_experiment` rather than `modify_experiments`, and this will override the models in every experiment in the list. Should we implement some experiment selection logic, and if so, what should it look like?


dagewa:
I think this also closes #2541 by allowing a workaround.

PR #2948, opened 2025-06-06
Support the xarray.Dataset format for reflection tables

yash4karan:


yash4karan:
> Hi @yash4karan please feel free to ping me if you need any help with this, or want advice on the best ways to do things in Xarray. I'm happy to help work through any confusion or issues you might run into with Xarray and to help test things out.
> 
> 
> 
> One note on availability: I am on vacation from June 9-> 17th and back to work on the 18th.

Hi @ianhi, thank you! I will let you know if I have any questions.

src/dials/array_family/flex_ext.py
```diff
@@ -13,15 +13,18 @@
 import operator
 import os
 import pickle
+from typing import List, Optional
 
 import numpy as np
 import pandas as pd
+import xarray as xr
```
ndevenish, line 20: While this is what xarray recommends for abbreviated usage, I wonder if it's worth it here because we have such a limited usage (twice that I can see!)

src/dials/array_family/flex_ext.py
```diff
@@ -375,6 +378,110 @@ def as_hdf5(self, filename):
...
+                    (index_name, "vec3_index"),
+                    flumpy.to_numpy(data).reshape(sz, 3),
+                )
+            elif isinstance(data, cctbx.array_family.flex.mat3_double):
+                ds[key] = (
+                    (index_name, "mat3_index1", "mat3_index2"),
+                    flumpy.to_numpy(data).reshape(sz, 3, 3),
+                )
+            elif isinstance(data, cctbx.array_family.flex.std_string):
+                ds[key] = ((index_name), list(data))
```
ndevenish, line 474: Is this for completion' sake or is there a usage of this that you found? My memory was that array_family string objects were _completely_ broken..

ndevenish:
This looks reasonable? I think the question about naming of coordinates and index columns is a reasonable one, are we really forced to have a specially named index per column/per data type?

ndevenish:
We can revive this if there is interest.

Issue #2920, opened 2025-05-09
Review slow tests


dagewa:
Now all tests are using `dials-data` we thought it might be a good time to review test durations. Here are the top twenty slowest DIALS tests:

```
206.40s call     tests/algorithms/integration/test_simple_integrate.py::test_against_dials_integrate
117.40s call     tests/algorithms/scaling/test_tof_scaling.py::test_tof_extract_shoeboxes
100.79s call     tests/algorithms/symmetry/cosym/test_cosym_analyse_datasets.py::test_cosym[I23-None-dimensions2-10-False-False-True]
91.56s call     tests/command_line/test_model_background.py::test_model_background
89.77s call     tests/algorithms/indexing/test_index.py::test_pink_indexer
55.71s call     tests/algorithms/indexing/test_index.py::test_index_small_molecule_multi_sequence_3
54.57s call     tests/command_line/test_integrate.py::test_multi_lattice
45.05s call     tests/command_line/test_ssx_reduction.py::test_ssx_reduction
42.57s call     tests/test_scan_varying_integration_bug.py::test
40.57s call     tests/algorithms/indexing/test_index.py::test_unconventional_P1_cell[cell_params1]
40.16s call     tests/algorithms/profile_model/ellipsoid/test_refiner.py::test_Refiner
39.26s call     tests/algorithms/indexing/test_index.py::test_unconventional_P1_cell[cell_params0]
38.81s call     tests/command_line/test_preservation_of_experiment_identifiers.py::test_for_preservation_of_identifiers_in_dials_processing
35.93s call     tests/algorithms/indexing/test_index.py::test_index_multi_lattice_multi_sweep
31.20s call     tests/command_line/test_refine.py::test_scan_varying_refinement_rmsds
30.10s call     tests/algorithms/indexing/test_index.py::test_index_4rotation
29.48s call     tests/command_line/test_spot_counts_per_image.py::test_spot_counts_per_image[thaumatin_grid_scan]
29.27s call     tests/command_line/test_detect_blanks.py::test_passing_still_images_raises_sysexit
28.71s call     tests/command_line/test_ssx_index.py::test_ssx_index_input_unit_cell
28.56s call     tests/algorithms/scaling/test_scale.py::test_incremental_scale_workflow
```

We might pick on one or two of these at a DIALS catch up in the near future.

dagewa:
[...]

PR #2949, opened 2025-06-06
Speed up tests

dagewa:
A branch for addressing #2920


dagewa:
So far,

- `test_index.py::test_index_small_molecule_multi_sequence_3` was sped up by setting `max_refine=10`
- `test_scan_varying_integration_bug.py` was removed as it does not seem very useful
- `test_index.py::test_unconventional_P1_cell` was sped up by setting `max_refine=5`
- `test_detect_blanks.py::test_passing_still_images_raises_sysexit` was sped up _9 times_ from 18s to 2s by simply passing in a stills `.{expt,refl}` rather than importing and finding spots from scratch.

PR #2950, opened 2025-06-10
Fill in empty xyzobs entries with xyzcal values during dials.import_xds

### LINKED ISSUE #2944: dials.import_xds issues with experiment model creation
Following updating the program, in further testing a couple of issues have become apparent:

1. When importing XDS data where the image data is from a h5 file, (sometimes/always?) the experiment created is understood by dials/dxtbx to have only one image, as seen in the output from `dials.import_xds`: 
```
Experiment 0
  format: <class 'dxtbx.format.Format.Format'>
  type: <class 'dxtbx_imageset_ext.ImageSequence'>
  num images: 1
```
This then leads to a crash when a subsequent program tries to load the `xds_models.expt` file:
```
File "/dials/modules/dxtbx/src/dxtbx/model/experiment_list.py", line 492, in decode
        imageset = self._imageset_from_imageset_data(imageset_data, models)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/dials/modules/dxtbx/src/dxtbx/model/experiment_list.py", line 412, in _imageset_from_imageset_data
        imageset.set_scan(scan)
    RuntimeError: dxtbx Internal Error: /dials/modules/dxtbx/src/dxtbx/imageset.h(365): DXTBX_ASSERT(index < scans_.size()) failure.
```
This is the error reported in https://github.com/dials/dials/pull/2938#issuecomment-2918442645 which I can reproduce on recent data processing at Diamond. I have a potential fix coming in dxtbx for this.

2. (Sometimes?) Issues with the geometry that manifest in `dials.two_theta_refine`/`dials.refine` not being able to sensibly refine the crystal model, with very large rmsds. Observed on the beta-lactamase (c2sum) and the semisynthetic_multilattice datasets. If testing xia2.multiplex, you can temporarily get around this by setting `unit_cell.refine=None`.

jbeilstenedmands:
Required to be able to run `dials.two_theta_refine` (and xia2.multiplex).

As described at https://xds.mr.mpg.de/html_doc/xds_files.html, the values of XOBS, YOBS and ZOBS in INTEGRATE.HKL are 0 if the reflection was "unobserved" (i.e. not found during spot finding). Several programs require valid values in this column (`dials.two_theta_refine`, `dials.scale`), so I propose we set the values for not strong reflections to be the `xyzcal` values. This is then not a problem for `dials.two_theta_refine`, which by default uses only the strong reflections. 

Fixes #2944


biochem-fan:
Isn't this confusing? If `dials.scale` needs spot positions, shouldn't we amend it to use xyzcal when xyzobs is invalid, rather than abusing the columns during import? What is stored in xyzobs when a weak reflection is integrated by DIALS?

jbeilstenedmands:
@dagewa That would work for `dials.two_theta_refine`, but would not work for other programs that use xyzobs e.g. `dials.scale`

jbeilstenedmands:
I am investigating whether we can leave these as zeroes, at least then they are obviously wrong and we can make sure the relevant dials programs handle this - I think this would be simple. @dagewa scaling filters out reflections with that flag so we can't use that.

dagewa:
I'm still not sure I like putting the obviously wrong data in these columns, and additional complexity in various programs required for the extra handling of these. I think that is also a confusing situation. My feeling remains that the best guess for `xyzobs` in the absence of that information from XDS is `xyzcal`, along with a particularly high centroid error estimate indicating our lack of certainty about that position. Any properly-written geometry refinement procedure using those positions would downweight those reflections based on the uncertainty. But @biochem-fan is right, looking at RMSDs resulting from this might still be a bit like looking at density maps where some missing Fobs are filled in with Fcalc. So I see the shortcomings with this approach too and feel we are stuck between a rock and a hard place :disappointed:

jbeilstenedmands:
I have now introduced a `not_suitable_for_refinement` reflection table flag, which should be used as a persistent flag (other refinement flags get reset during refinement so they were unsuitable). This flag is used as a filter in the refinement reflection manager, so no dials refinement programs inheriting from that will use these reflections. This now at least gives a codified trace in the reflection table as to what has happened.

src/dials/command_line/import_xds.py
```diff
@@ -237,6 +237,30 @@ def __call__(self, params, options):
...
+        unobserved = (x == 0) and (y == 0) and (z == 0)
+        table["xyzobs.px.value"].set_selected(
+            unobserved, table["xyzcal.px"].select(unobserved)
+        )
+        table["xyzobs.mm.value"].set_selected(
+            unobserved, table["xyzcal.mm"].select(unobserved)
+        )
+        table.set_flags(unobserved, table.flags.not_suitable_for_refinement)
+        table.set_flags(~unobserved, table.flags.strong)
+        logger.info(f"""Warning: {unobserved.count(True)} unobserved reflections did not have an xyzobs value: for
```
dagewa, line 260: Adding a warning is a very good idea, thanks. I think perhaps also we should note the limitations of this program in its help string. So, after `"This program imports xds processed data for use in dials."` maybe we should note that it is our best effort at mapping XDS data to DIALS types, but this is not a perfect process. We don't need to list the limitations, but for posterity I know we have found: 

1. `xyzobs` missing from weak reflections from XDS, so we have to copy from `xyzcalc` to allow scaling to work
2. Imprecise centroid position data coming from the XDS text-based data format, visible as steps in centroid residual plots from refinement
3. centroid pixel variances unavailable, so assumed to be 1,1,1

Maybe we can say that the intention for this tool is to enable interoperability between XDS and DIALS for analyses like `xia2.multiplex`, but that it should not be used to inform comparisons between the programs as the translation is necessarily inexact? Or some similar words to that effect :smile:

dagewa (approved):
Yes, thank you. I am happy this does enough to provide a cautionary "here be dragons" :+1:

PR #2964, opened 2025-07-03
Add ellipse and seed skewness masking algorithms for time-of-flight data

toastisme:
This adds two methods for calculating foreground/background masks for time-of-flight data.

- `tof_calculate_ellipse_shoebox_mask(reflections, experiment)`
- `tof_calculate_seed_skewness_shoebox_mask(reflections, experiment, d_skewness_threshold, min_iterations)`

The ellipse method maps all shoebox pixels to reciprocal space and does a eigenvalue decomposition of the weighted distribution of points to generate an ellipse (unlike the X-ray approach it does not map to Kabsch space).

The seed skewness method is based on J. Peters, The 'seed-skewness' integration method generalized for three-dimensional Bragg peaks, (2003), 36, 1475-1479 https://doi.org/10.1107/S0021889803021939. `d_skewness_threshold` sets the test for convergence when searching for new pixels, and `min_iterations` sets the fewest number of iterations before convergence is checked. Examples of how to use both methods are given in `test_tof_mask_calculator.py`.

Some examples below. For plotting purposes the time-of-flight dimension has been flattened.

example_masks_1

example_masks_2

example_masks_3

example_masks_4

example_masks_5

example_masks_7

example_masks_8


dagewa:
Interesting that the ellipse method produces systematically larger shoeboxes than seed-skewness. Do you see any evidence for seed-skewness being too tight - that is, systematically underestimated intensities? Or maybe ellipse is too wide, producing lower $I/\sigma$ due to including too much background? Maybe the optimum is somewhere in between, in which case tuning the parameters to obtain somewhat similarly-sized masks might be useful.

toastisme:
> Interesting that the ellipse method produces systematically larger shoeboxes than seed-skewness. Do you see any evidence for seed-skewness being too tight - that is, systematically underestimated intensities? Or maybe ellipse is too wide, producing lower I / σ due to including too much background? Maybe the optimum is somewhere in between, in which case tuning the parameters to obtain somewhat similarly-sized masks might be useful.

Yeah the images were just with one value of `d_skewness_threshold`. Downstream I'm finding seed-skewness generally gives better R1 values. The seed skewness method is actually completely generic and could be used with X-ray data. I guess having a benchmarking project would help answer if it could be useful more broadly.

toastisme:
> This is very interesting, how much work would it be to make the two methods available for Xray data too? I would be interested to know more about the ellipse method, sounds very similar to what we did for still shot data with the ellipsoid integration algorithm. Are you optimizing a set of reference ellipsoid parameters throughout the dataset that you fit to each reflection, or one global set of parameters, or per reflection?

The seed skewness algorithm can be used as is with any data currently using `tof_calculate_seed_skewness_shoebox_mask(reflections, experiment, d_skewness_threshold, min_iterations)`.  For the ellipse method, the only change would be having a different method for mapping all the points to reciprocal space.

At the moment both methods calculate everything per reflection. There looks to be a ToF dependence for `d_skewness_threshold` and ellipse scaling that I wanted to investigate later for optimisation. That was one motivation for keeping things isolated as ToF methods (but these could just be wrapper layers to keep the actual methods agnostic). The other reasons were not wanting to wade into the current integration interface (e.g. reflections.compute_mask goes through expt.profile.compute_mask etc.), and not having a good set of X-ray datasets to compare to current methods.

toastisme:
@jbeilstenedmands are you okay with me merging this for now for ToF data, then thinking about a PR in the future for generalising to other types of data?

PR #2972, opened 2025-07-18
Add spot finder wrapper for Laue data 

toastisme:
This adds `LaueSpotFinder` that does nothing except override `SpotFinder._post_process` to add initial `wavelength` and `s0` columns to the reflection table. The initial wavelength can be set using, e.g.

`dials.find_spots imported.expt laue.initial_wavelength=3`

If no initial wavelength is set, the average of `PolychromaticBeam.wavelength_range` is used.
These columns are required in place of any parts of the workflow that require `beam.wavelength` or `beam.s0`, e.g. for things like mapping spots to reciprocal space and initial values for refinement.


toastisme:
@PrinceWalnut can you confirm this would still make sense in the laue-dials workflow, if you switched to using a PolyChromatic beam?

Issue #2960, opened 2025-06-26
Error in dials.export for XDS


aglie:
I am running in a few problems with 

```
dials.export *SWEEP1.expt *SWEEP1.refl format=xds
```

Which produces `XDS.INP`, `XPARM.XDS`, and `SPOT.XDS`. The problems that I could I identify so far are the following:

in XDS.INP incident beam direction is set to wavevector, rather than the direction vector

```
INCIDENT_BEAM_DIRECTION= -0.001 0.004 2.058
```

instead, it should be vector with length 1.

In `XPARM.XDS` the problem is the opposite:

```
XPARM.XDS
     1     -175.0000    0.2000  0.000539  0.999999 -0.000868
       0.485900      -0.000690       0.002134       0.999997 <--- THIS LINE IS WRONG
```
The line 3 should contain wavelength, wavevector. Wavevector should have the length of 1/wavelength, but is instead length 1.

Additionally, the unit cell axes are reported incorrectly, but I am not sure what the exact problem is, most likely the values of x,y, and z reversed, but possibly there is something else on top of that.

The third problem is that in `SPOT.XDS` some of the intensities are weak

```
 1147.64 1700.17 0.50 11.37  -1 -3 -1 
 191.27 1140.90 0.50 17.79  1 3 7 
 1967.38 789.79 0.50 0.29  8 -1 -5 <----problematic line
```
Because every intensity which is smaller than 1 XDS reads as a value with intensity 0, and stops reading further. In xds-generated SPOT.XDS the reflections are sorted, and the line with intensity 0 has a special meaning of a stopline.

Could you kindly fix the above errors?
In XDS.INP report insident beam direction with norm 1
in XPARM.XDS report insident beam wavevector
in SPOT.XDS add 1 to each intensity, in cases they are less than 1.

PS:
Tested dials version is from https://anaconda.org/conda-forge/dials, version 3.24.1

jbeilstenedmands:
Hi @aglie, thanks for making us aware of these issues.

The incident beam direction and wavelength issues sounds like bugs and should be simple for us to address.

The cell axes could just be a different choice of cell setting - what space group/cell is your dataset? In dials we don't enforce strict conventions on unit cell axis ordering so this could be causing an incompatibility.

[...]

PR #2988, opened 2025-07-23
`dials.export format=xds`: ensure `SPOT.XDS` only contains `strong` reflections, in reverse intensity order

dagewa:
Part fix for #2960

### DIFF
--- a/newsfragments/xxx.bugfix
+++ b/newsfragments/xxx.bugfix
@@ -1,0 +1,1 @@
+``dials.export format=xds``: Ensure only ``strong`` reflections are exported in ``SPOT.XDS`` and reverse-sort by intensity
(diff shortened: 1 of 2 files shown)


src/dials/command_line/import_xds.py
```diff
@@ -91,6 +91,7 @@ def __call__(self, params, options):
             table["miller_index"] = flex.miller_index(miller_index)
         table["xyzobs.px.value"] = flex.vec3_double(centroid)
         table["intensity.sum.value"] = flex.double(intensity)
+        table.set_flags(flex.bool(len(table), True), table.flags.strong)
```
jbeilstenedmands, line 94: What is the motivation for setting all as strong?

dagewa: `SPOT.XDS` is supposed to be for spot-finding results, thus all strong. The issue #2960 pointed out that XDS chokes when an intensity is < 1, which we were hitting before by exporting even weak reflections in this format

jbeilstenedmands (approved):
Ok, I have convinced myself that these only affect importing/exporting of pre-integration data, so looks good, thanks.

PR #2997, opened 2025-08-04
Replace io_service with io_context for boost >=1.87 compatibility

dwpaley:
io_service has been deprecated for many versions: https://github.com/boostorg/asio/issues/110 and is finally removed in 1.87. It's starting to affect some tests where newer boost versions have been pulled in, see here for example: https://dev.azure.com/cctbx/cctbx_project/_build/results?buildId=39058&view=logs&jobId=57cf1bb1-62be-5ff8-45d1-8f14e5cb17f4&j=57cf1bb1-62be-5ff8-45d1-8f14e5cb17f4&t=c8b2d589-c48b-5836-1713-26aea8773c6d This PR converts thread_pool.h to the new pattern. It should work identically.


dwpaley:
draft for a sec while i check something

dwpaley:
Ok, this should be ready. I have removed a data member `boost::asio::io_service::work work_` from the ThreadPool class which was not used anywhere in the module or elsewhere in the dials codebase. I can't rule out that it does something magical, but if this still works in `algorithms.integration.parallel_integrator` as exercised in `tests/algorithms/test_parallel_integrator.py` then I propose leaving it removed.

dwpaley:
not super confident in all this refactoring with no real test so i would like to add one before merging this...

dwpaley:
oh, this must be it...
```
def test_cpp_program(executable):
    if "LIBTBX_BUILD" not in os.environ:
        pytest.skip("LIBTBX_ENV is unset; don't know how to find test executable")
```
I'll look in the libtbx build results.

dwpaley:
Thanks. All tests pass now including the new one. The cctbx_project XFEL CI build error on MacOS is also resolved.

PR #2999, opened 2025-08-05
Optimise db score

amyjaynethompson:
A few cases were found where the OPTICS analysis did a poor job of identifying clusters in the cosym coordinate plot. This is a significant concern as we are running this analysis automatically and providing clusters to users, but they may not have experience interpreting the clustering plots (or even know where to look for them!). 

See this example with 4 obvious clusters that OPTICS decided were actually 3. This is about the most text book, clear cut case, and we should be able to separate these groups automatically.  
before_optimise
This came about because the OPTICS analysis is quite sensitive to the initial choice of the min_samples parameter - ie what is the minimum number of samples for a cluster to be a cluster. 

As the number of groups is linked to the number of systematic differences (ie dimensions), we had been using a heuristic that the min_samples parameter should be equal to the number of datasets / number of dimensions * some buffer to account for groups of different sizes. This was a reasonable starting point, as we have no idea in advance of the analysis what the size of the smallest cluster is. However, in this case, the size of the smallest cluster was BELOW where this heuristic had set the min_samples parameter. Therefore, it was not identified as a separate cluster. 

This PR introduces a new package (optuna), which essentially brute forces the testing of different parameters. The benefit of optuna (https://github.com/optuna/optuna) is that it doesn't just randomly pick values to try, instead it uses results of previous trials to inform new parameters to test. The brute force approach is not a significant downside, given the OPTICS analysis is already much faster than the rest of the clustering analysis, and in this case, such an obviously wrong result would require re-running the entire clustering analysis from scratch (including the calculation of rij - the most time intensive part), to get the right groups separated out. Therefore, sacrificing a small amount of time to optimise in the first instance saves significant computation time later.

The Davies-Bouldin Score (https://scikit-learn.org/stable/modules/generated/sklearn.metrics.davies_bouldin_score.html) is a measure of how well defined the clusters are, therefore optimising min_samples to minimise this parameter results in clusters that are denser and better separated. To avoid tiny "super-dense" groups being identified with most of the data being classified as noise, a penalty was added for noise. Currently, the DB-Score and Noise penalty are weighted equally, though the penalty applied for noise has been added as a phil parameter. 

With all this, OPTICS automatically classifies the data as expected:
after_optimise

This approach also gives equivalent results when analysing the cows, pigs and people data from the recent publication.

### DIFF
--- a/src/dials/algorithms/correlation/analysis.py
+++ b/src/dials/algorithms/correlation/analysis.py
@@ -7,9 +7,11 @@ import sys
 from collections import OrderedDict
 
 import numpy as np
+import optuna
 import scipy.spatial.distance as ssd
 from scipy.cluster import hierarchy
 from sklearn.cluster import OPTICS
+from sklearn.metrics import davies_bouldin_score
 
 import iotbx.phil
 from dxtbx.model import ExperimentList
@@ -55,10 +57,13 @@ dimensionality_assessment {
 significant_clusters {
   min_points_buffer = 0.5
     .type = float(value_min=0, value_max=1)
-    .help = "Buffer for minimum number of points required for a cluster in OPTICS algorithm: min_points=(number_of_datasets/number_of_dimensions)*buffer"
+    .help = "Buffer for minimum number of points required for a cluster in OPTICS algorithm: min_points=(number_of_datasets/number_of_dimensions)*buffer - INITIAL GUESS ONLY"
   xi = 0.05
     .type = float(value_min=0, value_max=1)
     .help = "xi parameter to determine min steepness to define cluster boundary"
+  noise_tolerance = 1
+    .type = float
+    .help = "multiplier to down-weight clustering results which contain lots of noise"
 }
 
 hierarchical_clustering {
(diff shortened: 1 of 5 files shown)


src/dials/algorithms/correlation/analysis.py
```diff
@@ -413,6 +629,16 @@ def cluster_cosine_coords(self):
         logger.info(
             f"OPTICS identified {len(unique_labels)} clusters and {outliers} outlier datasets."
         )
+        if len(unique_labels) > 1:
+            logger.info(f"Davies-Bouldin Score of OPTICS result is {density_score}.")
```
jbeilstenedmands:
```suggestion
            logger.info(f"Davies-Bouldin Score of OPTICS result is {density_score:.4f}.")
```

src/dials/algorithms/correlation/analysis.py
```diff
@@ -413,6 +629,16 @@ def cluster_cosine_coords(self):
...
+        else:
+            logger.info(
+                f"Median reachability of the identified cluster is {density_score}."
+            )
+
+        logger.info(
+            f"Overall score including noise penalty (using tolerance of {self.params.significant_clusters.noise_tolerance}) is {overall_score}."
```
jbeilstenedmands:
```suggestion
            f"Overall score including noise penalty (using tolerance of {self.params.significant_clusters.noise_tolerance}) is {overall_score:.4f}."
```

src/dials/algorithms/correlation/analysis.py
```diff
@@ -413,6 +629,16 @@ def cluster_cosine_coords(self):
         logger.info(
             f"OPTICS identified {len(unique_labels)} clusters and {outliers} outlier datasets."
         )
+        if len(unique_labels) > 1:
+            logger.info(f"Davies-Bouldin Score of OPTICS result is {density_score}.")
+        else:
+            logger.info(
+                f"Median reachability of the identified cluster is {density_score}."
```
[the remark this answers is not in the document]
amyjaynethompson: Removed as this left over from testing...

src/dials/algorithms/correlation/analysis.py
amyjaynethompson: Added formatting :)

PR #3000, opened 2025-08-05
Added deposition and citation metrics to dials website

yangha7:


yangha7:
> Interesting, how do can we add additional papers to the plot in future?

We can always add new papers in the future. All I need is a google scholar link to scrape the citations

Issue #2215, opened 2022-08-26
We should not import code from command line programs?


graeme-winter:
At least, perhaps we should avoid doing so - if it is needed in > 1 `command_line` then it should be under the source tree somewhere else 

```
Grey-Area dials :) [main] $ git grep command_line src/dials | grep import 
src/dials/algorithms/background/modeller.py:        from dials.util.command_line import heading
src/dials/algorithms/indexing/bravais_settings.py:from dials.command_line.check_indexing_symmetry import (
src/dials/algorithms/indexing/ssx/processing.py:from dials.command_line.combine_experiments import CombineWithReference
src/dials/algorithms/integration/integrator.py:from dials.util.command_line import heading
src/dials/algorithms/integration/integrator_stills.py:        from dials.util.command_line import Command
src/dials/algorithms/integration/integrator_stills.py:        from dials.util.command_line import Command
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/parallel_integrator.py:            from dials.command_line.integrate import phil_scope
src/dials/algorithms/integration/ssx/stills_integrate.py:from dials.command_line.integrate import process_reference
src/dials/algorithms/scaling/algorithm.py:from dials.command_line.compute_delta_cchalf import phil_scope as deltacc_phil_scope
src/dials/algorithms/scaling/algorithm.py:from dials.command_line.cosym import cosym
src/dials/algorithms/scaling/algorithm.py:from dials.command_line.cosym import phil_scope as cosym_phil_scope
src/dials/algorithms/simulation/generate_test_reflections.py:    from dials.util.command_line import ProgressBar
src/dials/algorithms/simulation/generate_test_reflections.py:    from libtbx.phil import command_line
src/dials/algorithms/simulation/reciprocal_space.py:        from dials.util.command_line import ProgressBar
src/dials/algorithms/simulation/reciprocal_space.py:        from dials.util.command_line import Command
src/dials/array_family/flex_ext.py:            from dials.command_line.find_spots import phil_scope
[...]

PR #3013, opened 2025-09-10
Consistent data preparation for `dials.symmetry` and `dials.cosym`

dagewa:
The programs `dials.symmetry` and `dials.cosym` had largely duplicated code in the setup prior to symmetry analysis. This PR unifies the set up steps:

- after reduction to minimum cell, if not all data sets are consistent, `symmetry` raises `ValueError` while `cosym` continues. This is now controlled (for both programs) using the PHIL option `exclude_inconsistent_unit_cells`
- `get_subset_for_symmetry` (#1116) is now used in `cosym` as well as `symmetry`
- the order of steps is made the same between between the programs

The unified set up is refactored to a new function, `prepare_datasets_for_symmetry_analysis`, in `dials.algorithms.symmetry`. Other utility functions that used to be in `dials.command_line.symmetry`, and were imported from there by `cosym` and some other programs, are now also moved to `dials.algorithms.symmetry` (to help a bit with #2215)

There are still areas where `symmetry` and `cosym` differ in the set up stages:
- They still use separate PHIL scopes, with some duplicated parameters
- `cosym` does unit cell clustering but this has not been added to `symmetry`
- `cosym` can take in reference reflections, while `symmetry` cannot
- `symmetry` does outlier rejection after reflection filtering whereas `cosym` does not
- `cosym` uses a class with the Observer pattern (for producing plots), while `symmetry` is just a function
- `symmetry` uses anomalous Miller arrays, while `cosym` uses non-anomalous

Some of these points are harder to address without more extensive changes that alter behaviour or interfaces, so I leave them to follow up work.

### DIFF
--- a/tests/command_line/test_symmetry.py
+++ b/tests/command_line/test_symmetry.py
@@ -9,7 +9,7 @@ import pytest
 
 import scitbx.matrix
 from cctbx import crystal, sgtbx, uctbx
-from dxtbx.model import Crystal, Experiment, ExperimentList, Scan
+from dxtbx.model import Crystal, Experiment, ExperimentList, GoniometerFactory, Scan
 from dxtbx.serialize import load
 from dxtbx.util import ersatz_uuid4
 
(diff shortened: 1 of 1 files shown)


dagewa:
XFEL failures are because of imports from `dials.command_line.symmetry`, which with this PR change to `dials.algorithms.symmetry`

dwpaley:
> @phyy-nx @dwpaley the only thing left here is to check the XFEL tests pass on this branch, if [cctbx/cctbx_project#1104](https://github.com/cctbx/cctbx_project/pull/1104) is merged. It will hopefully be fine, as it's just a change of import location, but there are a few behaviour changes in this PR too, so it is worth a check. I'm afraid I'm not set up to run the XFEL tests locally, so I'm wondering if one of you guys might be able to check that for me? Thanks!

This test (which I hope I setup correctly) is running here: https://github.com/dials/dials/actions/runs/18922139289

dwpaley:
Hi @dagewa, ok, the XFEL tests ran here: https://github.com/dials/dials/pull/3035 and they look good. There is one failure on `tst_small_cell_process.py` that was fixed this afternoon in https://github.com/cctbx/cctbx_project/pull/1106 but your cctbx branch for cctbx/cctbx_project#1104 does not include that commit. WRT the XFEL tests I believe we are all set here.

PR #3023, opened 2025-10-08
Fixed pinkIndexer to return UB matrices in descending order

PrinceWalnut:
Classes like [`StillsIndexer`](https://github.com/dials/dials/blob/dcb165478a10ac6d570ace3061fa6357689d9dbb/src/dials/algorithms/indexing/stills_indexer.py#L503-L505) use only the first N orientation matrices returned by the lattice search algorithm. 

Formerly, `pinkindexer.py` would return these UB matrices in the order found in the `voxels` array. In cases where `min_lattices > max_refine`, only the first subset of matrices returned would be refined, and these matrices were not necessarily the best ones. This would lead to behavior where increasing `min_lattices` beyond `max_refine` would inadvertently often yield worse results. 

This patch fixes `pinkindexer.py` to return UB matrices in order of descending voxel solution density to fix this issue.

Requesting review and testing from @kmdalton and @phyy-nx.

### DIFF
--- a/newsfragments/3023.bugfix
+++ b/newsfragments/3023.bugfix
@@ -1,1 +1,1 @@
-``dials.index``: return best pinkIndexer orientation matrices when `min_lattices` > `max_refine`.
+``dials.index``: return best pinkIndexer orientation matrices when `target_lattices` > `max_refine`. Deprecate the min_lattices parameter in favour of target_lattices.


src/dials/algorithms/indexing/lattice_search/pinkindexer.py
```diff
@@ -36,9 +36,10 @@
     voxel_grid_points=150
         .type = int(value_min=10, value_max=1000)
         .help = "Controls the number of voxels onto which the rotograms are discretized"
-    min_lattices=1
+    target_lattices=1
         .type = int(value_min=1, value_max=100)
-        .help = "The minimum number of candidate lattices to generate."
+        .alias = min_lattices
```
jbeilstenedmands: Note that this phil alias mechanism from cctbx seems to be broken. So if you want to keep scripts working that use the old parameter, you need to keep it as its own phil option, print a deprecation warning and use it to set the new parameter, see this for an example:
https://github.com/dials/dials/blob/40bd3f46eee7a733c5188d5f404f5982bd5c65b2/src/dials/algorithms/indexing/stills_indexer.py#L676

PR #3030, opened 2025-10-24
Remove IOTA and Prime from default install

ndevenish:
As discussed in https://dials.github.io/kb/core/2025-10-23, both are now unmaintained, and prime will be shortly removed from cctbx.


phyy-nx:
Looks like prime and iota are also in ./installer/dials_installer.py, should they be removed there as well?

Issue #2236, opened 2022-10-03
dials.index / dials.refine: figure a way of rejecting "obviously bad" solutions


graeme-winter:
e.g. 

```
RMSDs by experiment:
+-------+--------+----------+----------+------------+
|   Exp |   Nref |   RMSD_X |   RMSD_Y |     RMSD_Z |
|    id |        |     (px) |     (px) |   (images) |
|-------+--------+----------+----------+------------|
|     0 |   1183 |  1.0997  |  0.32837 |    0.23099 |
|     1 |   1026 |  0.44609 |  0.33887 |    0.19208 |
|     2 |     76 | 40.254   |  8.0826  |   10.711   |
|     3 |   1047 |  0.58122 |  0.44348 |    0.2149  |
|     4 |    706 |  0.56193 |  0.32431 |    0.1902  |
+-------+--------+----------+----------+------------+
Updating predictions for indexed reflections
Saving refined experiments to refined.expt
Saving reflections with updated predictions to refined.refl
```

To a human being the second experiment is obviously bad: including it messes with things later -> would be really nice to be able to filter this early in the processing in a clever way e.g. in indexing

```
RMSDs by experiment:
+-------+--------+----------+----------+------------+
|   Exp |   Nref |   RMSD_X |   RMSD_Y |     RMSD_Z |
|    id |        |     (px) |     (px) |   (images) |
|-------+--------+----------+----------+------------|
|     0 |   1000 |  1.361   |  0.28802 |    0.31579 |
|     1 |   1000 |  0.54578 |  0.28479 |    0.25664 |
|     2 |     79 | 42.502   |  8.097   |   11.378   |
|     3 |   1000 |  0.63283 |  0.48488 |    0.26573 |
|     4 |    751 |  0.64144 |  0.33877 |    0.24854 |
+-------+--------+----------+----------+------------+
```

graeme-winter:
Another example of this - 

```
RMSDs by experiment:
+-------+--------+----------+----------+------------+
|   Exp |   Nref |   RMSD_X |   RMSD_Y |     RMSD_Z |
|    id |        |     (px) |     (px) |   (images) |
|-------+--------+----------+----------+------------|
|     0 |    694 |  3.0876  |  1.7548  |    1.1178  |
|     1 |   1844 |  0.64703 |  0.39619 |    0.41411 |
|     2 |   3535 |  0.89459 |  0.5629  |    0.26988 |
[...]

PR #3032, opened 2025-10-28
Reject obviously bad models from indexing

dagewa:
This PR introduces two new parameters, `xy_rmsd_threshold=4.0` (pixels) and `n_indexed_threshold=5` (reflections), which are used by the `ModelEvaluation` object in indexing. Prior to candidate solution refinement, if a solution indexes fewer than `n_indexed_threshold` reflections in total, then it is rejected. After refinement, if the solution has overall spatial RMSDs greater than `xy_rmsd_threshold` then it is rejected.

The intention behind this PR is to avoid provide obviously bad/wrong indexing solutions. This is ultimately the root cause behind many failures in `dials.integrate` reported by users, where a solution with very poor RMSDs results in enormous memory usage. In most of those cases, even if integration succeeded, data processing results would be low quality or just plain wrong, in the case of misindexing.

This actually addresses the original issue #2236, but that has since been changed (by me) to a request for a new program, `dials.filter_experiments`. I still see a use for that program to more interactively select a set of good experiments, so I'll leave the issue open.

### DIFF
--- a/newsfragments/3032.feature
+++ b/newsfragments/3032.feature
@@ -1,1 +1,2 @@
-``dials.index``: two new parameters, ``xy_rmsd_threshold`` and ``n_indexed_threshold`` are introduced to exclude obvious bad indexing solutions, with defaults set to 4 pixels and 5 reflections respectively.
+``dials.index``: two new parameters, ``xy_rmsd_threshold`` and ``n_indexed_threshold`` are introduced to exclude obvious bad indexing solutions.
+The default ``xy_rmsd_threshold`` is 6.0 pixels for rotation data and 2.0 pixels for stills, while ``n_indexed_threshold=5`` in either case.


dagewa:
I'm afraid I don't have a strong defensible justification for this value. I started with 0.5 mm, because that rejected all solutions for the data I got from https://github.com/xia2/xia2/issues/889, which is highly disordered. Then I realised the check should be in pixels, not mm, and I erroneously converted to 3.0 pixels assuming a Pilatus chip whereas actually the data was from an Eiger. Then I found this caused some tests to fail so I increased it to 4.0 pixels.

I don't think I have ever seen an indexing solution with RMSDs > 4 pixels that has gone on to provide a useful result, but I am happy to be corrected and could raise this threshold. Has anybody got a data set where the indexing solution is correct, useful and should not be rejected, but in fact the RMSDs are greater than 4 pixels?

dagewa:
Ok, judging by the test `test_index.py::test_rigaku_hypix_arc_100`, the 4 pixel RMSD threshold is indeed too tight. It does still index in that case, but only indexes up to 70% of the spots. So, it seems one of the better solutions is rejected from the candidates. If I increase to 5.0 pixels then it indexes 97% of spots, but I can see that three solutions are still rejected from the list with RMSDs up to 5.3 pixels.

A 6.0 pixel threshold might be okay? I think you're right @graeme-winter, it does depend on properties of the lattice. Not sure how best to bring that into the check. Maybe starting with a liberal threshold for now is at least a step in the right direction?

phyy-nx:
What about an auto parameter that's 6 for sweeps and 2 for stills?

biochem-fan:
> In practice I've only gone above 2px cutoff in limited cases, and it's usually been because something else was wrong, like geometry or bad data that is giving too high a mosaic estimate.

At SACLA, the facility-provided initial geometry is not very accurate but we don't normally run powder standard for calibration. To bootstrap indexing, I usually have to increase it. In addition, the RMSD might be inflated due to multiple lattices.

>  I think you're right @graeme-winter, it does depend on properties of the lattice. 

How about using fractional RMSDs? Fractional errors are calculated during index assignment (alien spots are rejected by `hkl_tolerance` there).

That being said, fractinoal errors also have cell dependency. The default `hkl_tolerance` is too large for small cells (e.g. inorganic materials). A fractional error of 0.3 (the default) can mean a spot displacement of > 20 px! in MicroED of inorganic materials, I usually decrease it to 0.15; otherwise, many spots from other lattices are included in refinement and the unit cell is distorted.

phyy-nx:
I think your proposed default of 5 for `n_indexed_threshold` is fine for this PR, since it's not something that will be used by small_cell at present.  But, if in the future, after #2477 is merged, and the user ran this command:

```
dials.index imported.expt strong.refl stills.indexer=sequences indexing.method=small_cell
```

Then the ModelEvaluator _would_ be used, which means `n_indexed_threshold` would be set too high (5) and should instead be 3.

Whew!  We can handle that case in #2477.

dagewa:
Ok, I've now run an A/B test using 232 Eiger data sets from [proteindiffraction.org](https://proteindiffraction.org/). No data sets failed to process (with xia2) on the branch compared to main. 82 of the data sets give identical results, as assessed by the contents of `xia2-summary.dat`. 150 data sets do give slightly different results. I have not checked if any of these are significantly different yet, but generally the differences look small. Here's one example:

```diff
--- ./trvaa_17083_b_b2-so4_9bky/data/PSL-0315_8439/A/xia2-summary.dat	2026-01-28 19:44:09.542021330 +0000
+++ ./trvaa_17083_b_b2-so4_9bky/data/PSL-0315_8439/B/xia2-summary.dat	2026-01-28 20:02:53.043111371 +0000
@@ -11,12 +11,12 @@
 For AUTOMATIC/DEFAULT/NATIVE:
 High resolution limit                           1.51    4.10    1.51
 Low resolution limit                           72.40   72.48    1.54
-Completeness                                  100.0   100.0    99.5
+Completeness                                  100.0   100.0    99.6
 Multiplicity                                   26.8    26.7    23.5
-I/sigma                                         8.8    47.8     0.4
-Rmerge(I+/-)                                  0.234   0.075  19.911
-CC half                                       0.999   0.998   0.433
-Anomalous completeness                        100.0   100.0    99.5
+I/sigma                                         8.8    47.3     0.4
+Rmerge(I+/-)                                  0.228   0.077  18.231
+CC half                                       0.998   0.997   0.436
+Anomalous completeness                        100.0   100.0    99.6
 Anomalous multiplicity                         14.1    15.2    12.2
-Cell:  86.406  86.406 132.637  90.000  90.000  90.000
+Cell:  86.403  86.403 132.634  90.000  90.000  90.000
 Spacegroup: P 41 21 2
```

PR #3038, opened 2025-11-01
LCLS psana2 support

phyy-nx:
One of three pull requests to add psana2 support to cctbx.xfel and dials.  The others are:

cctbx/cctbx_project#1111
cctbx/dxtbx#860

For more information, see the dxtbx PR.

Specifics for dials, accounting for psana2 requiring the use of MPI and specifically reserving ranks 0 and 1:
- Allow zero-length experiments during import (ranks 0 and 1 will have no data)
- Check for psana2, and if present, use rank 2 or higher after loading the data (have to load the data first so ranks 0 and 1 can see it).

Lots of folks worked on this, special thanks to @fredericpoitevin and @monarin!


fredericpoitevin (approved):
Thanks @phyy-nx

phyy-nx:
> LGTM. I don't know if that XFEL CI failure is related or an old one?

Error is from not having the latest cctbx code which renamed some parameters

PR #3040, opened 2025-11-03
Refstat - a tool to check space group symmetry

pcxod:
He, never tried myself outside of cctbx, but give it a try!

### DIFF
--- a/src/dials/command_line/refstat_symmetry_analysis.py
+++ b/src/dials/command_line/refstat_symmetry_analysis.py
@@ -51,6 +51,10 @@ phil_scope = libtbx.phil.parse(
         .help="Path to an .ins or .res file to calculate statistics for a"
               "particular example"
 
+    sigma_level = 5.0
+        .type = float
+        .help = "Sigma level to use to identify systematic absences"
+
     output {
         log = dev.dials.refstat_symmetry_analysis.log
             .type = str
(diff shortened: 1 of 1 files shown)


dagewa:
Thanks! I will have time to look at this during next week.

dagewa:
I still want to work on extending the tests

jbeilstenedmands (changes requested):
On the whole, the symmetry analysis looks good, but I'm not sure we want to include the command-line tool in its current form.
Long term we are not planning to need this (as it will be part of dials.symmetry/dials.cosym), and if we want it to exist as a "dev.dials" tool for short-term testing, then it should match the expected usage of other dials programs - i.e. it works on dials reflection and experiment files. In its current form, it looks to want either/or a test directory of olex data or .ins and .res files. But the program should just require intensity and miller index data from a dials reflection table and metadata available in the experiment object, unless I am missing something? We shouldn't need to create .ins and .res files, which would not be part of a typical dials workflow. And then we should add a small test to run the command line tool too.

jbeilstenedmands:
Thanks David. I think for this PR, just getting it in as a standalone `dev.dials` program is sufficient and quicker, and then we can immediately follow up with work to integrate into `dials.symmetry`. But for that testing it will be much better to be able to run on `.refl` and `.expt` files.

PR #3051, opened 2025-11-11
Require joint_index option to be explicit

### LINKED ISSUE #3050: joint_index should not have a default
Following from a discussion on slack with @graeme-winter and @ndevenish.  Enough users get tripped up by joint_index=Auto that we should change the default.  Recall that joint_index=True will try to find a common indexing solution for all experiments, while False determines a crystal model for each experiment.

Current behavior: Auto
- If stills: False
- If sweeps: True
- If mix: error

Suggested behavior: Auto
- If stills: False
- If sweeps: error, user must specify True or False explicitly
- If mix: error

Other suggested ideas:
- Change the Auto for sweeps to False, but enough users use disjoint sweeps that that's not ideal either
- Try and guess for sweeps.  If most of the indexed reflections are in one sweep or if only one sweep indexes and the rest fail, then fall back to joint_index=False.  This seems too complicated.

graeme-winter:
Rather than assuming multi-sweep from single crystal or multiple

New output:

```console
ethics-gradient:C graeme$ dials.index imported.expt strong.refl 
DIALS 3.dev.1386-g1a2d56924

input {
  experiments = imported.expt
  reflections = strong.refl
}

Unable to set joint_indexing automatically: set to True for multi-axis data from a single crystal or False for data from multiple samples
```

Resolves #3050

### DIFF
--- a/tests/algorithms/indexing/test_index.py
+++ b/tests/algorithms/indexing/test_index.py
@@ -345,6 +345,7 @@ def test_index_insulin_multi_sequence(insulin_spotfinding, tmp_path, method):
         f'known_symmetry.space_group="Hall: {expected_hall_symbol}"',
         f"indexing.method={method}",
         "treat_single_image_as_still=False",
+        "joint_indexing=True",
     ]
     run_indexing(
         reflections,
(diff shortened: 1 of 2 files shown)


graeme-winter:
I am now playing `pytest` whak-a-mole

PR #3059, opened 2025-12-03
remove duplicate line

rachel-dls:
Within the function `create_datastructures_for_reference_file`, the line `table.experiment_identifiers()[len(experiments)] = expt.identifier` was used twice.


jbeilstenedmands (changes requested):
Good spot!

Please add your name to the authors file to recognise your contribution:
https://github.com/dials/dials/blob/main/AUTHORS

PR #3061, opened 2025-12-05
Rework dials.stills_process output

phyy-nx:
- Convert output.XXX_filename to output XXX_suffix
- Standardize output so all the files are `*indexed*` or `*refined*`, not a mix
- Only save experiments if there are reflections in the matching reflection tables

I expect tons of tests and pipelines to break so this is very much a draft, but it's the direction I'd like to go for d.sp.

I'd like to make the transition easier by catching the renamed parameters using deprecations when the program starts, such as `indexed_filename = %s_indexed.refl` becoming `indexed_suffix=indexed`

### DIFF
--- a/src/dials/algorithms/integration/sublattice_helper.py
+++ b/src/dials/algorithms/integration/sublattice_helper.py
@@ -115,32 +115,21 @@ def integrate_coset(self, experiments, indexed):
     if self.params.integration.debug.delete_shoeboxes and "shoebox" in integrated:
         del integrated["shoebox"]
 
-    if self.params.output.composite_output:
-        if (
-            self.params.output.coset_experiments_filename
-            or self.params.output.coset_filename
-        ):
-            assert (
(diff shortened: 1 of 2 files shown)


phyy-nx:
Specifically:

- `experiments_filename` becomes `imported_suffix`, only outputs experiments
- `strong_filename` becomes `strong_suffix`, now outputs reflections and experiments, such that imported and strong are separate from each other
- Instead of `refined_experiments_filename` and `indexed_filename` being used for the output of both indexing and refinement, there is a separate `indexed_suffix` and `refinement_suffix`
- `integrated_experiments_filename` and `integrated_filename` becomes `integrated_suffix`
- `coset_experiments_filename` and `coset_filename` becomes `coset_suffix`

Additionally, the `%s_indexed.expt` and `%s_indexed.refl` convention is dropped.  Instead the suffixes are simply `indexed` or `refined`, etc., and the filenames are assembled like `idx-0053_indexed.expt`.

PR #3072, opened 2025-12-17
Reduce memory footprint of the `ArgumentParser`

dagewa:
I don't pretend to _fully_ understand this, but this small change is enough to reduce the memory footprint of DIALS programs by a significant amount.

memory-profiler plots for `dials.refine`, as an example:

## main
main

## PR
branch


ndevenish:
Hmm, we've had problems before where copies of all the data were kept around, maybe the system relies on errors happening during parsing and it's keeping copies of all the stack objects e.g. traceback, possibly sources?

dagewa:
Yeah, looks like we always try to read reflections as an experiment list (and vice versa), and catch the exceptions. But it gets bad when we then store the exception because it seems to hide hundreds of megabytes of data

PR #3074, opened 2025-12-18
Improve `dials.integrate` required memory estimate

dagewa:
Following on from https://github.com/dials/dials/issues/659#issuecomment-3656396834, this PR includes the overhead of the current process when calculating the total memory required per subprocess in integration. This gives a much more accurate estimate when compared with the actual memory per process as assessed by `memory-profiler`.

Here is a recalculation of the data and plot from https://github.com/dials/dials/issues/659#issuecomment-3656396834, now using this PR, and including also the change from #3072, which reduces the overhead.

| nproc | peak memory (MB) | average mem per process (MB) | estimated mem per process (MB) |
| ----- | ---------------- | ----------------------- | ------------------------- |
| 16    | 57362.929688     | 3585.183                | 3488                      |
| 12    | 44280.488281     | 3690.041                | 3492                      |
| 8     | 30841.175781     | 3855.147                | 3538                      |
| 4     | 17003.437500     | 4250.859                | 3546                      |
| 1     | 4889.597656      | 4889.598                | 3766                      |

memusage

### DIFF
--- a/newsfragments/3074.bugfix
+++ b/newsfragments/3074.bugfix
@@ -1,1 +1,1 @@
-``dials.integrate``: Include overhead in the calculation of memory required per process
+``dials.integrate``: Include overhead in the calculation of memory required per process, report memory calculations in MB rather than GB, and reduce the default ``max_memory_usage`` parameter to 0.8 rather than 0.9.


dagewa:
I think this one would be worth testing on other platforms and with a variety of data sets before merging. I'll try to come up with a simple test script that automates this.

PR #3082, opened 2026-01-21
Allow `dials.rs_mapper` to work with `double` raw image data

dagewa:
Previously this assumed `int` raw data only. This PR makes it work with double precision raw data (as found in e.g. some MRC format files) by templating the `fill_voxels` function.


biochem-fan:
Thank you very much for fixing this.

While testing this, I realized one more subtle issue. The program outputs an empty volume even when no image and experiment is given.

phyy-nx:
> Thank you very much for fixing this.
> 
> While testing this, I realized one more subtle issue. The program outputs an empty volume even when no image and experiment is given.

Noting this.  @biochem-fan if that's still the case after @dagewa's fixes could you create an issue?

phyy-nx:
Never mind, shoulda read more.  Looks like it's fixed.

PR #3093, opened 2026-01-28
Use sphinx-design for dropdowns in tutorial pages. 

### LINKED ISSUE #3078: "Show/Hide Log" buttons in the Processing in Detail tutorial do not show/hide the log.
The online [Processing in Detail tutorial](https://dials.github.io/documentation/tutorials/processing_in_detail_betalactamase.html) automatically shows the full log files created by running each command. Some of these are quite long, so there's text just above the log that says "Show/Hide Log" with a little triangle to indicate whether the full text is being displayed or not.

Unfortunately, this "Show/Hide Log" is not actually a clickable button (I opened the page in Chrome, Firefox, and Safari to confirm that it wasn't a browser issue somehow), and so there's no way to actually hide the log. This isn't a big issue - it would be much more problematic if the logs were automatically hidden with no way to show them - but it's slightly annoying when trying to quickly scroll through the tutorial to locate a particular piece of information, especially with the knowledge that hiding the log is _meant_ to be possible...

jbeilstenedmands:
Our CSS/JS container toggle patterns are no longer supported with more modern sphinx themes, instead the recommended approach now seems to be to use the sphinx-design package, rather than implementing our own CSS/JS components.

Fixes #3078

This is how it now looks by default (closed):
Screenshot 2026-01-28 at 10 45 18

And open:
Screenshot 2026-01-28 at 10 45 58

### DIFF
--- a/newsfragments/3100.misc
+++ b/newsfragments/3100.misc
@@ -1,0 +1,1 @@
+Update integration processor test
--- a/newsfragments/3101.misc
+++ b/newsfragments/3101.misc
@@ -1,0 +1,1 @@
+Fix refinement constraints manager test precision
--- a/tests/algorithms/integration/test_processor.py
+++ b/tests/algorithms/integration/test_processor.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import math
+import os
 from unittest import mock
 
+import psutil
 import pytest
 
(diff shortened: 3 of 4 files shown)


dagewa:
Nice! Looks better than the previous as a side-effect of fixing the functionality :+1:

PR #3102, opened 2026-02-09
Switch html cluster output

amyjaynethompson:
To make the output of dials.correlation_matrix easier to read, the correlation/cosine-angle tables have been moved below the heat map/dendrogram plots. Often, the graph is the critical part, though this was previously below the table listing the clusters. For large datasets, this required a significant amount of scrolling, which is not the most friendly... 

New format looks like this - @dagewa is this what you were hoping for? 

Screenshot 2026-01-30 at 08 45 37


dagewa (approved):
Yes! That's exactly what I wanted. Thanks :+1:

Issue #3117, opened 2026-02-23
The `Shoebox` struct is overweight


dagewa:
`Shoebox` contains 3 arrays of a size determined by the bounding box: data, background, and mask. The data and background arrays are both 4 byte `float`, while the mask array is a 4 byte `int`.

There are two obvious issues.

The simplest one is that `int` is unnecessary for the mask. There are [only 6 bits](https://github.com/dials/dials/wiki/shoebox_pixel_masks) for shoebox masks, so we could store this as a `char` array.

The second is that there is no need to store all the background pixels. This array is allocated with zeroes when the data array is read in, and then filled with background values according to the model. However, rather than storing background values only for pixels in the peak region, we store values right out to the bounding box limit.

Changing this in the general case to store only values for pixels in the peak might be somewhat complicated, however for the standard `constant3d` model there is a shortcut. This model assigns the same value, `mean_background`, to every pixel in the shoebox:

https://github.com/dials/dials/blob/c183a352983bf456d2cfb210d8e6755c2030ec8b/src/dials/algorithms/background/glm/creator.h#L279-L285

We might be able to make use of this shortcut to simply keep hold of this single value and never allocate the background array.

Taken together, these changes might be able to reduce the memory requirements of the shoebox from 12 bytes per pixel down to just 5 bytes (for the usual `constant3d` code path). This could be significant given how large we see these arrays can be https://github.com/dials/dials/issues/2849

dagewa:
I think #3119 and #3148 address this sufficiently. The idea in my original post to use only a single element for the `constant3d` code path would reduce the shoebox size further, but I think the additional complexity of handling a special case is not worth it. The other idea, using `float` images everywhere, is not related to the shoebox, so no need to continue tracking it here (I will still try it out at some point).

PR #3119, opened 2026-02-25
One-byte-per-pixel shoebox mask

dagewa:
WIP attempt to change the `Shoebox` mask array to use a single byte per pixel (`uint8_t` elements) rather than 4 bytes (`int` elements), for #3117.

The branch builds fine, but `dials.integrate` currently fails with a traceback ending with

```
  File "/home/fcx32934/sw/cctbx/modules/dials/src/dials/algorithms/profile_model/gaussian_rs/calculator.py", line 82, in _beam_direction_variance_list
    mask = shoebox[r].mask != 0
           ^^^^^^^^^^^^^^^
TypeError: Please report this error at https://github.com/dials/dials/issues or to dials-user-group@jiscmail.ac.uk: No to_python (by-value) converter found for C++ type: scitbx::af::versa<unsigned char, scitbx::af::c_grid<3ul, unsigned long> >
```

The problem appears to be that the export for the `Shoebox` class does not make a proper getter for the `mask`:

https://github.com/dials/dials/blob/b22b82aa45f695c00c63ebb9965ccce7b41deec4/src/dials/model/data/boost_python/shoebox.cc#L174-L177

If anyone knows how to make this work I would really appreciate it! The right flex type (`flex.uint8`) exists since  https://github.com/cctbx/cctbx_project/pull/533. How do I make a getter work with this type?


dwpaley:
XFEL CI passes for this branch on cctbx_project Azure: https://github.com/cctbx/cctbx_project/tree/test_dials_3119

Issue #3125, opened 2026-03-19
Error integrating `four_circle_eiger` data set: `Unsupported integer dtype int64`


dagewa:
I get a stacktrace trying to process the `four_circle_eiger` data with DIALS 3.27.0. These commands:
```bash
dials.import $(find $(dials.data get -q four_circle_eiger) -name *.nxs)
dials.find_spots imported.expt 
dials.index imported.expt strong.refl joint_indexing=True space_group=P4
dials.refine indexed.{expt,refl}
dials.integrate refined.{expt,refl}
```

lead to this error:

```
easy_mp crash detected; subprocess trace: ----
Traceback (most recent call last):
  File "/home/fcx32934/sw/dials-v3-27-0/modules/cctbx_project/libtbx/scheduling/job_scheduler.py", line 74, in job_cycle
    value = target( *args, **kwargs )
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dials/src/dials/algorithms/integration/processor.py", line 214, in execute_parallel_task
    result = task()
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dials/src/dials/algorithms/integration/processor.py", line 469, in __call__
    image = imageset.get_corrected_data(i)
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dxtbx/src/dxtbx/format/FormatMultiImage.py", line 47, in read
    return format_instance.get_raw_data(index)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dxtbx/src/dxtbx/format/FormatNXmx.py", line 164, in get_raw_data
    raw_data = dxtbx.nexus.get_raw_data(
        nxdata, nxdetector, index, bit_depth=self._bit_depth_readout
    )
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dxtbx/src/dxtbx/nexus/__init__.py", line 587, in get_raw_data
    data_as_flex = _dataset_as_flex(
        sliced_outer, tuple(module_slices), bit_depth=bit_depth
    )
  File "/home/fcx32934/sw/dials-v3-27-0/modules/dxtbx/src/dxtbx/nexus/__init__.py", line 548, in _dataset_as_flex
    raise TypeError(f"Unsupported integer dtype {data_np.dtype}")
Unsupported integer dtype int64
----------------------------------------------


Traceback (most recent call last):
[...]

PR #3126, opened 2026-03-19
Add refstat algorithm to `dials.symmetry`

dagewa:
Work in progress to bring the refstat algorithm into `dials.symmetry`, activated using a new `small_molecule=True` parameter. I'm not completely happy with the parameter name, but it is consistent with xia2.

At the moment I have not attempted to change the scoring system for detected symmetry elements and suggested space groups to a likelihood-based score, like the existing systematic absence detection code in `dials.symmetry`. One reason is that I had an example (unfortunately I've forgotten which) in which refstat's current scoring correctly identified a screw axis that `dials.symmetry` rejected. Another reason is that the current scoring method in refstat, while simplistic, has been tested against numerous data sets. Switching that to a different method would require a lot of re-testing to make sure it is actually better.

The way refstat scores symmetry elements (screw axes or glide planes) is direct, by looking at the mean intensity of putative absences and marking the element as present if the mean intensity is 0.0 within $n \sigma$, where $n$ is set by `sigma_level=5` (default). Elements that are present are marked with `+`, elements that are not found are marked with `-`, and elements where absences are shadowed by other symmetry elements, or there are no reflections to assess are marked with `+-`.

All space groups are then assessed, filtering out any with the wrong centring type or those that are not compatible with the unit cell, and then ranking according to:
1. Lower $R_{int}$
2. If $R_{int}$ is equal within a small tolerance then a higher number of observed present (`+`) elements matching is preferred
3. If the number of matching elements is the same then a higher number of weak reflections is preferred.

The resulting list of space groups is presented in a table like this (using the `quartz_processed` data set):

```
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
| Space group   | Centric   |   Matches (%) |   Incons. |   Rint |   #Weak |   Weak I/σ(I) |   #Strong |   Strong I/σ(I) |
|               |           |               |    equiv. |        |         |               |           |                 |
|---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------|
| P 31          |           |            50 |         2 |  7.229 |      12 |          1.26 |         4 |           14.76 |
| P 32          |           |            50 |         2 |  7.229 |      12 |          1.26 |         4 |           14.76 |
| P 31 2 1      |           |            50 |         0 |  7.527 |      12 |          1.26 |        10 |           19.46 |
| P 32 2 1      |           |            50 |         0 |  7.527 |      12 |          1.26 |        10 |           19.46 |
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
```

But unlike the existing `dials.symmetry` algorithms, these results are not limited to a single Laue group. As we only intend to do a systematic absence check here, I added a further filtering step, to select only the results in the Laue group already chosen by `dials.symmetry`. In this case, `dials.symmetry` selected `Laue group: P -3 m 1`, so the table is further filtered to leave just $P\  3_1\ 2\ 1$ and $P\  3_2\ 2\ 1$, which are indistinguishable so the first is selected.

The `quartz_processed` example is useful to demonstrate that Laue group filtering behaviour, however I would really like to add a test that involves a space group with a glide plane. At the moment I don't think we have a suitable data set in dials-data. I tried processing the `four_circle_eiger` data set, which I believe should be in $P\ 4/n\ m\ m$, but I hit an issue with that: https://github.com/dials/dials/issues/3125


jbeilstenedmands:
I am happy with the general form of this PR - I think it is fine to use a different scoring method for this different algorithm. The "small_molecule=True/False" parameter name is also acceptable and familiar to users from its use in xia2. I think this can pretty much go in as-is, although locally the test output seemed to be different for me, so interested to see what the CI jobs find.

jbeilstenedmands:
Ok, I updated the test based on the output I get - as this is a regression test, it won't be run by the CI, so could someone else please check? (`tests/command_line/test_symmetry.py::test_small_molecule`)

dagewa:
I get a failure on that test now. It seems that on my system there are fewer reflections that qualify as "strong" for the last two space groups:
```
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
| Space group   | Centric   |   Matches (%) |   Incons. |   Rint |   #Weak |   Weak I/σ(I) |   #Strong |   Strong I/σ(I) |
|               |           |               |    equiv. |        |         |               |           |                 |
|---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------|
| P 31          |           |            33 |        61 | 14.891 |      12 |           1.1 |         4 |            14.4 |
| P 32          |           |            33 |        61 | 14.891 |      12 |           1.1 |         4 |            14.4 |
| P 31 2 1      |           |            33 |        47 | 16.219 |      12 |           1.1 |         4 |            14.4 |
| P 32 2 1      |           |            33 |        47 | 16.219 |      12 |           1.1 |         4 |            14.4 |
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
Selected results in Laue group P -3 m 1: P 31 2 1, P 32 2 1
Saving reindexed experiments to symmetrized.expt in space group P 31 2 1
Saving 1653 reindexed reflections to symmetrized.refl
```
whereas the test checks for:
```
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
| Space group   | Centric   |   Matches (%) |   Incons. |   Rint |   #Weak |   Weak I/σ(I) |   #Strong |   Strong I/σ(I) |
|               |           |               |    equiv. |        |         |               |           |                 |
|---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------|
| P 31          |           |            50 |        61 | 14.891 |      12 |           1.1 |         4 |           14.4  |
| P 32          |           |            50 |        61 | 14.891 |      12 |           1.1 |         4 |           14.4  |
| P 31 2 1      |           |            50 |        47 | 16.219 |      12 |           1.1 |        10 |           20.32 |
| P 32 2 1      |           |            50 |        47 | 16.219 |      12 |           1.1 |        10 |           20.32 |
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
Selected results in Laue group P -3 m 1: P 31 2 1, P 32 2 1
Saving reindexed experiments to symmetrized.expt in space group P 31 2 1
Saving 1653 reindexed reflections to symmetrized.refl
```
Finding 10 rather than 4 seems quite a big difference - if it was just borderline reflections I might expect 1 or 2. We'll have to dig into this to understand the discrepancy.

dagewa:
My Windows build agrees with the previous test output, that is 4 strong reflections on each row

jbeilstenedmands:
I have been looking into this again, if we set `laue_group=P321` (and `normalisation=ml_iso` for stability), then I get the same results on both platforms:
```
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
| Space group   | Centric   |   Matches (%) |   Incons. |   Rint |   #Weak |   Weak I/σ(I) |   #Strong |   Strong I/σ(I) |
|               |           |               |    equiv. |        |         |               |           |                 |
|---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------|
| P 31          |           |            50 |        64 | 15.422 |      12 |           1.1 |         6 |           16.28 |
| P 32          |           |            50 |        64 | 15.422 |      12 |           1.1 |         6 |           16.28 |
| P 31 2 1      |           |            50 |        48 | 16.906 |      12 |           1.1 |        13 |           19.47 |
| P 32 2 1      |           |            50 |        48 | 16.906 |      12 |           1.1 |        13 |           19.47 |
+---------------+-----------+---------------+-----------+--------+---------+---------------+-----------+-----------------+
Selected results in Laue group P -3 m 1: P 31 2 1, P 32 2 1
Saving reindexed experiments to symmetrized.expt in space group P 31 2 1
Saving 1653 reindexed reflections to symmetrized.refl
```
Without specifying the `laue_group`, a different reindex operator is chosen during the initial laue group analysis on each platform. Are we missing a step where we map back to a standard setting before calling the refstat analysis?

jbeilstenedmands:
I'm wondering if we need this kind of thing in dials.symmetry too:
https://github.com/dials/dials/pull/2548

dagewa:
Hmm yes, maybe we do. I think in general we also had the intention of convergence for `dials.cosym` and `dials.symmetry`, so we should try to make these as consistent as possible. Perhaps we should make a separate issue for that, and then in this PR set `laue_group=P321` and `normalisation=ml_iso` for `test_small_molecule` with a comment pointing at the new issue number?

jbeilstenedmands (approved):
I now see consistent test results on mac and linux, @dagewa could you check the new test gives you the same result on windows?

dagewa:
Thanks! Confirmed, passes for me :+1:

PR #3143, opened 2026-04-09
Correct various problems with elliptical distortion maps

### LINKED ISSUE #3124: `dials.generate_distortion_maps` ellipse mode parameters are incorrect
I got a MicroED dataset from a badly aligned scope.

After data collection, the scope manager examined powder rings from a standard grid and discovered that the mean camera length was 578.4 mm but 559 mm along the short axis and 599 mm along the long axis.

I tried @dagewa's new feature, ellipse fitting tool (https://github.com/dials/dials/pull/3067), which gave me something like this (I averaged results from fitting on different rings):
```
mode = ellipse
ellipse
{
  phi = 173.7
  l1 = 1.000000
  l2 = 0.9344
  centre_xy = 31.752853 32.103444
}
```

I ran `dials.generate_distortion_maps imported.expt ellipse.phil` and imported images with `distance=599 dx=dx.pickle dy=dy.pickle`.

Unfortunately, this disabled profile fitting in integration and both `dials.symmetry` and `dials.scale` complained. Is this expected?

`dials.symmetry`:
```
Mapping all input cells to a common minimum cell
Filtering reflections for dataset 0
Read 15848 predicted reflections
WARN: WARNING: No profile-integrated reflections found
Attempting to reprocess with intensity choice: sum
Read 15848 predicted reflections
Selected 15848 summation integrated reflections
Removed 15848 intensity.sum.value reflections with negative variance
```

`dials.scale`:
```
Applying filter of min_isigi > -5.0, partiality > 0.4
Read 7660 predicted reflections
WARN: WARNING: No profile-integrated reflections found
```

Looking at `integrated.refl`, all `intensity.prf.value` is 0, `intensity.prf.variance` -1, `intensity.sum.value` 0, `intensity.sum.variance` 0. **So summation integration is also broken.**

I thought the distortion map only changed the Px to Mm strategy (i.e. centroid coordinates) and profile fitting was not affected. Am I wrong? As long as local distortion is small, profile fitting should be possible (as in XDS).

(Tested on 3.dev.1437-gd1df3f6b2 and DIALS 3.dev.1339-gc44e33dc4).

dagewa:
As detailed in #3124, there were various problems with elliptical distortion maps. Points were rotated when transformed between circle and ellipse, and the ellipse angle calculated by the `dials.image_viewer` ellipse tool did not match the angle required by `dials.generate_distortion_maps`. The existing tests were insufficient to identify these problems.

This PR corrects those issues and adds a comprehensive test that emulates the steps a researcher would take to correct a distortion map. In that test, an image is first created containing 100 points around an ellipse, and those points are found using `dials.find_spots`:

image

These points are then passed into the code used by `dials.image_viewer`'s ellipse tool to extract ellipse parameters, which in this case are `phi=115.9`, `l1=1.0`, and `l2=0.9`. As part of this PR, the meaning of the angle `phi` is clarified to be the rotation from the fast axis of the major axis in an anticlockwise direction when viewing the image in `dials.image_viewer` (in which typically fast points right and slow points down).

Distortion maps are created with those parameters, the image imported again with those maps, and the spot pixel positions are corrected via the `pixel_to_millimeter` function.

Plotting code (now commented out in the test) shows that the mm positions are visually corrected back to a circle:

mm_intersections

Looking at the original pixel positions (blue) and the _subtracted_ directions from the distortion maps (red arrows) produces the corrected pixel positions (orange).
pixel_intersections

During testing I discovered that there is an error in the correction that increases with the degree of ellipticity. This is small for mild distortion (`l2` >= 0.95) but increases sharply when the elliptical distortion is more severe. The test is run with `l2=0.9`, which is likely already greater distortion than is reasonable to expect for a real microscope. At that point, the error in radial positions of the corrected circle is about 1.2%

l2 | radial error (%)
-- | --
0.99 | 0.2
0.95 | 0.4
0.9 | 1.2
0.8 | 4.2

The cause of this error is not clear to me, but I believe it is acceptable for the levels of ellipticity seen in real data sets.


dagewa:
Note this branch needs a build because C++ code is changed

biochem-fan:
> During testing I discovered that there is an error in the correction that increases with the degree of ellipticity.

> The cause of this error is not clear to me

This is indeed very puzzling. I don't see any reason why we cannot get subpixel accuracy.

dagewa:
Yes, I can't make sense of it. There is some expected error in the test because it uses spot-finding results rather than precise positions, but this should not scale with the ellipticity. I was also considering that the ellipse parameters are determined in pixel space, but then applied by `PlaneLinearTransformationMaps` in mm lab space (because it should work for multiple panel detectors where not all fast, slow coordinate systems are aligned), but I can't see how this should affect it either.

biochem-fan:
@dagewa

> because it uses spot-finding results rather than precise positions

Can you check the results using precise positions? Your older test codes worked in that way.

>  this should not scale with the ellipticity

Not necessarily. Because you render only a single pixel at a rounded coordinate to mimic a diffraction spot, the precision of the pixel position is only 1 px. With Gaussian blobs, the centroid coordinates would be more precise.

https://github.com/dials/dials/blob/e6d58a197e81fa7cbb1f934caec077b307114b69/tests/command_line/test_generate_distortion_maps.py#L339-L340

In the worst case, this introduces an error of $$\sqrt 2$$ px. Consider rendering a pixel at (100.99, 100.99) in (100.0, 100.0).

When we undistort the position, this error is inflated. For example, an observation of a radius 450 px can actually mean 448.6 to 451.4 px. By undistorting 10 %, this becomes 448.6 / 0.9 ~ 498.4 to 451.4 / 0.9 ~ 501.6. The error is up to 1.6 px. The other axis is distortion free but still has the rounding error of up to $$\sqrt 2$$ px. Thus, the total error is about 1.6 + 1.4 = 3 px. This is consistent with your observation "With l2=0.9 we seem to get radial errors of up to around 3 pixel".

dagewa:
I added a precise version of the same test that does not incur round-off error due to spot-finding. This has smaller, but still significant errors in the corrected points that scales with ellipticity:


l2 | radial error (%)
-- | --
0.99 | 0.0
0.95 | 0.3
0.9 | 1.0
0.8 | 4.1

I don't think the problem is with interpretation of the `skimage.measure.EllipseModel` results because this recovers the `l2` and `phi` parameters used to create the ellipse almost exactly. I also don't think the problem is with the ellipse<-->circle transforms, because these are tested by `test_ellipse_transforms`.

I think the problem lies in the creation or application of the `dx` and `dy` maps, but I haven't been able to work out what it is. Perhaps your analysis suggests an off-by-one error in pixel position values somewhere.

dagewa:
I have a solution :grinning:. It turns out we did need the `ellipse_to_circle_transform` rather than the `circle_to_ellipse_transform` after all, and as a result we have to invert the direction of the map shifts. I will commit a further update to tests and comments and then I deeply hope I never have to touch this again. It has somehow been the most confusing feature I've worked on.

dagewa:
Done, I think :sweat_smile: 

Here are the new errors for various values of `l2` used in the test that goes through the spot-finding step. While Gaussian blobs would be more precise, the test was intended to emulate what a user might do using the ellipse fitting tool in `dials.image_viewer`, and I think it is unlikely that a user would click points with better than 1 pixel precision (it is also unlikely a user would bother to click 100 equally-spaced points either but never mind).


l2 | radial error (%)
-- | --
0.99 | 0.2
0.95 | 0.2
0.9 | 0.2
0.8 | 0.3

biochem-fan:
This is wonderful! Thank you very much for your detective work!

dagewa:
I'm sorry it was such a mess and took so long :weary:

biochem-fan:
No problem at all! Distortion correction is always tricky to get right. I remember we took several months to get RELION's anisotropic magnification correction to work with higher-order aberration corrections.

PR #3147, opened 2026-04-14
Option to disable `nproc` throttling in `dials.integrate`

dagewa:
For A/B testing it is desirable to be sure that the same value of `nproc` was used by `dials.integrate` in both arms of the test. This PR interprets `max_memory_usage=None` to mean "do not reduce `nproc`", in which case `dials.integrate` will respect the provided value of `nproc` (or default to using all available processors).

@graeme-winter this also effectively enables the "lucky punk" processing mode you suggested in the user group Slack :gun:


graeme-winter:
@dagewa just to note that I am happily using lucky punk mode to successfully warm up all 16 cores of my M4 where it would only use 11 before

dagewa:
Great! I do think the standard `nproc` throttling is now too aggressive, and it is something I'd like to revisit once the current performance push is wrapping up.

PR #3148, opened 2026-04-14
Just-in-time background allocation

dagewa:
Part of the response to #3117.

The `Shoebox` contains data, mask, and background arrays. As `dials.integrate` iterated through images, shoeboxes were allocated as soon as they come "into view", the data and mask arrays were filled with values in z slices, and as soon as the final image they are present on is read they are sent to integration then deallocated.

This arrangement meant that for most of the life of the shoebox, the background array was allocated with zeroes, because that array is not actually used until integration.

This PR splits the allocation of shoeboxes into `allocate_data` and `allocate_background`. The background array is not allocated until it is needed. In that way, only a single image's worth of shoeboxes needs allocated background arrays at any one time. 

The result is a significant reduction in peak memory requirements. Here is the memory-profiler plots for the [EscoA](https://proteindiffraction.org/project/escoa_17938_a_ae1-apo_8v8w/) data:

## main
main

## branch
branch

and here are the results of an A/B test over the IRRMC data:
peak-mem-A-B-boxplot
peak-mem-A-B-ratio
peak-mem-A-B-steps

The reductions seem similarly as significant as #3119, but these are orthogonal approaches so the memory savings should stack :grinning:


dagewa:
@dwpaley Following your lead with #3119, I made https://github.com/cctbx/cctbx_project/tree/test_dials_3148 to test this

dagewa:
Looks like the xfel tests all passed on that branch.

dagewa:
The XFEL test failure is appearing on other PRs too, and this branch was already [tested successfully](https://github.com/cctbx/cctbx_project/tree/test_dials_3148), so I'm going to merge this

PR #3156, opened 2026-04-23
Return None from do_work() when using mp.method=multiprocessing

dwmoreau:
When mp.method=multiprocessing, easy_mp tries to pickle the return value of do_work() to pass it back to the parent process. The Processor object contains C++ extension objects that are not picklable.

The Processor return value is used only to hand off cached state between calls in MPI/serial mode; under multiprocessing each worker is independent and the return value is unused. Return None in that case.

### DIFF
--- a/newsfragments/3156.bugfix
+++ b/newsfragments/3156.bugfix
@@ -1,0 +1,1 @@
+Fix a pickling error when running dials.stills_process with mp.method=multiprocessing.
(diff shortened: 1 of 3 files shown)


phyy-nx (approved):

dwmoreau:
The processor object is returned by the do_work method, which now retains a cached spotfinding object. These objects, DispersionExtendedThreshold for example, have no pickle support. Instead of returning the processor, which does not get used, return None.

PR #3158, opened 2026-04-23
Cache Indexer objects and PHIL scopes in stills_process Processor

dwmoreau:
Processor.index() was called once per still. Each call constructed a fresh Indexer via Indexer.from_parameters(), which involves PHIL scope extraction, reflection-table setup, and internal solver setup. For large MPI jobs (tens of stills per rank) this overhead dominates.

Indexer caching:
- update_indexer() (nested function inside index()) re-populates experiments/reflections and calls setup_indexing() without the full construction cost of Indexer.from_parameters().
- self.idxr, self.idxr_known_crystal_models, self.idxr_method_list cache one Indexer per code path; each is lazily created on the first still and updated on subsequent ones.

Params mutation refactor:
- The prior code called copy.deepcopy(self.params) once per still to isolate the scan_varying=False override applied during indexing. deepcopy of a large PHIL scope is expensive; replaced with explicit save/restore of scan_varying and max_refine (the only two fields modified). The max_refine=5 override for stills was previously applied inside Indexer.from_parameters (when the value was libtbx.Auto); it is now hoisted to the call site and explicitly restored. Note: this unconditionally sets max_refine=5 for stills regardless of any user-supplied value.

PHIL caching (stills_indexer, fft1d):
- StillsIndexer and FFT1D each called iotbx.phil.parse(indexing_api_defs) on every still. The result is constant for a given indexer instance. Moved to __init__ and stored as self.hardcoded_phil.
- StillsIndexer.index_reflections(): remove copy.deepcopy(all_params); the copy was used only to read immutable fields.

### DIFF
--- a/src/dials/algorithms/indexing/basis_vector_search/fft1d.py
+++ b/src/dials/algorithms/indexing/basis_vector_search/fft1d.py
@@ -55,6 +55,12 @@ class FFT1D(Strategy):
                 reflections, using at most 0.029 radians.
         """
         super().__init__(max_cell, params=params, *args, **kwargs)
+        import iotbx.phil
+        from rstbx.phil.phil_preferences import indexing_api_defs
+
+        self._hardcoded_phil = iotbx.phil.parse(
+            input_string=indexing_api_defs
+        ).extract()
 
     def find_basis_vectors(self, reciprocal_lattice_vectors):
         """Find a list of likely basis vectors.
@@ -67,12 +73,7 @@ class FFT1D(Strategy):
             A tuple containing the list of basis vectors and a flex.bool array
             identifying which reflections were used in indexing.
         """
-        import iotbx.phil
-        from rstbx.phil.phil_preferences import indexing_api_defs
-
         used_in_indexing = flex.bool(reciprocal_lattice_vectors.size(), True)
-
-        hardcoded_phil = iotbx.phil.parse(input_string=indexing_api_defs).extract()
 
         # Spot_positions: Centroid positions for spotfinder spots, in pixels
         # Return value: Corrected for parallax, converted to mm
@@ -88,7 +89,7 @@ class FFT1D(Strategy):
         DPS = DPS_primitive_lattice(
             max_cell=self._max_cell,
             recommended_grid_sampling_rad=self._params.characteristic_grid,
-            horizon_phil=hardcoded_phil,
+            horizon_phil=self._hardcoded_phil,
         )
 
         # transform input into what Nick needs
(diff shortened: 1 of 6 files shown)


src/dials/algorithms/indexing/stills_indexer.py
```diff
@@ -100,6 +100,20 @@ def __init__(self, reflections, experiments, params=None):
         super().__init__(reflections, experiments, params)
         self.warn_if_setting_unused_refinement_protocol_params()
 
+        import iotbx.phil
```
jbeilstenedmands: Is there a reason why this needs to be parsed within the class?
When I process SSX data, I loop through images and process them, so when I initialize the StillsIndexer for each image the parsing is repeated and it is still quite expensive. Instead could we define "hardcoded_phil" as a variable outside of any classes, so that it is run on import once rather than repeatedly?

src/dials/algorithms/indexing/basis_vector_search/fft1d.py
```diff
@@ -55,6 +55,12 @@ def __init__(self, max_cell, params=None, *args, **kwargs):
                 reflections, using at most 0.029 radians.
         """
         super().__init__(max_cell, params=params, *args, **kwargs)
+        import iotbx.phil
```
jbeilstenedmands: Should this also be moved outside of the class to avoid re-parsing? Or is the strategy only ever initialized once? Regardless, we could probably combine both hardcoded phil parsing into one location?

jbeilstenedmands (changes requested):
I have tested ``dials.ssx_index`` and these changes are fine, but I think we can still do better by moving the hardcoded phil imports out of the classes. I haven't tested the changes to `stills_process.py` as I don't have enough knowledge of the ins and outs of how everything should work there.

dwmoreau:
@jbeilstenedmands Thanks for the comments. I like your suggestion to "define "hardcoded_phil" as a variable outside of any classes, so that it is run on import once rather than repeatedly" and will try to incorporate it into this edit. I am waiting to move on this PR until the transition of dials.stills_process to ImageSequence has been fully explored.

src/dials/algorithms/indexing/basis_vector_search/fft1d.py
dwmoreau: See the new commit. These are moved out of the class and unified in __init__.py. The caching is handled via the functools.cache decorator.

jbeilstenedmands (approved):
Thanks for addressing the suggestions, looks good to go.

PR #3159, opened 2026-04-23
Vectorize RealSpaceGridSearch.score_vectors() and candidate_orientati…

dwmoreau:
…on_matrices()

RealSpaceGridSearch.score_vectors():
- The prior implementation looped over every search vector calling compute_functional() one at a time (O(N_sv) Python calls per still).
- New: precompute the (N_sv, 3) search vector array once in __init__ (constant for a given target cell). score_vectors() becomes a single matrix multiply: scores = cos(2pi * sv_array @ rlv.T).sum(axis=1).
- The companion flex.vec3_double of the same vectors is also precomputed once and returned directly, avoiding per-call conversion.
- Import switches from scitbx.array_family.flex to scitbx.matrix for constructing the per-triplet col() vectors passed to Crystal(); flex is still used indirectly via the return value.

candidate_orientation_matrices():
- The prior code built an n^3 flex.vec3_int of all (i, j, k) index triples, sorted by vector norm, then applied three sequential Python filters.
- New: itertools.combinations generates i<j<k triplets directly, avoiding the n^3 allocation. Triplets are sorted by squared index-norm (np.argsort(norms_sq, kind="stable")) to preserve the original ordering -- important because downstream code breaks after max_refine candidates. All three filter passes (angle(a,b), angle(a x b, c), right-handedness) are applied as numpy boolean masks over the full candidate batch before entering the C++ Crystal/Niggli-reduction loop, which runs only over the fraction that passes all filters.
- Bug fix: the original code computed alpha = b.angle(c, deg=True) but then compared it to half_pi (radians), causing some valid candidates to be flipped incorrectly. The numpy path uses radians throughout.

### DIFF
--- a/src/dials/algorithms/indexing/basis_vector_search/combinations.py
+++ b/src/dials/algorithms/indexing/basis_vector_search/combinations.py
@@ -5,6 +5,7 @@ import math
 from itertools import combinations as iter_combinations
 
 import numpy as np
+
 from cctbx.sgtbx.bravais_types import bravais_lattice
 from cctbx.uctbx.reduction_base import iteration_limit_exceeded
 from dxtbx.model import Crystal
--- a/src/dials/algorithms/indexing/basis_vector_search/combinations.py
+++ b/src/dials/algorithms/indexing/basis_vector_search/combinations.py
@@ -2,11 +2,13 @@ from __future__ import annotations
 
 import logging
 import math
+from itertools import combinations as iter_combinations
 
+import numpy as np
 from cctbx.sgtbx.bravais_types import bravais_lattice
 from cctbx.uctbx.reduction_base import iteration_limit_exceeded
 from dxtbx.model import Crystal
-from scitbx.array_family import flex
+from scitbx import matrix as scitbx_matrix
 
 from dials.algorithms.indexing import DialsIndexError
 from dials.algorithms.indexing.compare_orientation_matrices import (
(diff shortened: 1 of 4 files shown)


src/dials/algorithms/indexing/basis_vector_search/combinations.py
jbeilstenedmands, line 28: I think we need this to avoid edge case crashes
```suggestion
    n = len(basis_vectors)
    if n < 3:
        return []
```

jbeilstenedmands (changes requested):
I was testing this out on rotation data and it looks to knock up to a second off the runtime, which is nice, however one of the tests is failing:
`tests/algorithms/indexing/test_index.py::test_multiple_experiments`
with the error:
```
File "/dials/modules/dials/src/dials/algorithms/indexing/basis_vector_search/combinations.py", line 38, in candidate_orientation_matrices
    norms_sq = (idxs**2).sum(axis=1)
  File "/dials/conda_base/lib/python3.13/site-packages/numpy/_core/_methods.py", line 49, in _sum
    return umr_sum(a, axis, dtype, out, keepdims, initial, where)
numpy.exceptions.AxisError: axis 1 is out of bounds for array of dimension 1
"""
```
This can happen when n < 3, so we need an earlier guard for that.

jbeilstenedmands (approved):

PR #3160, opened 2026-04-23
Simplify groups_cache() and avoid space_group_type construction

dwmoreau:
The old implementation keyed the cache on
  group_info.type().lookup_symbol()
which constructs a space_group_type object on every lookup just to get a string key. The new implementation keys on
  (bravais_t, group_info.group())
where group_info.group() returns an sgtbx.space_group, which is hashable via its C++ operator==. The nested MultiClassCache descriptor is replaced by a single flat dict closed over in the decorator, removing a layer of indirection.

### DIFF
--- a/newsfragments/3160.misc
+++ b/newsfragments/3160.misc
@@ -1,0 +1,1 @@
+Simplify the symmetry groups cache to key directly on the space_group object, avoiding redundant space_group_type construction on each lookup.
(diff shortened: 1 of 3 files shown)


jbeilstenedmands (approved):
Nice, tested and in my test case this saves about 10% of ssx indexing runtime, by reducing the number of calls to `group_info.type()` from 232K to 42K. This is a better version of the first implementation of caching that I introduced and looks like it should give equivalent function, and the tests pass.

Issue #3139, opened 2026-03-27
DIALS processing synchrotron stills: add option to retain experiments through the process


graeme-winter:
In discussion with @maggie-klureza

At the moment when we work through a set of still shots we treat it like a filter, with only "successful" experiments being passed onto the next stage. This is sane. It could however be sometimes useful to retain _all_ the experiments through the process to simplify book keeping for more challenging (insane) experiments where we are trying to do weird and unique things which can only be done with DIALS.

Would it be possible to add a `retain_experiments=True` to `dials.ssx_index` and `dials.ssx_integrate` such that the "shape" of the data remains the same going through the pipe. This would also require ensuring that experiments with no reflections do not kill `dials.refine` etc.

I note with some regret that there is prior art here https://github.com/dials/dials/issues/1029 and recognise that this could be termed my white whale. Adding this option to routine processing could also be useful, but I am happy it is an option for people doing weird things.

Confession: I am now building beamlines to do weird things so have more skin in the game than before.

jbeilstenedmands:
Hi @graeme-winter 
Happy to look into adding this as an option to the dials.ssx_* programs, shouldn't be too much change to the bookkeeping. I think the tricky bit would then be getting other programs like dials.refine to accept these files. I was never fully happy with my attempted solutions for issue 1029, however just trying it optionally for the dials.ssx_* programs could be a gentler way into tackling this.

jbeilstenedmands:
Ok @graeme-winter I'm thinking about this a bit more, let's clarify what we want exactly.
Currently, after import/spot finding N ssx images, our ExperimentList (EList) contains N experiment objects, sharing a single imagesequence with N scan objects indexing into that sequence.
If retain_experiments=True, the EList after indexing would look like this:
- if an image is not indexed, the EList contains the same single experiment object from import (up from 0 currently)
- if a single lattice is indexed, the EList two experiment objects, which share a common scan object, one of which contains a crystal.
- similarly, if N>1 lattices are indexed, we have N+1 experiment objects sharing a common scan, N of which contain crystals.

Then the reflection table will contain unindexed and indexed reflections (i.e. no reflections lost from strong.refl), with suitable identifier mapping. The only tricky thing would be if all spots are indexed on an image, then you would have an experiment with no corresponding reflections - we would need this to be mapped in the reflection table's experiment identifiers map, but there would be no corresponding id in the "id" column - this is correct but we would need to make sure the downstream programs could respect this.

Would this enable you to do what you were planning?

PR #3161, opened 2026-04-24
Add option to retain experiments in dials.ssx_index

jbeilstenedmands:
For https://github.com/dials/dials/issues/3139

Add `retain_unindexed_experiments=False/True` option. This retains all the input models and unindexed reflection data.

i.e. if 7 crystals are indexed on 10 images, and there are 200 input strong spots total, the output experiment list contains 17 experiments, 7 of which have crystal models, and the reflection table contains 200 reflections which are a mix of indexed and unindexed.


src/dials/command_line/ssx_index.py
```diff
@@ -83,6 +83,13 @@
     .type = int
     .expert_level = 2
     .help = "Images with fewer than this number of strong spots will not be indexed"
+retain_experiments = False
+    .type = bool
```
graeme-winter, line 87: Practically speaking: for this to be useful the follow-on programs _will_ need to work - `dials.ssx_integrate` and `dials.scale` which in practice means all the follow on programs... I appreciate that this is making extra work.

src/dials/algorithms/indexing/ssx/processing.py
```diff
@@ -247,10 +254,13 @@ def index_all_concurrent(
     n = 0
     original_isets = list(experiments.imagesets())
     identifiers_to_scans = {expt.identifier: expt.scan for expt in experiments}
+    filtered_refls = {}
```
graeme-winter: everywhere else these are `_reflections` - also filtered how?

I suspect you are looking for `rejected_reflections` or similar

src/dials/algorithms/indexing/ssx/processing.py
```diff
@@ -306,16 +326,89 @@ def _join_indexing_results(
...
-    indexed_reflections = flex.reflection_table()
-
     use_gonio = None
     if len(experiments.goniometers()):  # need a placeholder gonio
         use_gonio = experiments.goniometers()[0]
-
     n_tot = 0
-    for res in results:
+
+    if retain:
```
graeme-winter: better to name this `retain_experiments` for consistency with the command line option? Otherwise a naive reader may get confused that:

```python
if retain:
    add_stuff_back()
```

which does not feel like retention - also where are the input (crystal-less) experiments in the structure? A naive reader (i.e. this reviewer) may expect them to be passed through then have the experiment list _extended_ to include the ones with crystals - here it looks like the input ones are appended to the output list?

src/dials/algorithms/indexing/ssx/processing.py
```diff
@@ -306,16 +326,89 @@ def _join_indexing_results(
...
+                    output["id"] = flex.int(output.size(), n_tot)
+                    output.experiment_identifiers()[n_tot] = (
+                        res.unindexed_experiment.identifier
+                    )
+                    n_tot += 1
+                    output_reflections.extend(output)
+            else:
+                # Either filtered out or empty.
+                assert i in filtered_refls
+                refls_this = filtered_refls[i]
```
graeme-winter: `refls_this`?

src/dials/algorithms/indexing/ssx/processing.py
```diff
@@ -426,11 +519,23 @@ def preprocess(
                 refl.centroid_px_to_mm(elist)
                 refl.map_centroids_to_reciprocal_space(elist)
             else:
+                # For filtered out or no refls, make a placeholder refl table
```
graeme-winter, line 538: ~filtered~ rejected I would suggest

graeme-winter (commented):
@jbeilstenedmands as discussed outside this review is really on the text of the changes and I do not want or expect you to implement them as the problem as posed in #3193 is fundamentally ill considered

I do however thank you for your efforts here - though I don't think it is worth merging these changes, the change set and comments in the PR made me realise the errors I had made in the approach to the problem. I am however sorry for wasting your time

jbeilstenedmands:
@graeme-winter  thanks for your input and comments.
As a side effect, this PR now enables one to run the image viewer on ssx data and see all images in the set, with unindexed and indexed spots correctly displayed across all images. For this feature alone, I will tidy up this work and look to add similar to dials.ssx_integrate, then we have the option to run the programs in this way for useful visualisation.

jbeilstenedmands:
The parameter name has now been updated to `retain_unindexed_experiments`, and the code structure and variable naming improved following review comments.

PR #3163, opened 2026-05-02
Fix filename for loading dy data

phyy-nx:
Spotted by Claude.  Totally untested.


dagewa:
The test `test_translate` goes through the creation of `dx.pickle` and `dy.pickle` maps, and re-import of the experiments with those lookups applied. The maps are different, with a 2 pixel correction in Y and a 1 pixel correction in X. These corrections are verified, so the dx and dy maps are correct. This test passes both on main and this branch, so while the fix on this branch does look _correct_, it seems not to affect actual processing, which I find puzzling :thinking: 
https://github.com/dials/dials/blob/5015d9cbacd30ce7d2352a25c4c2d769314fa2c0/tests/command_line/test_generate_distortion_maps.py#L66-L103

dagewa:
This is definitely a bug, but I've figured out why it doesn't affect processing. This script counts the number of times that `dx.pickle` and `dy.pickle` are opened by `dials.import`, and then again by `dials.find_spots` on Linux:

```bash
dials.generate_distortion_maps $(dials.data get -q centroid_test_data)/centroid_0001.cbf dx=1 dy=2
strace -e trace=open,openat -f\
    dials.import $(dials.data get -q centroid_test_data)/centroid_0001.cbf\
    dx=dx.pickle dy=dy.pickle > trace1.txt 2>&1 
echo "Number of times dx.pickle and dy.pickle are opened by dials.import"
grep "dx.pickle" trace1.txt | grep "openat" | wc -l
grep "dy.pickle" trace1.txt | grep "openat" | wc -l

echo "Number of times dx.pickle and dy.pickle are opened by dials.find_spots"
strace -e trace=open,openat -f\
    dials.find_spots imported.expt > trace2.txt 2>&1 
grep "dx.pickle" trace2.txt | grep "openat" | wc -l
grep "dy.pickle" trace2.txt | grep "openat" | wc -l
```

Run on `main`, the output ends with:
```
Number of times dx.pickle and dy.pickle are opened by dials.import
2
0
Number of times dx.pickle and dy.pickle are opened by dials.find_spots
3
3
```

So, `dials.import` does the wrong thing and opens `dx.pickle` twice. However, when it comes to actually using the experiments in downstream programs, the maps are opened again (3 times each!) when loading the experiments, and this rigmarole actually corrects for the error :sweat_smile:

dagewa (approved):
Good spot. Confusingly, thankfully harmless, but definitely a bug. Please add a newsfragment then this is good to go.

PR #3153, opened 2026-04-15
Refactor reflection prediction to avoid try/catch overhead

dwpaley:
As noted in dials#3146, this PR saves 105 s on the multicore benchmark:
```
Single-core (360 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              3.37s      2.72s     -19.3%
  find_spots         35.63s     35.19s      -1.2%
  index              34.16s     34.19s      +0.1%
  refine             14.30s     14.60s      +2.1%
  integrate         139.75s    141.27s      +1.1%
  total             227.21s    227.98s      +0.3%

Multi-core (1800 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              2.05s      2.05s      -0.1%
  find_spots         47.24s     50.99s      +7.9%
  index              75.96s     76.84s      +1.2%
  refine             48.53s     50.01s      +3.1%
  integrate         452.09s    347.90s     -23.0%
  total             625.87s    527.78s     -15.7%
  ```

Note that this is not strictly a pure refactor. The previous version was treating `get_ray_intersection` exceptions as "no intersection", but any other exception (e.g. in `panel.get_D_matrix()` was silently caught the same way. This version will raise those other exceptions. I would claim this way is more correct, but not sure if it really matters too much.

Issue #3146, opened 2026-04-13
DIALS performance tracking 2026


dwpaley:
Trying something new here. This is an issue to collect benchmarking results.

With much appreciation to @jmholton, https://github.com/dwpaley/dials_bench contains scripts that lightly modify James's DIALS benchmarks on synthetic lysozyme data (original benchmark scripts obtained from [here](https://bl831.als.lbl.gov/~jamesh/benchmarks/index.html)).

The benchmarks run in two modes, 360-degree single-core and 1800-degree multi-core (default 10). `run_bench.py` accepts the names of dials, cctbx_project and dxtbx branches as arguments. 

```
git clone git@github.com:dwpaley/dials_bench
python3 setup_bench.py # requires ~8GB storage
python3 run_bench.py --dials "dp2026_2/dials-inline-panel-ray-intersection-5arg"
[...]
=== Benchmark Results ===

Single-core (360 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              3.37s      2.72s     -19.3%
  find_spots         35.63s     35.19s      -1.2%
  index              34.16s     34.19s      +0.1%
  refine             14.30s     14.60s      +2.1%
  integrate         139.75s    141.27s      +1.1%
  total             227.21s    227.98s      +0.3%

Multi-core (1800 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              2.05s      2.05s      -0.1%
  find_spots         47.24s     50.99s      +7.9%
  index              75.96s     76.84s      +1.2%
  refine             48.53s     50.01s      +3.1%
  integrate         452.09s    347.90s     -23.0%
  total             625.87s    527.78s     -15.7%
```

...thus branch `dp2026_2/dials-inline-panel-ray-intersection-5arg` reduces integration runtime by 105 s on the multicore benchmark.

I have an incredible number of performance branches (a self-burial in slop PRs) and I will be using this issue to prioritize them.

dwpaley:
To build suspense, here are results for all improvements combined, as seen on the `profiling_20260220` branches of dials, dxtbx and cctbx_project. 

```
=== Benchmark Results ===

Single-core (360 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              2.76s      1.69s     -38.9%
  find_spots         34.96s     27.96s     -20.0%
  index              35.15s     11.58s     -67.1%
[...]

PR #3177, opened 2026-05-21
Adopt non-throwing intersection checks

dwpaley:
This PR adopts the changes in https://github.com/cctbx/dxtbx/pull/884/changes and supersedes #3153. Performance gains measured on #3146 are comparable to the previous approach. The multicore benchmark is a bit compromised by inconsistent # of cores selected for integration (7 instead of 5). The big picture is that this PR (like 3153) eliminates about 50% of the total dials.integrate execution time which had been spent on try/catch overhead in the multicore test. 

I hope this design will seem more justifiable than the previous inlining approach.

```
=== Benchmark Results ===

Single-core (360 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import             10.45s      3.93s     -62.4%
  find_spots         37.14s     36.63s      -1.3%
  index              37.42s     35.91s      -4.0%
  refine             14.94s     14.62s      -2.2%
  integrate         142.89s    108.90s     -23.8%
  total             242.85s    199.99s     -17.6%

Multi-core (1800 img):
  Step             Baseline    Feature      Delta
  -------------- ---------- ---------- ----------
  import              2.61s      2.48s      -4.8%
  find_spots         51.98s     50.78s      -2.3%
  index              76.01s     77.00s      +1.3%
  refine             49.43s     49.29s      -0.3%
  integrate         453.37s    183.37s     -59.6%
  total             633.41s    362.92s     -42.7%
```


dagewa:
This looks very promising! BTW, if it is useful, you can now enforce a particular number of processes in `dials.integrate` by using `integration.block.max_memory_usage=None nproc=5`. That avoids throttling and will use `nproc=5`, or die trying.

graeme-winter:
I can measure the performance improvement here with a "boring, normal" data set

Main:

```console
History (1 lines):
From DIALS 3.dev.1465-g69ae1ec17, run on 2026-05-21 at 08:32:31 CDT  (2026-05-21 at 13:32:31 UTC)
Writing html report to dials.merge.html

real	5m0.378s
user	16m1.262s
sys	1m36.827s
```

Branch:

```console
History (1 lines):
From DIALS 3.dev.1432-g980c55258, run on 2026-05-21 at 08:52:35 CDT  (2026-05-21 at 13:52:35 UTC)
Writing html report to dials.merge.html

real	3m44.949s
user	14m55.294s
sys	1m35.144s
```

Running:

```bash
dials.import ../Insulin_6_2.nxs
dials.symmetry integrated.expt integrated.refl
dials.scale symmetrized.expt symmetrized.refl absorption_level=medium
dials.export scaled.expt scaled.refl
dials.merge scaled.expt scaled.refl
```

Ths is a nontrivial improvement in performance. The results are subtly different, suggest you merge `main` on both repos into the branches

dwpaley:
@dagewa Thanks and I will start using that option in the benchmark suite.

@graeme-winter Nice! Both branches are up to date now if you would be so kind as to rerun or tell me where to find your insulin dataset.

graeme-winter:
Something does not smell right: we really should not be able to save 75s of wall clock time just from a few mis-predicted reflections which makes me think we are predicting way too many reflections somehow.

Needs deeper investigation.

graeme-winter:
Are we setting a sensible `d_min` for prediction?

graeme-winter:
Trivially tested by setting the same `d_min` for integration runs

graeme-winter:
@dwpaley will count how many cases we throw this exception

jbeilstenedmands (approved):
Thanks @dwpaley ! This feels like the right approach now.

In terms of expected performance improvements, I think it is more helpful to think of this as taking a fixed chunk of time off any integration job, rather than percentages.

For a dials.integrate job, the time is predominantly taken by: predict time + sigma estimation time + profile creation + integration.
The first two steps are done serially and so aren't affected by multiprocessing. The predict time is proportional to the number of reflections, so for average unit cells the benefits will be similar across datasets. The last two steps benefit from multiprocessing and can be thought of in terms of an approximately constant read time and a processing time that scales with the number of pixels in each shoebox (affected by `(sigma_b, sigma_m)`).

So we are taking a large fixed chunk off of the initial serial part of processing. On multi-core processing with small `(sigma_b, sigma_m)` this will look like a high % reduction in runtime, but less so for datasets with high `(sigma_b, sigma_m)` values, which are typically our longest running integration jobs. This is not intending to detract from the excellent improvement here, just to set expectations for what improvements we will see across different datasets.

dwpaley:
> @dwpaley will count how many cases we throw this exception

In the single-panel, 5-turn benchmark case, essentially all prediction misses are reflections between the edge and corner resolutions. With a strategic debug print:

```
dwpaley@Mac:~/daily/20260411/testing/dials_bench
$ dials.show refined.expt |grep 'Max resolution'
Max resolution (at corners): 1.540141
Max resolution (inscribed):  2.120664
$ dials.integrate refined.* > lim_inf.out
$ dials.integrate refined.* prediction.d_min=1.54 > lim_154.out
$ dials.integrate refined.* prediction.d_min=2.13 > lim_213.out
$ grep 'miss in append_for_ray' lim_inf.out |wc
 1038952 3116856 23895896
$ grep 'miss in append_for_ray' lim_154.out |wc
 1039492 3118476 23908316
$ grep 'miss in append_for_ray' lim_213.out |wc
       0       0       0
```

dwpaley:
@jbeilstenedmands Thanks and good point that this cuts integration time by 50% 💫on the benchmark💫. Would be good in the future to increase the number and variety of performance benchmark cases.

I would like to merge this and cctbx/dxtbx#884 by end of day Tuesday if there are no objections. @ndevenish ok with you?

PR #3178, opened 2026-05-21
Bug fix in testing

dwmoreau:
There was a bug in generating the test experiment for this test that affected dials/tests/algorithms/profile_model/ellipsoid/test_refiner.py. The test experiment is created with 0.01 degree rotation, however, it got no gonio. In the test, the experiment is then identified as an ImageSet instead of an ImageSequence, which effectively labels it as a still image. Running dials tests on my ImageSet -> ImageSequence work revealed this issue because the check for still frames is different on my local branch.

The solution was to create an ImageSequence with a default goniometer to get the test experiment.


dwmoreau:
This PR doesn't seem to do much. The test passes either way. It does raise a question, should there be a guard against creating data with a rotation, but no goniometer?

dwmoreau:
That seemed to cause more collateral damage than I expected.

PR #3179, opened 2026-05-22
Ellipsoid profile test bug

dwmoreau:
Tests should pass now


dwmoreau:
There was a bug in generating the test experiment for this test that affected dials/tests/algorithms/profile_model/ellipsoid/test_refiner.py. The test experiment is created with 0.01 degree rotation, however, it got no gonio. In the test, the experiment is then identified as an ImageSet instead of an ImageSequence, which effectively labels it as a still image. Running dials tests on my ImageSet -> ImageSequence work revealed this issue because the check for still frames is different on my local branch.

The solution was to create an ImageSequence with a default goniometer to get the test experiment.

This PR doesn't seem to do much. The test passes either way. It does raise a question, should there be a guard against creating experiments with a rotation, but no goniometer?

jbeilstenedmands:
I'm confused as to which part of `test_refiner` this was affecting? It looks like the test experiment was not created with a goniometer or imageset/sequence because they are not relevant for the code being tested (seems like everything still works with scan=None too). Your comment implies that there is some code branching here that depends on the type of imageset/sequence? Or are you saying you need to introduce some branching as part of the stills_process conversion work?
The changeset looks harmless, just trying to understand what the issue was.

dwmoreau:
There is not necessarily an issue with the current code, the execution doesn't seem to behave as expected. After thinking a bit about this, I don't believe this is worth pursuing. I'm closing the PR.

PR #3181, opened 2026-05-26
Stills process imagesequence

dwmoreau:
# `stills_process_imagesequence`: architectural narrative

Branches: `stills_process_imagesequence` in **dials**, **dxtbx**, and **cctbx_project**. A prototype, not intended for merge. The purpose of this document is to give senior DIALS / dxtbx developers enough of the design to reason about it without reading the code, and to anchor the group conversation that produces the production version.

Associated edits can be found on dxtbx and cctbx_project:
- https://github.com/cctbx/dxtbx/compare/main...stills_process_imagesequence
- https://github.com/cctbx/cctbx_project/compare/master...stills_process_imagesequence

## 1. Motivation

`dials.stills_process` has historically converted every imported image into a fresh
per-frame `ImageSet` inside `do_import()`. That conversion destroyed *object identity*:
imageset, beam, and detector were rebuilt as new Python objects for every frame.

Two consequences followed, both of which this branch removes:

- The JSON serializer deduplicates models by identity; with per-frame copies it could not
  deduplicate. Experiments from one source file each carried their own imageset, beam,
  and detector — N copies for N frames.
- The pipeline had to rebuild per-frame bookkeeping (and accumulate compensating caches)
  precisely because nothing was shared across frames. Once one `ImageSequence` shares a
  single imageset/beam/detector/static-mask per source file, most of that bookkeeping —
  and the per-frame caches that worked around its absence — simply becomes unnecessary
  (§6).

The branch keeps the data as an `ImageSequence` for the entire pipeline. Everything
else in this document is either a direct consequence of removing the conversion, or a
consequence of the new contracts that removal forces on the rest of DIALS.

## 2. The pivot: `do_import()` keeps data as `ImageSequence`

`do_import()` calls `format_class.get_imageset(as_sequence=True)` instead of
`ImageSetFactory.imageset_from_anyset()`. Per-frame slices are `imageset[i:i+1]` and
are then rebuilt with a fresh `Scan((1, 1))` to normalize the absolute frame range to
`(0, 1)`. The slice is itself an `ImageSequence`, sharing the parent imageset's beam
and detector.

This is the root architectural change of the branch: object identity (the same Python
object visible across frames) is what JSON dedup needs, and what lets the pipeline shed
the per-frame caches it previously needed to compensate for the lack of sharing (§6).
No other piece of the work functions without it.

Every slice is rebuilt regardless of whether the original scan was a still or a
rotation. Rotation CBF frames processed via `convert_sequences_to_stills` also carry
absolute scan ranges (e.g., frame 2 → `array_range=(1, 2)`) that confuse the
integrator's 0-based z assumption just as much as XFEL absolute ranges. The
normalization is upstream of XFEL-specific concerns.

## 3. New dxtbx primitives for XFEL data

A generalized beam hierarchy plus three small XFEL primitives in dxtbx; each carries an
explicit decision. The hierarchy change (§3a) is the structural one — the rest hangs off
it: per-shot wavelengths on the scan (§3b), a capability-based per-frame beam accessor
(§3c), and the format plumbing that produces both (§3d).

### 3a. `XFELBeam` — lightweight guard / marker on a generalized beam hierarchy

C++ class in `model/beam.h` / `model/boost_python/beam.cc`. It is **not** a `Beam`
subclass. The beam hierarchy was generalized for this work: `BeamBase` is now an
abstract base holding the universal geometry data and concrete accessors plus a
**pure-virtual spectral interface** (`get/set_wavelength`, `get/set_s0`, the scan-point
family, `operator==`, `is_similar_to`), and the three spectral models are **direct
siblings** under it:

```
BeamBase  (abstract: universal geometry + pure-virtual spectral interface)
   ├─ Beam              monochromatic                __id__="monochromatic"
   ├─ PolychromaticBeam (+ wavelength_range)         __id__="polychromatic"
   └─ XFELBeam          (no wavelength; per-shot λ in scan)   __id__="xfel"
```

This replaces the earlier `PolychromaticBeam`/`XFELBeam : public Beam` chain. The
consequence that matters downstream: `isinstance(poly_or_xfel, Beam)` is now **False**
(it was True under the old chain), so generic "accept any beam" code binds
`const BeamBase&` (e.g. masking, `Experiment`), and `Format.setup()` beam asserts check
`BeamBase`, not `Beam`. The generalization is deliberate — it is the seam the open
"can the beam model subsume electron / Laue?" question (§8) would build on, rather than
a bespoke XFEL hack bolted onto `Beam`.

`XFELBeam` itself is stateless with respect to wavelength: it carries direction,
divergence, polarization, flux, transmission, probe, sample-to-source distance.
`get_wavelength()`, `set_wavelength()`, `get_s0()`, `set_s0()` and all scan-point
methods **throw**. Serializes with `"__id__": "xfel"`.

- **Why not `PolychromaticBeam`.** `PolychromaticBeam` models a source with an
  intrinsic spectral bandwidth (Laue / ToF) and stores a `wavelength_range`. XFEL is a
  sequence of monochromatic pulses with varying energies. The per-shot energies are
  *measurements*, not a property of the source. Conflating the two is incorrect physics.
- **Why a throwing guard rather than a default wavelength.** Code that calls
  `beam.get_s0()` on an XFEL beam has a real semantic gap — it has not chosen a frame.
  A silent fallback (mean, first, nominal) would mask bugs at exactly the boundaries
  where they need to surface.
- **Why `get_num_scan_points()` returns 0 instead of throwing.** Generic `Experiment`
  copy / compare / serialize code probes scan-varying state as a benign check.
  Returning 0 matches the real semantics — an `XFELBeam` genuinely has no scan points —
  and avoids forcing XFEL-awareness into generic code.

### 3b. Per-frame wavelengths in the scan properties table

*Design due to David Waterman and David McDonagh; implementation by McDonagh.*

Wavelengths are stored as `scan.set_property("wavelength", flex.double([...]))`. The
property is written by `FormatXFEL.get_imageset()` for XFEL data, and read by
`ImageSequence.get_beam(i)` (§3c), by `do_import` (which stamps a single-entry array
onto each per-frame scan so the spotfinder can detect XFEL data on the in-flight path),
and by `_rebuild_shared_imageset_output` (which stamps the full per-frame array onto
the output scan so the property survives JSON round-trip).

The design principle behind this is that **the beam should not need to carry
per-image information**. The beam represents the source; per-frame wavelength is a
per-image measurement and belongs on the scan alongside other per-image data such as
timestamps and exposure times. The same model handles the constant-wavelength case
without modification — the `"wavelength"` property is simply absent — and the
property gets consolidated serialization for free (§5c).

### 3c. `ImageSequence.get_beam(index)` override

Per-frame beam access is handled by a Python method override injected onto the
existing `ImageSequence` class (in `imageset.py`). The dispatch is **capability-based,
not type-based**: when the imageset's beam exposes `get_monochromatic_beam`
(`hasattr(beam, "get_monochromatic_beam")`) *and* the scan has a `"wavelength"`
property, `get_beam(i)` returns `beam.get_monochromatic_beam(scan.get_property("wavelength")[i])`
— a regular monochromatic `Beam` built from the source beam's direction / divergence /
polarization tuple and the per-frame wavelength. All other paths (no index, a beam
without the capability, non-XFEL imagesets) fall through unchanged. Duck-typing on the
capability rather than `isinstance(beam, XFELBeam)` keeps the per-shot-resolve seam
independent of the concrete beam class — it survives the beam-hierarchy generalization
(§3a) and would extend to any future spectral beam that grows a `get_monochromatic_beam`.

- **Why a method override rather than an `XFELImageSequence` subclass.** An earlier
  iteration of this work introduced an `XFELImageSequence` subclass. The dxtbx
  maintainers (Waterman, McDonagh) pushed back: a new ImageSet class proliferates type
  checks across dxtbx and DIALS. The override is invisible to non-XFEL callers, and
  the XFEL-specific dispatch is encapsulated inside dxtbx — it never leaks into DIALS.
- **No public per-frame beam API beyond `get_beam(i)`.** `imageset.get_beam(i)` is the
  existing contract. Code that already called it works without modification.

### 3d. `FormatXFEL` mixin and the two concrete formats

The mixin's `get_imageset()` lets the parent format build the raw `ImageSet`, reads
per-frame wavelengths via a subclass-provided `get_wavelengths()` returning
`list[float]` (Ångström), constructs the `XFELBeam` from the first frame's geometry
(passing the source beam's full polarization tuple through, see §7), and returns a
regular `ImageSequence` with that `XFELBeam` and a zero-oscillation `Scan` carrying
the `"wavelength"` property.

`FormatNXmxXFEL` delegates wavelength reading to
`CachedWavelengthBeamFactory.get_wavelengths()` (the 1-D `incident_wavelength` HDF5
dataset). Its `understand()` requires `wl.ndim > 0 and wl.size > 1`: scalar wavelengths
and 1-element arrays represent a single fixed wavelength for all frames; those files
belong to `FormatNXmx`, not the XFEL specialization. `FormatXTCXFEL` delegates
`understand()` to `FormatXTC` because XTC streams are always XFEL stills with
per-event wavelengths.

## 4. Cross-cutting contract: stills detection by scan, not imageset type

The old detection idiom — `isinstance(imageset, ImageSequence)` is False ⇒ still — no
longer separates stills from rotations, because both are now `ImageSequence`. The new
contract is:

> A still is `expt.scan is None or expt.scan.is_still()` (zero oscillation width).

Every check has been routed through `Experiment.is_still()`, the C++ accessor that
already encodes the full scan-is-still / scan-is-None / goniometer-is-None contract.
The sites touched include the indexer, the integrator's memory estimator, the
reflection predictor's static-vs-scan-varying dispatch, the refinement crystal
parameterisation, the spotfinder factory, and `combine_experiments`.

This is the widest blast radius in the branch. Centralizing through `is_still()` is
load-bearing: the 2026-05-24 review pass found that `spot_finding/factory.py`'s
`is_stills=False` fallback was still using the old isinstance check, silently
inverting `no_shoeboxes_2d` for stills entering via non-`stills_process` callers. The
question of whether scan-based detection is the right pipeline-wide contract is a
decision the group should ratify; this branch's position is that there should be a
single check, and it should be `Experiment.is_still()`.

A separate, narrower question is *XFEL* detection in the spotfinder. That uses
`scan.has_property("wavelength")` rather than `isinstance(beam, XFELBeam)`, because
the in-flight path carries monochromatic beams built by `do_import` — an
`XFELBeam` check would always be False there.

## 5. Output format: shared models, sparse imagesets, consolidated scans

Three layered transformations on the way out, each addressing a distinct cause of
file bloat.

### 5a. Shared models per source file

`_rebuild_shared_imageset_output()` groups experiments by source file and rebuilds them
so every experiment from one file points at one `ImageSequence` / beam / detector
object. Multi-lattice experiments from the same frame share one `Scan`. A no-op for
rotation data. This is the change that actually shrinks the file.

### 5b. Sparse imageset layout

The shared `ImageSequence` is built with indices `sorted(set(frame_indices))` — only
the frames that were successfully integrated — and a compact `Scan((1, N))`. The
alternative (a contiguous `range(min_f, max_f + 1)`) would show every frame in the
source-file range in `image_viewer`, including frames that never produced a result.

Allowing non-contiguous indices required three downstream changes in dxtbx: the C++
`ImageSequence` constructor's sequential-indices assertion is now conditioned on
`!scan->is_still()`; the Python-level assertion in `FormatMultiImage.get_imageset()`
is similarly conditioned; and `ExperimentListDict._make_sequence()` bypasses
`ImageSetFactory.make_sequence()` when `single_file_indices` is present in the JSON,
because `make_sequence()` recomputes a contiguous index range from the scan's
`image_range` and asserts equality with `array_range`, both of which fail for a sparse
imageset with compact scan.

### 5c. Consolidated stills scan, re-consolidated `XFELBeam`

When every scan in an `ExperimentList` is a single-frame still (oscillation width 0)
**and** every still-scan imageset is a single-file reader, `ExperimentList.to_dict()`
collapses the N scan dicts into one object:

```json
{"__stills_consolidated": true, "batch_offset": 0,
 "properties": {"wavelength": [...]}, "valid_image_ranges": {}}
```

Per-frame identity is **not** stored on the consolidated scan. It lives in the 0-based
`single_file_indices` array already carried by each imageset; each experiment's per-frame
`image_range` is reconstructed on load as `(fi+1, fi+1)`. (An earlier design carried a
redundant 1-based `frame_numbers` array on the scan; it was removed once
`single_file_indices` was made the single, zero-indexed source of truth.) All-zero
property arrays (epochs, exposure time, oscillation) are omitted and reconstructed as
zeros on load; angle-valued arrays are written in degrees to match the `Scan` dict
convention.

Each experiment gains `"scan_point": i` and `"scan": 0`. **`scan_point` is assigned per
(imageset, unique frame), not per `Scan` object** — multi-lattice experiments indexed on
one frame share a `scan_point`, which is exactly what dedups the wavelength array to one
entry per physical frame (the `split→combine` path produces separate-but-equal `Scan`
objects, so an identity-based assignment would over-count). On the read side,
`ExperimentListDict._expand_consolidated_scans()` runs in `__init__` before any model
extraction (grouping experiments by imageset, deduping `scan_point`, and zipping against
the imageset's sorted `single_file_indices`), so the rest of `decode()` sees the standard
format. Measured reduction on a 764-experiment file: about 8 % of total size.

The single-file-reader requirement is an **eligibility gate** (`all_single_file`), not an
assertion: when any still-scan imageset is *not* a single-file reader — multi-file CBF
stills, where each shot is its own `.cbf` and there is no `single_file_indices` to
reconstruct from — consolidation is skipped and the normal per-scan write emits each
`image_range`. No data is lost; the gate simply declines to compact a structure it cannot
losslessly reconstruct.

The same `to_dict()` block re-consolidates per-frame monochromatic beams back into a
single `XFELBeam` when every experiment beam is a plain `Beam` sharing geometry
(direction, divergence, polarization). This preserves the canonical single-beam
representation through `refine` / `split` / `combine` round-trips. If beams have
diverged (e.g., genuinely refined to different geometries), all N beams are written —
no data is lost.

Consolidation runs automatically. An earlier opt-in flag (`compact_stills_scans=True`)
was removed: requiring 200+ downstream call sites to opt in is fragile, and the guard
(all scans single-frame, oscillation 0) makes the transformation a true no-op for
rotation data. Backward compatibility: a new reader handles old files transparently;
old readers fail on new files.

## 6. Performance: native sharing over per-frame caches

Nick Devenish proposed exploring the `ImageSequence` conversion specifically so that the
per-frame caches an earlier profiling branch (`stills_process_profiling`) carried *could
be simplified*. The outcome went further than "simplified": once one `ImageSequence`
shares a single imageset/beam/detector/static-mask per source file, **most of those
caches became unnecessary and were removed** — keeping them would have been bookkeeping
to preserve identity that the data model now preserves natively. The savings now come
from three places: format-level read fast-paths that exploit shared identity, one kept
resolution-mask cache, and one kept indexer cache.

**Format-level read fast-paths.** Two read optimizations the `ImageSequence` model
enables on the XFEL path:
- `FormatXFEL.get_imageset()` constructs its `ImageSequence` directly instead of routing
  through the parent's generic per-frame model loop. That loop built a fresh beam (and
  detector / goniometer / scan) for *every* frame, only for the XFEL mixin to discard
  them in favor of one shared `XFELBeam` and shared detector — pure waste on a many-frame
  composite. The local construction also re-attaches the static pixel mask
  (`_add_static_mask_to_iset`); dropping it would let the spotfinder pick up hot pixels.
- `FormatNXmx._start` caches the NXmx tree wrapper, the `nxdata`/`nxdetector` handles, and
  the per-module slice tuples once at open time, since they are pure functions of
  file-level metadata that does not change after open. `get_raw_data` previously rebuilt
  all of them per frame; it now takes an optional `module_slices=` kwarg (recompute as
  fallback for non-cached callers).

**Spotfinder — caches removed.** The per-frame spotfinder caches an earlier branch carried
(`ExtractPixelsFromImage.update_imageset`, and the `ExtractSpots` object cache) were
**removed**. `ExtractPixelsFromImage` is now rebuilt per frame: its `__init__` only stores
references and the mask is recombined per frame inside `__call__` regardless, so
reconstruction is ~0 cost, and the static (lookup mask + untrusted-rectangle) component now
rides on `imageset.get_mask(index)`, shared natively across the `ImageSequence` via
`_add_static_mask_to_iset`. The one spotfinder cache **kept** is
`imageset_mask = mask_generator(imageset)` — the resolution mask, the only genuinely
expensive per-frame recompute. It benefits non-XFEL/SSX stills with resolution masking and
is **suppressed on XFEL** (`scan.has_property("wavelength")` forces a per-frame rebuild,
because the per-shot wavelength changes the resolution mask). On the branch's primary (XFEL)
path the spotfinder therefore runs cache-free.

**Indexer — one cache kept.** `Processor` caches the `Indexer` instance;
`update_indexer()` resets per-frame state instead of reconstructing (read-confirmed
orthogonal to `ImageSequence` sharing — what it skips is param-driven, not imageset-driven,
so native sharing does not make it redundant). `copy.deepcopy(all_params)` per image was
replaced with direct mutate-and-restore of `refinement.parameterisation.scan_varying`
(forced False for stills). `max_refine` is no longer touched at the call site:
`Indexer.from_parameters` already resolves the `libtbx.Auto` sentinel to 5 for stills, and
the cached `Indexer` inherits the resolved value — matching `main`, where each frame's
`copy.deepcopy` reached the same end-state through the same resolver. An explicit
user-supplied `max_refine` is preserved.

A correctness item the cached `Indexer` introduced — `Indexer.refined_experiments` being
nulled only in `__init__`, allowing a frame whose indexing failed before the success block
of `index()` to silently inherit the previous frame's refined result — has been fixed:
`index()` now resets `self.refined_experiments = None` at entry, so the
`if self.refined_experiments is None: raise DialsIndexRefineError(...)` guard fires
correctly on per-frame failures.

## 7. Downstream contract — what consumers must do

The XFEL portion of the contract — `Experiment.beam` may be an `XFELBeam` that throws
on `get_s0()`, with per-frame wavelengths living on the scan — has a non-trivial blast
radius. The branch addresses it with one central mitigation and a small set of
targeted patches in DIALS tools that bypass the central path.

**Central mitigation.** `ExperimentList.decode()` auto-resolves each `Experiment.beam`
from `XFELBeam` to a per-frame monochromatic `Beam` after `_expand_consolidated_scans`
has restored the per-frame scans. Any consumer that loads a composite `.expt` via
`ExperimentListFactory.from_*` sees a regular `Beam` on `Experiment.beam` and requires
no modification. Crucially, the imageset's `XFELBeam` and the scan's `"wavelength"`
property are left untouched, so `imageset.get_beam(i)` continues to dispatch through
the per-frame override, and the write-side re-consolidation in `to_dict()` is
unaffected.

**Targeted patches** were required where the auto-resolve does not apply — either
because the tool copies experiments instead of reloading them, or because the original
behavior was wrong in a way the auto-resolve does not catch:

- `dials.refine` builds an experiment *copy* for refinement
  (`_copy_experiments_for_refining`). The copy is now resolved via
  `Experiment.get_monochromatic_beam()` at that single chokepoint, so every
  refinement-internal consumer (`StillsReflectionManager`, `StillsReflectionPredictor`,
  `BeamParameterisation`, the prediction parameterisations) sees an ordinary
  monochromatic beam with no code change. Refined output flows back through the same
  helper and is re-consolidated to one `XFELBeam` on write. A separate fix in
  `_parameterise_crystals` stops classifying a zero-oscillation still scan as a
  rotation experiment.
- `dials.combine_experiments` needed four fixes spanning both repos: a
  `FormatMultiImage` guard against `format_instance=None` with `check_format=False`;
  a `FormatXFEL` wavelength-subset fix when loading composite output; an
  `ExperimentListDict.decode` change to not overwrite the format-provided scan with
  the JSON-decoded one; and a `CombineWithReference` change that keeps per-frame
  scans on stills and relaxes `wavelength_tolerance` to infinity for stills↔stills
  comparisons (after the auto-resolve, every loaded `Beam.wavelength` is its own
  per-frame value). A later split→combine audit added two more: `CombineWithReference`
  now propagates `profile` / `scaling_model` / `history` (previously dropped — history
  via `OrderedSet` line-level dedup plus a shared write-timestamp, so combining N
  same-provenance experiments no longer multiplies history lines); and when filtering
  drops experiments (`min/max_reflections_per_experiment`, `n_subset`, and the
  structurally-identical `split_still_data` path), the shared `ImageSequence` is now
  **pruned** to the surviving frames — otherwise its `single_file_indices` still
  advertise all N source frames while only the survivors contribute `scan_point`s, and
  the read-side alignment check fails. The prune helper (`consolidate_stills_imagesets`)
  is shared across the combine and split write sites.
- `dials.image_viewer` overlays on composite stills required an order-independent
  `_identifiers_for_frame` (the previous binary search assumed sorted frame order,
  which neither `_rebuild_shared_imageset_output` nor the multiprocessing-combine
  path guarantee); a `_still_scan_frame_list` helper that recovers absolute frame
  indices from per-experiment scans (the shared imageset's scan loses its start
  offset on JSON round-trip); per-overlay color reassignment for composite-id
  reflection tables; and a per-frame beam via `image_set.get_beam(self.index)`.
- `dials.split_experiments` → `dials.combine_experiments` round-trip required
  decoupling `to_dict()` beam re-consolidation from the existing `len(beams) > 1`
  guard (so single-experiment and single-shared-beam stills writes also re-consolidate
  to `XFELBeam`), and passing the full polarization tuple through
  `BeamFactory.make_xfel_beam` at re-consolidation (a silent drop of the LCLS ~0.999
  polarization fraction was a P0 regression caught in the review pass).

**The pattern downstream tools must follow** is: if a tool builds experiments at
runtime *without* going through `decode()`, it must run
`expt.beam = expt.get_monochromatic_beam()` after assembly.
`Experiment.get_monochromatic_beam()` is a no-op for non-XFEL beams, so the call is
safe to make unconditionally. For tools that load experiments via `decode()`, no
change is required.

**Downstream tool status.** Exercised end-to-end against output produced by this branch:

- `cctbx.xfel.filter_experiments_by_rmsd` — verified, correct results.
- `cctbx.xfel.detector_residuals` — verified, correct results (full per-panel-group
  residual statistics).
- `cctbx.xfel.merge` — **partially verified.** 15 of 22 merge variants reproduce the
  reference within the test's tolerances on converted input. The remaining failures are
  dominated by **one known bug class** (plus one benign relabel, below):
  `cctbx.xfel.merge`'s `filter` / `filter_global` /
  `modify_reindex_to_reference` workers are another *unpruned shared-imageset write
  site* — a worker writes out a *subset* of experiments sharing one consolidated stills
  `ImageSequence`, but the output keeps the full `single_file_indices` while only the
  survivors contribute `scan_point`s, so reloading the merge output trips the read-side
  alignment check (the same defect already patched per-site in `combine_experiments` and
  `split_still_data`). This is **reported, not yet fixed** — the systemic cure is the
  `scan_point` re-scope in §8; a per-site fix would call the shared
  `consolidate_stills_imagesets` helper before the merge `.expt` write. (One further
  variant differs only in a benign CC½ odd/even relabel — identical merged intensities.)

The auto-resolve in `decode()` means no per-tool *beam* patch was required for any of
these; the merge gap above is a scan/imageset-consolidation issue, orthogonal to the
beam contract.

## 8. Open architectural questions

- **Generalize the beam model — and whether `XFELBeam` survives it.** The `BeamBase`
  refactor (§3a) is a partial step toward a general spectral-beam abstraction: the
  spectral interface is now pure-virtual on `BeamBase`, with monochromatic / polychromatic
  / XFEL as siblings. The open question is whether to push that to a single generalized
  model (designed with anticipated electron / Laue beams in mind) that *subsumes* the XFEL
  case and retires `XFELBeam` as a bespoke guard class — versus keeping `XFELBeam` as is.
  The maintainers' explicit steer: do **not** simply expand `PolychromaticBeam` (XFEL
  pulses are individually monochromatic with per-shot energies on the scan, not a
  wavelength array). This is a group design decision, not a code gap.
- **Systemic `scan_point` representation (the prune patches' root cause).** Per-frame
  identity currently survives as an *ordinal* `scan_point` that must be sorted-and-zipped
  against the imageset's `single_file_indices` on read, guarded by a brittle
  `len(single_file_indices) == unique scan_points` assertion. Every write site that emits
  a *subset* of a shared `ImageSequence` (combine-after-filter, `split_still_data`, and
  the still-unfixed `cctbx.xfel.merge` workers, §7) must prune the imageset to keep that
  invariant — fixed per-site today. The systemic cure is to store `scan_point` *as* the
  direct 0-based frame index (`single_file_index`), deleting the alignment dance and the
  assert. Not done in this branch; it is a coordinated dxtbx+dials format change.
- **Beam refinement for XFEL stills is not designed.** Currently `refine_level0.phil`
  fixes the beam (`beam.fix=*all`) and the write-side re-consolidation produces one
  `XFELBeam` as expected. A phil that refines beam parameters would produce N
  differing beams that the re-consolidation guard would refuse to collapse — the
  result would be a much larger file with no clear consumer story. Beam refinement
  for XFEL stills needs separate design before the group should consider this
  contract for production.
- **`combine_all_ranks` output semantics.** The new phil option merges per-rank (MPI)
  or per-worker (multiprocessing) composite files into a single combined file with
  shared models. The current semantics — including how merging interacts with
  per-source-file model identity — have not been settled with the group.
- **B-covariance on the stills write path (investigated, not acted on).** A proposal to
  zero-and-omit the per-crystal `B_covariance` table was **refuted** as a blanket change:
  it feeds cell-parameter ESDs that `dials.show` and the SHELX export report for
  rotation / scan-varying data. A narrower *stills-scoped* omission (no stills consumer
  reads it; it would shrink composite files) is viable and back-compatible but was left
  unimplemented by maintainer decision.

**Flagged as unverified (not yet closed):**

- **Per-frame wavelength squash — which layer.** A fix corrected the *in-memory* per-frame
  beam on the eager model-load path, but whether any residual squash exists in the
  *persisted* output requires running `dials.import` on data with genuine per-frame
  wavelength variation and diffing the output consolidated-scan `"wavelength"` array. Until
  that diff is captured this is not declared resolved. (`stills_process` itself stamps each
  per-frame scan explicitly and is unaffected either way.)
- **`cspad_cbf_metrology` λ-refinement drift.** A 300-experiment metrology regression shows
  the freshly-refined per-experiment wavelengths differ from the stored reference by
  ~4e-4 Å (300 still-distinct values — *not* a collapse). It fails at baseline too, so the
  branch-vs-`main` attribution is unresolved; flagged for maintainer attention as a
  possible real refinement regression, **not** dismissed as environmental.

## 9. Pre-publication checklist

Before these branches are pushed:

1. Run the pinned `ruff format` and `ruff check` (the repo's canonical formatter is
   ruff-format, pinned at `v0.15.9` in `.pre-commit-config.yaml` — **not** `black`;
   black disagrees with ruff-format on string hugging and would introduce non-canonical
   churn) across all three repos, so the published diff is canonical style only, with no
   formatter noise to distract reviewers from the architectural diff.

## 10. Testing: fixture alignment, not production weakening

The DIALS test suite contains fixtures that pre-date this branch and construct
`Experiment` objects without an imageset — supplying only `{beam, detector,
crystal}`. The contract change in §4 (stills detection via scan) and the
auto-routed sort path added to `dials.combine_experiments` in §5 make
`expt.imageset.paths()[0]` an unconditional call on the still-detection path,
so these fixtures surface as `AttributeError: 'NoneType' object has no
attribute 'paths'` at first contact with the new code.

The policy on this branch is **update the fixture, not the production code.**
Weakening upstream call sites with `if iset is None: ...` guards would let
genuinely malformed experiment data propagate silently through the pipeline;
the loud crash at the boundary is the correct behavior. The proper fix is to
bring the fixture into alignment with the new contract that every experiment
carries an imageset.

**Fixtures updated:**

- `tests/algorithms/profile_model/ellipsoid/conftest.py` — the `test_experiment`
  fixture, previously `Experiment(beam=..., crystal=..., scan=..., detector=...)`
  with no imageset, is now rebuilt to construct a real `ImageSequence` with a
  `Format.Reader` and to build `Experiment` from the imageset's own models.
  Commit `ae871b468`.
- `tests/command_line/test_combine_experiments.py::test_min_max_reflections_per_experiment`
  — the test loads `multi_stills_combined.json` from `dials_data`, an
  old-format file with no imagesets. Since the `dials_data` fixture itself is
  not editable, a new test-setup helper rewrites the `.expt` into `tmp_path`,
  attaching **one** shared zero-oscillation `ImageSequence` (dummy reader)
  covering all N frames and a single-frame per-experiment scan per experiment.
  The shared-object reuse satisfies the `id()` identity check in
  `_consolidate_stills_imagesets`, matching the in-flight state §5a produces,
  so the dummy file is never opened. Commit `963d588ea`.

In both cases the rewrites are test-side preprocessing: the test bodies,
expected results, parametrize matrices, and the `dials_data` files themselves
are unchanged. The same pattern applies to any future `None.x` failure on a
model accessor in a still path — attach a shared `ImageSequence` with the
right beam/detector/scan structure and feed the rewritten file to the code
path under test.

**New positive coverage added by the branch** (not just fixture alignment):
`dxtbx/tests/model/test_beam.py` exercises the generalized hierarchy — sibling
relationships under `BeamBase`, the `XFELBeam` guard throws, and the capability-based
per-shot resolve; `test_experiment_list.py` adds consolidated-scan round-trip tests
(including the oscillation degrees-vs-radians regression and the history dedup /
shared-timestamp collapse); and `test_stills_imageset_convert.py::test_converter_synthesizes_frames_without_sfi`
covers the converter's synthesized-frame path for legacy inputs that lack
`single_file_indices`.

### `test_sort_by_imageset_path_and_image_index` — resolved: ascending order is the stills contract

`test_combine_experiments.py::test_sort_by_imageset_path_and_image_index` (PR #3036,
pre-branch) recombines four lyso JF16M stills in shuffled order `(0, 2, 3, 1)` and
originally asserted that *without* `sort_by_imageset_path_and_image_index` input order is
preserved, and *with* it the output sorts to `(0, 1, 2, 3)`.

This branch makes the `(source path, frame index)` sort **unconditional** for stills in
`dials.combine_experiments` and `stills_process` composite output (motivation:
`dials.image_viewer` overlay correctness — the combine paths leave experiments in
worker-encounter order, and several overlay paths assumed ascending frame order; sorting
at write is the chokepoint fix). That makes the first assertion also see `(0, 1, 2, 3)`.

This was an open policy question in earlier drafts; **it has been resolved** by adopting
that contract: ascending `(path, frame_index)` is now a hard guarantee for stills, and
`sort_by_imageset_path_and_image_index` is effectively a no-op for stills (retained for
rotation data). The test's default-case assertion was updated to `(0, 1, 2, 3)`
accordingly (`test_combine_experiments.py:550,569`), and the test passes. The branch's
position — a single, unconditional stills ordering — is the one the group is asked to
ratify, not re-litigate.


dwmoreau:
@phyy-nx suggested that creating an XFEL beam, in addition to the new polychromatic beam, sets precedent for new models to be developed in an-unchecked manner. A better approach would be to see how the XFEL beam could be unified with the polychromatic beam. That way future efforts with polychromatic XFEL beams, electron beams, etc... don't also require new models to be made.

phyy-nx:
We worked up all the use cases we could think of in [NXmx](https://manual.nexusformat.org/classes/applications/NXmx.html#nxmx-entry-instrument-beam-incident-wavelength-field):

- Static beam
- Polychromatic beam
- Per-shot beam
- Polychromatic per-shot beam
- Polychromatic beam that varies shot-to-shot and where the channels also vary non-linearly

Now one thing I don't understand is how the polychromatic beam is different than the spectrum object, but I think it needs to be incorporated into this discussion of rationalizing all these ideas as well, including perhaps deprecating the spectrum object.

PR #3185, opened 2026-05-29
Add a link to our Slack workspace to dials.github.io

phyy-nx:


phyy-nx:
Closing, enough reservations with public links that makes sense to table this.

PR #3188, opened 2026-06-05
use twin laws for lattice angle calculation

### LINKED ISSUE #3110: dials.index with max_lattices >= 2: rejection criteria
If I have a split crystal and index, the second lattice will be rejected if closer than some small angle from the first - this does not however take into consider rotational symmetry as 

```console
Finish searching for more lattices: 12650 unindexed reflections remaining.
Change of basis op: a,b,c
Rotation matrix to transform crystal 1 to crystal 2:
{{-0.305, 0.475, 0.825},
 {0.472, -0.677, 0.564},
 {0.827, 0.562, -0.018}}
Rotation of -179.892 degrees about axis (0.589, 0.402, 0.701)
```

This is the same as about 0.11° and thus should be excluded but it was not

Remedy: the check should consider rotational symmetry in the angular misalignment

kmdalton:
Attempt to resolve #3110 . Use the twin laws from the lattice and space group rather than the operators from the laue group to calculate the minimum angle between lattices. This accounts for indexing ambiguity, psuedo-merohedral twinning, and cases where the spacegroup is unknown during indexing and set to P1. 

@graeme-winter -- can you check to see whether this helps on your test data set from #3110?

### DIFF
--- a/src/dials/algorithms/indexing/basis_vector_search/combinations.py
+++ b/src/dials/algorithms/indexing/basis_vector_search/combinations.py
@@ -198,7 +198,8 @@ def filter_similar_orientations(
         cell = gemmi.UnitCell(*cryst.get_unit_cell().parameters())
         sg = gemmi.SpaceGroup(cryst.get_space_group().type().lookup_symbol())
         gemmi_ops = gemmi.find_twin_laws(cell, sg, max_obliq=max_obliq, all_ops=True)
-        sgtbx_ops = [rt_mx(op.triplet()) for op in gemmi_ops]
+        sgtbx_ops = [rt_mx("x,y,z")]
+        sgtbx_ops.extend([rt_mx(op.triplet()) for op in gemmi_ops])
 
         for cryst_a in other_crystal_models:
             R_ab, axis, angle, cb_op_ab = difference_rotation_matrix_axis_angle(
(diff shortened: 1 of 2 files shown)


dagewa:
I'm wondering if `max_obliq` should be linked somehow to `minimum_angular_separation`? My understanding is that `max_obliq` sets the tolerance for finding pseudo-symmetric cases, but we are kind of defining that tolerance already by the `minimum_angular_separation` parameter. Maybe `max_obliq` should just be high to avoid missing any possibilities, and then filter out based on just the `minimum_angular_separation`? Just a thought to avoid bringing another parameter into the mix.

kmdalton:
> I'm wondering if `max_obliq` should be linked somehow to `minimum_angular_separation`? My understanding is that `max_obliq` sets the tolerance for finding pseudo-symmetric cases, but we are kind of defining that tolerance already by the `minimum_angular_separation` parameter. Maybe `max_obliq` should just be high to avoid missing any possibilities, and then filter out based on just the `minimum_angular_separation`? Just a thought to avoid bringing another parameter into the mix.

there is probably a way to set `max_obliq` more intelligently. it is a number in degrees that sets the "fuzziness" of the pseudo twin laws. i believe 3.0 is a fairly permissive setting, but it's probably worth looking more deeply into this. i agree it'd be a confusing option to expose to users.

PR #3193, opened 2026-06-09
Reduce precision of values for plotly json to reduce HTML file sizes

### LINKED ISSUE #3190: HTML reports are very large due to over-precise values in plotly plots
Something I noticed with the A/B runs. Sometimes the scaling HTML output from xia2 is very large. The attached file is 26M uncompressed. 

[AUTOMATIC_DEFAULT_SCALE.html.zip](https://github.com/user-attachments/files/28682876/AUTOMATIC_DEFAULT_SCALE.html.zip)

It looks like most of this is in two variables with JSON contents, `graphs_anom_scatter_plot_lowres` and `graphs_outlier_plot_0`. I think this is mostly due to the values being stored with extremely high precision. Here's the first part of one of those definitions

```
        var graphs_anom_scatter_plot_lowres = {"data": [{"mode": "markers", "name": "half-set anomalous differences (acentrics)", "size": 1, "type": "scatter", "x": [-7.985473584597473, 6.176636087866069, 15.823998502610175, -6.707774848471679, -2.6504068696450283, -16.11571640559372, 22.623129846530304, -7.265811214521779, 12.73683263069193, -8.494078427028024, -2.688411033845176, -13.873663954107318, -15.919465472456974, -5.535893754193629, -11.820880017452168, -13.085083662563353, 2.840956525271876, -9.906590387153953, 13.20680990765717,
```

dagewa:
This began as a limited fix to reduce the data size of a couple of scaling plots for https://github.com/dials/dials/issues/3190, but I soon realised that lots of other plots contain a large number of overly-precise text values. So I'm trying a more extensive change here, where most of the float data in scatter plots is reduced to 3 d.p. precision. In one case where the data are fractions of pixels and images, I reduced to 2 d.p. It is possible that some of the other data could be reduced to lower than 3 d.p. precision, and perhaps some others should be increased again from 3.

I had to add a couple of guards to avoid calling `round` on `None`, which was revealed by tests. Perhaps that guard should be added to all of these list comprehensions?


jbeilstenedmands:
Thanks @dagewa , I will review, but out of interest, how much does this change reduce the size of your example dials.scale.html ?

dagewa:
With just the original change to those two plots, the one worst file went from 26M to 12M. I'm now doing another run of all the data sets to see what the overall effect of all the changes is.

dagewa:
I think I want to go beyond scatterplots with this PR. Lots of other plots use large numbers of points with overly-precise values.

dagewa:
This is more comprehensive now. I've been through most plots, using the rounding function only for data that looks like it should be float. I adjusted rounding precision in some cases where it is required (hopefully appropriately). I have not checked every plot visually, but I checked a lot of the main ones in `.html` files generated by a `pytest` run, and they look very similar. There are 117 `.html` files generated by the tests. On `main` they total 105M, whereas on the branch they total 82M. So, a decent saving, though not quite as much as I'd hoped.

The largest files from the A/B test runs are now `xia2.json` files. Something similar to this might be appropriate in xia2, though in that case the data are not plotly plots.

src/dials/algorithms/integration/ssx/ssx_integrate.py
```diff
@@ -295,7 +296,7 @@ def make_plots(self):
                 data_angular.append(
                     {
                         "x": n,
-                        "y": list(v),
+                        "y": round_for_json(v),
```
jbeilstenedmands:
```suggestion
                        "y": round_for_json(v, ndigits=6),
```

src/dials/algorithms/integration/ssx/ssx_integrate.py
```diff
@@ -306,7 +307,7 @@ def make_plots(self):
                 data.append(
                     {
                         "x": n,
-                        "y": list(v),
+                        "y": round_for_json(v),
```
jbeilstenedmands:
```suggestion
                        "y": round_for_json(v, ndigits=6),
```

src/dials/pychef/damage_series.py
```diff
@@ -139,35 +143,35 @@ def _add_data_to_plots_dict(self, plots_dict, data, low, upper):
...
                     "name": label,
                     "mode": "lines",
                 }
             )
             plots_dict["cc_half"]["data"].append(
                 {
-                    "x": d_star_sq_bins_this,
-                    "y": [b.cc_one_half for b in result.bins],
+                    "x": round_for_json(d_star_sq_bins_this, ndigits=6),
+                    "y": round_for_json([b.cc_one_half for b in result.bins]),
```
jbeilstenedmands, line 156: Looking at some these plots some more, I'm thinking that maybe we should increase to 4 digits for statistics that are likely to be in the range of 0-1, which would include cc1/2, Rpim, completeness? This would also need to be changed in other places like dials.report and dials.scale code. What do you think?

jbeilstenedmands:
I have reviewed the scaling heatmap style plots, I don't see any that can obviously reduced further. They all seem to use binning of around 200 points in x and y, so although not small, these should be fairly constant in size, independent of dataset size. Whereas the scatter type plots will grow in size with dataset size, so there should be good savings there. This seems to particularly help with the ssx html file sizes.

src/dials/pychef/damage_series.py
dagewa: Yes, that sounds sensible. I'll go through the changes and increase precision on those types of plots.

dagewa:
Yes, that's a good point. I was focused on the heatmaps because when many of the values are the same (like "NaN") then the JSON array is obviously not an efficient way to carry data. But that's a plotly limitation I guess. The amount of data is fixed and not huge, as you point out. The scattering vector plot takes about 644 KB, so almost possible to hold in memory on an IBM PC :wink:

src/dials/pychef/damage_series.py
dagewa: Actually I found that if I just increase the default precision of `round_for_json` from 3 to 4 digits then the total weight of HTML files produced by tests increases only by 1M. This seems an easier approach than looking for all the plots where values are usually in [0,1], and is a safer default for adding new plots in future

jbeilstenedmands (approved):
Thanks for adding this and changing the default, if we need to adjust any plots in future it will be easy to just change the parameter in the call to `round_for_json` 👍

PR #3195, opened 2026-06-16
Make sync_geometry hierarchy-topology-safe.

dwmoreau:
I had an issue where I was constructing a detector with a different hierarchy than the applied reference geometry. sync_geometry failed silently and wasted several hours of beamtime. This hardening is meant to safe guard against this in the future.

Replace the local-frame, zip-by-child copy (which silently mis-pairs nodes when src/dest tree depths differ) with a positional leaf-panel global-frame transfer via set_frame, plus a ValueError guard on panel-count mismatch.

### DIFF
--- a/newsfragments/3195.misc
+++ b/newsfragments/3195.misc
@@ -1,0 +1,1 @@
+Raise in sync_geometry if the reference geometry has a different hierarchy than the source.
--- a/src/dials/command_line/stills_process.py
+++ b/src/dials/command_line/stills_process.py
@@ -388,39 +388,31 @@ def do_import(filename, load_models=True):
     return all_experiments
 
 
-def _collect_leaf_panels(node):
-    """Depth-first list of leaf panel nodes under (and including) `node`."""
-    if node.is_panel():
-        return [node]
-    panels = []
-    for child in node:
-        panels.extend(_collect_leaf_panels(child))
-    return panels
-
-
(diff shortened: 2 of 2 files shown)


phyy-nx:
This does do the thing but it also leaves behind the refined hierarchy.  Example:

Old detector origins:
level 0: local: (0,0,0), global: (0,0,0)
level 1: local: (10,10,10), global: (10,10,10)
level 2: local: (10,10,10), global: (20,20,20)

Refined detector origins (refining level0, then 1, then 2 in sequence)
level 0: local: (1,1,10), global: (1,1,1)
level 1: local: (11,11,11), global: (12,12,12)
level 2: local: (13,13,13), global: (24,24,24)

With the old code, the sync detector function reproduces this hierarchy.  With the new code that uses set_frame on the leaves you get a hierarchy like this:

level 0: local: (0,0,0), global: (0,0,0)
level 1: local: (10,10,10), global: (10,10,10)
level 2: local: (14,14,14), global: (24,24,24)

In practice for downstream applications it's identical, but it is a distortion of what was provided.

What about raising an error if the two models aren't the same 'shape'?

dwmoreau:
I reverted the behavior of this function and added Raise's in two cases:

1. If the src and dst are a mix of panels and groups
2. If the number of children of the between the src and dst are different.

phyy-nx (approved):

PR #3198, opened 2026-06-18
Break tied peak_coordinates selection toward the shoebox centroid

### LINKED ISSUE #3014: Max peak selection may drop useful reflections on equal intensities
A problem I have stumbled across while working on the [fast feedback service](https://github.com/DiamondLightSource/fast-feedback-service) has to do with the way that we are deterministically selecting the max peak signal in connected component analysis.

As it stands, we are using a canonical search for the max peak signal by searching through all pixels in z, y, x ordering and strictly taking the first value greater than the current maximum:
```C++
std::size_t result = 0;
    for(std::size_t i=1;i<a.size();i++) {
      if (a[result] < a[i]) result = i;
    }
```
Snippet from [max_index in `scitbx`](https://github.com/cctbx/cctbx_project/blob/16d2b3950cc497d7e38a877f77b6d881d10ade58/scitbx/array_family/ref_reductions.h#L30-L42) which gets used here:
https://github.com/dials/dials/blob/1e0a33a688e9848691a0bacc808071e0b7e3aaa9/src/dials/array_family/boost_python/flex_shoebox.cc#L452-L467

# Problem
This has the (presumably) unintended side effect that if two pixels have **identical intensity values**, the algorithm _always_ selects the one encountered first in z, y then x order.

For example, if two pixels have **identical** intensity values: Peak1 which has a distance above the limit and Peak2 which is within the limit. The loop:
```C++
values = [ ..., Peak1=245, ..., Peak2=245, ... ]
max = values[0]

for v in values:
    if v > max:
        max = v     # first 245 (Peak1) accepted
                    # second 245 (Peak2) skipped (245 > 245 == false)
```
would always select Peak1 and ignore Peak2, since `current_pixel > max_peak` returns `false` for the second peak, even if it lies within the filter threshold.

Image

# Next steps
This raises the question: is this the best strategy for peak selection? We may be discarding useful data that would otherwise pass the filter if we instead chose the peak closest to the centroid. As mentioned in https://github.com/DiamondLightSource/fast-feedback-service/issues/58, other strategies non-exhaustive include:
- Nearest peak to centroid distance
- Average peak to centroid distance
- Furthest peak to centroid distance

A brief discussion in the last DIALS meeting led to the conclusion that selecting the peak nearest to the centroid would - in fact - be a more accurate measure. I would like to open the discussion here and invite opinions on the matter.

dimitrivlachos:
This PR changes how `Shoebox.peak_coordinates()` resolves ties between
equally-bright pixels. It previously used `af::max_index`, which returns
the first pixel in z, y, x order at the maximum value, so when several
pixels shared that maximum the geometrically-first one always won. That
biases the reported peak away from the spot centroid, and a
poorly-placed peak can push the peak-centroid distance over the
threshold in `PeakCentroidDistanceFilter` and drop an otherwise-valid
reflection. The tie is now broken toward the pixel nearest the shoebox
centroid.

The z, y, x rule is the same deterministic tie-break used by the
fast-feedback-service 3D connected-component analysis, and whether it is
the right choice was initially raised in
DiamondLightSource/fast-feedback-service#58 and discussed in
dials/dials#3014 This moves DIALS to the nearest-to-centroid option
discussed and agreed there.

Closes #3014.

### Changes:

- Find the maximum value, its first index, and the tie count in a single
  pass over the shoebox data. The untied case (almost every reflection)
  is unchanged and costs the same as the old `af::max_index` scan; a
  `DIALS_ASSERT` restores the empty-array guard `af::max_index` enforced
  internally.
- Only when there is a tie compute the shoebox `centroid_all()` and pick
  the tied pixel nearest it. The tie-break lives in
  `peak_index_nearest_centroid`, which compares squared distances (no
  per-candidate `sqrt`), keeps the first pixel on equal distances so the
  result stays deterministic, and falls back to the first maximum when
  the centroid is non-finite (e.g. an all-zero shoebox).
- Add two `test_flex_shoebox` tests: an untied regression that pins the
  existing fast-path behaviour, and the tie case that fails on the old
  implementation.

### DIFF
--- a/newsfragments/3014.bugfix
+++ b/newsfragments/3014.bugfix
@@ -1,0 +1,1 @@
+``Shoebox.peak_coordinates``: when several pixels share the maximum value, break the tie by selecting the pixel nearest the shoebox centroid rather than the first one in z, y, x order. This stops well-placed reflections being rejected by the peak-centroid distance filter.
--- a/src/dials/array_family/boost_python/flex_shoebox.cc
+++ b/src/dials/array_family/boost_python/flex_shoebox.cc
@@ -473,15 +473,99 @@ namespace dials { namespace af { namespace boost_python {
   }
 
   /**
+   * Break a tie between equally-bright pixels by choosing the one nearest the
+   * shoebox centroid, rather than relying on pixel ordering (see
+   * https://github.com/dials/dials/issues/3014). Falls back to first_max (the
+   * first pixel at the maximum, in z, y, x order) when the centroid is
+   * degenerate / non-finite, e.g. for an all-zero shoebox.
+   */
+  template <typename FloatType>
+  std::size_t peak_index_nearest_centroid(
+    const Shoebox<FloatType>& shoebox,
+    const af::const_ref<FloatType, af::c_grid<3> >& data,
+    FloatType max_value,
+    std::size_t first_max) {
+    // Intensity-weighted centroid of the shoebox, in global pixel coordinates.
+    vec3<double> c = shoebox.centroid_all().px.position;
+    if (!std::isfinite(c[0]) || !std::isfinite(c[1]) || !std::isfinite(c[2])) {
+      return first_max;
+    }
+
+    af::c_grid<3> accessor = data.accessor();
+    std::size_t chosen = first_max;
+    double best = -1.0;  // negative sentinel: no candidate measured yet
+
+    // No tied pixel can precede the first maximum, so start the scan there.
+    for (std::size_t j = first_max; j < data.size(); ++j) {
+      // Only the pixels tied at the maximum value are candidates.
+      if (data[j] != max_value) {
+        continue;
+      }
+
+      // Pixel centre in global (x, y, z); the accessor is z, y, x ordered,
+      // hence cd[2]/cd[1]/cd[0] map to x/y/z.
+      tiny<std::size_t, 3> cd = accessor.index_nd(j);
+      vec3<double> p((double)shoebox.bbox[0] + cd[2] + 0.5,
+                     (double)shoebox.bbox[2] + cd[1] + 0.5,
+                     (double)shoebox.bbox[4] + cd[0] + 0.5);
+
+      // Squared distance to the centroid (squared avoids a per-pixel sqrt).
+      double d2 = (p - c).length_sq();
+
+      // Keep the closest candidate; the strict comparison means an equal
+      // distance does not displace an earlier pixel, so ties stay deterministic.
+      if (best >= 0.0 && d2 >= best) {
+        continue;
+      }
+      best = d2;
+      chosen = j;
+    }
+    return chosen;
+  }
+
+  /**
    * Get the maximum index of each shoebox
    */
   template <typename FloatType>
   shared<vec3<double> > peak_coordinates(ref<Shoebox<FloatType> > a) {
     shared<vec3<double> > result(a.size(), af::init_functor_null<vec3<double> >());
     for (std::size_t i = 0; i < a.size(); ++i) {
-      std::size_t index = af::max_index(a[i].data.const_ref());
-      af::c_grid<3> accessor = a[i].data.accessor();
-      tiny<std::size_t, 3> coord = accessor.index_nd(index);
+      af::const_ref<FloatType, af::c_grid<3> > data = a[i].data.const_ref();
+      // The loop below seeds from data[0], so requires a non-empty
+      // array. This restores the empty-array guard that af::max_index
+      // enforced internally throwing on size 0
+      DIALS_ASSERT(data.size() > 0);
+
+      // Single pass to find the maximum value, the first index
+      // attaining it (preserving the historic z, y, x ordering for the
+      // common, untied case) and how many pixels share that maximum
+      // value. This is the same scan af::max_index performed (update
+      // only on a strictly greater value, so the first maximum is
+      // kept), with the sole addition of counting ties in n_max so we
+      // know whether the centroid tie-break below is needed.
+      FloatType max_value = data[0];
+      std::size_t max_index = 0;
+      std::size_t n_max = 1;
+      for (std::size_t j = 1; j < data.size(); ++j) {
+        if (data[j] > max_value) {
+          max_value = data[j];
+          max_index = j;
+          n_max = 1;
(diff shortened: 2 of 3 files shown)


dagewa:
Looks sensible. If you like I can do an A/B test on the proteindiffraction.org Eiger data to check the effect on performance and data quality (both of which I expect to be minimal).

dimitrivlachos:
That would be greatly appreciated @dagewa!

jbeilstenedmands:
Interestingly, this PR causes a few test failures as there are now more reflections existing than previously (the format of these tests are `assert len(reflections) == expected`):
```
FAILED test_find_spots.py::test_find_spots_radial_profile[None-559] - assert 560 == 559
FAILED test_find_spots.py::test_find_spots_radial_profile[narrow-721] - assert 730 == 721
FAILED test_find_spots.py::test_find_spots_radial_profile[wide-739] - assert 750 == 739
FAILED test_find_spots.py::test_find_spots_with_xfel_stills - assert 1155 == 1148
```
So we are now throwing away fewer reflections.

dimitrivlachos:
That would be an expected consequence of this change! In fact it's nice to see. Though I defer to you the determination of correctness here.

dagewa:
After all the issues on SCARF recently I finally got an A/B test run with this branch. The dashboard isn't really ready to explore data quality in detail, but using the proxy of differences in resolution limit as determined by `dials.estimate_resolution`, this looks neutral or slightly good. The histogram shows the resolution limit of the B arm (the branch) minus the A arm (main), so values to the left of 0.0 mean the A resolution limit was numerically higher, and therefore the B processing is "better" than the A processing.

res-A-B


In terms of memory usage the branch seems slightly favourable. We could speculate why (more spots --> better refined model?)
peak-mem-A-B-ratio

There were a few weird individual data sets that show a big difference though. I'm looking into why.
peak-mem-A-B-steps

dagewa:
I looked into the 3 data sets where peak memory deviates by a large amount on the B arm vs A arm.

- Data set 27, [KlaeA_00906_a_B1-PLP-SER_8u99](https://proteindiffraction.org/project/KlaeA_00906_a_B1-PLP-SER_8u99/): got a different indexing solution (`mC 167.48 131.19  87.81  90.00 106.67  90.00` vs `mC  87.81 131.21  82.66  90.00 103.92  90.00`) with twice the volume on the B arm, hence processed twice the number of spots
- Data set 51, [nsls2_fmx_20161122_lys_266](https://proteindiffraction.org/project/nsls2_fmx_20161122_lys_266/): this one is strange. Essentially no difference in indexing solution or data quality from the processing. Possibly the peak memory comparison here is spurious. To dig deeper this would have to be processed again as most of the intermediate files are deleted during the A/B test
- Data set 92, [8xj5](https://proteindiffraction.org/project/8xj5/): similar indexing solutions, but the refined model is better for B with lower `sigma_b` and `sigma_m`, resulting in lower memory requirements and better merging statistics

dagewa:
In addition to the 3 data sets that showed large differences in peak memory during processing, there were also 4 data sets where the A/B test failed, so they are not included in the graphs above. In 3 of these cases this was due to being OOM killed during the B arm processing.

- [8r5r](https://proteindiffraction.org/project/8r5r/): fairly different indexing solution on the B arm (`oP  89.66 130.17 134.74  90.00  90.00  90.00` vs `oP  91.57 132.53 137.21  90.00  90.00  90.00`) ultimately leading to much worse `sigma_m`, which causes the job to be killed
- [6s0r](https://proteindiffraction.org/project/6s0r/): completely different indexing solution on the B arm (`aP  50.36 110.15 185.96  91.42  94.96  94.22` vs `mP 100.53 183.82 117.40  90.00 110.99  90.00`). This solution has half the volume and therefore we process half the spots. However the `sigma_b` and `sigma_m` model is larger so it uses more RAM and is killed
- [KlaeA_00906_a_B1-PLP-oxamate_8sa9](https://proteindiffraction.org/project/KlaeA_00906_a_B1-PLP-oxamate_8sa9/): in this case xia2 is not sure of the indexing solution. On the A arm it tries `mC  87.31 131.50  82.57  90.00 103.80  90.00`, decides it doesn't like it and goes back to reprocess with `aP  78.93  78.92  82.57  97.59  97.58 112.84`. On the B arm it picks a similar solution to the initial A arm solution `mC  87.43 132.29  82.79  90.00 104.00  90.00`, but this seems different enough that the `sigma_b` and `sigma_m` are large enough that the job is OOM-killed.

For the 4th case, [BumuA_00117_a_B1-45C_9dra](https://proteindiffraction.org/project/BumuA_00117_a_B1-45C_9dra/), this was OOM killed even on the A arm (this is a data set that frequently fails to process and probably I should exclude it)

dagewa:
A further point of interest: the median increase in the number of found spots is about 0.2%, but there are some data sets where the change finds a somewhat larger proportion of spots, up to about 1.75% more

nspots-boxplot

I found this a bit surprising - given that it should only affect spots that have tied peaks I thought the number of additionally-found spots would be smaller than this. This is not necessarily a problem, I just found it a bit counterintuitive.

In conclusion, the branch looks neutral or good for most data sets, but for about 3% of data sets the processing pipeline is unstable enough that a small change in the spot-finding results leads to a very different data processing outcome.

graeme-winter:
@dagewa 

> ... but for about 3% of data sets the processing pipeline is unstable enough that a small change in the spot-finding results leads to a very different data processing outcome.

these are not happy words to me: extracting those sets into a collection we should stare hard at feels useful

dagewa:
Yes, there are also quite a few data sets I have already excluded from processing because they are in some way problematic (do not process right with a standard xia2 command). This tells me I need to keep track of the good, bad, and ugly sets...

graeme-winter:
> Yes, there are also quite a few data sets I have already excluded from processing because they are in some way problematic (do not process right with a standard xia2 command). This tells me I need to keep track of the good, bad, and ugly sets...

That feels like something for the greater good as these are worthy of closer inspection

phyy-nx:
+1 for some kind of repository of knowledge about good/bad/ugly datasets, and access to your A/B testing methods :)

(did you post those somewhere? apologies if I missed it)

dagewa:
It's still under active development with zero documentation, but I will get it into a shareable state eventually

PR #3228, opened 2026-07-24
Add dials.ssx_refine program

### LINKED ISSUE #3216: Add dials.ssx_refine front end
We have `dials.ssx_index` which is functionally different to `dials.index` then (for calibration) we run `dials.refine` with e.g.

```console
refinement {
  parameterisation {
    auto_reduction {
      action = fail fix *remove
    }
    beam {
      fix = *all in_spindle_plane out_spindle_plane wavelength
    }
    detector {
      fix_list = "Tau1"
    }
  }
  refinery {
    engine = SimpleLBFGS LBFGScurvs GaussNewton LevMar *SparseLevMar
  }
  reflections {
    outlier {
      algorithm = null auto mcd tukey *sauter_poon
    }
  }
}
```

It would feel ... internally consistent to offer a front-end command line program which had "suitable" defaults for SSX refinement with `dials.ssx_refine`

jbeilstenedmands:
Add a small wrapper program, ``dials.ssx_refine``, which runs ``dials.refine`` with suitable defaults for ssx joint detector geometry refinement. Fixes #3216 .

This supports the workflow `dials.import -> dials.find_spots -> dials.ssx_index -> dials.ssx_refine -> dials.import with reference geometry -> dials.find_spots -> dials.ssx_index`.

Ideally what we want is to be able to skip two steps and just run `dials.ssx_index refined.expt strong.refl`, but to enable this I think we need to be able to propagate unindexed reflections and experiment models through `dials.ssx_index` (✅ ) and `dials.refine` (❌), so that is a separate piece of work.
Alternatively, we can use `dials.modify_experiments` on the original imported.expt to set updated detector parameters, but they need to be manually specified. So an alternative general solution would be to add the ability for `dials.modify_experiments` to use a reference detector from an experiment file. i.e. the workflow would be `dials.import -> dials.find_spots -> dials.ssx_index -> dials.ssx_refine -> dials.modify_experiments -> dials.ssx_index`.


jbeilstenedmands:
Ok I have a working update to `dials.modify_experiments` so that you can specify a `reference_geometry` in the same way as for `dials.import`, to enable the latter workflow suggested, PR will be incoming soon.

PR #3231, opened 2026-07-27
update unit cell clustering plots

amyjaynethompson:
Pairs with an upcoming xia2 PR to update multiplex graphs. 

The capability to highlight a specific region of the unit cell clustering graph is added. The purpose of this is to quickly visualise which datasets are accepted and taken on for processing. Here, it was also realised that the proper dataset labels weren't carried through (ie if datasets 0,1,4,5,6 were accepted, they were later being relabelled to 0,1,2,3,4). This was fixed by removing the commented out labels input to the dendrogram function. 

Also found that some of the rij histograms were getting too many bins, so have added an upper limit to prevent this.


src/dials/algorithms/correlation/analysis.py
```diff
@@ -921,15 +921,18 @@ def convert_to_html_json(self):
...
+                20,
+                (
+                    int(
+                        (self.correlation_matrix.max() - self.correlation_matrix.min())
+                        / 0.005
+                    )
+                    * 2
+                ),
             ),
+            60,
```
jbeilstenedmands: If the potential data range is -1 to 1, would it make sense to use a rounder number like 100, so that the smallest possible bin width is 0.02 - would this give us enough resolution in seeing the histogram features?

jbeilstenedmands (changes requested):
Approved once we increase the max number of bins slightly. Current main branch allows up to 800 bins in the potential data range of -1 to 1, which does feel too high. I think we should look to restrict based on a sensible bin width in this case, round numbers like 0.01 (200 bins) or 0.02 (100 bins) feel like sensible choices. We might have that full range in dials.cosym, less likely in dials.correlation_matrix, so maybe 100 is fine, but it would be good to check on a few cosym and correlation_matrix examples.

jbeilstenedmands (approved):

PR #3232, opened 2026-07-27
Add reference_geometry option to dials.modify_experiments

jbeilstenedmands:
Rather than having to specify a lot of parameters, enable the use of a reference experiment file to provide detector, beam and goniometer information. With the use case in mind of updating an SSX import.expt file following joint detector refinement.

One thing that struck me about dials.modify_experiments (after comparing to dials.modify_geometry), is that we update imagesets but loop over experiments rather than imagesets, which could get slow for ssx files with thousands of experiments that share a single imagesequence.


ndevenish:
An obvious improvement!

dagewa (approved):
This looks straightforwardly sensible :+1:

Regarding the loop over experiments, apparently I needed that to work around some "deeper issue with dxtbx" (https://github.com/dials/dials/issues/2927#issuecomment-2894510327). I guess we could look at this further if the performance becomes a concern

jbeilstenedmands:
I just tried on a 1000 image ssx expt file, `dials.modify_experiments` takes 5s compared to 3s with an equivalent modification with `dials.modify_geometry` (one of my earlier commits). I can live with that for now.

PR #3248, opened 2026-08-27
Determine the number of dimensions from one minimisation, by projection

graeme-winter:
`_determine_dimensions` minimises the target once for every candidate number of dimensions, each time starting again from random coordinates.  With 24 symmetry operations that is 24 minimisations, and it is the slowest part of a cosym run.

A solution in d dimensions is also a solution in d + 1 with a column of zeros, so the curve can be traced downwards from a single solution instead. Minimising once at the largest dimension and rotating the coordinates so that their principal axes come first, the leading d of them are the best d-dimensional representation of that solution, and the functional evaluated there is a point on the curve.  The rotation is cheap: the coordinates are dims_to_test by N, so it is their Gram matrix, of side dims_to_test, that is decomposed rather than anything the size of Rij.

Like the values the present loop produces these are upper bounds on the minimum at each dimension rather than the minimum itself.  Unlike them they are monotonic, as a functional minimised over a widening space must be; the present curve is not, because twenty calls from random coordinates do not converge, and the elbow is fitted partly to that.

Measured on synthetic problems of 60 to 120 data sets and 24 operations at 10-12% density, planting two, three and four clusters and two seeds each: 15 times faster, and agreeing with the present method wherever the present method is right.  Where they differ it is the present method that is wrong: with four clusters planted it returns five dimensions, twice, where projection returns four.

Selected by dimensions_method=projection; the default is unchanged.


graeme-winter:
Test case: took CIX data sets from cows, pigs, people, imported with split=10 i.e. into 120 x 1 degree wedges, processed as usual.

Before:

```console
real	6m35.873s
user	21m16.786s
sys	7m17.620s
```

After a few dead ends and other approaches tested:

```console
real	0m35.779s
user	2m22.657s
sys	0m34.678s
```

graeme-winter:
# Determining the number of dimensions in cosym

## Status: the eigenvalue shortcut does not work, and is withdrawn

The idea below was that the functional-versus-dimension curve could be had in
closed form from the eigenvalues of Rij, replacing 24 minimisations with one
decomposition. It is fast and it is wrong. Tested on real data it chose 7
dimensions where the existing code chooses 2, and the reason is the one this
document originally listed as a risk and then under-weighted: **the mask**.

Entries below `min_pairs` are never computed. They sit as zeros in
`rij_matrix`, with zeros in `wij_matrix` to match, and the weighted target
ignores them. The unweighted eigenvalue problem cannot ignore them: it reads
them as correlations that genuinely are zero and fits them. The diagonal is
excluded in the same way and has the same effect, and being the entries that
directly control the length of each coordinate vector, it has an influence out
of proportion to its size.

Fitting all those zeros flattens the curve. On real data:

```
              patch (eigenvalue)     original
  dim 1             1.04584e+06       579263
  dim 2                  842998       274268
  drop                    19.4%        52.7%
  elbow                       7            2
```

Reproduced on a synthetic matrix 39% measured with the diagonal excluded: the
plain eigenvalue curve falls 18.4% from one dimension to two, where both the
weighted minimisation and a mask-respecting eigenvalue calculation fall 73.5%.
The elbow survives at that density; at the sparsity of real narrow-wedge data
it does not.

Respecting the mask fixes the answer but removes the reason for doing it. The
soft-impute iteration needed to handle missing entries costs an
eigendecomposition per pass, and measured against the minimisation it was ten
times *slower*, not fifty times faster. There is no shortcut here.

The patch is kept in `patches/withdrawn/` as a record. What follows is the
original argument, which is correct for a dense Rij and wrong for cosym's.

---


## What the current code does

`CosymAnalysis._determine_dimensions` chooses how many dimensions the
multidimensional scaling should use. It minimises the target

```
f = 0.5 * sum_ij W_ij (R_ij - x_i . x_j)^2
```

once for every candidate number of dimensions from 1 up to the number of
symmetry operations, records the value reached, and takes the elbow of the
resulting curve by the method of Zhang et al. (2006).

For a cubic lattice group that is 24 minimisations. Each function evaluation
forms `x.T @ x`, an N × N matrix where N is the number of data sets times the
number of symmetry operations, so with 347 data sets and 24 operations N = 8328
and every call touches 69 million elements. It is the slowest part of a run.

## The observation

Set the weights aside for a moment. The matrix `x.T @ x`, where `x` has `d`
rows, is a symmetric positive semi-definite matrix of rank at most `d`. So
minimising `f` over `x` is exactly the problem of finding the best positive
semi-definite rank-`d` approximation to `R` in the Frobenius norm.

That problem has a closed-form solution — the Eckart–Young theorem. The best
rank-`d` approximation keeps the `d` largest eigenvalues and their
eigenvectors. The positive semi-definite constraint adds one wrinkle: a
negative eigenvalue of `R` can never be reproduced by `x.T @ x`, so negative
eigenvalues are always left behind whatever `d` is.

The residual is the sum of the squares of the eigenvalues left behind. Since
the squares of *all* the eigenvalues sum to `||R||_F^2`, that gives

```
f(d) = 0.5 * ( ||R||_F^2 - sum_{i <= d} lambda_i^2 )
```

where `lambda_1 >= lambda_2 >= ...` are the largest **positive** eigenvalues.

So the entire functional-versus-dimension curve is a running total over the
eigenvalues of `R`. One decomposition gives every point on it, and no
minimisation is needed at all. Only the largest `dims_to_test` eigenvalues are
needed, so a truncated solver can be used: `O(N^2 m)` rather than `O(N^3)`.

## Why this is not only faster

The minimisation is capped at 20 function calls per dimension. In a space of
`d * N` parameters started from random coordinates, 20 L-BFGS calls do not
converge. The consequence is visible in the curve itself:

```
 dim   minimisation (as now)   closed form
   1              1371118.8      1371118.8
   2               418887.0       418887.0
   3               419903.1       418304.3
   4               419915.2       417726.0
   5               428993.0       417155.2
```

The current curve **rises** from dimension 2 to 3, and again at 5. A converged
`f(d)` cannot do that: any solution in `d` dimensions is also a solution in
`d + 1` with a column of zeros, so the minimum is non-increasing in `d` by
construction. What is being fitted for the elbow is therefore partly noise. It
still lands on the right answer here, because the first drop dominates
everything else, but it is not doing what it appears to be doing.

The closed form is monotonic by construction, because it is a running total of
squares.

On a synthetic problem of 120 data sets and 24 operations, both find dimension
2; the closed form takes 1.6 s against 82 s for the minimisations.

## Where the approximation lies

Two differences from the target that cosym actually minimises, both of which
matter only through their effect on the choice of an integer:

**Weights.** With `weights=count` (the default) each term is weighted by the
number of reflections the correlation was computed over. Eckart–Young applies
to the unweighted problem; weighted low-rank approximation has no closed-form
solution. The eigenvalue curve is therefore an approximation whenever weights
are on.

**Entries that were never measured.** Where a pair has fewer than `min_pairs`
reflections in common the entry is skipped and stays zero in `rij_matrix`, with
a zero in `wij_matrix` to match. The weighted target ignores those terms; the
unweighted eigenvalue problem treats them as correlations that genuinely are
zero. Those are different questions, and they diverge as the matrix gets
sparser.

This is the objection to test rather than argue about. On a synthetic matrix
that is 6.3% measured, both approaches still find dimension 2, but the
unweighted curve falls only 3% from one dimension to two where a mask-respecting
calculation falls 71%. The elbow is in the same place and much less firmly
determined. On a matrix that is essentially complete the two agree closely.

If the difference does matter on real data, the fix is a soft-impute iteration:
fill the unmeasured entries with the current rank-`d` approximation, take the
eigendecomposition, repeat. That respects the mask, converges in a few passes,
and still costs far less than a minimisation per dimension.
`tools/cosym_dimension_prototype.py --impute` implements it, so the comparison
can be made before deciding whether the patch needs it.

## What is worth doing instead

**Trace the curve downwards from one minimisation.**
`patches/0004-cosym-dimensions-by-projection.patch`, and this supersedes the
sampled grid below.

A solution in d dimensions is also a solution in d + 1 with a column of zeros.
So the curve does not need a fresh minimisation at every dimension: minimise
once at the largest, rotate the coordinates so their principal axes come first,
and the leading d of them are the best d-dimensional representation of that
solution. The functional evaluated there is a point on the curve. The rotation
costs nothing worth counting, because the coordinates are dims_to_test by N and
it is their Gram matrix, of side dims_to_test, that gets decomposed — not
anything the size of Rij.

Like the values the present loop reports, these are upper bounds on the true
minimum at each dimension rather than the minimum itself. Unlike them they are
monotonic, as a functional minimised over a widening space must be.

Measured on synthetic problems of 60 to 120 data sets and 24 operations at
10-12% density, planting two, three and four clusters, two seeds each:

| planted | current | projection | speedup |
| --- | --- | --- | --- |
| 2 | 2, 2 | 2, 2 | 16.7x, 14.0x |
| 3 | 3, 3 | 3, 3 | 15.0x, 14.6x |
| 4 | **5, 5** | 4, 4 | 15.7x, 15.0x |

Where the two differ it is the present method that is wrong: twenty calls from
random coordinates do not converge, the curve is inflated at the dimensions
where the fit is hardest, and the elbow moves. Fifteen times faster and more
accurate is an unusual combination, and it comes from the same cause.

## Superseded and secondary

**Sample the dimensions, do not scan them.**
`patches/0003-cosym-sampled-dimension-grid.patch`. The cost of one evaluation
grows with the dimension, so a full scan of 24 spends most of its time at the
large end where the curve is flat and the elbow never falls: the work is
proportional to the sum of the dimensions tested, 300 for a full scan. The elbow
search takes whatever pairs it is given and does not need them consecutive.
Keeping every dimension up to eight and then a widening selection, always
including the largest since it anchors the search, brings 300 down to 118.
Measured on 120 data sets with 24 operations at 19% density: **138 s to 55 s,
same answer.** `dimensions_grid=full` restores the old behaviour.

This helps, but it is a way of not doing expensive work rather than a way of
doing it better; with the projection method above the curve is cheap enough that
sampling is unnecessary.

**Solve by soft-impute rather than by minimising from random coordinates.**
Prototype in `tools/cosym_softimpute_prototype.py`, not yet a patch. This is
what recovers some of the speed the eigenvalue shortcut appeared to offer,
without its error. The masked low-rank problem is solved by filling the absent
entries with the current rank-d approximation and taking the eigendecomposition,
which is what the withdrawn patch should have done; the trick that makes it
affordable is that the filled matrix is never formed. It is a sparse residual
plus a low-rank term, so a matrix-vector product costs O(nnz + N d) and a
truncated solver can work on it as a linear operator. The coordinates that come
out are then scored with cosym's own weighted functional, so the reported curve
is directly comparable.

On the same problem that is 74 s against 138 s, the same elbow, and functional
values within 0.6% of what the minimiser reaches. Combined with the sampled grid
it is 19 s, a 7x speedup overall — but at that setting the elbow moved from 2 to
3, because cutting the number of impute iterations to six leaves the low
dimensions short of converged. It is promising and it is not yet trustworthy:
the iteration count needs to be chosen on evidence rather than guessed, and the
whole thing solves the unweighted masked problem rather than the weighted one.

Two things came out of the exercise that do stand up.

**Start each dimension from the previous one.**
`patches/0001-cosym-warm-start-dimensions.patch` does this. The curve cosym
computes now is not monotonic - it can rise from one dimension to the next,
which a converged f(d) cannot do, since any solution in d - 1 dimensions is
also a solution in d with a column of zeros. Twenty L-BFGS calls from random
coordinates in a space of dim * n * m parameters do not converge, so the elbow
is fitted partly to non-convergence. Starting from the previous dimension's
answer with a small random column appended makes the curve monotonic at no
extra cost. It is a correctness fix rather than a speed one: cutting the call
count to 5 on the back of it moved the elbow from 2 to 5, so the calls are
genuinely needed.

**Keep the correlation matrices sparse — withdrawn.**
`patches/0002-cosym-sparse-target.patch` is not recommended. On a real 120 data
set case it had not finished after twenty minutes. The gather `x[k][rows]` is a
random access over the stored entries, and against dense BLAS it only wins when
the matrix really is sparse; the benchmark below was taken at 12% density, and
on a denser matrix the gathers dominate and it is far slower. If it is ever
revisited it needs a density test that keeps the dense path above perhaps 30%.
The measurements below stand for the sparse regime only. `_compute_rij_wij` builds
Rij and Wij as scipy COO matrices and then calls `.toarray()` on both. At 347
data sets and 24 operations that is two dense 8328 x 8328 arrays, 555 MB each,
and `compute_functional` forms a third as a temporary on every call.

The weights are zero wherever a correlation was not computed, so those entries
contribute nothing to the functional, the gradient or the curvatures. Evaluating
over the stored entries alone is therefore **not an approximation but the same
number**, which is what makes this worth doing rather than a trade.

Checked against the dense implementations at 6%, 50% and 95% density: the
functional agrees exactly, the gradients and curvatures to 4e-15. At 12% density
and 2400 coordinates:

| | dense | sparse | |
| --- | --- | --- | --- |
| functional | 42.5 ms | 23.6 ms | 1.8x |
| gradient | 56.3 ms | 25.7 ms | 2.2x |
| curvatures | 5.1 ms | 2.0 ms | 2.6x |
| storage | 92 MB | 13 MB | 7x |

The gain grows as the matrix gets sparser, which is the case that is slow. Two
details worth a reviewer's attention. The two matrices are given a common
sparsity pattern so that the stored values of one index the other; and
`compute_functional_score_for_dimension_assessment` takes quantiles over every
element including the zeros that are no longer held, so the quantile is located
in a notional sorted array of that many zeros followed by the stored values
rather than materialising them.

## Testing the withdrawn patch

`patches/withdrawn/0001-cosym-eigenvalue-dimensions.patch` adds a phil choice:

```
dimensions_method = *minimization eigenvalue
```

The default is unchanged, so nothing behaves differently until asked. Apply it
with

```
cd /path/to/dials/modules/dials
git am /path/to/mxoverlap/patches/0001-cosym-eigenvalue-dimensions.patch
```

and then run the same data both ways:

```
dials.cosym integrated.expt integrated.refl
dials.cosym integrated.expt integrated.refl dimensions_method=eigenvalue
```

The tabulated `Dimensions / Functional` table is written to the log by both, so
the two curves can be compared directly, along with `Best number of dimensions`.

To look at the curves without running cosym twice, or to try the mask-respecting
variant:

```
dials.python tools/cosym_dimension_prototype.py integrated.* --save run1 --compare
python tools/cosym_dimension_prototype.py --rij run1_rij.npy --wij run1_wij.npy --impute
```

`--save` writes the matrices out, so building them — the other expensive stage —
is paid for once.

## What to check before proposing it upstream

1. Do the two methods choose the same number of dimensions on real data,
   across a few data sets of different sparsity? The sparse cases are where
   they might not.
2. Does `--impute` agree with both? If it agrees with the plain eigenvalue
   curve, the mask is not an issue for that data and the simple version is
   enough.
3. Does the truncated solver ever fail to converge on a real Rij? It falls back
   to a full decomposition if `scipy.sparse.linalg.eigsh` raises, but a silent
   near-miss would be worse than a failure. Comparing against
   `np.linalg.eigvalsh` on one real matrix would settle it.

If all three hold, the follow-up is a one-line change of the default, and the
existing path can stay for a release as a way of reproducing old results.

## What this does not address

The correlation matrix itself still has to be built, and on sparse data that is
the other expensive stage. Two things stand out there, neither touched by this
patch:

- In `_compute_rij_matrix_one_row_block`, `cb_op_kk.apply(...)` followed by
  `miller.map_to_asu(...)` for data set `j` sits inside the loop over `i`, so
  every data set is re-mapped once per partner: about `n^2 m / 2` mappings where
  `n m` would do. Precomputing the mapped and sorted indices once per
  (data set, operation) would cut that by a factor of roughly `n / 2`.
- `compute_functional` allocates a full N × N temporary per call. Unweighted,
  the expansion `f = 0.5 (||R||^2 - 2 tr(X R X^T) + ||X X^T||^2)` needs one
  `N^2 d` product and no N × N temporary; the last term costs only `d^2 N`.

## References

- Gildea, R. J. & Winter, G. (2018). *Acta Cryst.* **D74**, 405–410.
- Brehm, W. & Diederichs, K. (2014). *Acta Cryst.* **D70**, 101–109.
- Eckart, C. & Young, G. (1936). *Psychometrika* **1**, 211–218.
- Zhang, Z., Sauter, N. K., van den Bedem, H., Snell, G. & Deacon, A. M. (2006).
  *J. Appl. Cryst.* **39**, 112–119.

graeme-winter:
More realistic / typical example that I frequently have to wait for when doing a demo, cows, pigs, people; before:

```console
real	0m49.244s
user	3m23.605s
sys	0m53.018s
```

and after:

```console
real	0m15.395s
user	1m4.420s
sys	0m12.501s
```

so a more modest 3x speedup but valid nonetheless. Protected behind `dimensions_method=projection` until we elect to change the default

graeme-winter:
OK more extreme example: cows pigs people, but with split=10 so we have 360 x 1 degree data sets

Before:

```console
real	68m47.211s
user	192m12.670s
sys	228m58.544s
```

After:

```console
real	5m2.278s
user	16m52.467s
sys	12m58.552s
```

And functional before / after are very similar:

```console
+--------------+--------------+
|   Dimensions |   Functional |
|--------------+--------------|
|            1 |  4.84353e+06 |
|            2 |  2.62463e+06 |
|            3 |  2.61231e+06 |
|            4 |  2.60141e+06 |
|            5 |  2.59641e+06 |
|            6 |  2.59153e+06 |

+--------------+--------------+
|   Dimensions |   Functional |
|--------------+--------------|
|            1 |  4.84877e+06 |
|            2 |  2.63185e+06 |
|            3 |  2.61976e+06 |
|            4 |  2.60905e+06 |
|            5 |  2.60288e+06 |
|            6 |  2.59705e+06 |
```

amyjaynethompson (approved):
Speed improvements will be very appreciated, and the approach seems sensible to me.  I'm happy for this to be merged as an available option. Over time, we can test against a wider range of cases and see if the default should be changed.

jbeilstenedmands:
Thanks Graeme, this is a really impressive speedup.
I went back and profiled the larger cow-pig-people dataset, which I profiled in a previos PR (https://github.com/dials/dials/pull/3062).

With your addition, we now have:

projections
Note the file from before no longer exists on disk, so don't compare the peak memory, but the effect on runtime is clear, with the dimension assessment now taking ~115s compared to ~2200s.

Combined with the previous work, we've gone from ~4600s total runtime to ~360s, very nice!

Through the inheritance of phil scopes, this PR makes the parameter available to `dials.correlation_matrix`, but I don't think it is applicable there. Indeed the new parameter breaks the 36 dataset cow-pig-people example, because it does the minimisaton in 36 dimensions (because max dim = no of datasets), where the solution has a functional of ~zero (because the dimensionality gives enough dof to fully reproduce all rij values), and then projecting that results in a meaningless answer. Whereas in cosym, you have Nsym x M xtal datapoints which is always greater than max dimensions, which is limited to Nsym. So I think this always makes sense for cosym, might it make sense for correlation matrix if we start at e.g. max_dim = N / 2 ? I feel like probably not in general, so we might need to keep the current approach there.

graeme-winter:
@jbeilstenedmands 

> Through the inheritance of phil scopes, this PR makes the parameter available to dials.correlation_matrix, but I don't think it is applicable there. Indeed the new parameter breaks the 36 dataset cow-pig-people example, because it does the minimisaton in 36 dimensions (because max dim = no of datasets), where the solution has a functional of ~zero (because the dimensionality gives enough dof to fully reproduce all rij values), and then projecting that results in a meaningless answer. Whereas in cosym, you have Nsym x M xtal datapoints which is always greater than max dimensions, which is limited to Nsym. So I think this always makes sense for cosym, might it make sense for correlation matrix if we start at e.g. max_dim = N / 2 ? I feel like probably not in general, so we might need to keep the current approach there.

Agreed - it is clearly applicable only to `cosym` - will figure out how to get it locked down to there. Good spot.

jbeilstenedmands:
I'm tempted to say that we should make your option the new default for cosym given the vast speedup, either now or after a few weeks of testing post-merge. Do you have some good examples that are not cubic, just to validate this further?

jbeilstenedmands:
Actually I have some tests I wrote for cosym with generated data that is pesudo-cubic, so you need to have at least 6 or 12 dimensions to solve properly, I will dig those out next week to check we don't miss the elbow in the functional curve with the new approach.

graeme-winter:
> Actually I have some tests I wrote for cosym with generated data that is pesudo-cubic, so you need to have at least 6 or 12 dimensions to solve properly, I will dig those out next week to check we don't miss the elbow in the functional curve with the new approach.

That would be a spectacular way to check that this is doing what we think it should be

graeme-winter:
@jbeilstenedmands this commit ☝️ moves the selection to the command line phil scope, thank you for highlighting that. It is a ... fragile ... way of checking it though so probably we want to refactor that sometime.

jbeilstenedmands:
Thanks, I have now made projection the default method, I think we should do this because the performance is so much better and we understand theoretically why it is a robust approach.
I tested on some synthetic pseudosymmetry cases, which still work fine, although that changeset includes changes to test data generation which breaks some dials.symmetry tests, so I am not going to include that here.

jbeilstenedmands (approved):

dagewa:
Nice :+1: The improvements here should help for the I19 serial small wedge project, especially when they have high symmetry cases. Multiplex speed was noted as an issue there, which was probably mainly cosym.

dagewa:
As a note to self, I should check this one: `/dls/i19-2/data/2025/nt43442-1/processing/crystallographic_data/5_deg_datasets/CuBTC/`, which took about 28 minutes prior to any multiplex performance improvements

jbeilstenedmands:
@dagewa Thanks for the reminder, I went back to check that dataset, which now takes 5m 20s total runtime for multiplex (with nproc=16).
Longest times from multiplex debug log:
```
158.2s:   1. dials.cosym  'tmp.expt' 'tmp.refl' 'nproc=16' 'output.experiments=1_reindexed.expt' 'output.reflections=1_reindexed.refl' 'relative_length_tolerance=0.05' 'absolute_angle_tolerance=2.0' 'output.json=1_dials.cosym.json' 'output.html=1_dials.cosym.html' 'best_monoclinic_beta=False' 'lattice_symmetry_max_delta=5.0'
 82.3s:   3. dials.scale  '2_refined_cell.expt' 'observations.refl' 'nproc=16' 'full_matrix=False' 'output.html=3_dials.scale.html' 'output.experiments=3_scaled.expt' 'output.reflections=3_scaled.refl'
 18.5s:  12. dials.correlation_matrix
  6.7s:   2. dials.two_theta_refine  'models.expt' 'observations.refl' 'combine_crystal_models=True' 'output.cif=2_dials.two_theta_refine.cif' 'output.mmcif=2_dials.two_theta_refine.mmcif' 'output.p4p=2_dials.two_theta_refine.p4p' 'output.experiments=2_refined_cell.expt'
  .....
```

PR #3254, opened 2026-08-27
Fix PHIL option type in dials.search_beam_position

### LINKED ISSUE #3253: PHIL definition for method needs to have a type, and some fix inside the code that handles this type
All this commands run well:

     dials.search_beam_position imported.expt strong.refl method=default
     dials.search_beam_position imported.expt method=midpoint
     dials.search_beam_position imported.expt method=maximum                
     dials.search_beam_position imported.expt method=inversion

It seem like the parameter `method` should be a _type_ `"choice"` but I need it defined in the _PHIL_ definition, so the auto-generated parameter editor in Dui2 reads it.

If I edit the file `... modules/dials/src/dials/command_line/search_beam_position.py` and I add the _type_ `"choice"` to it, the function `resolve_projection_methods`  brakes.

Please fix!

mpks:
The "method" option needs .type = choice to work with DUI2. With that type, PHIL returns the selected value as a single string rather than a list, so `resolve_projection_methods` has been updated to handle a string instead of a list of strings.

Fixes #3253


jbeilstenedmands:
@mpks This PR is now causing the tests in `tests/command_line/test_search_beam_position.py` to fail, could you update that please. Thanks.

mpks:
@jbeilstenedmands Sorry, I should have run pytest with --regression. The file has 9 tests, but 6 are skipped without it, and those were the failing ones. They were skipped in the CI log too, so nothing flagged it before the merge.

The fix is in the linked PR above (#3261).  Could you take a look?
Also, should I always request a review on my PRs, or is it OK to merge them myself?

jbeilstenedmands:
@mpks I'm happy for simple PRs that affect a small amount of code to be merged without review, but we're always happy to take a look over if you want a sanity check.

No worries about the missed regression tests, we've all been caught out by that before.

PR #3255, opened 2026-08-27
Use hdf5 file format as default on-disk storage format for reflection tables

jbeilstenedmands:
As discussed at a previous dials-core meeting, we should consider changing to hdf5 as the default on-disk storage format for reflection tables.

@dagewa provided a nice A/B test output, could you post the results as a comment here?

This feature has been available for a while in dials, this PR makes it the default. I have used this feature on and off over the past year and not run into any issues but would welcome other people testing it out. In recent versions of dials, this can be turned on with the environment variable "DIALS_USE_H5=TRUE".

Immediate benefits are generally faster read/writes, ~~slightly reduced file size~~ and ability to interact with dials data files outside of the dials ecosystem.
Longer term benefits are that this will enable routes for performing selective loading of subsets of data, which could significantly reduce memory usage and runtime for big datasets (e.g. multi-crystal, SSX, XFEL) by not requiring all reflection data to be loaded into memory at the start of a dials program.


graeme-winter:
A question around this

Is the file structure set in stone for these?

Asked as we have recently discussed the ideas around reflex files and having some internal consistency could be useful

jbeilstenedmands:
I don't wish for it to be set in stone, current structure is:
```
/dials                   Group
/dials/processing        Group
/dials/processing/group_0 Group
/dials/processing/group_0/background.mean Dataset {212296}
/dials/processing/group_0/background.sum.value Dataset {212296}
.....
/nx_reflections          Group
/nx_reflections/group_0  Group
/nx_reflections/group_0/background.mean Dataset, same as /dials/processing/group_0/background.mean
/nx_reflections/group_0/bounding_box Dataset, same as /dials/processing/group_0/bbox
.....
```
with a little bit of experiment identifier mapping metadata in /dials/processing/group_0.

All data is currently saved in group_0, I did it this way to leave the option to save data into multiple groups for e.g. multi-crystal data or parallel writers, but maybe this was overkill.

Code for reading/writing is in `dials/util/table_as_hdf5_file.py`. @graeme-winter Should we add some versioning string in, so we can change structure in future and keep compatibility?

dagewa:
Here's some plots from @faiadhussain's dashboard.

### Memory comparison
A is the default with MessagePack `.refl`
B is the HDF5 `.refl`

image

The line is very close to Y=X. There are a handful of datasets where HDF5 output uses a chunk more RAM than MessagePack. It is not clear to me why, but this was reproducible, at least for the dataset with the biggest difference.

### Wallclock time comparison

image

This is the accumulated time for all jobs recorded in xia2's timing data structure. With HDF5 the overall processing time is significantly shorter (approx 0.8 on average). Drilling down to an individual job we see the main saving is in faster spot-finding (though integration is a bit quicker too)

image

(ignore the X axes being at different scales - this is fixed on a branch, but I forgot and used the main branch)

### Data quality

Results are identical for all data sets, showing that choice of HDF5 produces no nasty surprises in data quality

graeme-winter:
> I don't wish for it to be set in stone, current structure is:

Good...

> Code for reading/writing is in `dials/util/table_as_hdf5_file.py`. @graeme-winter Should we add some versioning string in, so we can change structure in future and keep compatibility?

Absolutely we should add a versioning string at the top level. For the `rflx` format I had always envisaged a fairly "flat" structure e.g. 

```console
/experiments/<structure-like-expts-json>
/reflections/<structure-like-refl-msgpack>
```

The additional layers you have which don't seem to be functional e.g. `/dials/processing` make me a little edgy - I suspect I missed the boat to offer an opinion, but the overall data model we have now is relatively clean and sane and mapping this to HDF5 future proofs us for C++ applications which could then collaborate with DIALS to get things done ... faster

I may have been working on this and ended up with that flatter / simpler structure

My personal measure of this is "you are standing in front of a class of students explaining it: are you comfortable?"

graeme-winter:
As an example:

```console
/                        Group
/experiments             Group
/experiments/beam        Group
/experiments/beam/0      Group
/experiments/crystal     Group
/experiments/crystal/0   Group
/experiments/crystal/0/B_covariance Dataset {81}
/experiments/detector    Group
/experiments/detector/0  Group
/experiments/detector/0/hierarchy Group
/experiments/detector/0/hierarchy/children Group
/experiments/detector/0/hierarchy/children/0 Group
/experiments/detector/0/hierarchy/mask Dataset {0}
/experiments/detector/0/hierarchy/px_mm_strategy Group
/experiments/detector/0/panels Group
/experiments/detector/0/panels/0 Group
/experiments/detector/0/panels/0/mask Dataset {0}
/experiments/detector/0/panels/0/px_mm_strategy Group
/experiments/experiment  Group
/experiments/experiment/0 Group
/experiments/goniometer  Group
/experiments/goniometer/0 Group
/experiments/imageset    Group
/experiments/imageset/0  Group
/experiments/imageset/0/params Group
/experiments/imageset/0/single_file_indices Dataset {1800}
/experiments/profile     Dataset {0}
/experiments/scaling_model Dataset {0}
/experiments/scan        Group
/experiments/scan/0      Group
/experiments/scan/0/properties Group
/experiments/scan/0/properties/epochs Dataset {1800}
/experiments/scan/0/properties/exposure_time Dataset {1800}
/experiments/scan/0/properties/oscillation Dataset {1800}
/experiments/scan/0/valid_image_ranges Group
/reflections             Group
/reflections/0           Group
/reflections/0/bbox      Dataset {146118, 6}
/reflections/0/entering  Dataset {146118}
/reflections/0/flags     Dataset {146118}
/reflections/0/id        Dataset {146118}
/reflections/0/imageset_id Dataset {146118}
/reflections/0/intensity.sum.value Dataset {146118}
/reflections/0/intensity.sum.variance Dataset {146118}
/reflections/0/miller_index Dataset {146118, 3}
/reflections/0/n_signal  Dataset {146118}
/reflections/0/panel     Dataset {146118}
/reflections/0/rlp       Dataset {146118, 3}
/reflections/0/s1        Dataset {146118, 3}
/reflections/0/shoebox   Group
/reflections/0/shoebox/bbox Dataset {146118, 6}
/reflections/0/shoebox/panel Dataset {146118}
/reflections/0/shoebox/shoebox_background Dataset {3806805}
/reflections/0/shoebox/shoebox_data Dataset {3806805}
/reflections/0/shoebox/shoebox_mask Dataset {3806805}
/reflections/0/xyzcal.mm Dataset {146118, 3}
/reflections/0/xyzcal.px Dataset {146118, 3}
/reflections/0/xyzobs.mm.value Dataset {146118, 3}
/reflections/0/xyzobs.mm.variance Dataset {146118, 3}
/reflections/0/xyzobs.px.value Dataset {146118, 3}
/reflections/0/xyzobs.px.variance Dataset {146118, 3}
```

jbeilstenedmands:
I'm not too fussed about the extact depth of the structure, looks like your example still retains options for multiple groups? e.g. `reflections/0`, `reflections/1`. but that would not have groups mapping one-to-one to experiments, I was envisioning that the default is that multi-crystal data is saved to a single group, as this is how we currently store in memory.

jbeilstenedmands:
To be added to this PR before merging - a top level version string. I also need to check how we're saving shoeboxes with respect to v1 and v2 of shoeboxes in reflection table code.

jbeilstenedmands:
Data files are now created with attributes at the top level:

```
handle.attrs["file_type"] = "dials_processed_data"
handle.attrs["file_version"] = 1
```

If there are any strong preferences for what to name the file_type, chip in. This was kept generic enough to cover adding in .expt data in future.

I am happy that the shoebox versioning is fine, as the shoebox code converts any legacy data on disk (from pickle/msgpack format) into uint8 on loading, so it will only ever be saved as uint8 for hdf5. We will not be able to read hdf5 format data with shoeboxes stored as int rather than uint8 - but I don't think anyone has been using this format so far outside of testing, so I don't see this as an issue.

graeme-winter:
> Data files are now created with attributes at the top level:
> 
> ```
> handle.attrs["file_type"] = "dials_processed_data"
> handle.attrs["file_version"] = 1
> ```
> 
> If there are any strong preferences for what to name the file_type, chip in. This was kept generic enough to cover adding in .expt data in future.
> 
> I am happy that the shoebox versioning is fine, as the shoebox code converts any legacy data on disk (from pickle/msgpack format) into uint8 on loading, so it will only ever be saved as uint8 for hdf5. We will not be able to read hdf5 format data with shoeboxes stored as int rather than uint8 - but I don't think anyone has been using this format so far outside of testing, so I don't see this as an issue.

I would have gone with `dials.reflection_table` as file type since the file type with experiment info _will_ be different - being explicit feels right here (I welcome other opinions)

jbeilstenedmands:
I was thinking that the experiment information would be included within this `file_type` in future versions, so the `file_type` would stay the same but `file_version` number would increase. Otherwise for a reflex file, then we create a new `file_type` and have to check between the two, which then feels redundant with the version numbering. So the name should not be overly constrained to reflections only.

PR #3259, opened 2026-08-28
Frame-sliced integration results

dagewa:
The intensities in 3DED data are highly dependent on orientation due to dynamic diffraction effects. This causes dramatic changes in rocking curves, which can be modelled e.g. by Bloch wave simulation. Comparison of experimental data with simulation is limited by the loss of the fine details of the rocking curve when integrating a single value for the intensity.

This PR adds optional per-frame ("frame sliced") shoebox data during integration, providing a breakdown of shoebox data frame-by-frame for each reflection, thus retaining the summation integration rocking curve. Storing this additional `"frame_sliced_shoebox"`column is considerably cheaper than outputting the entire array of shoeboxes. The additional output is switched on using `integration.frame_slice_shoeboxes=True`, and is limited to use with rotation data.

This feature was developed in collaboration for [GROND](https://github.com/Martynowycz-Lab/GROND), however this can easily be used for comparison with other dynamic diffraction simulators, or for other circumstances where reconstruction of the rocking curves might be useful.

Features added:

- New `FrameSlicedShoebox` C++ class (exposed to Python) holding per-frame data for a single reflection: `frames`, `phi`, `excitation_error`, `partiality`, `summation_intensity`, `summation_intensity_variance` and `summation_intensity_valid`.
- New reflection table column type for `FrameSlicedShoebox`, with flex array support and serialisation to msgpack, pickle and HDF5.
- New `FrameSlicedIntegratorExecutor`, a subclass of `IntegratorExecutor`, which populates the column during integration.
- New `integration.frame_slice_shoeboxes` PHIL parameter to switch it on.
- `FrameOrientations` gains `phi_scan_points`, the rotation angles at the frame boundaries (the phi values are their midpoints).
- Tests at each level: the C++ class, the reflection table column and its serialisation, the executor, and an end-to-end test through dials.integrate.

Two consistency properties are enforced and tested end to end: the per-frame summation intensities and variances sum to `intensity.sum.value` and `intensity.sum.variance`, and the per-frame partialities sum to the `partiality` column. The excitation error is measured along the beam direction rather than radially from the centre of the Ewald sphere.


dagewa:
I plan to merge this by the end of the week, unless anyone has any objections. It is optional additional information calculated during integration so does not affect any existing standard code path. It also works with HDF5 output, so should not be affected by the switch to that as default.

Issue #1363, opened 2020-07-30
dials.scale exclude_images syntax is different to image range


graeme-winter:
This is confusing to users ... and demonstrators. 

Should it be like this? 

```
image_range = 1, 100
exclude_images = 201:300
```

dagewa:
I note that the `dials.scale` syntax includes an optional third element that specifies the experiment, so `201:300` is actually equivalent to `0:201:300`. One could make this something like `0:201,300` or `0,201,300`,  but it is awkward when you want to specify multiple ranges.

I _prefer_ the `dials.scale` way of specifying a range. `exclude_images=201,300` to me looks like you want to exclude two images, not the range from 201 to 300. In the DIALS scale syntax you could express a complicated set of ranges:

`exclude_images=0:201:300,1:50:90,2:175:180`.

I can't remember if it is actually done like this or by passing multiple `exclude_images=` definitions

biochem-fan:
@dagewa 

> exclude_images=0:201:300,1:50:90,2:175:180.
> I can't remember if it is actually done like this or by passing multiple exclude_images= definitions

Actually your single option example with commas does not work (gives an error). We have to pass "multiple exclude_images= definitions".

Very confusingly, if we specify `exclude_images="0:201:300 1:50:90 2:175:180"`, the program **silently** ignores the second and the third exclusion ranges, taking only "0:201:300".

I think we should either:
- give a friendly error to `exclude_images="0:201:300 1:50:90 2:175:180"`
- or (I prefer this) allow the single option syntax with spaces and/or commas.

jbeilstenedmands:
Thanks for the suggestions @biochem-fan 
I think it should be possible to allow both of those options, and I think that it's clunky/annoying to have to specify `exclude_images=` multiple times. Shouldn't be too difficult to implement so i'll see what I can do.

biochem-fan:
Another related issue:

![image](https://user-images.githubusercontent.com/1611483/149059448-287bb617-243a-47b7-93e8-515b726b96c4.png)

Suppose I am merging multiple crystals and want to truncate damaged later frames from each crystal. By hovering the mouse cursor on the "scale vs batch" plot, I see "1:54" and decide to truncate at the 54th frame of the second crystal. I am inclined to say `--exclude_images=1:54:999` (the last number does not matter) but this does not do what I want.

In this case, the scan range happened to start from 31 for the second crystal.
[...]

PR #3260, opened 2026-08-28
exclude_images: allow * to mean every dataset

### LINKED ISSUE #2560: dials.scale: allow wildcard for experiment in exclude_images
Just had to write a script to

```
dials.scale ../work/symmetrized.* d_min=1.5 \
exclude_images=0:21:100,1:21:100,2:21:100,3:21:100,4:21:100,5:21:100,6:21:100,7:21:100,8:21:100,9:21:100,10:21:100,11:21:100,12:21:100,13:21:100,14:21:100,15:21:100,16:21:100,17:21:100,18:21:100,19:21:100
```

but it feels like a legitimate use case for modest-large-N data sets to say exclude imges `a:b` for all runs.

Propose syntax: `exclude_images=*:21:100` in above case (since this is a common sense expectation)

graeme-winter:
Excluding the same image range from every dataset currently means spelling out one range per experiment:

  dials.scale symmetrized.* exclude_images=0:21:100,1:21:100,2:21:100,...

which for modest-to-large N is a job for a script, not a person. Accept * in place of the experiment number, so the above becomes

  dials.scale symmetrized.* exclude_images=*:21:100

The wildcard expands to the identifier of every experiment (falling back on the identifiers recorded in the reflection tables if no experiments are given), and may be mixed with explicit per-dataset ranges. As the expansion is in _parse_exclude_images_commands this works anywhere exclude_images is honoured - dials.scale, dials.integrate, dials.find_spots.

While here, extend the error message raised on a malformed range to spell out that ranges use start:stop while commas separate one range from the next. This is the point at which users hit the difference from image_range=start,stop noted in #1363; the two cannot be made to agree, since a comma already means something else here.

Fixes #2560
Refs #1363

### DIFF
--- a/src/dials/util/exclude_images.py
+++ b/src/dials/util/exclude_images.py
@@ -23,10 +23,12 @@ phil_scope = iotbx.phil.parse(
     .multiple = True
     .help = "Input in the format exp:start:end"
             "Exclude a range of images (start, stop) from the dataset with"
-            "experiment identifier exp  (inclusive of frames start, stop)."
+            "experiment identifier exp  (inclusive of images start, stop)."
             "Multiple ranges can be given in one go, e.g."
             "exclude_images=0:150:200,1:200:250"
             "exclude_images='0:150:200 1:200:250'"
+            "The experiment may be given as * to apply the same range to"
+            "every dataset, e.g. exclude_images=*:21:100"
     .short_caption = "Exclude images"
     .expert_level = 1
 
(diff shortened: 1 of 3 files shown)


graeme-winter:
Force push was I unintentionally rebased to main and pushed, then realised this would break the tests so dropped the last two commits from main and push -f

src/dials/util/exclude_images.py
```diff
@@ -96,11 +98,38 @@ def get_selection_for_valid_image_ranges(reflection_table, experiment):
...
+Ranges are given as start:stop, using colons: note that this differs from
+image_range=start,stop as used by dials.import and dials.slice_sequence,
+because here a comma separates one range from the next, e.g.
+exclude_images=0:100:150,1:120:200. Multiple ranges may also be separated by
+spaces, or given as repeated exclude_images= parameters. The experiment may be
+given as * to apply the same range to every dataset, e.g. exclude_images=*:21:100
+"""
+
+
+def _all_identifiers(experiments, reflections):
```
jbeilstenedmands: I don't think this practically gives us anything more than just using `experiments.identifiers()`?
I would favour keeping it simple here.

graeme-winter: Good spot, removed in d17db5d1cd462627defe5b9af92a8735eb4aaaf9

PR #3261, opened 2026-08-28
Fix "method" option dispatch in dials.search_beam_position

mpks:
The PHIL option "method" was changed from a list of strings into a single string in #3254, but one line in
`search_beam_position.py` still assumed it was a list of strings.

The error was not caught by CI, which skipped the regression tests for `test_search_beam_position.py`.


jbeilstenedmands (approved):
Thanks for the fix

PR #3269, opened 2026-09-04
Add `dials.export format=cif` for small molecule CIF output

### LINKED ISSUE #3009: `dials.export format=cif` for small molecule data
Currently the main route into structure solution for small molecule data is to export with `format=shelx`. However, this is not as metadata-rich as CIF, and has some issues (such as scaling values to fit within limited text field precision). It would be useful to implement a CIF export option. For 3D ED applications, this should include the agreed [ED-CIF additions,](https://github.com/COMCIFS/cif_ed).

dagewa:
Adds a core-CIF exporter using gemmi, for small molecule data reduction, giving a metadata-rich alternative to `format=shelx` that can be taken straight into structure solution (targeting Olex2).

**Additions**

- `dials.export format=cif` - writes a single data block with cell, symmetry, data reduction metadata and an unmerged `_diffrn_refln_ intensity` loop.
- Intensities are written at full precision, not rescaled to fit SHELX's F8.2 field.
- Reflection loop uses the `_intensity_net` / `_intensity_u` column pair and a `_space_group_symop_operation_xyz` loop, which is what Olex2 reads on import.
- Cell parameters carry standard uncertainties, suppressed for angles fixed by the space group symmetry, and for cells supplied with `best_unit_cell=` or derived from experiments with differing cells.
- Merging statistics (Rmerge, Rmeas, CC½, <I/σ>, multiplicity, completeness) are taken from `iotbx.merging_statistics`, in the same way as done for `format=mmcif`.
- `cif.chemical_formula=` writes `_chemical_formula_sum`, `_chemical_formula_weight` and `_cell_formula_units_Z`. The parameter name `chemical_formula` is used instead of `composition` to avoid collision with `format=shelx` (this is maybe a slightly annoying inconsistency)
- `cif.scale_group_code=True` writes the per-sweep SHELX batch number; off by default.
- Electron diffraction items of the core dictionary (COMCIFS release 1.0.0) when the beam probe is `electron`: `_diffrn_measurement_rotation_mode` from the scan oscillation width, `_diffrn_source_voltage` derived from the wavelength, and the source/device/method descriptions.
- `cif.extra=_data_name=value`, repeatable, for metadata a data reduction cannot know (e.g. precession angle, gun type, crystal description), overriding derived values.
- Shared `unit_cell_and_esds()`, `single_wavelength()` and `parse_compound()` factored out of `export_shelx` so both exporters report the same cell.

**Verification**

Tested end-to-end on the public NiDppeCl2 dataset (zenodo 20041091):

```bash
#!/bin/bash
curl https://zenodo.org/records/20041091/files/001_NiDppeCl2_01.tar.bz2 -o 001_NiDppeCl2_01.tar.bz2
tar xvjf 001_NiDppeCl2_01.tar.bz2
mkdir dials && cd dials
dials.import ../001_NiDppeCl2_01.nxs
dials.symmetry integrated.expt integrated.refl small_molecule=True
dials.two_theta_refine symmetrized.expt symmetrized.refl
dials.scale refined_cell.expt symmetrized.refl
dials.export scaled.expt scaled.refl format=cif chemical_formula=C26H24Cl2NiP2
```

the exported CIF loads in Olex2 and solves with ShelXT.

Unit and command line tests cover the loop contents, cell SUs, symmetry, formula, ED items and cif.extra.

Not included here: polychromatic (Laue/ToF) data, and merged data in an optional `merged _refln_F_squared_meas` block for archiving.

Closes #3009


dwmoreau:
I attended the Olex2 workshop at the IUCR, Horst Puschmann was adamant that data reduction applications should export a single .cif file instead of the .ins & .hkl format.

dagewa:
I've heard him say the same too. This should complete that 😊

dagewa:
> cif.chemical_formula= writes _chemical_formula_sum, _chemical_formula_weight and _cell_formula_units_Z. The parameter name chemical_formula is used instead of composition to avoid collision with format=shelx (this is maybe a slightly annoying inconsistency)

I think actually this is slightly too annoying. I propose moving `composition` out of the `shelx` scope and into a new `small_molecule` scope. Then both `format=shelx` and `format=cif` can read `composition` from here rather than having a different parameter in the `cif` scope.

PR #3270, opened 2026-09-04
dials.countrate: Countrate analysis tool

pblowey:
Photon counting detectors such as the Dectris Eiger2, employ countrate corrections to account for the under-reporting of counts caused by deadtime. The manufacturers define a maximum trusted range, which corresponds to the range that the manufacturer deems these countrate corrections to be valid and values that exceed this range are returned as invalid pixels.

It has been reported that, towards the upper limit of the trusted range, the countrate corrections can be inaccurate, which has an impact on data processing, particularly data processing involving anomalous diffraction data.

dials.countrate takes the "shoeboxes" (raw pixel data) from a reflection table (e.g. strong.refl from dials.find_spots) and calculates a histogram of pixel intensities. A maximum recommended transmission is then calculated to keep a specified percentile of pixel intensities below a target percentage of the detector trusted range. The idea of this is to try and keep most of the pixel intensities in the linear range of the detector and limit reliance on the detector countrate correction.

Default parameters of 99.9% as a reference percentile and 10% as a target percentage of the detector trusted range. 10% is selected as a safe upper limit for the linear response of the detector. The default reference percentile of 99.9% was selected as a balance between ensuring the maximum amount of pixels are within the linear response of the detector and keeping low exposure times for fast data collections. 

Relies on https://github.com/cctbx/dxtbx/pull/896 and https://github.com/dials/nxmx/pull/78 to allow the experiment transmission to be read in from nxmx format files.

### DIFF
--- a/src/dials/command_line/countrate.py
+++ b/src/dials/command_line/countrate.py
@@ -1,0 +1,197 @@
+# DIALS_ENABLE_COMMAND_LINE_COMPLETION
+
+
+from __future__ import annotations
+
+import json
+import logging
+import pathlib
+import time
+from collections import Counter
+from itertools import accumulate
+
+import h5py
+
+import iotbx.phil
+
+from dials.util import log, show_mail_handle_errors
+from dials.util.options import ArgumentParser, reflections_and_experiments_from_files
+from dials.util.version import dials_version
+
+logger = logging.getLogger("dials.commandline.countrate")
+
+help_message = """
+dials.countrate: Analyse pixel intensities from reflection shoebox data.
+
+Usage examples:
+    dials.countrate imported.expt strong.refl
+"""
+
+phil_scope = iotbx.phil.parse(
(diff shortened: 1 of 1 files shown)


phyy-nx:
Cool!  Is the idea that the histogram would inform a new trusted range for the data that would be enforced in dials.import and then downstream through integration?  Or is this a calibration tool for beamline scientist to keep exposure below a threshhold?

pblowey:
Hi! It's the latter rather than the former. The original idea for this was to be part of a screening/strategy pipeline for small wedge (15°) data, where the pipeline would read in the data, find spots and warn/advise if the data was likely to be affected by heavy countrate corrections and suggest a transmission to take the full data collection with. There was some interest in having this as a standalone dials tool (hence this PR).

phyy-nx:
Got it.  Sounds good to me!

PR #3275, opened 2026-09-16
Update micromamba URL

ndevenish:
The old URL has an expired certificate.


ndevenish:
All CI has made it past the broken point, merging

PR #3276, opened 2026-09-16
Add dials.remove_columns

graeme-winter:
New command line program to remove named columns from a reflection table, for example to reduce the size of a file on disk by discarding data which are no longer needed for the analysis to be performed:

  dials.remove_columns indexed.refl \
    remove=xyzobs.mm.value,xyzobs.mm.variance output=small.refl

Columns may be given comma separated, repeated as separate remove= options, or matched with shell-style wildcards against the columns present. A pattern which matches nothing warns; a command line where nothing at all matches is an error, so that a mistyped column name does not silently do nothing. Removing every column is refused, and removing id or panel warns that the output is unlikely to be usable by other DIALS programs.

The output parameter is a flat definition rather than an output.reflections scope, following dials.merge_reflection_lists and dials.predict.


dagewa:
Perhap this should be merged into `dials.filter_reflections` in some way? 

I note also that we wanted to expand that program to handle experiments, and call it `dials.filter_experiments` https://github.com/dials/dials/issues/2236

Just a thought. I'm not totally against adding a simple single-use jiffy, if the use case is different enough to warrant a separate tool.

graeme-winter:
> Perhap this should be merged into `dials.filter_reflections` in some way?
> 
> I note also that we wanted to expand that program to handle experiments, and call it `dials.filter_experiments` #2236
> 
> Just a thought. I'm not totally against adding a simple single-use jiffy, if the use case is different enough to warrant a separate tool.

This is not filtering reflections _per se_ because all the reflections remain just with less data attached. It's a single use jiffy. I will listen to push back, though. I just found myself repeatedly writing and running the same script.

jbeilstenedmands:
We had a brief discussion on this, perhaps it would be best to have it as a `dev.dials` tool, as it would be unlikely something a standard user would use?

PR #3277, opened 2026-09-16
Adding PCA graphs to dials.cosym

amyjaynethompson:
This adds the matrix-style plot from dials.correlation matrix into dials.cosym to better visualise higher dimension symmetry cases. This involved adding some of the core rotation logic into dials.cosym from dials.correlation matrix. The pca plots function was also generalised to accept input without cluster classification.


jbeilstenedmands (approved):
Thanks for adding this feature, the newsfragment needs manually renaming before merge.

jbeilstenedmands:
I renamed it.

